PR #1, opened 2016-02-01
Upgrade winrm4j to version 0.3.1

bostko:
- added config key for use Https or not
- added config key for using NTLM
- refactor WINRM_PORT usage

### DIFF
--- a/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
+++ b/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
@@ -731,7 +731,7 @@ public class JcloudsLocation extends AbstractCloudMachineProvisioningLocation im
 
             boolean windows = isWindows(node, setup);
             if (windows) {
-                int newLoginPort = node.getLoginPort() == 22 ? 5985 : node.getLoginPort();
+                int newLoginPort = node.getLoginPort() == 22 ? (getConfig(WinRmMachineLocation.USE_HTTPS_WINRM) ? 5986 : 5985) : node.getLoginPort();
                 String newLoginUser = "root".equals(node.getCredentials().getUser()) ? "Administrator" : node.getCredentials().getUser();
                 LOG.debug("jclouds created Windows VM {}; transforming connection details: loginPort from {} to {}; loginUser from {} to {}", 
                         new Object[] {node, node.getLoginPort(), newLoginPort, node.getCredentials().getUser(), newLoginUser});
@@ -1535,7 +1535,7 @@ public class JcloudsLocation extends AbstractCloudMachineProvisioningLocation im
         boolean windows = isWindows(template, config);
         if (windows) {
             if (!(config.containsKey(JcloudsLocationConfig.USER_METADATA_STRING) || config.containsKey(JcloudsLocationConfig.USER_METADATA_MAP))) {
-                config.put(JcloudsLocationConfig.USER_METADATA_STRING, WinRmMachineLocation.getDefaultUserMetadataString());
+                config.put(JcloudsLocationConfig.USER_METADATA_STRING, WinRmMachineLocation.getDefaultUserMetadataString(config()));
             }
         }
                
@@ -2297,7 +2297,7 @@ public class JcloudsLocation extends AbstractCloudMachineProvisioningLocation im
                     .configure("jcloudsParent", this)
                     .configure("displayName", vmHostname)
                     .configure("address", address)
-                    .configure(WinRmMachineLocation.WINRM_PORT, sshHostAndPort.isPresent() ? sshHostAndPort.get().getPort() : node.getLoginPort())
+                    .configure(WinRmMachineLocation.WINRM_CONFIG_PORT, sshHostAndPort.isPresent() ? sshHostAndPort.get().getPort() : node.getLoginPort())
                     .configure("user", getUser(setup))
                     .configure(WinRmMachineLocation.USER, setup.get(USER))
                     .configure(WinRmMachineLocation.PASSWORD, setup.get(PASSWORD))
@@ -2476,7 +2476,7 @@ public class JcloudsLocation extends AbstractCloudMachineProvisioningLocation im
                     hostAndPortOverride = ((SshMachineLocation)machine).getSshHostAndPort();
                 } else if (machine instanceof WinRmMachineLocation) {
                     String host = ((WinRmMachineLocation)machine).getAddress().getHostAddress();
-                    int port = ((WinRmMachineLocation)machine).config().get(WinRmMachineLocation.WINRM_PORT);
+                    int port = ((WinRmMachineLocation)machine).getPort();
                     hostAndPortOverride = HostAndPort.fromParts(host, port);
                 } else {
                     LOG.warn("Unexpected machine {} of type {}; expected SSH or WinRM", machine, (machine != null ? machine.getClass() : null));
@@ -2644,7 +2644,8 @@ public class JcloudsLocation extends AbstractCloudMachineProvisioningLocation im
         String user = (users.size() == 1) ? Iterables.getOnlyElement(users) : "{" + Joiner.on(",").join(users) + "}";
         String vmIp = hostAndPortOverride.isPresent() ? hostAndPortOverride.get().getHostText() : getFirstReachableAddress(node, setup);
         if (vmIp==null) LOG.warn("Unable to extract IP for "+node+" ("+setup.getDescription()+"): subsequent connection attempt will likely fail");
-        int vmPort = hostAndPortOverride.isPresent() ? hostAndPortOverride.get().getPortOrDefault(5985) : 5985;
+        int defaultWinRmPort = getConfig(WinRmMachineLocation.USE_HTTPS_WINRM) ? 5986 : 5985;
+        int vmPort = hostAndPortOverride.isPresent() ? hostAndPortOverride.get().getPortOrDefault(defaultWinRmPort) : defaultWinRmPort;
 
         String connectionDetails = user + "@" + vmIp + ":" + vmPort;
         final HostAndPort hostAndPort = hostAndPortOverride.isPresent() ? hostAndPortOverride.get() : HostAndPort.fromParts(vmIp, vmPort);
--- a/software/base/src/main/java/org/apache/brooklyn/entity/software/base/VanillaWindowsProcess.java
+++ b/software/base/src/main/java/org/apache/brooklyn/entity/software/base/VanillaWindowsProcess.java
@@ -26,7 +26,9 @@ import org.apache.brooklyn.api.entity.ImplementedBy;
 import org.apache.brooklyn.api.sensor.AttributeSensor;
 import org.apache.brooklyn.config.ConfigKey;
 import org.apache.brooklyn.core.config.ConfigKeys;
+import org.apache.brooklyn.core.sensor.AttributeSensorAndConfigKey;
 import org.apache.brooklyn.core.sensor.Sensors;
+import org.apache.brooklyn.util.core.internal.winrm.WinRmTool;
 import org.apache.brooklyn.util.time.Duration;
 
 import com.google.common.collect.ImmutableSet;
@@ -37,7 +39,7 @@ public interface VanillaWindowsProcess extends AbstractVanillaProcess {
     // 3389 is RDP; 5985 is WinRM (3389 isn't used by Brooklyn, but useful for the end-user subsequently)
     ConfigKey<Collection<Integer>> REQUIRED_OPEN_LOGIN_PORTS = ConfigKeys.newConfigKeyWithDefault(
             SoftwareProcess.REQUIRED_OPEN_LOGIN_PORTS,
-            ImmutableSet.of(5985, 3389));
+            ImmutableSet.of(5986, 5985, 3389));
     
     @CatalogConfig(label = "Install PowerShell command", priority=5.5)
     ConfigKey<String> INSTALL_POWERSHELL_COMMAND = ConfigKeys.newStringConfigKey("install.powershell.command",
@@ -103,5 +105,5 @@ public interface VanillaWindowsProcess extends AbstractVanillaProcess {
             "duration to wait whilst waiting for a machine to finish rebooting, and thus to become available again", Duration.minutes(30));
     
     AttributeSensor<Integer> RDP_PORT = Sensors.newIntegerSensor("rdpPort");
-    AttributeSensor<Integer> WINRM_PORT = Sensors.newIntegerSensor("winrmPort");
+    AttributeSensor<Integer> WINRM_PORT = Sensors.newIntegerSensor(WinRmTool.PROP_PORT.getName());
 }
--- a/software/base/src/main/java/org/apache/brooklyn/entity/software/base/VanillaWindowsProcessImpl.java
+++ b/software/base/src/main/java/org/apache/brooklyn/entity/software/base/VanillaWindowsProcessImpl.java
@@ -19,6 +19,10 @@
 package org.apache.brooklyn.entity.software.base;
 
 
+import com.google.common.collect.Iterables;
+import org.apache.brooklyn.api.location.Location;
+import org.apache.brooklyn.location.winrm.WinRmMachineLocation;
+
 public class VanillaWindowsProcessImpl extends SoftwareProcessImpl implements VanillaWindowsProcess {
     @Override
     public Class getDriverInterface() {
@@ -29,7 +33,10 @@ public class VanillaWindowsProcessImpl extends SoftwareProcessImpl implements Va
     protected void preStart() {
         super.preStart();
         sensors().set(RDP_PORT, 3389);
-        sensors().set(WINRM_PORT, 5985);
+        WinRmMachineLocation loc = Iterables.getFirst(Iterables.filter(getLocations(), WinRmMachineLocation.class), null);
+        if (loc != null) {
+            sensors().set(WINRM_PORT, loc.getPort());
+        }
     }
     
     @Override
--- a/software/winrm/src/main/java/org/apache/brooklyn/location/winrm/AdvertiseWinrmLoginPolicy.java
+++ b/software/winrm/src/main/java/org/apache/brooklyn/location/winrm/AdvertiseWinrmLoginPolicy.java
@@ -27,6 +27,7 @@ import org.apache.brooklyn.api.sensor.SensorEventListener;
 import org.apache.brooklyn.core.entity.AbstractEntity;
 import org.apache.brooklyn.core.policy.AbstractPolicy;
 import org.apache.brooklyn.core.sensor.Sensors;
+import org.apache.brooklyn.util.core.internal.winrm.WinRmTool;
 import org.slf4j.Logger;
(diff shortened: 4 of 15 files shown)


software/base/src/main/java/org/apache/brooklyn/entity/software/base/VanillaWindowsProcess.java
```diff
@@ -103,5 +105,5 @@
             "duration to wait whilst waiting for a machine to finish rebooting, and thus to become available again", Duration.minutes(30));
     
     AttributeSensor<Integer> RDP_PORT = Sensors.newIntegerSensor("rdpPort");
-    AttributeSensor<Integer> WINRM_PORT = Sensors.newIntegerSensor("winrmPort");
+    AttributeSensorAndConfigKey<Integer, Integer> WINRM_PORT = WinRmTool.PROP_PORT;
```
neykov: Use `PortAttributeSensorAndConfigKey`, same for `RDP_PORT`.

software/base/src/main/java/org/apache/brooklyn/entity/software/base/VanillaWindowsProcessImpl.java
```diff
@@ -29,7 +33,12 @@ public Class getDriverInterface() {
     protected void preStart() {
         super.preStart();
         sensors().set(RDP_PORT, 3389);
-        sensors().set(WINRM_PORT, 5985);
+        Location loc = Iterables.getFirst(getLocations(), null);
+        if (loc != null) {
+            if(loc.getConfig(WinRmMachineLocation.USE_SECURE_WINRM)) {
+                sensors().set(WINRM_PORT, 5986);
+            }
+        }
```
neykov: Why not handle the  `!USE_SECURE_WINRM` case? Should be possible to use both protocols.

bostko: I fixed it thank you!

software/winrm/src/main/java/org/apache/brooklyn/location/winrm/WinRmMachineLocation.java
```diff
@@ -197,7 +198,15 @@ public String getHostname() {
...
         String host = getHostname();
-        return (host == null) ? null : host + ":" + config().get(WINRM_PORT);
+        return (host == null) ? null : host + ":" + (config().get(USE_SECURE_WINRM) ? 5986 : 5985);
+    }
+
+    public int getPort() {
+        if (getConfig(WINRM_PORT) != null) {
+            return getConfig(WINRM_PORT);
+        } else {
+            return getConfig(USE_SECURE_WINRM) ? 5986 : 5985;
```
neykov: Where do `WINRM_PORT` & `USE_SECURE_WINRM` get set on the machine location?

bostko: From the location config from the blueprint.
If you set a location config `winrm.useHttps` to the JcloudsLocation from the blueprint, the config is inherited for the MachineLocation.

software/base/src/main/java/org/apache/brooklyn/entity/software/base/VanillaWindowsProcess.java
bostko: The port is a constant in this code which is retrieved by the location config.
It is not a trivial case where an entity should allocate the next available port.
I do not think this change is needed.

software/winrm/src/main/java/org/apache/brooklyn/location/winrm/WinRmMachineLocation.java
aledsage: It seems very strange to have `VanillaWindowsProcess.WINRM_PORT` and also a completely separate `WINRM_PORT` config on the location. I see that in a `VanillaWindowsProcessImpl.preStart` it reads the `machine. USE_SECURE_WINRM` but it then makes its own decision about the port number, rather than using the value from the machine. I can see how this works because we're using standard port numbers, but it's confusing (and would likely get very confusing if non-standard port numbers were every used).

software/winrm/src/main/java/org/apache/brooklyn/util/core/internal/winrm/winrm4j/Winrm4jTool.java
```diff
@@ -75,7 +78,11 @@ public Winrm4jTool(Map<String,?> config) {
     public Winrm4jTool(ConfigBag config) {
         this.bag = checkNotNull(config, "config bag");
         host = getRequiredConfig(config, PROP_HOST);
-        port = getRequiredConfig(config, PROP_PORT);
+        port = config.get(PROP_PORT.getConfigKey()) != null
```
aledsage: This will almost never return non-null, because there is a default value - and that default value is 5985. As long as we've overridden that value with 5986 for https by the time we get here then I can see this will work.

software/winrm/src/main/java/org/apache/brooklyn/util/core/internal/winrm/winrm4j/Winrm4jTool.java
```diff
@@ -184,7 +191,13 @@ public Winrm4jTool(ConfigBag config) {
     }
 
     private io.cloudsoft.winrm4j.winrm.WinRmTool connect() {
-        return io.cloudsoft.winrm4j.winrm.WinRmTool.connect(host+":"+port, user, password);
+        WinRmTool.Builder builder = WinRmTool.Builder.builder(host, user, password);
+        builder.setAuthenticationScheme(authenticationScheme);
+        builder.useHttps(useSecureWinrm);
+        if (useSecureWinrm) {
+            builder.disableCertificateChecks(true);
```
aledsage: This deserves a TODO in the code, I think. That we disable certificate checks when told to "use secure" is concerning, but I understand why you do it (see new FIXME in `WinRmMachineLocation.java`).

software/winrm/src/main/java/org/apache/brooklyn/util/core/internal/winrm/winrm4j/Winrm4jTool.java
bostko: The logic I follow here is:
When initializing the `WinRmTool` this could be set explicitly from the WinRmTool properties
however if not the default value varies to be either 5985 or 5986.
It cannot be set from https://github.com/bostko/brooklyn-server/blob/upgrade_winrm4j/software/winrm/src/main/java/org/apache/brooklyn/location/winrm/WinRmMachineLocation.java#L324
The most proper way is to set this from the entity as we already have this sensor set with this logic in the VanillaSoftwareProcess, however there is no link between the entity and the machine location so I do this check `config.get(USE_SECURE_WINRM) ` once again.

bostko: The best option for supporting https is to have two separate parameters 
for `port`
`winrm.port` and `winrm.https.port`. However one port to open/enable is enough for Brooklyn to work.

software/base/src/main/java/org/apache/brooklyn/entity/software/base/VanillaWindowsProcess.java
neykov: `PortAttributeSensorAndConfigKey` works with fixed port as well, see analogous [SSH_PORT](https://github.com/apache/brooklyn-server/blob/master/core/src/main/java/org/apache/brooklyn/core/entity/Attributes.java#L75). But fine to leave as is for this PR.

software/winrm/src/main/java/org/apache/brooklyn/location/winrm/WinRmMachineLocation.java
neykov: +1 for getting the port from the location.

neykov: The config key duplication comes from the location deciding which login port to use, but the entity configuring the inbound ports. Seems there's no way for the location to influence the inbound ports. For ssh this works because 22 is the default for [`REQUIRED_OPEN_LOGIN_PORTS`](https://github.com/apache/brooklyn-server/blob/master/software/base/src/main/java/org/apache/brooklyn/entity/software/base/SoftwareProcess.java#L66). 

Getting the port from the location (above comment) will allow configuring the port at a single place.  

Cleaner solution will be to allow the location to add inbound ports and remove the keys from `VanillaWindowsProcessImpl`, but not in scope for this PR.

software/winrm/src/main/java/org/apache/brooklyn/location/winrm/WinRmMachineLocation.java
```diff
@@ -388,20 +401,38 @@ public static String getDefaultUserMetadataString() {
...
 
+        // FIXME USE_SECURE_WINRM
+        // Missing generate certificate step.
+        // https://support.microsoft.com/en-us/kb/2019527
+        //
+        // One possible approach is to generate a certificate and append it to this command.
+        // http://stackoverflow.com/questions/1615871/creating-an-x509-certificate-in-java-without-bouncycastle
+        boolean useSecureWinrm = config.getBag().get(USE_SECURE_WINRM);
+
+        boolean credSsp = config.getBag().get(USE_NTLM);
```
neykov: SSP is not NTLM. Better leave it alone (or use separate config key).
`Negotiate` includes both `Kerberos` and `NTLM` (but not sure what happens when `Kerberos` is explicitly disabled).

software/winrm/src/main/java/org/apache/brooklyn/location/winrm/WinRmMachineLocation.java
```diff
@@ -388,20 +401,38 @@ public static String getDefaultUserMetadataString() {
...
-        return "winrm quickconfig -q & " +
-                "winrm set winrm/config/service/auth @{Basic=\"true\"} & " +
-                "winrm set winrm/config/service/auth @{CredSSP=\"true\"} & " +
-                "winrm set winrm/config/client/auth @{CredSSP=\"true\"} & " +
-                "winrm set winrm/config/client @{AllowUnencrypted=\"true\"} & " +
-                "winrm set winrm/config/service @{AllowUnencrypted=\"true\"} & " +
+        return String.format("winrm quickconfig -q & " +
+                "winrm set winrm/config/service/auth @{Basic=\"%1$s\"} & " +
+                "winrm set winrm/config/service/auth @{CredSSP=\"%2$s\"} & " +
+                "winrm set winrm/config/client/auth @{CredSSP=\"%2$s\"} & " +
```
neykov: Why configure the winrm client? Not needed.

software/winrm/src/main/java/org/apache/brooklyn/location/winrm/WinRmMachineLocation.java
```diff
@@ -388,20 +401,38 @@ public static String getDefaultUserMetadataString() {
...
+        // FIXME USE_SECURE_WINRM
+        // Missing generate certificate step.
+        // https://support.microsoft.com/en-us/kb/2019527
+        //
+        // One possible approach is to generate a certificate and append it to this command.
+        // http://stackoverflow.com/questions/1615871/creating-an-x509-certificate-in-java-without-bouncycastle
+        boolean useSecureWinrm = config.getBag().get(USE_SECURE_WINRM);
+
+        boolean credSsp = config.getBag().get(USE_NTLM);
+        boolean basicAuth = !credSsp, allowEncrypted = !useSecureWinrm;
```
neykov: `allowEncrypted` should be `allowUnencrypted`.

software/winrm/src/main/java/org/apache/brooklyn/location/winrm/WinRmMachineLocation.java
bostko: You are right.
I tested it connecting to a machine with disabled CredSSP and it was working.

software/base/src/main/java/org/apache/brooklyn/entity/software/base/VanillaWindowsProcess.java
neykov: Scratch that. Login port can't be configured at the entity (for both SSH and WINRM). Even if a config value is set it will be ignored. Leave it as sensor, getting the value from the location.

software/winrm/src/main/java/org/apache/brooklyn/location/winrm/WinRmMachineLocation.java
```diff
@@ -72,8 +72,10 @@
             "address",
             "Address of the remote machine");
 
-    public static final ConfigKey<Integer> WINRM_PORT = WinRmTool.PROP_PORT;
-    
+    public static final ConfigKey<Integer> WINRM_PORT = WinRmTool.PROP_PORT.getConfigKey();
```
neykov: Better make this key prefixed, analogously to the `SSH_PORT` for consistency.

software/winrm/src/main/java/org/apache/brooklyn/location/winrm/WinRmMachineLocation.java
```diff
@@ -72,8 +72,10 @@
...
+    public static final ConfigKey<Boolean> USE_SECURE_WINRM = WinRmTool.USE_SECURE_WINRM;
```
neykov: Minor, we want `HTTPS` transport here, so a `USE_HTTPS_WINRM` could be a better name.

software/winrm/src/main/java/org/apache/brooklyn/location/winrm/WinRmMachineLocation.java
```diff
@@ -72,8 +72,10 @@
...
+    public static final ConfigKey<Boolean> USE_NTLM = WinRmTool.USE_NTLM;
```
neykov: In an ideal world this would be a list of white listed authentication protocols - 'Kerberos, NTLM' by default. But good enough for this PR - could mark it `@Beta`, subject to change.

software/winrm/src/main/java/org/apache/brooklyn/location/winrm/AdvertiseWinrmLoginPolicy.java
```diff
@@ -68,7 +69,7 @@ public void onEvent(SensorEvent<Location> event) {
     protected void advertiseUserAsync(final Entity entity, final WinRmMachineLocation machine) {
         String user = machine.getUser();
         String hostname = machine.getHostname();
-        int port = machine.config().get(WinRmMachineLocation.WINRM_PORT);
+        Integer port = machine.getPort();
```
neykov: `getPort` returns `int` and can't be null - keep the `int` type here as well.

software/winrm/src/main/java/org/apache/brooklyn/location/winrm/WinRmMachineLocation.java
```diff
@@ -388,20 +401,38 @@ public static String getDefaultUserMetadataString() {
                 "}"
         ));
 
+        // FIXME USE_SECURE_WINRM
+        // Missing generate certificate step.
+        // https://support.microsoft.com/en-us/kb/2019527
+        //
+        // One possible approach is to generate a certificate and append it to this command.
+        // http://stackoverflow.com/questions/1615871/creating-an-x509-certificate-in-java-without-bouncycastle
```
neykov: The certificate is best generated on the machine, without leaving it ever (for self-signed case). On the other hand it's not possible to get the public part at this step. 

I see this setup step dissappearing longer term (or keeping it minimalistic). Instead do something like the jclouds init sequence where it connects with whatever is provided by the cloud, then configuring it to our liking.

software/winrm/src/main/java/org/apache/brooklyn/util/core/internal/winrm/WinRmTool.java
```diff
@@ -43,7 +44,9 @@
             "static final initializer classload ordering problem");
     
     ConfigKey<String> PROP_HOST = newStringConfigKey("host", "Host to connect to (required)", null);
-    ConfigKey<Integer> PROP_PORT = newIntegerConfigKey("port", "WinRM port to use when connecting to the remote machine", 5985);
+    AttributeSensorAndConfigKey<Integer, Integer> PROP_PORT = ConfigKeys.newIntegerSensorAndConfigKey("port", "WinRM port to use when connecting to the remote machine");
```
neykov: Doesn't need to be a sensor?

software/winrm/src/main/java/org/apache/brooklyn/util/core/internal/winrm/WinRmTool.java
```diff
@@ -43,7 +44,9 @@
...
+    ConfigKey<Boolean> USE_SECURE_WINRM = ConfigKeys.newBooleanConfigKey("winrm.useHttps", "The parameter configures tells the machine sensors whether the winrm port is over https. If the parameter is true then 5986 will be used as a winrm port.", false);
```
neykov: `SECURE`->`HTTPS`

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -2297,7 +2297,6 @@ protected JcloudsWinRmMachineLocation createWinRmMachineLocation(ComputeService
                     .configure("jcloudsParent", this)
                     .configure("displayName", vmHostname)
                     .configure("address", address)
-                    .configure(WinRmMachineLocation.WINRM_PORT, sshHostAndPort.isPresent() ? sshHostAndPort.get().getPort() : node.getLoginPort())
```
neykov: Keep this, used by `WinRmMachineLocation.getPort()`. The override (port forwarding) port would be lost otherwise.

neykov:
Finished review.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
bostko: @neykov there is no WinRmMachineLocation instance here?

neykov: This is creating the `WinRmMachineLocation` and its `getPort` will look for `WINRM_CONFIG_PORT` key, but you are not passing it any.

software/base/src/main/java/org/apache/brooklyn/entity/software/base/VanillaWindowsProcessImpl.java
```diff
@@ -29,7 +33,14 @@ public Class getDriverInterface() {
     protected void preStart() {
         super.preStart();
         sensors().set(RDP_PORT, 3389);
-        sensors().set(WINRM_PORT, 5985);
+        Location loc = Iterables.getFirst(getLocations(), null);
+        if (loc != null) {
+            if(loc.getConfig(WinRmMachineLocation.USE_HTTPS_WINRM)) {
+                sensors().set(WINRM_PORT, 5986);
+            } else {
+                sensors().set(WINRM_PORT, 5985);
```
neykov: Get the port from the location instead.

neykov:
Looks good now, waiting for confirmation from tests before merging.

PR #4, opened 2016-02-03
Initial support for SaltStack.

geomacy:
See https://issues.apache.org/jira/browse/BROOKLYN-217.

This commit merges code from incubator-brooklyn, following the move to the new repos
after graduation. As such, it unfortunately loses the commit history, but the code
is based on the work done by Hadrian Zbarcea.  See the Jira link above for history
and discussion.

### DIFF
--- a/launcher/pom.xml
+++ b/launcher/pom.xml
@@ -73,6 +73,11 @@
         </dependency>
         <dependency>
             <groupId>org.apache.brooklyn</groupId>
+            <artifactId>brooklyn-cm-salt</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.brooklyn</groupId>
             <artifactId>brooklyn-utils-common</artifactId>
             <version>${project.version}</version>
         </dependency>
--- a/utils/common/src/main/java/org/apache/brooklyn/util/yaml/Yamls.java
+++ b/utils/common/src/main/java/org/apache/brooklyn/util/yaml/Yamls.java
@@ -151,14 +151,14 @@ public class Yamls {
 
     /** simplifies new Yaml().loadAll, and converts to list to prevent single-use iterable bug in yaml */
     @SuppressWarnings("unchecked")
-    public static Iterable<Object> parseAll(String yaml) {
+    public static List<Object> parseAll(String yaml) {
         Iterable<Object> result = new org.yaml.snakeyaml.Yaml().loadAll(yaml);
         return (List<Object>) getAs(result, List.class);
     }
 
     /** as {@link #parseAll(String)} */
     @SuppressWarnings("unchecked")
-    public static Iterable<Object> parseAll(Reader yaml) {
+    public static List<Object> parseAll(Reader yaml) {
         Iterable<Object> result = new org.yaml.snakeyaml.Yaml().loadAll(yaml);
         return (List<Object>) getAs(result, List.class);
     }
--- a/pom.xml
+++ b/pom.xml
@@ -194,6 +194,7 @@
         <module>locations/jclouds</module>
 
         <module>software/base</module>
+        <module>software/cm/salt</module>
         <module>software/winrm</module>
 
         <module>storage/hazelcast</module>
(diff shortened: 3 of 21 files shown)


utils/common/src/main/java/org/apache/brooklyn/util/yaml/Yamls.java
```diff
@@ -151,14 +151,14 @@ public static void dump(int depth, Object r) {
 
     /** simplifies new Yaml().loadAll, and converts to list to prevent single-use iterable bug in yaml */
     @SuppressWarnings("unchecked")
-    public static Iterable<Object> parseAll(String yaml) {
+    public static List<Object> parseAll(String yaml) {
```
ahgittin: prefer the return type being `Iterable` as that allows lazy eval.  callers who need a list can do `MutableList.copyOf(Iterable)`.

launcher/pom.xml
```diff
@@ -73,6 +73,11 @@
         </dependency>
         <dependency>
             <groupId>org.apache.brooklyn</groupId>
+            <artifactId>brooklyn-cm-salt</artifactId>
```
ahgittin: launcher should be quite bare.  i think the `all` project is where things like this are brought in.

ahgittin:
The error looks like a problem with the jenkins/dev setup / assumption in where the UI is.  Probably better to declare a maven dependency and get it from the classpath. /cc @aledsage @ygy 

In any case the failure is unrelated.

```
2016-02-03 17:01:57,763 INFO  TESTNG INVOKING: "Surefire test" - org.apache.brooklyn.rest.jsgui.BrooklynJavascriptGuiLauncherTest.testJavascriptWithoutRest()
2016-02-03 17:01:57,768 WARN  Web application not found ../../brooklyn-ui/src/main/webapp
2016-02-03 17:01:57,774 WARN  Failed startup of context o.e.j.w.WebAppContext@6009cdc6{/,null,null}{../../brooklyn-ui/src/main/webapp}
java.io.FileNotFoundException: ../../brooklyn-ui/src/main/webapp
	at org.eclipse.jetty.webapp.WebInfConfiguration.unpack(WebInfConfiguration.java:493) ~[jetty-webapp-9.2.13.v20150730.jar:9.2.13.v20150730]
2016-02-03 17:01:57,775 INFO  JS GUI server started (no REST) at  http://localhost:8080/
```

ahgittin:
@geomacy - i think this should go in `brooklyn-library` instead of `server` as it is optional.  `chef` and `winrm` should probably also be moved /cc @aledsage @hzbarcea .

probably also use package `o.a.b.entity.cm.salt` (add `entity`) wdyt?

otherwise - looks good at a high level - a couple minor comments i've made on lines of code. have **not** yet tested it though!

grkvlt:
@ahgittin @aledsage @geomacy @hzbarcea Remember that there is that old SaltStack hackathon code in [brooklyn-library/sandbox/extra](https://github.com/apache/brooklyn-library/tree/master/sandbox/extra/src/main/java/org/apache/brooklyn/entity/salt) so when this gets moved there we should delete the old sandbox code, to prevent confusion.

hzbarcea:
I am testing and reviewing now.
+1 on moving to `brooklyn-library`.
-0 on adding `entity`, don't see much value there.
+1 @grkvlt on removing the sandbox piece, obviously.

geomacy:
If it helps for testing I have pushed the files I used to create a mysql formula and pillar data to a repository at https://github.com/geomacy/salt-ent-test.  I tarred these up as

```
$ tar tf my-mysql-formula.tar.gz
my-mysql-formula/
my-mysql-formula/mysql/
my-mysql-formula/mysql/files/
my-mysql-formula/mysql/files/classicmodels.schema
my-mysql-formula/mysql/files/foo.schema
```

and 
```
$ tar tf my-mysql-pillar.tar.gz
mysql/
mysql/init.sls
```

Then simply served them up from localhost with Node http-server.

The classicmodels.schema is from http://www.mysqltutorial.org.

geomacy:
Closing this PR - as per the [comment above](https://github.com/apache/brooklyn-server/pull/4#issuecomment-179360438) I have opened a PR on the brooklyn-library repository instead, see https://github.com/apache/brooklyn-library/pull/1.

geomacy:
Oops hit the wrong button- closing this again.

PR #5, opened 2016-02-03
Brooklyn API utilities

sjcorbett:
Useful methods for working with the Brooklyn API programatically.

Borne out of experiences with `brooklyn-maven-plugin` and `brooklyn-tosca`.


robertgmoss:
@sjcorbett All seems sensible.  Jenkins failure is unrelated.

sjcorbett:
@robertgmoss Thanks. Have verified that tests pass locally.

PR #6, opened 2016-02-05
[BROOKLYN-183] Add karaf features

kcris:


ahgittin:
@kcris can you elaborate on why this is needed?

also it seems wrong that mentions to `brooklyn-library` repo items (e.g. `software-*`) should be in this repo (`brooklyn-server`) which is an upstream dependency.  would they be better in a `brooklyn-dist` file?

ahgittin:
(the jenkins failure is unrelated, intermittent unrelated failing test; it should now be fixed so @kcris a close/reopen should give us a chance at a green run)

ahgittin:
btw when i said close/reopen i meant you can *reopen* the same PR.  no harm opening a new one, just wanted to clarify.

PR #7, opened 2016-02-05
Improve how properties to pass to jclouds are selected

ahgittin:
including b.l.j.jclouds.xxx properties (which jclouds understands) and known multi-word


ahgittin:
triggering rebuild with some test fixes merged

neykov:
Looks good, merging.

PR #9, opened 2016-02-08
[BROOKLYN-183] Fix brooklyn shutdown while running in karaf

kcris:


CMoH:
Tested fine both in karaf and with brooklyn launcher. Merging.

core/src/main/java/org/apache/brooklyn/util/core/osgi/Osgis.java
```diff
@@ -340,10 +340,6 @@ public static void ungetFramework(Framework framework) {
     }
 
 
-    /** Tells if Brooklyn is running in an OSGi environment or not. */
-    public static boolean isBrooklynInsideFramework() {
```
ahgittin: should deprecate for one version rather than remove a `public` method

CMoH:
I think the original was added in 0.9.0 and was not yet released. I'll re-add it if you think it is needed

PR #10, opened 2016-02-08
[BROOKLYN-183] Add karaf features

kcris:


neykov:
Looks good and builds fine, merging.

ahgittin:
as per comment on #6 

> it seems wrong that mentions to brooklyn-library repo items (e.g. software-*) should be in this repo (brooklyn-server) which is an upstream dependency. would they be better in a brooklyn-dist file?

/cc @neykov

neykov:
Thanks @ahgittin, didn't notice the previous PR.
I think this file will eventually be split a feature per module, is this correct @kcris, @CMoH?

Also worth discussing on the mailing list what are the benefits of maintaining it manually vs having it generated automatically from the maven dependencies.

CMoH:
I started the `brooklyn-features` karaf repo using the module structure of brooklyn's build system, but I believe they will drift as we progress. In my mind the milestone for evolving features further is the when brooklyn runs in karaf.

Maven projects are meant to control the build process and distribution of artifacts during development, while karaf features are about deployment into a karaf container during operation. Besides this, the arguments against basing `features.xml` on maven are as follows:

1. Features need not match maven projects. Some features can rely on multiple jars generated during the build. There might be multiple features based on the same set of jars for various deployments (e.g. brooklyn-minimal, brooklyn-full, brooklyn-library or just the jclouds location etc)

2. Not all maven runtime dependencies can be installed in karaf. Not all need to be installed. Take, for instance, java-validation-api. Or apache-felix. On the other hand, some non-OSGi jars need to be wrapped or matched to servicemix counterparts.

3. The `features.xml` file can do more than handle bundle dependencies and versions: they can write configuration files, or might have conditional sections. This is something I've come across recently by looking at the [karaf standard feature repo](https://github.com/apache/karaf/blob/master/assemblies/features/standard/src/main/feature/feature.xml), and also at [cxf feature repo](https://github.com/apache/cxf/blob/master/osgi/karaf/features/src/main/resources/features.xml)

However, if the `features.xml` file remains a close representation of the current maven build, then generating dependencies automatically would be a good idea.

CMoH:
@ahgittin @neykov I was planning to move the features that refer to `brooklyn-library` modules in a `brooklyn-library-features` module inside the `brooklyn-library` project, and depend on those in the `brooklyn-features` project inside `brooklyn-server`. We could also rename this latter project as `brooklyn-server-features` for consistency.

The karaf distro project might be better suited in `brooklyn-dist`, along with the `brooklyn-itest` project, and refer both `brooklyn-server-features` and `brooklyn-library-features`. I am not sure if we should also have a top-level `brooklyn-features` project, since I have not seen that practice so far.

ahgittin:
a `brooklyn-library-features` makes a lot of sense

on the fence about karaf dist being in `server` or `dist` or even one in each.  i'll defer to someone who knows features better than i do, or until we better understand usage models.

not sure we need a toplevel `brooklyn-features` project

PR #12, opened 2016-02-09
Inject WebAppContexts into BrooklynWebServer

sjcorbett:
`WebAppContexts` for things running alongside `brooklyn.war` can be configured externally. This can be used to configure JSP support in certain contexts.

### DIFF
--- a/launcher/src/main/java/org/apache/brooklyn/launcher/BrooklynWebServer.java
+++ b/launcher/src/main/java/org/apache/brooklyn/launcher/BrooklynWebServer.java
@@ -40,7 +40,6 @@ import javax.servlet.DispatcherType;
 import org.apache.brooklyn.rest.filter.SwaggerFilter;
 import org.eclipse.jetty.server.Connector;
 import org.eclipse.jetty.server.Server;
-import org.eclipse.jetty.server.SessionManager;
 import org.eclipse.jetty.servlet.FilterHolder;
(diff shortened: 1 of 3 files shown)


launcher/src/main/java/org/apache/brooklyn/launcher/BrooklynWebServer.java
```diff
@@ -296,7 +298,12 @@ public BrooklynWebServer setWar(String url) {
     /** specifies a WAR to use at a given context path (only if server not yet started);
      * cf deploy(path, url) */
     public BrooklynWebServer addWar(String path, String warUrl) {
-        wars.put(path, warUrl);
+        contextProviders.put(path, new WebAppContextProvider(path, warUrl));
+        return this;
+    }
+
+    public BrooklynWebServer addWar(WebAppContextProvider contextProvider) {
```
Graeme-Miller: should this be called addWebAppContextProvider?

Graeme-Miller:
Looks good, just one more general comment.
Maybe WebAppContextProvider should be an interface? With what is currently called WebAppContextProvider becoming a concrete implementation of it.

It seems like it could be useful to switch out ContextProviders, which is easier to do if it's an interface.

sjcorbett:
@Graeme-Miller Considered an interface but felt that with the way `BrooklynWebServer` handles additional wars we _always_ want to use `WebAppContextProvider`. If BWS were refactored I think it may become useful.

PR #14, opened 2016-02-09
Better error messages

ahgittin:


locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -1510,7 +1510,7 @@ public Template buildTemplate(ComputeService computeService, ConfigBag config) {
             if (image==null) throw new NullPointerException("Template does not contain an image (templateBuilder.build returned invalid template)");
         } catch (AuthorizationException e) {
             LOG.warn("Error resolving template: not authorized (rethrowing: "+e+")");
```
aledsage: Would prefer that something (e.g. this log.warn message) says what actually failed (i.e. includes the `templateBuilder`, or some other appropriate description). Otherwise it can be very hard to figure out from a customer's logs what failed and why, when there are multiple apps being deployed at the same time.

Agree that removing it from the propagated exception makes sense.

test-framework/src/test/java/org/apache/brooklyn/test/framework/InfrastructureDeploymentTestCaseTest.java
```diff
@@ -184,12 +180,9 @@ public void testNoInfrastructureSpec() {
...
         try {
             app.start(ImmutableList.of(app.newSimulatedLocation()));
-            fail("Should have thrown execption");
+            Asserts.shouldHaveFailedPreviously();
         } catch (Throwable throwable) {
-            Throwable firstInteresting = Exceptions.getFirstInteresting(throwable);
-            assertThat(firstInteresting).isNotNull();
-            assertThat(throwable).isNotNull();
-            assertThat(firstInteresting).isInstanceOf(IllegalArgumentException.class);
+            Asserts.expectedFailure(throwable);
```
aledsage: Can we say anything about the expected exception? This would currently pass if we had a rouge NullPointerException etc.

rest/rest-server/src/main/java/org/apache/brooklyn/rest/resources/ApplicationResource.java
```diff
@@ -266,27 +268,48 @@ protected Response createFromAppSpec(ApplicationSpec applicationSpec) {
     @Override
     public Response createFromYaml(String yaml) {
         // First of all, see if it's a URL
-        URI uri;
+        Preconditions.checkNotNull(yaml, "Blueprint cannot be null");
```
aledsage: Personal preference for "Blueprint must not be null", given that it is null. No strong feelings though. Happy for you to ignore this :-)

rest/rest-server/src/main/java/org/apache/brooklyn/rest/resources/ApplicationResource.java
```diff
@@ -266,27 +268,48 @@ protected Response createFromAppSpec(ApplicationSpec applicationSpec) {
...
 
         log.debug("Creating app from yaml:\n{}", yaml);
-        EntitySpec<? extends Application> spec = createEntitySpecForApplication(yaml);
+        EntitySpec<? extends Application> spec;
+        try {
+            spec = createEntitySpecForApplication(yaml);
+        } catch (Exception e) {
+            Exceptions.propagateIfFatal(e);
+            Throwable root = Throwables.getRootCause(e);
+            if (root instanceof UserFacingException) throw (UserFacingException) root;
```
aledsage: Do we really want to compare just the root cause against `UserFacingException`? For example, a few lines above we create a `UserFacingException` with the cause `e`. I'd have thought we'd just want to look if `Exceptions.firstInterestingException(e) instanceof UserFacingException`, and throw that if it is.

rest/rest-server/src/main/java/org/apache/brooklyn/rest/resources/ApplicationResource.java
```diff
@@ -266,27 +268,48 @@ protected Response createFromAppSpec(ApplicationSpec applicationSpec) {
...
         if (!Entitlements.isEntitled(mgmt().getEntitlementManager(), Entitlements.DEPLOY_APPLICATION, spec)) {
             throw WebResourceUtils.unauthorized("User '%s' is not authorized to start application %s",
                 Entitlements.getEntitlementContext().user(), yaml);
         }
 
-        return launch(yaml, spec);
+        try {
+            return launch(yaml, spec);
+        } catch (Exception e) {
+            throw WebResourceUtils.badRequest(e, "Error launching blueprint");
```
aledsage: Do we want the same logic as above, to look for `UserFacingException`? Or do we not expect that ever from `launch`?

rest/rest-server/src/main/java/org/apache/brooklyn/rest/resources/ApplicationResource.java
```diff
@@ -339,14 +362,20 @@ public Response createPoly(byte[] inputToAutodetectType) {
 
         //TODO infer encoding from request
         String potentialYaml = new String(inputToAutodetectType);
-        EntitySpec<? extends Application> spec = createEntitySpecForApplication(potentialYaml);
+        EntitySpec<? extends Application> spec;
+        try {
+            spec = createEntitySpecForApplication(potentialYaml);
+        } catch (Exception e) {
+            // TODO if not yaml/json - try ZIP, etc
```
aledsage: Add an `Exceptions.propagateIfFatal(e);`?

rest/rest-server/src/main/java/org/apache/brooklyn/rest/util/DefaultExceptionMapper.java
```diff
@@ -92,13 +93,17 @@ public Response toResponse(Throwable throwable1) {
                 LOG.warn("REST call generated exception type "+throwable2.getClass()+" unrecognized in "+getClass()+" (subsequent occurrences will be logged debug only): " + throwable2, throwable2);
             }
         }
+
+        // Before saying server error, look for a user-facing exception anywhere in the hierarchy
```
aledsage: Should this not be `Exceptions.getFirstThrowableOfType(e, UserFacingException.class)` if we're looking for it anywhere in the hierarchy?

utils/common/src/main/java/org/apache/brooklyn/util/exceptions/Exceptions.java
```diff
@@ -46,12 +47,19 @@
 
 public class Exceptions {
 
-    private static final List<Class<? extends Throwable>> BORING_THROWABLE_SUPERTYPES = ImmutableList.<Class<? extends Throwable>>of(
-        ExecutionException.class, InvocationTargetException.class, PropagatedRuntimeException.class, UndeclaredThrowableException.class);
+    private static final List<Class<? extends Throwable>> ALWAYS_BORING_THROWABLE_SUPERTYPES = ImmutableList.<Class<? extends Throwable>>of(
+        ExecutionException.class, InvocationTargetException.class, UndeclaredThrowableException.class);
```
aledsage: I think it's ok so far with where this is being used, but we need to be very careful to not accidentally imply that there was an exception in the current thread, when in fact it was thrown in a different thread. That kind of thing can be extremely misleading and a huge pain when debugging. We really do want to keep ExecutionException to indicate it was an exception from (probably) a different thread when we're logging.

But when it's to truncate a message for the user, then that's absolutely fine.

aledsage: A concrete example of really bad behaviour would be if `task.get()` finds an exception (wrapped in an `ExecutionException`), unwraps it, and then re-throws this "nicer" exception. At that point, we've lost the stacktrace that got us to the code which called `task.get()`, and only have the stacktrace of the other thread. Also, if the other thread's failure was an `InterruptedException` which we then re-throw, then we could end up interrupting our own thread even though we weren't supposed to!

I'm not saying we're doing this anywhere (I hope not!), but we need to be very careful about how we use `isBoring(t)` and `getFirstInteresting(t)` - both for logging and for propagating.

utils/common/src/main/java/org/apache/brooklyn/util/exceptions/RuntimeInterruptedException.java
```diff
@@ -33,18 +33,14 @@
 
     private static final long serialVersionUID = 915050245927866175L;
 
-    public RuntimeInterruptedException(InterruptedException cause) {
+    public RuntimeInterruptedException(Exception cause) {
```
aledsage: Slightly uncomfortable about it taking `Exception`, if we intend that it is only ever constructed with an `InterruptedException` or another `RuntimeInterruptedException`.

aledsage:
Looks good. Only minor comments. Branch has merge conflicts.

@ahgittin will you resolve merge conflicts and look through my comments, and then merge?

ahgittin:
Agree with all comments -- will do.  Thanks.

rest/rest-server/src/main/java/org/apache/brooklyn/rest/resources/ApplicationResource.java
ahgittin: doing

            UserFacingException userFacing = Exceptions.getFirstThrowableOfType(e, UserFacingException.class);
            if (userFacing!=null) {
                log.debug("Throwing "+userFacing+", wrapped in "+e);
                throw userFacing;
            }

`UserFacing` is a stronger assertion of interest than an outer "interesting" (aka "non-boring") in most cases, i think ... not perfect by any stretch but i think the best balance

rest/rest-server/src/main/java/org/apache/brooklyn/rest/resources/ApplicationResource.java
ahgittin: not sure if it's used in launch, unless internally in an application ... FWIW there is some similar `UserFacing` preference in `DefaultExceptionMapper` (which I've updated similarly) -- let's see how this works and tweak it or overhaul it as we discover we need to

utils/common/src/main/java/org/apache/brooklyn/util/exceptions/Exceptions.java
ahgittin: completely agree, have beefed up javadoc significantly

utils/common/src/main/java/org/apache/brooklyn/util/exceptions/RuntimeInterruptedException.java
ahgittin: that was the intention.  i have changed to expose more constructors accepting only those types.

rest/rest-server/src/main/java/org/apache/brooklyn/rest/util/DefaultExceptionMapper.java
ahgittin: yes

ahgittin:
All addressed.  If build passes, let's merge.

ahgittin:
familiar unrelated nondet failure

```
org.apache.brooklyn.AssemblyTest.checkBrooklynCoreFeature

Failing for the past 1 build (Since Unstable#86 )
Took 5.8 sec.
Error Message

gave up waiting for service org.apache.karaf.features.BootFinished
Stacktrace

org.ops4j.pax.swissbox.tracker.ServiceLookupException: gave up waiting for service org.apache.karaf.features.BootFinished
	at org.ops4j.pax.swissbox.tracker.ServiceLookup.getService(ServiceLookup.java:199)
	at org.ops4j.pax.swissbox.tracker.ServiceLookup.getService(ServiceLookup.java:136)
	at org.ops4j.pax.exam.inject.internal.ServiceInjector.injectField(ServiceInjector.java:89)
	at org.ops4j.pax.exam.inject.internal.ServiceInjector.injectDeclaredFields(ServiceInjector.java:69)
	at org.ops4j.pax.exam.inject.internal.ServiceInjector.injectFields(ServiceInjector.java:61)
	at org.ops4j.pax.exam.invoker.junit.internal.ContainerTestRunner.createTest(ContainerTestRunner.java:61)
	at org.junit.runners.BlockJUnit4ClassRunner$1.runReflectiveCall(BlockJUnit4ClassRunner.java:266)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.BlockJUnit4ClassRunner.methodBlock(BlockJUnit4ClassRunner.java:263)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
	at org.ops4j.pax.exam.invoker.junit.internal.ContainerTestRunner.runChild(ContainerTestRunner.java:68)
```

passes locally.  merging.

PR #15, opened 2016-02-09
REST API using CXF as JAX-RS implementation

neykov:
Use CXF for the JAX-RS implementation because it's better suited for use in Karaf.
Provide Jersey alternative, to be removed in next release.

Other changes:
  * move resources into `brooklyn-rest-resources`, leave servlet code in `brooklyn-server` (this is where the bulk of the changes come from)
  * Filter approach for installing the REST API no longer available, servlet is the default
  * `/v1` prefix removed from resource paths and added at the servlet. User URLs remain the same.
  * Resource implementations can't assume the url at which they live, need to use `@Context UriBuilder uriBuilder` to construct urls returned to the user.
  * remove multipart rest call for creating apps (file upload)
  * TODO (also add to release notes)

To build a dist working with the Jersey dependencies use the `-Djax-rs=jersey` command line option with the maven build.


aledsage:
I don't think we need the "REST API optional Jersey compatibility". What is the benefit to keeping that in the short-term?

neykov:
@aledsage I see the transitive dependency change from CXF to Jersey as disruptive due to the incompatibility between the two. If a downstream project depends on jersey in some way then the migration might not be smooth for them.

On the other it brings additional support burden so I'd be happy to drop it. May be keep it only if someone shows interest on the mailin list? It's a nice to have, but not required feature for the release.

neykov:
Ready for review.

CMoH:
Looks good - I suggest we merge this!

ahgittin:
reviewing this now.  merge conflicts were simple to resolve btw, at https://github.com/ahgittin/brooklyn-server/tree/osgi (if review seems fine, i'll merge)

i'm getting errors in eclipse:

* The project was not built since its build path is incomplete. Cannot find the class file for com.sun.jersey.spi.container.ResourceFilterFactory. Fix the build path then try building this project	brooklyn-rest-server		Unknown	Java Problem
* The project was not built since its build path is incomplete. Cannot find the class file for com.sun.jersey.spi.inject.SingletonTypeInjectableProvider. Fix the build path then try building this project	brooklyn-launcher		Unknown	Java Problem
* The type com.sun.jersey.spi.container.ResourceFilterFactory cannot be resolved. It is indirectly referenced from required .class files	HaMasterCheckFilter.java	/brooklyn-rest-server/src/main/java/org/apache/brooklyn/rest/filter	line 75	Java Problem
* The type com.sun.jersey.spi.inject.SingletonTypeInjectableProvider cannot be resolved. It is indirectly referenced from required .class files	BrooklynWebServer.java	/brooklyn-launcher/src/main/java/org/apache/brooklyn/launcher	line 1	Java Problem

i can probably solve them but any pointers appreciated.

so far looks very thorough and good.  having URI injected (rather than assuming it) is a good practice in any case.

(a lot more to do though so do not merge)

ahgittin:
pragmatic question:  as we want to get an 090 release out soon, do you think the benefit of including this is worth the risk as it won't get a huge amount of testing.  i'm inclined to yes but would revise that if there is a large chance of hidden gremlins!

ahgittin:
`com.sun.jersey.spi.container.ResourceFilterFactory` seems a real problem, we are accessing classes where that var is used from contexts where it might not be (i think); easily refactored to avoid it, which i've done.  the other one may be an env issue (ide clean required), checking that now.

neykov:
re build problems - worth a clean rebuild
re release - Doesn't add any new functionality, needs additional work on the OSGi front (context init), so doesn't bring much to the release. Can be skipped (for just in case reasons). But would be nice to have it  merged as soon as the release branch is cut.

ahgittin:
yep i see this leaves jersey as the default so actually very little is changed.  looks pretty sound to me.

fixed some integration tests which were missing `v1`, aside from that looking good so far.

assuming additional checks pass expect i'll merge soon

neykov:
@ahgittin note that by default CXF is used (https://github.com/neykov/brooklyn-server/blob/cxf/parent/pom.xml#L56), did I commit something by mistake?

ahgittin:
aha thanks @neykov - i missed the conditional activation of the jersey classifier.  looks like jersey should kick in only if you builds with `-Djax-rs=jersey` .

neykov:
hmm how did I skip that part from the PR description, will update.

ahgittin:
mvn build is fine, but failures in IDE (which includes integration failures):

```
FAILED: ApplicationResourceIntegrationTest.testListEntities // and others
FAILED: UsageResourceTest.testGetApplicationUsage // and others
FAILED: PolicyResourceTest.testReconfigureConfig
```

and a lot skipped following failure in `ErrorResponseTest.setUp`

investigating these

ahgittin:
actually all the above are fine in the IDE when i run them class by class so it is something in how the TestNG plugin combines classes; possibly it's ignoring the `suiteName` and so interleaving stuff.  ignoring that.

ahgittin:
runs well and easy to work with

unf this breaks our license generation because CXF upstream dep `xml-resolver:xml-resolver:1.2` gives a model error due to unexpected xml `deploymentStatus/status` being set.  i'm investigating; should be able to ignore this error.

ahgittin:
model error resolved.  new licenses all look fine.  merging.

PR #16, opened 2016-02-09
Add entitlements checks for sensor/config lookup

aledsage:


rest/rest-server/src/main/java/org/apache/brooklyn/rest/resources/EntityConfigResource.java
```diff
@@ -56,6 +58,11 @@
...
     public List<EntityConfigSummary> list(final String application, final String entityToken) {
         final Entity entity = brooklyn().getEntity(application, entityToken);
+        if (!Entitlements.isEntitled(mgmt().getEntitlementManager(), Entitlements.SEE_ENTITY, entity)) {
+            throw WebResourceUtils.unauthorized("User '%s' is not authorized to see entity '%s'",
+                    Entitlements.getEntitlementContext().user(), entity);
+        }
+
         // TODO merge with keys which have values
         return Lists.newArrayList(transform(
                 entity.getEntityType().getConfigKeys(),
```
neykov: Skip items which are `!SEE_CONFIG`?

rest/rest-server/src/main/java/org/apache/brooklyn/rest/resources/SensorResource.java
```diff
@@ -140,6 +167,11 @@ public void set(String application, String entityToken, String sensorName, Objec
     @Override
     public void delete(String application, String entityToken, String sensorName) {
         final Entity entity = brooklyn().getEntity(application, entityToken);
+        if (!Entitlements.isEntitled(mgmt().getEntitlementManager(), Entitlements.MODIFY_ENTITY, entity)) {
```
neykov: Would be nice to have the same for `config.set`.

rest/rest-server/src/main/java/org/apache/brooklyn/rest/resources/SensorResource.java
```diff
@@ -54,6 +55,10 @@
     @Override
     public List<SensorSummary> list(final String application, final String entityToken) {
         final Entity entity = brooklyn().getEntity(application, entityToken);
+        if (!Entitlements.isEntitled(mgmt().getEntitlementManager(), Entitlements.SEE_ENTITY, entity)) {
+            throw WebResourceUtils.unauthorized("User '%s' is not authorized to see entity '%s'",
+                    Entitlements.getEntitlementContext().user(), entity);
+        }
 
         return Lists.newArrayList(transform(filter(entity.getEntityType().getSensors(), AttributeSensor.class),
                 new Function<AttributeSensor, SensorSummary>() {
```
neykov: Don't include summary for `!SEE_SENSOR`.

rest/rest-server/src/test/java/org/apache/brooklyn/rest/entitlement/AuthenticateAnyoneSecurityProvider.java
```diff
@@ -0,0 +1,23 @@
+package org.apache.brooklyn.rest.entitlement;
+
+import javax.servlet.http.HttpSession;
+
+import org.apache.brooklyn.rest.security.provider.SecurityProvider;
+
+public class AuthenticateAnyoneSecurityProvider implements SecurityProvider {
```
neykov: Why not use `AnyoneSecurityProvider`?

rest/rest-server/src/main/java/org/apache/brooklyn/rest/resources/EntityConfigResource.java
```diff
@@ -56,6 +58,11 @@
     @Override
     public List<EntityConfigSummary> list(final String application, final String entityToken) {
         final Entity entity = brooklyn().getEntity(application, entityToken);
+        if (!Entitlements.isEntitled(mgmt().getEntitlementManager(), Entitlements.SEE_ENTITY, entity)) {
+            throw WebResourceUtils.unauthorized("User '%s' is not authorized to see entity '%s'",
```
neykov: Use `forbidden` instead.

neykov:
Good to merge after the rat error is addressed and `unauthorized` changed to `forbidden`.

It'd be nice to have consistency between `list` and `batchRead` in skipping the items, but no strong feelings.

rest/rest-server/src/main/java/org/apache/brooklyn/rest/resources/EntityConfigResource.java
ahgittin: we use `401 unauthorized` throughout.  but you are right, and we should fix in another PR -- https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.4.2 says `401 unauthorized` is actually for cases where a user has provided no authentication or wrong credentials.  (it should probably be called `unauthenticated` not `unauthorized`!)

best practise i think is:
* `404 not found` when we really should not be giving any info on the existence of a resource (in this case)
* `403 forbidden` -- when a user is able to see something but not act on it in the manner they are attempting

rest/rest-server/src/main/java/org/apache/brooklyn/rest/resources/SensorResource.java
aledsage: Agreed long-term, but don't think I'll address it here. Adding more types like that would also make it a bit more verbose to configure, unless we do some kind of hierarchy (e.g. would enabling `entity.modify` mean that `config.set` is implicitly enabled, unless it is explicitly overridden).

rest/rest-server/src/test/java/org/apache/brooklyn/rest/entitlement/AuthenticateAnyoneSecurityProvider.java
aledsage: With `AnyoneSecurityProvider`, the user wasn't set. I needed to have authenticate impl as `user != null` before the user was actually set in the entitlements context.

Not sure if we could change `AnyoneSecurityProvider`, as that would presumably require username+password to always be passed in for all the other tests.

aledsage:
@neykov I've incorporated all your changes, and squashed + rebased those into the first commit. I've also included two new commits:

* Changing unauthorized to forbidden
* Adding entitlements for activity streams (and task lookup)

Can you take another look please?

core/src/main/java/org/apache/brooklyn/core/mgmt/entitlement/Entitlements.java
```diff
@@ -155,6 +161,20 @@ public static EntitlementClassesEnum of(EntitlementClass<?> entitlementClass) {
         }
     }
     
+    public static class TaskAndItem<T> extends Pair<Task<?>,T> {
+        public static <TT> TypeToken<TaskAndItem<TT>> typeToken(Class<TT> type) {
+            return new TypeToken<Entitlements.TaskAndItem<TT>>() {
```
neykov: Is the `TT` type actually included in the anonymous class? I'd expect not since the information is not available at compile time, so no use of the `type` argument. Just curious, I see it's the same in similar pairs.

neykov:
Thanks for the updates, merging.

PR #17, opened 2016-02-10
Generate config/effector descriptions for yaml-only blueprints

iyovcheva:


tbouron:
Doing a `/path/to/brooklyn list-objects --yaml <path-to-yaml/bom>` works, I can see my `catalog.bom` YAML blueprint listed into the generated `items.(js|json)` with my configuration keys. However I was expected to see only my YAML. Currently, all entities and catalog items are listed along my YAML blueprint.

I reckon that's logical as by default, the command grab all the jars within the classpath. I tried to specified `--jar` with nothing behind but it's not accepted. I would allow that for the command to become: `brooklyn/list-objects --yaml <path/to/bom> --jars`

WDYT?

iyovcheva:
retest this please

tbouron:
Just did, the output is the same, it contains all entities

iyovcheva:
@tbouron this message makes Jenkins to run tests again :)

ahgittin:
This looks good.  Making a few changes:

* the `type` of a yaml item should be its id not its underlying java type
* `itemDescriptor` generation for yaml items moved to `ItemDescriptors`
* support `yaml` as a list, like `jars`; and allow a user to supply both; and if just yaml is supplies, it does not scan classes; if neither supplied, as before, it looks at `INITIAL_CLASSPATH`
* share more code between yaml and java block
* use `filename.toLowerCase().endsWith(".yaml")` rather than case-sensitive `contains`
* it just builds one catalog, even with multiple items; if user wants multiple catalogs, they should just run this multiple times
* use our `Strings` class to test for null/empty

then i'll merge

PR #18, opened 2016-02-10
Fixes port inference for dynamically added ConfigKeys

robertgmoss:


aledsage:
Jenkins build failure looks unrelated, and looks caused by something wrong in hudson/environment rather than our code: `hudson.remoting.ChannelClosedException: channel is already closed`

software/base/src/main/java/org/apache/brooklyn/entity/software/base/InboundPortsUtils.java
```diff
@@ -75,6 +76,8 @@
         if (portsAutoInfer == null || portsAutoInfer.booleanValue()) { // auto-infer defaults to true if not specified
             Set<ConfigKey<?>> configKeys = Sets.newHashSet(extraConfigKeys);
             configKeys.addAll(entity.getEntityType().getConfigKeys());
+            // Also add dynamically added config keys
+            configKeys.addAll(((EntityInternal)entity).config().getBag().getAllConfigAsConfigKeyMap().keySet());
```
aledsage: I think this works fine (because it's a HashSet).


My slight hesitation was because of the contract of `Set.addAll()`: "Adds all of the elements in the specified collection to this set if they're not already present (optional operation)." I interpret that as an implementation can optionally replace the previous value with the equal new value. Our new values are all of type `ConfigKey<Object>` (but are equal because the name is equal).

Fine to leave as you have it.

aledsage:
Looks good. Would be nice to have a test though. I notice we don't have any test for `InboundPortsUtils` just now. 

@robertgmoss do you want to add a test now, or are you in a rush to have it merged as-is and then a test added later?

robertgmoss:
thanks @aledsage, I've added a test

sjcorbett:
@robertgmoss Thanks. Will merge.

PR #20, opened 2016-02-11
Add policy to stop applications after a period has elapsed

sjcorbett:
Aka "the money saver".

`DurationSinceSensor` records the duration since it as added to an entity.

`StopAfterDurationPolicy` waits for an entity to reach a lifecycle, adds a duration-since sensor to it, waits for the duration to be greater than the configured lifetime, then stops the entity.

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/sensor/DurationSinceSensor.java
+++ b/core/src/main/java/org/apache/brooklyn/core/sensor/DurationSinceSensor.java
@@ -41,13 +41,15 @@ public class DurationSinceSensor extends AddSensor<Duration> {
 
     public static final ConfigKey<Supplier<Long>> EPOCH_SUPPLIER = ConfigKeys.builder(new TypeToken<Supplier<Long>>() {})
             .name("duration.since.epochsupplier")
-            .description("The source of time from which durations are measured. Defaults to System.currentTimeMillis.")
+            .description("The source of time from which durations are measured. Defaults to System.currentTimeMillis when " +
+                    "if no supplier is given or the configured supplier returns null.")
             .defaultValue(CURRENT_TIME_SUPPLIER)
             .build();
 
     public static final ConfigKey<Supplier<Long>> TIME_SUPPLIER = ConfigKeys.builder(new TypeToken<Supplier<Long>>() {})
             .name("duration.since.timesupplier")
-            .description("The source of the current time. Defaults to System.currentTimeMillis.")
+            .description("The source of the current time. Defaults to System.currentTimeMillis if unconfigured or the " +
+                    "supplier returns null.")
             .defaultValue(CURRENT_TIME_SUPPLIER)
             .build();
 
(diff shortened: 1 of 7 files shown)


sjcorbett:
Failure irrelevant: org.apache.brooklyn.AssemblyTest.checkBrooklynCoreFeature

sjcorbett:
Example blueprint:
```yaml
location: localhost
services:
  brooklyn.policies:
  - type: org.apache.brooklyn.policy.action.StopAfterDurationPolicy
    brooklyn.config:
      lifetime: 10s
      pollPeriod: 2s
  brooklyn.children:
```

core/src/main/java/org/apache/brooklyn/core/sensor/DurationSinceSensor.java
```diff
@@ -0,0 +1,108 @@
...
+        timeSupplier = params.get(TIME_SUPPLIER);
+        epochSensor = Sensors.newLongSensor(sensor.getName() + ".epoch");
+    }
+
+    @Override
+    public void apply(final EntityLocal entity) {
+        super.apply(entity);
+
+        if (entity.sensors().get(epochSensor) == null) {
+            entity.sensors().set(epochSensor, epochSupplier.get());
```
Graeme-Miller: what if epochSupplier returns a null value here? Looks like null values are checked for in UpdateTimeSince , but is there any point creating function feed?

core/src/test/java/org/apache/brooklyn/core/sensor/DurationSinceSensorTest.java
```diff
@@ -0,0 +1,73 @@
...
+import org.apache.brooklyn.core.test.entity.TestEntity;
+import org.apache.brooklyn.util.core.config.ConfigBag;
+import org.apache.brooklyn.util.time.Duration;
+import org.testng.annotations.Test;
+
+import com.google.common.base.Supplier;
+import com.google.common.base.Suppliers;
+import com.google.common.collect.ImmutableMap;
+
+public class DurationSinceSensorTest extends BrooklynAppUnitTestSupport {
```
Graeme-Miller: maybe add a test for when the epochSupplier/timeSupplier return null?

policy/src/main/java/org/apache/brooklyn/policy/action/StopAfterDurationPolicy.java
```diff
@@ -0,0 +1,135 @@
...
+            }
+        }
+    }
+
+    private class TimeIsUpListener implements SensorEventListener<Duration> {
+        @Override
+        public void onEvent(SensorEvent<Duration> event) {
+            synchronized (eventLock) {
+                if (!config().get(FIRED_STOP)) {
+                    if (!config().get(LIFETIME).subtract(event.getValue()).isPositive()) {
```
Graeme-Miller: It's a niggle, but is there an "isNegative" method and would this be clearer:
`config().get(LIFETIME).subtract(event.getValue()).isNegative()`

Graeme-Miller:
couple of minor comments but LGTM

core/src/main/java/org/apache/brooklyn/core/sensor/DurationSinceSensor.java
sjcorbett: I think if epochSupplier returns null it should default to using CURRENT_TIME_SUPPLIER. It's surely a programming error. I've updated the description of the config key to indicate this.

sjcorbett:
@Graeme-Miller Thanks for your comments.

PR #21, opened 2016-02-11
Fix/rest api

sjcorbett:


rest/rest-api/src/main/java/org/apache/brooklyn/rest/domain/TaskSummary.java
```diff
@@ -82,8 +82,8 @@ public TaskSummary(
             @JsonProperty("endTimeUtc") Long endTimeUtc,
             @JsonProperty("currentStatus") String currentStatus,
             @JsonProperty("result") Object result,
-            @JsonProperty("error") boolean error,
-            @JsonProperty("cancelled") boolean cancelled,
+            @JsonProperty("isError") boolean error,
```
andreaturli: we were discussing about using isError also as var name?

rest/rest-api/src/main/java/org/apache/brooklyn/rest/domain/TaskSummary.java
```diff
@@ -82,8 +82,8 @@ public TaskSummary(
...
+            @JsonProperty("isCancelled") boolean cancelled,
```
andreaturli: same here?

PR #11, opened 2016-02-09
Initial work to simplify the rest server

andreaturli:
- remove JsonNode from API interfaces
- refactor consistently toStringand hashCode and equals for the domain objects
- remove deprecated code
- simplified pom dependencies
- remove com.codehaus.jackson dependency and promote com.fasterxml.jackson usage
- adjust BrooklynJacksonSerializer to use jackson 2 api
- fix web.xml for rest-api and rest-client
- update feature.xml dependencies, following maven dependency changes

PR #23, opened 2016-02-12
Remove unused dependency

neykov:
The code is already refactored to use fasterxml jackson provider, remove leftover dependency.

Follow up to https://github.com/apache/brooklyn-server/pull/11.


andreaturli:
lgtm, thanks @neykov

PR #24, opened 2016-02-12
Adds equals and hashcode to DslComponent and related classes

nakomis:


camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/methods/BrooklynDslCommon.java
```diff
@@ -231,6 +233,16 @@ public DslFormatString(String pattern, Object ...args) {
         }
 
         @Override
+        public int hashCode() {
+            return Objects.hashCode(pattern, args);
+        }
+
+        @Override
+        public boolean equals(Object obj) {
+            return EqualsBuilder.reflectionEquals(this, obj);
```
sjcorbett: Personally I feel `reflectionEquals` is fragile - if someone adds another field it's going to be automatically incorporated into `equals` but not into `hashCode` and we've broken the contract. Is there any reason to prefer the shortcut over just defining `equals` normally?

nakomis:
PR Comments addressed

camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/methods/BrooklynDslCommon.java
aledsage: @sjcorbett that's not quite true, is it? The contract is just that two equal methods always have the same hashcode: you could implement a terrible `hashCode() { return 1; }` that would obey the contract. Adding additional fields shouldn't cause the contract to be broken.

But agree that implementing a simple non-reflective equals is good, mostly because it's consistent with the rest of Brooklyn.

aledsage:
LGTM - merging.

PR #28, opened 2016-02-15
winrm4j: exclude slf4j transitive dependency

aledsage:
Depends on org.slf4j 1.7.12 (via org.apache.cxf), whereas Brooklyn depends on 1.6.6

### DIFF
--- a/software/winrm/pom.xml
+++ b/software/winrm/pom.xml
@@ -29,7 +29,6 @@
         <relativePath>../../parent/pom.xml</relativePath>
     </parent>
 
-    <groupId>org.apache.brooklyn</groupId>
     <artifactId>brooklyn-software-winrm</artifactId>
 
     <name>Brooklyn WinRM Software Entities</name>
@@ -44,6 +43,26 @@
             <groupId>io.cloudsoft.windows</groupId>
             <artifactId>winrm4j</artifactId>
             <version>${winrm4j.version}</version>
+            <exclusions>
+                 <exclusion>
+                    <groupId>org.slf4j</groupId>
+                    <artifactId>jcl-over-slf4j</artifactId>
+                </exclusion>
+                 <exclusion>
+                    <groupId>org.slf4j</groupId>
+                    <artifactId>slf4j-api</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+        <!--
+            Excluded from winrm4j the wrong jcl-over-slf4j version; include the right one below.
+            Note use of explicit exclude, rather than wildcard, because maven-enforcer-plugin 1.4.1
+            does not support wildcards: https://issues.apache.org/jira/browse/MENFORCER-195
+         -->
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>jcl-over-slf4j</artifactId>
+            <version>${slf4j.version}</version>
         </dependency>
 
         <!-- test -->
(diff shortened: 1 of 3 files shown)


bostko:
Looks good.

API of slf4j i.e LOG.warn, LOG.info etc won't change so the PR looks ok.
However if it is non-obvious API it will be good if there is tests which verifies used signatures in winrm4j and cxf.

software/winrm/pom.xml
```diff
@@ -44,6 +43,12 @@
             <groupId>io.cloudsoft.windows</groupId>
             <artifactId>winrm4j</artifactId>
             <version>${winrm4j.version}</version>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.slf4j</groupId>
+                    <artifactId>*</artifactId>
+                </exclusion>
+            </exclusions>
```
neykov: Shouldn't you include `jcl-over-slf4j` explicitly with the Brooklyn's version after excluding the incorrect version?

aledsage: Very good point! I'll add that.

aledsage: When I add to software/winrm/pom.xml:

    
    <dependency>
        <groupId>org.slf4j</groupId>
        <artifactId>jcl-over-slf4j</artifactId>
        <version>${slf4j.version}</version>
    </dependency>
 
Then the maven-enforcer fails with the error below. Does it not understand exclusions?! Or is it just very very strict?

    [WARNING] 
    Dependency convergence error for org.slf4j:jcl-over-slf4j:1.7.12 paths to dependency are:
    +-org.apache.brooklyn:brooklyn-software-winrm:0.9.0-SNAPSHOT
      +-io.cloudsoft.windows:winrm4j:0.3.1
        +-io.cloudsoft.windows:winrm4j-client:0.3.1
          +-org.apache.cxf:cxf-rt-transports-http-hc:3.1.4
            +-org.slf4j:jcl-over-slf4j:1.7.12
    and
    +-org.apache.brooklyn:brooklyn-software-winrm:0.9.0-SNAPSHOT
      +-org.slf4j:jcl-over-slf4j:1.6.6
    
    [WARNING] Rule 2: org.apache.maven.plugins.enforcer.DependencyConvergence failed with message:
    Failed while enforcing releasability the error(s) are [
    Dependency convergence error for org.slf4j:jcl-over-slf4j:1.7.12 paths to dependency are:
    +-org.apache.brooklyn:brooklyn-software-winrm:0.9.0-SNAPSHOT
      +-io.cloudsoft.windows:winrm4j:0.3.1
        +-io.cloudsoft.windows:winrm4j-client:0.3.1
          +-org.apache.cxf:cxf-rt-transports-http-hc:3.1.4
            +-org.slf4j:jcl-over-slf4j:1.7.12
    and
    +-org.apache.brooklyn:brooklyn-software-winrm:0.9.0-SNAPSHOT
      +-org.slf4j:jcl-over-slf4j:1.6.6
    ]

When I check with `mvn dependency:tree`, it only mentions `org.slf4j:jcl-over-slf4j:jar:1.6.6`.

Any ideas @neykov ?

aledsage:
@neykov I've pushed a fix to add `org.slf4j:jcl-over-slf4j:jar:1.6.6` - can you take a quick look please?

ahgittin:
@aledsage did you forget to push?  i see nothing about `jcl` in the diff

aledsage:
oops, thanks @ahgittin - have pushed now (with `git push -f`, so just the one commit).

neykov:
Failure looks unrelated, good to merge.

PR #30, opened 2016-02-18
Feature/paas cf support

kiuby88:
Please, note that this PR is extended by [brooklyn-library#8](https://github.com/apache/brooklyn-library/pull/8).

For adding PaaS support, a `CloudFoundryLocation` is added which is based on `PaasLocation`.
This new location contains a API client for CloudFoundry Paas Services, (CloudFoundry Java Client), that allows to use the services of this provider.

Extracting ssh behaviour from `SoftwareProcessImpl`: provisioning flags
For adding paas support, `SoftwareProcess` entities should to be agnostics in order to allows to be deployed on SSH and PaaS locations. So, the machine behaviour and management should be decoupled of the entities and it should be moved to an isolate class.

`SoftwareProcessImpl` contained a pair of method for obtaining the needed flags to create and configure a VM. However as it was mentioned previously, this entity shouldn't contain any ssh (machine) behaviour. Then, this code is moved to an external class,  `MachineProvisioningLocationFlagsSupplier` that is managed in `SoftwareProcessImpl` as a `LocationFlagSupplier` interface. Moreover, an equivalent class is added to represent this behaviour (obtain flags location) for PaaS. `FlagsSuppliers` are used to separate some entity's logic to an external class but this object could be avoided and its logic could be moved to `MachineLifecycleEffectorTasks` or `PaasLifecycleEffectorTasks` .

`SoftwareProcessImpl` will select a behaviour for obtaining the flags depending on the selected location for deploying the entity.
Adding `BehaviourFactory` for selecting PaaS or SSH behaviour in run time.

Following the previous commits, `SoftwareProcessImpl` entities should select PaaS or SSH behaviour in run time, depending on the target location. An entity know the target location when it is started (calling to start effector), then just when this effector is called the entity will able to know the selected location, so in this moment the entity will able to find the required behaviour, SSH or PaaS.
The whole behaviour is composed by several classes, then `SoftwareProcessImpl` uses a factory (BehaviourFactory) which is selected according to the target location. This factory allows to select in run time all behavior focus in a location.

A new hierarchy of drivers is added to Brooklyn, these new drivers is oriented to PaaS but they follow the hierarchy defined for SSH. For example, `AbstractSoftwareProcessDriver`, `AbstractSoftwareProcessSshDriver` allow to `SoftwareProcess` entities use drivers for location based on  machines. Then, to enable the PaaS support for `SoftwareProcess` entities a set of drivers is added such as `AbstractSoftwareProcessCloudFoundryDriver` and `AbstractApplicationCloudFoundryDriver`.
Fixing dependency convergence.

### DIFF
--- a/locations/cloudfoundry/pom.xml
+++ b/locations/cloudfoundry/pom.xml
@@ -37,7 +37,7 @@
 
     <properties>
         <cloudfoundry.client.version>1.1.2</cloudfoundry.client.version>
-        <springframework-webmvn.version>4.0.5.RELEASE</springframework-webmvn.version>
+        <springframework-webmvc.version>4.1.4.RELEASE</springframework-webmvc.version>
         <springframework-security.version>2.0.4.RELEASE</springframework-security.version>
     </properties>
 
@@ -78,12 +78,16 @@
                     <groupId>org.springframework.security.oauth</groupId>
                     <artifactId>spring-security-oauth2</artifactId>
                 </exclusion>
+                <exclusion>
+                    <groupId>org.springframework</groupId>
+                    <artifactId>spring-webmvc</artifactId>
+                </exclusion>
             </exclusions>
         </dependency>
         <dependency>
             <groupId>org.springframework</groupId>
             <artifactId>spring-webmvc</artifactId>
-            <version>${springframework-webmvn.version}</version>
+            <version>${springframework-webmvc.version}</version>
             <exclusions>
                 <exclusion>
                     <groupId>commons-logging</groupId>
--- a/locations/cloudfoundry/pom.xml
+++ b/locations/cloudfoundry/pom.xml
@@ -1,0 +1,164 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+    Licensed to the Apache Software Foundation (ASF) under one
+    or more contributor license agreements.  See the NOTICE file
+    distributed with this work for additional information
+    regarding copyright ownership.  The ASF licenses this file
+    to you under the Apache License, Version 2.0 (the
+    "License"); you may not use this file except in compliance
+    with the License.  You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing,
+    software distributed under the License is distributed on an
+    "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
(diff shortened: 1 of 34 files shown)


locations/cloudfoundry/pom.xml
```diff
@@ -0,0 +1,168 @@
...
+            <artifactId>guava</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-core</artifactId>
+            <version>${fasterxml.jackson.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.cloudfoundry</groupId>
+            <artifactId>cloudfoundry-client-lib</artifactId>
```
robertgmoss: Can this PR be based on the newer [CF Java client](https://github.com/cloudfoundry/cf-java-client)?

kiuby88: Sure, let me update

ahgittin:
As discussed at https://github.com/apache/brooklyn-library/pull/8 I'm not convinced the driver is the most natural place for paases to kick in.  Having a `PaasLocation` feels absolutely right but the more we can abstract from software the better.

I'd also prefer to see this entirely in `brooklyn-library`, e.g. `/brooklyn-library/systems/cloudfoundry/` containing the `CloudFoundryLocation` and the `CloudFoundryXxxApp` in suitable packages e.g. `org.apache.brooklyn.cloudfoundry.{location,entities}`.  When there is another PaaS then it might make sense to introduce a `PaasLocationConfig` as a core piece in `brooklyn-server` -- but that can be done without breaking compatibility for people who start with `CloudFoundryLocationConfig`.

The subject of "smart entities" -- which inspect the location given (e.g. CF or VM) and choose an appropriate entity -- is one we should discuss separately and on the mailing list once we have the basics working.  I can think of a few ways to do that, but trying to do it with the driver feels too brittle (though clever!).

andreaturli:
@kiuby88 this is a long lasting PR which I don't think can be merged easily now.

I also know this work has been very likely superseded by https://github.com/kiuby88/brooklyn-cloudfoundry - shall you close this PR and open a new one to donate brooklyn-cloudfoundry to `apache/brooklyn-cloudfoundry`? Thanks!

kiuby88:
@andreaturli I agree with you, it should be closed.

About donating brooklyn-cloudfoundry, I TOTALY agreed. If you opened the PR I would open the PR immediately

Thanks a lot

PR #31, opened 2016-02-18
[OSGi] JAAS LoginModule delegating to Brooklyn's SecurityProviders 

neykov:
Depends on #15.

An implementation of a JAAS LoginModule, delegating to the SecurityProvider configured in brooklyn.properties. Used as the authentication mechanism for the REST API when running in Karaf.
Currently configured as a separate realm, used only for web.

### DIFF
--- a/karaf/itest/pom.xml
+++ b/karaf/itest/pom.xml
@@ -30,6 +30,11 @@
         <relativePath>../pom.xml</relativePath>
     </parent>
 
+    <properties>
+        <includedTestGroups />
+        <excludedTestGroups>org.apache.brooklyn.test.IntegrationTest</excludedTestGroups>
+    </properties>
+
     <dependencies>
         <!-- Pax Exam Dependencies -->
         <dependency>
@@ -155,6 +160,15 @@
 
         <!-- project deps -->
         <dependency>
+          <groupId>${project.groupId}</groupId>
+          <artifactId>brooklyn-features</artifactId>
+          <version>${project.version}</version>
+          <type>xml</type>
+          <classifier>features</classifier>
+          <scope>test</scope>
+        </dependency>
+
+        <dependency>
             <groupId>${project.groupId}</groupId>
             <artifactId>apache-brooklyn</artifactId>
             <version>${project.version}</version>
@@ -171,6 +185,12 @@
             <groupId>${project.groupId}</groupId>
             <artifactId>brooklyn-rest-resources</artifactId>
             <version>${project.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>brooklyn-rest-resources</artifactId>
+            <version>${project.version}</version>
             <classifier>tests</classifier>
             <scope>test</scope>
         </dependency>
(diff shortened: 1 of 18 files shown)


CMoH:
Looks good to merge (after conflict resolution). A few comments though, all of them addressable either here on in another PR - up to you.

- `AssemblyTest.configuration()` can be made to use `KarafUtils.DEFAULT_OPTIONS`
- remove `keepRuntimeFolder()` from `KarafUtils.DEFAULT_OPTIONS` (the target folder keeps getting bigger with this, which might be offensive to some - but it is useful for pin-pointed debugging)
- not sure if the jaas realm name should be `webconsole`; I think `brooklyn` might be a better choice (see realm-list below why)
- would be interesting to have the jaas backend implemented, so that the session below works; but that's a point of discussion outside the scope of this PR

```
karaf@root()> jaas:realm-list 
Index | Realm Name | Login Module Class Name                                                                                                                                                                                                 
-----------------------------------------------------------------------------------                                                                                                                                                          
1     | karaf      | org.apache.karaf.jaas.modules.properties.PropertiesLoginModule                                                                                                                                                          
2     | karaf      | org.apache.karaf.jaas.modules.publickey.PublickeyLoginModule                                                                                                                                                            
3     | karaf      | org.apache.karaf.jaas.modules.audit.FileAuditLoginModule                                                                                                                                                                
4     | karaf      | org.apache.karaf.jaas.modules.audit.EventAdminAuditLoginModule                                                                                                                                                          
5     | webconsole | org.apache.brooklyn.rest.security.jaas.BrooklynLoginModule                                                                                                                                                              
karaf@root()> jaas:realm-manage --index 5
karaf@root()> jaas:group-list 
Can't get the list of users (no backing engine service registered)
karaf@root()> jaas:user-list 
Can't get the list of users (no backing engine service found)
karaf@root()> jaas:cancel 
karaf@root()> jaas:realm-manage --index 1
karaf@root()> jaas:user-list 
User Name | Group      | Role
--------------------------------------
karaf     | admingroup | admin
karaf     | admingroup | manager
karaf     | admingroup | viewer
karaf     | admingroup | systembundles
karaf@root()> jaas:cancel 
karaf@root()> jaas:realm-manage --index 2
karaf@root()> jaas:user-list 
User Name | Group | Role
------------------------
karaf@root()> 
```

ahgittin:
@neykov bump

neykov:
Thanks @CMoH. Addressed pts 1 & 2.
The main purpose of the `BrooklynLoginModule` is to bring Brooklyn's `SecurityProvider` implementations to the Karaf world and allow for easier migration. Long term I'd expect that we drop it in favour of re-implementing the existing `SecurityProviders` "the Karaf way". As a wrapper of `SecurityProvider`s it has some limitations:
  * fixed and single role
  * no access to the backing store of the `SecurityProvider` implementation

Because of the first I'd be wary of using it in non-web context - the REST API supports entitlements which can limit what a user can do inside brooklyn, but if the same user is allowed to ssh into Karaf's console without any limitations then any other security measures are meaningless.
The second makes it impossible to implement the backend, which would allow for user management inside Karaf's console.

neykov:
Rebased & solved conflicts.

neykov:
Build is successful, jenkins didn't update status.

PR #32, opened 2016-02-20
Pick up root wars when passed to BrooklynLauncher

neykov:
wars passed to BrooklynLauncher get their context path stripped of the leading "/", resulting in empty string for root apps. The web server expects a "/" context path for a root war so it uses the default war instead.

/cc @sjcorbett


sjcorbett:
@neykov Thanks.

PR #34, opened 2016-02-23
Initialization of brooklyn within karaf

CMoH:
Initializes brooklyn using a serial initialization sequence in a new class `org.apache.brooklyn.core.OsgiLauncher` in `brooklyn-core`. The purpose of this PR is to stay close to the existing `BrooklynLauncher` initialization code, while discussing the location of various settings within karaf's etc folder. Further work should move individual bean initialization into corresponding  `init-method`s of each bean specified in `OSGI-INF/blueprint/blueprint.xml`


CMoH:
I have been unable to determine why I had to drop a test in 05518e7; from my investigation the test setup does not call `setManagementContext` on the `ServerResource`, but I was not able to find a solution so far. Can anybody help?

CMoH:
Fixed failing test. Please review and merge.

core/src/main/java/org/apache/brooklyn/core/OsgiLauncher.java
```diff
@@ -0,0 +1,557 @@
...
+//
+//            computeAndSetApp(launcher, utils, loader);
+//            customize(launcher);
+//        } catch (FatalConfigurationRuntimeException e) {
+//            throw e;
+//        } catch (Exception e) {
+//            throw new FatalConfigurationRuntimeException("Fatal error configuring Brooklyn launch: " + e.getMessage(), e);
+//        }
+//
+//        try {
```
neykov: Lots of commented out code in this file. Can you remove the parts which are not needed? For the rest perhaps leave a comment how should be solved or why can't be used directly.

core/src/main/java/org/apache/brooklyn/core/OsgiLauncher.java
```diff
@@ -0,0 +1,557 @@
...
+            // for other exceptions we log it, possibly redundantly but better too much than too little
+            Exceptions.propagateIfFatal(e);
+            log.error("Error launching brooklyn: " + Exceptions.collapseText(e), e);
+            try {
+                shutdown();
+            } catch (Exception e2) {
+                log.warn("Subsequent error during termination: " + e2);
+                log.debug("Details of subsequent error during termination: " + e2, e2);
+            }
+            Exceptions.propagate(e);
```
neykov: How would an exception in this code manifest in the karaf dist? Does it just prevent the bundle from loading, leading to failures in all dependent bundles?

core/src/main/java/org/apache/brooklyn/core/OsgiLauncher.java
```diff
@@ -0,0 +1,557 @@
...
+//            } catch (Exception e) {
+//                handleSubsystemStartupError(ignoreWebErrors, "core web apps", e);
+//            }
+//        }
+        // not needed: camp is a separate karaf feature - let it initialize itself
+//        // Add a CAMP platform
+//        campPlatform = new BrooklynCampPlatformLauncherNoServer()
+//                .useManagementContext(managementContext)
+//                .launch()
+//                .getCampPlatform();
```
neykov: The CAMP platform is injecting itself in the ManagementContext on launch. Leaving this out will prevent us from using yaml blueprints. Do you think we should initialize it here, before there's a better OSGi alternative?

core/src/main/java/org/apache/brooklyn/core/OsgiLauncher.java
```diff
@@ -0,0 +1,557 @@
...
+                    log.warn("Loading catalog for " + state + " as part of launch sequence (it was not loaded as part of the rebind sequence)");
+                    catInit.populateCatalog(state, true, true, null);
+                }
+            }
+        } catch (Exception e) {
+            handleSubsystemStartupError(ignoreCatalogErrors, "initial catalog", e);
+        }
+        catInit.setStartingUp(false);
+
+        // no need for command-line locations
```
neykov: As a general question, what do you think about supporting command line arguments? Is it something that's not recommended in karaf world? How do other apps handle it?

core/src/main/java/org/apache/brooklyn/core/OsgiLauncher.java
```diff
@@ -0,0 +1,557 @@
...
+//        if (webServer != null) {
+//            try {
+//                webServer.stop();
+//            } catch (Exception e) {
+//                LOG.warn("Error stopping web-server; continuing with termination", e);
+//            }
+//        }
+
+        // TODO Do we want to do this as part of managementContext.terminate, so after other threads are terminated etc?
+        // Otherwise the app can change between this persist and the terminate.
```
neykov: Yes. Needs to be executed as part of the shut down sequence.
On a related note, any recommendation on how to go about supporting the `shutdownHandler` in OSGi?

karaf/features/src/main/feature/feature.xml.orig
```diff
@@ -0,0 +1,288 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
neykov: Probably committed by mistake, remove.

karaf/itest/src/test/java/org/apache/brooklyn/rest/BrooklynRestApiLauncherTest.java
```diff
@@ -0,0 +1,98 @@
...
+import static org.ops4j.pax.exam.karaf.options.KarafDistributionOption.features;
+import static org.ops4j.pax.exam.karaf.options.KarafDistributionOption.karafDistributionConfiguration;
+import static org.ops4j.pax.exam.karaf.options.KarafDistributionOption.logLevel;
+import org.ops4j.pax.exam.karaf.options.LogLevelOption;
+import org.ops4j.pax.exam.spi.reactors.ExamReactorStrategy;
+import org.ops4j.pax.exam.spi.reactors.PerClass;
+
+@RunWith(PaxExam.class)
+@ExamReactorStrategy(PerClass.class)
+@Ignore // TODO: re-enable after brooklyn is properly initialized within the OSGI environment
```
neykov: Can you list what's missing for this to work?

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/resources/AbstractBrooklynRestResource.java
```diff
@@ -49,12 +50,18 @@
 
     
     private ManagementContext managementContext;
+    private ManagementContextInternal managementContextInternal;
```
neykov: After discussing this change still don't see the reason for having both properties. If the resources require `ManagementContextInternal`, then while not change the type of `managementContext` and inject the appropriate type? Or leave the type as is and keep the castings.

rest/rest-resources/src/main/resources/OSGI-INF/blueprint/service.xml
```diff
@@ -29,58 +29,78 @@ limitations under the License.
...
     </cxf:bus>
 
     <reference id="localManagementContext"
                interface="org.apache.brooklyn.api.mgmt.ManagementContext" />
+    <reference id="localManagementContextInternal"
+               interface="org.apache.brooklyn.core.mgmt.internal.ManagementContextInternal" />
 
     <bean id="accessResourceBean" class="org.apache.brooklyn.rest.resources.AccessResource">
         <property name="managementContext" ref="localManagementContext" />
+        <property name="managementContextInternal" ref="localManagementContextInternal" />
```
neykov: Following the previous comment, why not inject `localManagementContextInternal` in the `managementContext` property?

rest/rest-server/src/main/java/org/apache/brooklyn/rest/RestApiSetup.java
```diff
@@ -54,6 +63,21 @@ public static void installRest(ServletContextHandler context, Object... provider
...
+
+    private static ManagementContext extractManagementContext(Object... providers) {
+        for (Object o : providers) {
+            if (o instanceof ManagementContextProvider) {
+                return ((ManagementContextProvider) o).getContext(ManagementContext.class);
+            }
+        }
+        return null;
+    }
+
```
neykov: This will break when combined with #41 because `ManagementContextProvider` no longer holds a reference to the `ManagementContext` in the classic launcher case. Instead it fetches it from `ServletContext`.
This change is needed because of the additional `managementContextInternal` change. Instead I suggest letting `ManagementContextProvider` handle the injection in resources (i.e. adding @Context .... to the property). This way we will have a unified code path in both classical launcher and OSGi.

neykov:
Replying to @CMoH's comment above:

Two points to tackle here - `ManagementContextHolder` and rest resource initialization
* `ManagementContextHolder` is meant to be used by `BrooklynLoginModule` exclusively That's the approach used by Karaf for passing `BundleContext` to its `ProxyLoginModule`. You've given me a great suggestion of looking for the `ManagementContext` service directly, but for the time being `BrooklynLoginModule` will be used by the classic brooklyn as well, so this is something that could be done once we switch to karaf completely.

* Agree that `ManagementContext` handling in REST needs a clean up.
  * All current uses of `OsgiCompat` can be replaced by alternatives. Classic-only code like servlet filters should look directly into the `ServletContext`. For JAX-RS code see below.
  * For consistency would be nice if all JAX-RS code (resources, filters, interceptors, providers) depends on the `ManagementContextProvider`. 
  * `ManagementContextProvider` could be used to abstract the environment - classical vs OSGi, the first one injecting through `ServletContext`, the other through aries blueprint injection. One provider could handle both cases or have a different one for each.
  * If there are tests still injecting the `ManagementContext` I'd consider this oversight, something that I've missed to migrate to one of the above aproaches.

neykov:
Finished review. Looks good. The web console is happy with the state.
Some points:
  * The catalog is empty on startup - is this something that needs to be added to init or an OSGi limitation preventing us from initializing it atm?
  * Can you share your view on next steps. What absolutely needs to be implemented, nice to have changes, anything on your mind.

CMoH:
@neykov Thanks! I'll try to reply in general, and next to line comments.

First, there were some aspects of `BrooklynLauncher`'s initialization code that were unclear to me. I've let them commented out in the `OsgiLauncher` for the purpose of further discussion. I think your line notes (and the other commented sections) can act as a checklist on what's still to be done to round the osgi initalization up.

Second, the catalog is indeed empty on startup, at least for now. I have started work on this, but there are some issues to clarify on how to leverage OSGi services in populating the catalog with existing entities, policies and the rest of the Java-based catalog items. The catalog items represented by blueprints is the other side of the catalog, and I think this won't change at all.

So, for next steps I think this empty catalog issue needs to be tackled. Finishing up the rest of the initialization code is also on the board. There is a particular issue with guava versions used in brooklyn and jclouds, which makes `brooklyn-location-jclouds` unusable in OSGi at the moment.

From my limited perspective, after these three steps we'll have a demo-grade brooklyn running in karaf.

core/src/main/java/org/apache/brooklyn/core/OsgiLauncher.java
CMoH: I am unsure in which direction brooklyn will evolve within karaf, so I left all these commented out for review. I'll refer to each aspect of the code commented above here:

- `initBrooklynProperties()` - right now it is not on the table, since we're trying to preserve the existing functionality within the normal launcher, but `brooklyn.properties` should disappear in favor of karaf's `ConfigAdmin` interface. By default this looks up `cfg` files in karaf's `etc` folder, but these are actually normal property files. Also, karaf can automatically inject properties from these `cfg` files in various java beans and notify when the files change on disk. Within karaf cellar we get these changes distributed on all nodes within a cluster group. Furthermore, going this path allows multiple instances of karaf on the same machine, easing development and management.
- *stopping apps on shutdown* does not really makes sense in karaf, where `brooklyn-core` can be stopped or started with various purposes (upgrades, cellar nodes being added or removed from the cluster etc). We might envision a concept of stopping all existing apps independently of brooklyn's lifecycle. Would that be acceptable?
- the same applies to `computeAndSetApp` from the command-line. One can envision karaf shell commands that take care of this. Or we could try and preserve the existing command-line functions, but these will be awkward to match the bundle lifecycle in the container (when do we really say it "started"? what action do we interpret as "shutdown"?)
- `customize(launcher)` - I think this applied command-line arguments to the launcher, so it ties to the above discussion. On the other hand, any change to karaf's `etc` folder will reload the configuration, so any customization is envisionable through this mechanism. Also, karaf shell commands can be used to change behavior as well.

I hope you will review the ideas above and help decide what to do in the short term and the long term.

core/src/main/java/org/apache/brooklyn/core/OsgiLauncher.java
CMoH: Exceptions should propagate to the blueprint container and prevent it from instantiating the corresponding bean. As a result the bundle should transition in the `Failed` state instead of `Started`. I think bundles depending on services of a failed bundle will enter a 5 minute `GracePeriod` state, and next fail themselves, but it might be they don't get to start at all - not really sure.

core/src/main/java/org/apache/brooklyn/core/OsgiLauncher.java
CMoH: Well, the camp bundles could simply bind to the management context service provided by `brooklyn-core` and initialize themselves. This would enable separate deployment of the `brooklyn-core` and `brooklyn-camp` features. If we initialize camp here we need to tie them closer, so basically `brooklyn-core` won't work without the `camp` feature on the classpath (instead of being decoupled through microservice calls).

@neykov I don't know which is best. What is your take on this?

core/src/main/java/org/apache/brooklyn/core/OsgiLauncher.java
CMoH: Personally I would not use command-line arguments within karaf, so I have not researched this. But I think it can be technically done if so wanted.

Here is why I want to stay away from command-line args:
- bundle lifecycle is different from karaf lifecycle. How do we handle command-line arguments if we restart `brooklyn-core`? Do we assume the previous life of `brookyn-core` handled those and we don't act on them? Or should they be re-applied? If so, we need to store this somehwere, so why not set our wishes in the config files from the beginning?
- how do we handle command-line arguments on different karaf cellar nodes?
- karaf has a command-line shell where we can place commands for users to play with. I think karaf defines a startup script (i.e. like bashrc) and this could be a place to issue startup commands

core/src/main/java/org/apache/brooklyn/core/OsgiLauncher.java
CMoH: Answered in a comment above. The first thing to define is what "shutdown" means in the OSGi context. My opinion is not to consider `brooklyn-core` bundle deactivation as shutdown, since that happens when we want to upgrade it, or might happen when manipulating karaf cellar nodes.

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/resources/AbstractBrooklynRestResource.java
CMoH: Karaf injects proxies to services, not the service objects themselves. This loses multiple inheritance of interfaces that act as services.

As a result, even if the bean from `brooklyn-core` implements both  `org.apache.brooklyn.api.mgmt.ManagementContext` and `org.apache.brooklyn.core.mgmt.internal.ManagementContextInternal`, the proxies for each service will only implement their respective interface, and cannot be cast.

Also the fact that two services are implemented by the same object is not the concern of the client of those services.

rest/rest-server/src/main/java/org/apache/brooklyn/rest/RestApiSetup.java
CMoH: I suggest we apply this PR after #41, so I can look at the failing test and see how to make it work with the `ManagementContextProvider`

core/src/main/java/org/apache/brooklyn/core/OsgiLauncher.java
neykov: * `initBrooklynProperties()` - agree long term, but it's a lot of work to have everything the "karaf way" from the beginning. I'd prefer if we split it in two stages: 1. have Brooklyn usable in Karaf, 2. Make Brooklyn a good OSGi citizen. This will allow us to work on gradually improving the code base instead of all or nothing approach. 
* Regular users won't be interacting directly with karaf, instead through the cli and webconsole. They will need the process started and shut down at some point. We need a way to shut the process down and optionally terminate all apps as part of that so we clean up cloud resources nicely. This is not so much about backwards compatibility but as having the feature available - don't think the launcher is the place to support this in karaf
* `computeAndSetApp` as part of the cli can be TBD later, not needed right away (but good to have a comment that it is missing)
* `customize(launcher)` - downstream projects have their own `Main` + `Launcher` implementations. This is a generic callback to custom code to do any additional initializations specific to the project. Would be nice to have an infrastructure for downstream projects to influence initialization, but not needed immediately.

To summarize:
  * CLI - needs additional discussion how it maps to the Karaf world, not needed immediately. Personally I believe that it would be nice to support Brooklyn's CLI in headless karaf.
  * `brooklyn.properties` - it would be easy to support most of it so I think we should support it as a transition to Karaf config.
  * next steps would be making sure we can deploy BasicEntity
  * deploy MachineEntity using a jclouds location (by using MachineEntity through a catalog item + explicit libraries bundle pointing at `brooklyn-software-base`)
  * catalog initialization (as in loading entities through reflection) I see as a thing that should be addressed later, i.e. non-essential. If we can have a catalog loaded from `default.catalog.bom`  would be nice, depending on the amount of work involved.

core/src/main/java/org/apache/brooklyn/core/OsgiLauncher.java
neykov: I like that - let the camp bundle execute the above code. CAMP support is essential as Brooklyn is not usable without it.

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/resources/AbstractBrooklynRestResource.java
neykov: `ManagementContextInternal` implements `ManagementContext` so it can be used at any place where `ManagementContext` is needed. And the code assumes that `ManagementContext` and `ManagementContextInternal` are backed by the same instance. So still think that having a single service and a single point to inject it is the way to go.

karaf/itest/src/test/java/org/apache/brooklyn/rest/BrooklynRestApiLauncherTest.java
CMoH: I've re-enabled the test after fixing the features for winrm4j, and here is the problem:

```
Tests run: 1, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 10.586 sec <<< FAILURE! - in org.apache.brooklyn.rest.BrooklynRestApiLauncherTest
testStart(org.apache.brooklyn.rest.BrooklynRestApiLauncherTest)  Time elapsed: 1 sec  <<< FAILURE!
java.lang.AssertionError: URL http://localhost:9998/v1/catalog/entities does not contain text: BrooklynNode
        at org.apache.brooklyn.test.Asserts.fail(Asserts.java:721)
        at org.apache.brooklyn.util.http.HttpAsserts.assertContentContainsText(HttpAsserts.java:191)
        at org.apache.brooklyn.rest.BrooklynRestApiLauncherTest.testStart(BrooklynRestApiLauncherTest.java:89)
```

This is due to the catalog being empty, which is expected at this stage.

neykov:
Merging, any TODO can be addressed in separate PRs.

geomacy:
Notes on key points still to address:

* Decide how to access management context from various places
* Read brooklyn.properties
* Make MachineEntity and BasicEntity deployable as a first step in getting the Catalog populated from OSGI
* CAMP - initialise CAMP platform in the launcher
* Define what shutdown means in the context of Karaf

PR #36, opened 2016-02-23
resolves static sensor value as a task, allowing DSLComponents to be set

robertgmoss:


ahgittin:
good fix.  consider in future doing TDD where you write a failing test first -- that way we'd have a test for this :)

PR #37, opened 2016-02-24
Added start latch to BasicStartable and StartableApplication

grkvlt:
Allows groups of components and applications to have their startup sequence controlled more easily. This includes only setting `service.isUp` to `true` on an `AbstractApplication` once it, and its children, have started. This is the same semantics as other entities, and allows a more consistent blueprint style.

Also reverts https://github.com/apache/brooklyn-server/commit/d7c3759a150f409d1674fdaf74117404d56a665d which prevents Clocker from re-using images.

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/entity/AbstractApplication.java
+++ b/core/src/main/java/org/apache/brooklyn/core/entity/AbstractApplication.java
@@ -21,6 +21,11 @@ package org.apache.brooklyn.core.entity;
 import java.util.Collection;
 import java.util.Map;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.common.collect.ImmutableSet;
+
 import org.apache.brooklyn.api.entity.Application;
 import org.apache.brooklyn.api.entity.Entity;
 import org.apache.brooklyn.api.location.Location;
@@ -35,10 +40,6 @@ import org.apache.brooklyn.util.exceptions.Exceptions;
 import org.apache.brooklyn.util.exceptions.RuntimeInterruptedException;
 import org.apache.brooklyn.util.text.Strings;
 import org.apache.brooklyn.util.time.Time;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
-import com.google.common.collect.ImmutableSet;
 
 /**
  * Users can extend this to define the entities in their application, and the relationships between
@@ -132,12 +133,12 @@ public abstract class AbstractApplication extends AbstractEntity implements Star
     @Override
     protected void initEnrichers() {
         super.initEnrichers();
-        
+
         // default app logic; easily overridable by adding a different enricher with the same tag
         ServiceStateLogic.newEnricherFromChildren().checkChildrenAndMembers().addTo(this);
         ServiceStateLogic.ServiceNotUpLogic.updateNotUpIndicator(this, Attributes.SERVICE_STATE_ACTUAL, "Application created but not yet started, at "+Time.makeDateString());
     }
-    
+
     /**
      * Default start will start all Startable children (child.start(Collection<? extends Location>)),
      * calling preStart(locations) first and postStart(locations) afterwards.
@@ -152,15 +153,19 @@ public abstract class AbstractApplication extends AbstractEntity implements Star
         setExpectedStateAndRecordLifecycleEvent(Lifecycle.STARTING);
         try {
             preStart(locationsToUse);
-            // if there are other items which should block service_up, they should be done in preStart
-            ServiceStateLogic.ServiceNotUpLogic.clearNotUpIndicator(this, Attributes.SERVICE_STATE_ACTUAL);
-            
+
+            // Opportunity to block startup until other dependent components are available
+            Object val = config().get(START_LATCH);
+            if (val != null) log.debug("{} finished waiting for start-latch; continuing...", this);
+
             doStart(locationsToUse);
             postStart(locationsToUse);
+
+            ServiceStateLogic.ServiceNotUpLogic.clearNotUpIndicator(this, Attributes.SERVICE_STATE_ACTUAL);
         } catch (Exception e) {
             // TODO should probably remember these problems then clear?  if so, do it here ... or on all effectors?
-//            ServiceProblemsLogic.updateProblemsIndicator(this, START, e);
-            
+            // ServiceProblemsLogic.updateProblemsIndicator(this, START, e);
+
             recordApplicationEvent(Lifecycle.ON_FIRE);
             // no need to log here; the effector invocation should do that
             throw Exceptions.propagate(e);
--- a/core/src/main/java/org/apache/brooklyn/entity/stock/BasicStartableImpl.java
+++ b/core/src/main/java/org/apache/brooklyn/entity/stock/BasicStartableImpl.java
@@ -22,6 +22,13 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.common.base.Predicates;
+import com.google.common.collect.Iterables;
+import com.google.common.collect.Lists;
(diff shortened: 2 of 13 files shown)


core/src/main/java/org/apache/brooklyn/entity/stock/BasicStartableImpl.java
```diff
@@ -51,6 +52,10 @@ public void start(Collection<? extends Location> locations) {
         try {
             ServiceStateLogic.setExpectedState(this, Lifecycle.STARTING);
 
+            // Opportunity to block startup until other dependent components are available
+            Object val = config().get(START_LATCH);
+            if (val != null) log.debug("{} finished waiting for start-latch; continuing...", this, val);
```
sjcorbett: There's no matching `{}` for `val` in the message.

core/src/main/java/org/apache/brooklyn/core/entity/AbstractApplication.java
```diff
@@ -148,19 +149,23 @@ public void start(Collection<? extends Location> locations) {
...
         setExpectedStateAndRecordLifecycleEvent(Lifecycle.STARTING);
         try {
             preStart(locationsToUse);
-            // if there are other items which should block service_up, they should be done in preStart
-            ServiceStateLogic.ServiceNotUpLogic.clearNotUpIndicator(this, Attributes.SERVICE_STATE_ACTUAL);
-            
+
+            // Opportunity to block startup until other dependent components are available
+            Object val = config().get(START_LATCH);
+            if (val != null) log.debug("{} finished waiting for start-latch; continuing...", this, val);
```
sjcorbett: And here.

sjcorbett:
Two trivial comments, otherwise good to merge.

core/src/main/java/org/apache/brooklyn/core/entity/AbstractApplication.java
aledsage: I see you've copy-pasted the log message from `MachineLifecycleEffectorTasks`. I don't think the `val` will be interesting (presumably it will simply be "true"). Worth tidying up this log statement and also those in `MachineLifecycleEffectorTasks`.

core/src/main/java/org/apache/brooklyn/core/entity/AbstractApplication.java
```diff
@@ -132,12 +133,12 @@ public AbstractApplication setParent(Entity parent) {
     @Override
     protected void initEnrichers() {
         super.initEnrichers();
-        
+
         // default app logic; easily overridable by adding a different enricher with the same tag
         ServiceStateLogic.newEnricherFromChildren().checkChildrenAndMembers().addTo(this);
-        ServiceStateLogic.ServiceNotUpLogic.updateNotUpIndicator(this, Attributes.SERVICE_STATE_ACTUAL, "Application created but not yet started, at "+Time.makeDateString());
+        ServiceStateLogic.ServiceNotUpLogic.updateNotUpIndicator(this, Attributes.SERVICE_STATE_ACTUAL, Lifecycle.CREATED);
```
aledsage: Why change these, out of interest? They are now inconsistent with the calls in `stop()` (which pass "Application stopping" etc).

grkvlt: Because the not-up indicator is checking a sensor against its value, and the `Attributes.SERVICE_STATE_ACTUAL` is of type `Lifecycle`. I'll also change `stop()` ten...

neykov: The attribute is passed just for the name, its type is not used in the not-up map. Human readable values are preferred as that's its primary target - troubleshooting by users.

grkvlt:
Updated an squashed commits

grkvlt:
Added another commit to fix issue with Clocker where `preInstall()` was being skipped.

software/base/src/main/java/org/apache/brooklyn/entity/software/base/AbstractSoftwareProcessDriver.java
```diff
@@ -108,23 +108,24 @@ public void start() {
...
+                DynamicTasks.queue("pre-install", new Runnable() { public void run() {
+                    preInstall();
+                }});
+
+                DynamicTasks.queue("pre-install-command", new Runnable() { public void run() {
+                    runPreInstallCommand();
+                }});
+
+                boolean skipInstall = locationInstalled.or(entityInstalled).or(false);
+                if (!skipInstall) {
```
bostko: Just to state the cases we need to test or to have automated test:

1. Install multiple apps on a BYON location
1. Perform a restart on a BYON location
1. Perform a restart with machineStop: NEVER/(or false)

grkvlt: What I really want is a stage that is always executed on entity startup, maybe not `preInstall()` but `prepare()` that doesn't do anything on the VM, but sets up the entity instance. This is where the setting of expandedInstallDir and so on would happen. I'll see if it's easy to make that change.

grkvlt:
@bostko I think this (and https://github.com/apache/brooklyn-library/pull/14) will do what you want. The `preInstall()` methods will be skipped if the `SKIP_INSTALL` key is set, and the new `prepare()` method will set up the entity internals every time.

sjcorbett:
@grkvlt Can you rebase this please?

grkvlt:
@sjcorbett done

sjcorbett:
Tests pass locally.

ahgittin:
@sjcorbett have to confirmed that downstream entities work with this PR and with no changes?

NB there are a lot of related changes in

* https://github.com/apache/brooklyn-library/pull/14

but (hopefully) these are cleaning it up, not requirements.  (if this PR breaks backwards compatibility then we have to document that very clearly!)

core/src/main/java/org/apache/brooklyn/core/entity/AbstractApplication.java
```diff
@@ -152,15 +153,19 @@ public void start(Collection<? extends Location> locations) {
...
-            
+
+            // Opportunity to block startup until other dependent components are available
+            Object val = config().get(START_LATCH);
+            if (val != null) log.debug("{} finished waiting for start-latch; continuing...", this);
+
             doStart(locationsToUse);
             postStart(locationsToUse);
+
+            ServiceStateLogic.ServiceNotUpLogic.clearNotUpIndicator(this, Attributes.SERVICE_STATE_ACTUAL);
```
ahgittin: feels odd to do this after start, we might be clearing important not-up indications which happened during `doStart` etc.  (ideally we probably shouldn't have to do it at all.)

not a big deal as logic has not been perfect but maybe worth a comment in the code

ahgittin: i guess this is how we do it elsewhere, we basically assume if `start` failed without errors then it should normally be up, and we'll get told subsequently + periodically if that isn't the case.

in which case it makes sense to me to put this after, else we might catch a rogue not-up indicator from before the start which isn't reset in time.

(but still worth a comment in the code!)

ahgittin:
don't like seeing expansion of root lifeycle methods (`provision` `prepare` `install` ...) but +1 feels like the best thing for now

nice @grkvlt

sjcorbett:
@ahgittin I ran the library/nosql+webapp integration tests. All nosql tests passed. The webapp tests that failed did not do so because of this change. As I understood it library-#14 shifts entity's previous behaviour from `preInstall` to `prepare`, so while one is merged and the other isn't Brooklyn will behave as before (and thus https://github.com/brooklyncentral/clocker/issues/245 still holds). @grkvlt Can you confirm?

ahgittin:
Thanks @sjcorbett .  I haven't noticed any problems either.

grkvlt:
@sjcorbett correct https://github.com/apache/brooklyn-library/pull/14 needs merged to fix Clocker

PR #38, opened 2016-02-24
Misc server changes for nicer UI experience

ahgittin:
rendering errors more helpfully, and waiting for app to be STARTING when deploying

### DIFF
--- a/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
+++ b/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
@@ -1347,6 +1347,7 @@ public class JcloudsLocation extends AbstractCloudMachineProvisioningLocation im
                     }})
             .put(AUTO_CREATE_FLOATING_IPS, new CustomizeTemplateOptions() {
                     public void apply(TemplateOptions t, ConfigBag props, Object v) {
+                        LOG.warn("Using deprecated "+AUTO_CREATE_FLOATING_IPS+"; use "+AUTO_ASSIGN_FLOATING_IP+" instead");
                         if (t instanceof NovaTemplateOptions) {
                             ((NovaTemplateOptions)t).autoAssignFloatingIp((Boolean)v);
                         } else {
@@ -1553,7 +1554,6 @@ public class JcloudsLocation extends AbstractCloudMachineProvisioningLocation im
     }
 
     protected String toStringNice() {
-        Entities.dumpInfo(this);
         String s = config().get(ORIGINAL_SPEC);
         if (Strings.isBlank(s)) s = config().get(NAMED_SPEC_NAME);
         if (Strings.isBlank(s)) s = config().get(FINAL_SPEC);
@@ -1561,13 +1561,13 @@ public class JcloudsLocation extends AbstractCloudMachineProvisioningLocation im
         
         String s2 = "";
         String provider = getProvider();
-        if (Strings.isNonBlank(s) && Strings.isNonBlank(provider) && !s.toLowerCase().contains(provider.toLowerCase()))
+        if (Strings.isBlank(s) || (Strings.isNonBlank(provider) && !s.toLowerCase().contains(provider.toLowerCase())))
             s2 += " "+provider;
         String region = getRegion();
-        if (Strings.isNonBlank(s) && Strings.isNonBlank(region) && !s.toLowerCase().contains(region.toLowerCase()))
+        if (Strings.isBlank(s) || (Strings.isNonBlank(region) && !s.toLowerCase().contains(region.toLowerCase())))
             s2 += " "+region;
         String endpoint = getEndpoint();
-        if (Strings.isNonBlank(s) && Strings.isNonBlank(endpoint) && !s.toLowerCase().contains(endpoint.toLowerCase()))
+        if (Strings.isBlank(s) || (Strings.isNonBlank(endpoint) && !s.toLowerCase().contains(endpoint.toLowerCase())))
             s2 += " "+endpoint;
         s2 = s2.trim();
         if (Strings.isNonBlank(s)) {
@@ -2714,7 +2714,10 @@ public class JcloudsLocation extends AbstractCloudMachineProvisioningLocation im
             waitForReachable(checker, connectionDetails, credentialsToTry, setup, timeout);
         } finally {
             for (WinRmMachineLocation machine : machinesToTry.keySet()) {
-                getManagementContext().getLocationManager().unmanage(machine);
+                if (getManagementContext().getLocationManager().isManaged(machine)) {
+                    // get benign but unpleasant warnings if we unmanage something already unmanaged
+                    getManagementContext().getLocationManager().unmanage(machine);
+                }
             }
         }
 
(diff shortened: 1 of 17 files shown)


locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -1551,6 +1552,37 @@ public Template buildTemplate(ComputeService computeService, ConfigBag config) {
         return template;
     }
 
+    protected String toStringNice() {
+        Entities.dumpInfo(this);
```
sjcorbett: Is this left intentionally?

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -1551,6 +1552,37 @@ public Template buildTemplate(ComputeService computeService, ConfigBag config) {
...
+        if (Strings.isNonBlank(s) && Strings.isNonBlank(endpoint) && !s.toLowerCase().contains(endpoint.toLowerCase()))
+            s2 += " "+endpoint;
+        s2 = s2.trim();
+        if (Strings.isNonBlank(s)) {
+            if (Strings.isNonBlank(s2)) {
+                return s+" ("+s2+")";
+            }
+            return s;
+        }
+        if (Strings.isNonBlank(s2)) {
```
sjcorbett: `s2` can only be non-blank if `s` is non-blank and in that case the method has already returned on line 1575.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -1551,6 +1552,37 @@ public Template buildTemplate(ComputeService computeService, ConfigBag config) {
...
 
+    protected String toStringNice() {
+        Entities.dumpInfo(this);
+        String s = config().get(ORIGINAL_SPEC);
+        if (Strings.isBlank(s)) s = config().get(NAMED_SPEC_NAME);
+        if (Strings.isBlank(s)) s = config().get(FINAL_SPEC);
+        if (Strings.isBlank(s)) s = getDisplayName();
+        
+        String s2 = "";
+        String provider = getProvider();
```
sjcorbett: Here up to ~line 1581 would be much clearer if `Strings.isNonBlank(s)` were checked once and the whole block skipped if not.

utils/common/src/main/java/org/apache/brooklyn/util/exceptions/Exceptions.java
```diff
@@ -95,7 +95,8 @@ public static boolean isPrefixBoring(Throwable t) {
         return false;
     }
 
-    private static String stripBoringPrefixes(String s) {
+    // not used, but kept in case we might want it
+    static String stripBoringPrefixes(String s) {
```
sjcorbett: Deprecate?

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -2782,7 +2814,10 @@ public Boolean call() {
             waitForReachable(checker, connectionDetails, credentialsToTry, setup, timeout);
         } finally {
             for (SshMachineLocation machine : machinesToTry.keySet()) {
-                getManagementContext().getLocationManager().unmanage(machine);
+                if (getManagementContext().getLocationManager().isManaged(machine)) {
+                    // get benign but unpleasant warnings if we unmanage something already unmanaged
+                    getManagementContext().getLocationManager().unmanage(machine);
```
sjcorbett: This seems like something the location manager should take care of. Either way I assume the equivalent logic should be applied to `waitForWinRmAvailable`.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocationConfig.java
```diff
@@ -77,6 +66,20 @@
...
+    public static final ConfigKey<String> KEY_PAIR = ConfigKeys.newStringConfigKey("keyPair",
+        "Custom keypair (name) known at the cloud to be installed on machines for initial login (selected clouds only); "
+        + "you may also need to set "+LOGIN_USER_PRIVATE_KEY_FILE.getName(), null);
+    public static final ConfigKey<Boolean> AUTO_GENERATE_KEYPAIRS = ConfigKeys.newBooleanConfigKey("jclouds.openstack-nova.auto-generate-keypairs",
+        "Whether to generate keypairs automatically (OpenStack Nova)");
+
+    /**
+     * @deprecated since 0.7.0 Use {@link #AUTO_ASSIGN_FLOATING_IP} instead
+     */
+    public static final ConfigKey<Boolean> AUTO_CREATE_FLOATING_IPS = ConfigKeys.newBooleanConfigKey("jclouds.openstack-nova.auto-create-floating-ips",
```
sjcorbett: Can this be deleted?

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/resources/ApplicationResource.java
```diff
@@ -315,6 +320,21 @@ private Response launch(String yaml, EntitySpec<? extends Application> spec) {
         }
     }
 
+    private void waitForStart(final Application app, Duration timeout) {
+        // without this UI shows "stopped" sometimes esp if brooklyn has just started,
+        // because app goes to stopped state if it sees unstarted children,
+        // and that gets returned too quickly, before the start has taken effect
```
sjcorbett: This feels like a treatment of the symptom rather than the cause (which is probably rather more involved).

software/base/src/main/java/org/apache/brooklyn/entity/software/base/lifecycle/ScriptHelper.java
```diff
@@ -425,11 +425,11 @@ public ScriptHelper environmentVariablesReset(Map<?,?> envVarsToSet) {
     }
     
     public String getResultStdout() {
-        if (stdout==null) throw new IllegalStateException("output not available on "+this+"; ensure gatherOutput(true) is set");
+        if (stdout==null) throw new IllegalStateException("stdour not available on "+this+"; ensure task has completed and gatherOutput(true) is set");
```
sjcorbett: `stdour`

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
ahgittin: good catch.  have changed the logic above, it should be

    if (Strings.isBlank(s) || (Strings.isNonBlank(endpoint) && !s.toLowerCase().contains(endpoint.toLowerCase())))

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocationConfig.java
ahgittin: no warning was being given :( so i've added that, let's remove in next version

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/resources/ApplicationResource.java
ahgittin: i'm not quite so down on this.  the problem is that the ui polls for status, and polls immediately after it gets confirmation that an application has been created.  since we almost always start right after creation, we now hold off on confirmation until it is started, so the ui doesn't see the misleading brief initial state transactions.

if we had real-time sockets going to the ui then there would be no need to wait for start but in a polling world this feels good to me.

ahgittin:
thanks sam, good feedback

updated, addressing all comments except where explicitly noted here

PR #39, opened 2016-02-24
Make destroy on stop configurable for apps

ahgittin:
should fix https://issues.apache.org/jira/browse/BROOKLYN-203 and BasicStartableTest.testTransitionsThroughLifecycles (https://builds.apache.org/job/brooklyn-master-build/16/console) failure

### DIFF
--- a/core/src/test/java/org/apache/brooklyn/entity/stock/BasicStartableTest.java
+++ b/core/src/test/java/org/apache/brooklyn/entity/stock/BasicStartableTest.java
@@ -32,6 +32,7 @@ import org.apache.brooklyn.api.location.LocationSpec;
 import org.apache.brooklyn.api.mgmt.ManagementContext;
 import org.apache.brooklyn.core.entity.Attributes;
 import org.apache.brooklyn.core.entity.Entities;
+import org.apache.brooklyn.core.entity.EntityAsserts;
 import org.apache.brooklyn.core.entity.RecordingSensorEventListener;
 import org.apache.brooklyn.core.entity.factory.ApplicationBuilder;
 import org.apache.brooklyn.core.entity.lifecycle.Lifecycle;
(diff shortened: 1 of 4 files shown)


ahgittin:
failure unrelated, infra problem https://issues.apache.org/jira/browse/INFRA-11316

core/src/test/java/org/apache/brooklyn/entity/stock/BasicStartableTest.java
```diff
@@ -148,11 +149,14 @@ public void testIgnoresUnstartableEntities() throws Exception {
...
     public void testTransitionsThroughLifecycles() throws Exception {
         startable = app.addChild(EntitySpec.create(BasicStartable.class));
+        EntityAsserts.assertAttributeEqualsEventually(app, Attributes.SERVICE_STATE_ACTUAL, Lifecycle.STOPPED);
+        
         RecordingSensorEventListener<Lifecycle> listener = new RecordingSensorEventListener<Lifecycle>(true);
         managementContext.getSubscriptionContext(startable)
                 .subscribe(startable, Attributes.SERVICE_STATE_ACTUAL, listener);
 
         app.start(ImmutableList.of(loc1));
+        app.config().set(StartableApplication.DESTROY_ON_STOP, false);
```
aledsage: The assertion below this one looks like it would fail non-deterministically. 

We have added a listener (with `managementContext.getSubscriptionContext(startable).subscribe(...)`, so that will get notifications asynchronously. Immediately after calling `app.stop()`, we assert that the listener has the expected events. I'd have though we'd need to use `Asserts.succeedsEventually(...)` when checking if the listener has received the expected notifications.

Graeme-Miller:
hi @ahgittin, we had this issue again with [this build](https://builds.apache.org/job/brooklyn-master-build/43/). Might be worth getting this PR merged in?

ahgittin:
good spot @aledsage.  fixing that then merging /cc @Graeme-Miller we'll nail these evil nondet fails!

PR #40, opened 2016-02-25
BROOKLYN-231 fix race where updating and deleting can cause deleted item not to be deleted

ahgittin:
the persists and deletions run simultaneously; now, don't write things which are deleted.
also suppress the immediate delta write that was previously being done;
this may impact tests if there relying on an immediate checkpoint, but that's a bad test,
for in normal operation checkpointing should be started when the mgmt node starts and there should be nothing to persist.
if we support turning on persistence later that should trigger a full write not a delta write.

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/mgmt/persist/BrooklynMementoPersisterToObjectStore.java
+++ b/core/src/main/java/org/apache/brooklyn/core/mgmt/persist/BrooklynMementoPersisterToObjectStore.java
@@ -24,6 +24,7 @@ import java.io.IOException;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executors;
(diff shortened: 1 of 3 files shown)


ahgittin:
failure unrelated, infra problem https://issues.apache.org/jira/browse/INFRA-11316

core/src/main/java/org/apache/brooklyn/core/mgmt/persist/BrooklynMementoPersisterToObjectStore.java
```diff
@@ -581,9 +583,16 @@ private Stopwatch deltaImpl(Delta delta, PersistenceExceptionHandler exceptionHa
             Stopwatch stopwatch = Stopwatch.createStarted();
             List<ListenableFuture<?>> futures = Lists.newArrayList();
             
+            Set<String> deletedIds = MutableSet.of();
```
aledsage: I'm surprised we need this. I can see that if one uses `ImmediateDeltaChangeListener` then it would happen - but that class is deprecated and not used (we should delete it).

If using `PeriodicDeltaChangeListener.persistNowInternal()` (which I thought all code paths would do), then it has already gone through this to remove the objects whose ids were removed.

ahgittin:
the problem is real so i'm going to merge but will review more after

core/src/main/java/org/apache/brooklyn/core/mgmt/persist/BrooklynMementoPersisterToObjectStore.java
ahgittin: looks like `persistNowInternal` doesn't actually do that, that's why we need it

PR #41, opened 2016-02-28
[OSGi] Support JAAS auth in Jetty

neykov:
Replace BrooklynPropertiesSecurityFilter with JAX-RS functionality.

Depends on #31, merge together with https://github.com/apache/brooklyn-ui/pull/10

### DIFF
--- a/rest/rest-server-jersey/src/test/java/org/apache/brooklyn/rest/util/HaMasterCheckResource.java
+++ b/rest/rest-server-jersey/src/test/java/org/apache/brooklyn/rest/util/HaMasterCheckResource.java
@@ -1,0 +1,41 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.brooklyn.rest.util;
+
+import javax.ws.rs.POST;
+import javax.ws.rs.Path;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.MediaType;
+
+@Path("/")
+@Produces(MediaType.APPLICATION_JSON)
+public class HaMasterCheckResource {
+
+    @POST
+    @Path("/server/shutdown")
+    public void shutdown() {
+
+    }
+
+    @POST
+    @Path("/ha/post")
+    public void post() {
+    }
+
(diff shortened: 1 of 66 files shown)


rest/rest-server-jersey/src/test/java/org/apache/brooklyn/rest/util/HaMasterCheckResource.java
```diff
@@ -0,0 +1,41 @@
...
+package org.apache.brooklyn.rest.util;
+
+import javax.ws.rs.POST;
+import javax.ws.rs.Path;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.MediaType;
+
+@Path("/")
+@Produces(MediaType.APPLICATION_JSON)
+public class HaMasterCheckResource {
```
ahgittin: maybe call this `TestingHaMasterCheckResource` (or even `DummyTestingResource` and change the paths so they don't mimic real ones?)

ahgittin:
looks reasonable.  should we deprecate the old mechanism for handling logout, or at least comment pointing to the new one?

rest/rest-server-jersey/src/test/java/org/apache/brooklyn/rest/util/HaMasterCheckResource.java
neykov: Renamed with a `Testing` prefix at https://github.com/neykov/brooklyn-server/commit/25ff23420e0b6a0bbe19ba31db8e3cdc88866b31. Can't use the more generic name because there are a few of them.

`/server/shutdown` needs to be fixed, because it's testing `HaHotCheckResourceFilter.isMasterRequiredForRequest` behaviour. The other paths are made up.

neykov:
Previous logout mechanism lives at `BrooklynPropertiesSecurityFilter` which is now deprecated and not used.

neykov:
Added two additional commits addressing problems that came up during testing.

neykov:
Fixed newly introduced conflicts here and #31

neykov:
Triggering a rebuild

neykov:
Jenkins didn't even start a build for this one, weird. #55 includes this changes from here and is successful.

m4rkmckenna:
Tested this along with #31 & #55 ... noticed the following

Works as expected with the following in your `brooklyn.properties`
```
brooklyn.webconsole.security.provider=org.apache.brooklyn.rest.security.provider.AnyoneSecurityProvider
```
**OR**
```
brooklyn.webconsole.security.users=superAdmin
brooklyn.webconsole.security.user.superAdmin.password=5up3r4dm1n
```

but leaving the above out `org.apache.brooklyn.rest.security.provider.BrooklynUserWithRandomPasswordSecurityProvider` is never invoked so you end up with the below error when login is attempted

```
2016-03-16 15:13:17,521 | INFO  | qtp26227812-675  | DelegatingSecurityProvider       | 55 - org.apache.brooklyn.rest-resources - 0.9.0.SNAPSHOT | REST using security provider org.apache.brooklyn.rest.security.provider.ExplicitUsersSecurityProvider
2016-03-16 15:13:17,521 | WARN  | qtp26227812-675  | ExplicitUsersSecurityProvider    | 55 - org.apache.brooklyn.rest-resources - 0.9.0.SNAPSHOT | REST has no users configured; no one will be able to log in!
```

neykov:
Thanks for testing this @m4rkmckenna. Addressed comment in https://github.com/apache/brooklyn-server/pull/69 (not directly related to this PR).

neykov:
Also rebased all related PRs to resolve merge conflicts.

PR #43, opened 2016-02-29
Fix hostname resolving for jclouds locations

neykov:
* don't initialize hostname directly from the node
* correct lazy hostname initialization logic

According to [this comment](
https://github.com/apache/brooklyn-server/blob/master/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsSshMachineLocation.java#L280-L283) we shouldn’t use `node.getHostname()`. But in practice that’s exactly what happens, because `hostname` is [initialized here]( 
https://github.com/apache/brooklyn-server/blob/master/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsSshMachineLocation.java#L200) which is called both on location init and rebind.

### DIFF
--- a/locations/jclouds/pom.xml
+++ b/locations/jclouds/pom.xml
@@ -206,5 +206,15 @@
             <artifactId>mockito-all</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <!-- 
+                jsr311 excluded above, this module is JAX-RS version independent. Version decided at user.
+                For tests running in the module still need to use a specific dependency - use JAX-RS 2.0 as
+                that's what is used in standard build.
+            -->
+            <groupId>javax.ws.rs</groupId>
+            <artifactId>javax.ws.rs-api</artifactId>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 </project>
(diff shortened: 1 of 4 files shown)


locations/jclouds/pom.xml
```diff
@@ -206,5 +206,15 @@
...
         </dependency>
+        <dependency>
+            <!-- 
+                jsr311 excluded above, this module is JAX-RS version independent. Version decided at user.
+                For tests running in the module still need to use a specific dependency - use JAX-RS 2.0 as
+                that's what is used in standard build.
+            -->
+            <groupId>javax.ws.rs</groupId>
+            <artifactId>javax.ws.rs-api</artifactId>
+            <scope>test</scope>
```
ahgittin: why this?  tests seem to have been running fine without this, and the tests in this PR don't seem to need it.

ahgittin: or presumably live tests haven't been working without this...

ahgittin:
fix looks good.  assuming the new dep is for live tests i think good to merge @neykov .

locations/jclouds/pom.xml
neykov: That's right, live tests were failing.

PR #44, opened 2016-03-01
WinRM operationTimeout parameter

bostko:
* Configurable  operationTimeout. 1 minute by default as the new winrm4j does regular status polling when a command is executed.
* WinRM set default execTries
  This enables easier debugging and immidiate feedback when a problem occurs.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -151,7 +151,7 @@
         <jax-rs-api.version>2.0.1</jax-rs-api.version>
         <maxmind.version>0.8.1</maxmind.version>
         <jna.version>4.0.0</jna.version>
-        <winrm4j.version>0.3.1</winrm4j.version>
+        <winrm4j.version>0.3.2</winrm4j.version>
         <!-- Transitive dependencies, declared explicitly to avoid version mismatch -->
         <clojure.version>1.4.0</clojure.version>
         <zookeeper.version>3.3.4</zookeeper.version>
--- a/software/winrm/src/main/java/org/apache/brooklyn/location/winrm/WinRmMachineLocation.java
+++ b/software/winrm/src/main/java/org/apache/brooklyn/location/winrm/WinRmMachineLocation.java
@@ -77,6 +77,7 @@ public class WinRmMachineLocation extends AbstractLocation implements MachineLoc
             "Address of the remote machine");
 
     public static final ConfigKey<Integer> WINRM_CONFIG_PORT = newConfigKeyWithPrefix(BrooklynConfigKeys.BROOKLYN_WINRM_CONFIG_KEY_PREFIX, WinRmTool.PROP_PORT);
+    public static final ConfigKey<String> OPERATION_TIMEOUT = newConfigKeyWithPrefix(BrooklynConfigKeys.BROOKLYN_WINRM_CONFIG_KEY_PREFIX, WinRmTool.OPERATION_TIMEOUT);
     public static final ConfigKey<Boolean> USE_HTTPS_WINRM = WinRmTool.USE_HTTPS_WINRM;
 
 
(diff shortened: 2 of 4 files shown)


pom.xml
```diff
@@ -151,7 +151,7 @@
         <jax-rs-api.version>2.0.1</jax-rs-api.version>
         <maxmind.version>0.8.1</maxmind.version>
         <jna.version>4.0.0</jna.version>
-        <winrm4j.version>0.3.1</winrm4j.version>
+        <winrm4j.version>0.4.0</winrm4j.version>
```
aledsage: winrm4j version `0.3.2` is on its way to maven central - can you update to use that instead please?

software/winrm/src/main/java/org/apache/brooklyn/util/core/internal/winrm/WinRmTool.java
```diff
@@ -57,12 +57,13 @@
     ConfigKey<Boolean> USE_NTLM = ConfigKeys.newBooleanConfigKey("winrm.useNtlm", "The parameter configures tells the machine sensors whether the winrm port is over https. If the parameter is true then 5986 will be used as a winrm port.", true);
     ConfigKey<String> PROP_USER = newStringConfigKey("user", "User to connect as", null);
     ConfigKey<String> PROP_PASSWORD = newStringConfigKey("password", "Password to use to connect", null);
+    ConfigKey<String> OPERATION_TIMEOUT = newStringConfigKey("winrm.operationTimeout", "WinRM OperationTimeout", "1m");
```
neykov: Now that we have a better idea of what `operationTimeout` is actually doing, doesn't look like it's needed as a user-facing configuration. Suggest that we don't add this, wdyt?

software/winrm/src/main/java/org/apache/brooklyn/location/winrm/WinRmMachineLocation.java
```diff
@@ -339,7 +340,6 @@ protected WinRmTool newWinRmTool(Map<?,?> props) {
             args.putAll(props);
             args.configure(SshTool.PROP_HOST, getAddress().getHostAddress());
             args.configure(WinRmTool.USE_NTLM, getConfig(WinRmMachineLocation.USE_NTLM));
-            args.configure(WinRmTool.USE_HTTPS_WINRM, getConfig(WinRmMachineLocation.USE_HTTPS_WINRM));
```
neykov: Does the HTTPS config still get to the WinRmTool with this removed?

software/winrm/src/main/java/org/apache/brooklyn/util/core/internal/winrm/WinRmTool.java
bostko: I am for keeping it. I think it is good to have more configurable environment, especially when it is  still not heavily tested.

software/winrm/src/main/java/org/apache/brooklyn/location/winrm/WinRmMachineLocation.java
bostko: Yes it works. Check https://github.com/apache/brooklyn-server/blob/master/software/winrm/src/main/java/org/apache/brooklyn/location/winrm/WinRmMachineLocation.java#L153-L157 and the loop above.

software/winrm/src/main/java/org/apache/brooklyn/util/core/internal/winrm/WinRmTool.java
neykov: In this case document it better, noting that it has nothing to do with user timeouts. Instead it's polling timeout.

PR #47, opened 2016-03-04
minor code cleanup

ahgittin:
the only vaguely interesting items:
* make `INSTALL_DIR` avail as a shell env var when doing customize and launch
* make `MachineInitTasks` transient, in case the outer class is ever serialized (it shouldn't be but it seems that it is)

### DIFF
--- a/software/base/src/main/java/org/apache/brooklyn/entity/software/base/lifecycle/MachineLifecycleEffectorTasks.java
+++ b/software/base/src/main/java/org/apache/brooklyn/entity/software/base/lifecycle/MachineLifecycleEffectorTasks.java
@@ -124,7 +124,7 @@ public abstract class MachineLifecycleEffectorTasks {
     public static final ConfigKey<Duration> STOP_PROCESS_TIMEOUT = ConfigKeys.newConfigKey(Duration.class,
             "process.stop.timeout", "How long to wait for the processes to be stopped; use null to mean forever", Duration.TWO_MINUTES);
 
-    protected final transient MachineInitTasks machineInitTasks = new MachineInitTasks();
+    protected final MachineInitTasks machineInitTasks = new MachineInitTasks();
     
     /** Attaches lifecycle effectors (start, restart, stop) to the given entity post-creation. */
     public void attachLifecycleEffectors(Entity entity) {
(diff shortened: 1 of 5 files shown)


neykov:
retest this please

neykov:
Looks good, thought the above comment will trigger a build...

software/base/src/main/java/org/apache/brooklyn/entity/software/base/lifecycle/MachineLifecycleEffectorTasks.java
```diff
@@ -124,7 +124,7 @@
     public static final ConfigKey<Duration> STOP_PROCESS_TIMEOUT = ConfigKeys.newConfigKey(Duration.class,
             "process.stop.timeout", "How long to wait for the processes to be stopped; use null to mean forever", Duration.TWO_MINUTES);
 
-    protected final MachineInitTasks machineInitTasks = new MachineInitTasks();
+    protected final transient MachineInitTasks machineInitTasks = new MachineInitTasks();
```
neykov: Thinking more about this it will do more damage than good:
  * `MachineInitTasks` doesn't have any properties (hence state), so is lightweight to serialize
  * On rebind the constructor of objects in config bag is not called by xstream, so this will be null. In most cases where the default value of the config key `SoftwareProcess.LIFECYCLE_EFFECTOR_TASKS` is used it will work fine, but when the user sets a value and it is serialized things will fall apart.

Tried adding `DoNothingSoftwareProcess` to the `BrooklynMementoPersisterInMemorySizeIntegrationTest`, but the OOME didn't surface. Will continue looking.

ahgittin:
strange test failure, maybe network flakiness?

```
2016-03-04 13:39:11,651 INFO  TESTNG FAILED: "Surefire test" - org.apache.brooklyn.util.core.ResourceUtilsHttpTest.testGet() finished in 1015 ms
org.apache.brooklyn.util.exceptions.PropagatedRuntimeException: Error getting resource 'http://10.10.3.184:50505/simple' for mycontext: ConnectException: Connection refused
	at org.apache.brooklyn.util.exceptions.Exceptions.propagate(Exceptions.java:137)
	at org.apache.brooklyn.util.core.ResourceUtils.getResourceFromUrl(ResourceUtils.java:296)
	at org.apache.brooklyn.util.core.ResourceUtilsHttpTest.testGet(ResourceUtilsHttpTest.java:78)
Caused by: java.net.ConnectException: Connection refused
	at java.net.PlainSocketImpl.socketConnect(Native Method)
```

closing + reopening

PR #48, opened 2016-03-05
for local address when used in tests, ports not contactable on some build machines; try loop back instead

ahgittin:
see if this fixes failed PR builds on `ubuntu-us-1` eg as described at #47

### DIFF
--- a/software/base/src/test/java/org/apache/brooklyn/feed/jmx/JmxFeedTest.java
+++ b/software/base/src/test/java/org/apache/brooklyn/feed/jmx/JmxFeedTest.java
@@ -113,7 +113,7 @@ public class JmxFeedTest {
     
     public static class TestEntityWithJmx extends TestEntityImpl {
         @Override public void init() {
-            sensors().set(Attributes.HOSTNAME, "localhost");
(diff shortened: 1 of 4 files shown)


ahgittin:
first commit seemed to fix the problem in core https://builds.apache.org/job/brooklyn-server-pull-requests/144/, revealing same issue w jmx.  hopefully second will tidy that up.

neykov:
Changes look good, but looks like there's something more going on. Last failure:
```
unable to connect to JMX url: service:jmx:rmi://127.0.0.1:40069/jndi/rmi://127.0.0.1:40069/jmxrmi
java.rmi.ConnectException: Connection refused to host: 10.10.3.184; nested exception is: 
	java.net.ConnectException: Connection refused
```

ahgittin:
yeah we're making progress.  something is jmx is still advertising the local ip.  can't look in to it just now but will later.

good to fix these tests, maybe also good to fix the machine.

ahgittin:
fixed when target server was rebooted.  there probably are still jmx failures if the machine's IP is not self-contactable - but immediate problem is resolved.

PR #49, opened 2016-03-06
Make sure no unwanted state is serialized

neykov:
Certain types of objects could bring a lot of unwanted baggage with them when serialized.
  * Check and warn if such objects are serialized
  * Test that serializing SoftwareProcessDriverLifecycleEffectorTasks and MachineInitTasks is lean, doesn't lead to OOME

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/mgmt/persist/UnwantedStateLoggingMapper.java
+++ b/core/src/main/java/org/apache/brooklyn/core/mgmt/persist/UnwantedStateLoggingMapper.java
@@ -1,0 +1,68 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/apache/brooklyn/core/mgmt/persist/UnwantedStateLoggingMapper.java
```diff
@@ -0,0 +1,61 @@
...
+        logIfInteresting(type);
+        return super.serializedClass(type);
+    }
+
+    private void logIfInteresting(Class<?> type) {
+        if (type != null) {
+            if (Task.class.isAssignableFrom(type)) {
+                long cnt = WARN_CNT.getAndIncrement();
+                if (cnt < 5 || cnt % 10000 == 0) {
+                    LOG.warn("Trying to serialize a Task object of type " + type + ", which could lead to OOME. " +
```
ahgittin: won't this `WARN` get invoked for every class, or at least every class that wants to use default mapping?

neykov: It's at the top of the mappers chain, so the method should be called for any object.
The `warn` will be invoked only for objects implementing `Task` - one class of objects which are not meant to be serialized. Could add even more checks - `ThreadLocal` comes to mind. I've seen it serialized.

ahgittin: whoops missed line 46.  good idea to add `static TYPES_TO_WARN_IF_SERIALIZED` containing `ThreadLocal` and `Thread`.

ahgittin:
LGTM, merging

neykov:
Thanks, missed my last update for a second or so :) - https://github.com/apache/brooklyn-server/pull/50

PR #50, opened 2016-03-07
Make unwanted state checks configurable

neykov:


ahgittin:
even better than #49 :)

PR #53, opened 2016-03-09
SoftLayer VLAN customizer

grkvlt:
Builds on code from https://github.com/brooklyncentral/clocker/pull/138 to create a `JcloudsLocationCustomizer` that can be used to ensure all VMs in a specific application or other scope are created on the same VLAN. Use by configuring `customizers` on the location with an instance of this class.

### DIFF
--- a/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/softlayer/SoftLayerSameVlanLocationCustomizer.java
+++ b/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/softlayer/SoftLayerSameVlanLocationCustomizer.java
@@ -325,14 +325,15 @@ public class SoftLayerSameVlanLocationCustomizer extends BasicJcloudsLocationCus
      */
     protected CountDownLatch createCountDownLatch(JcloudsLocation location, String scopeUid) {
         synchronized (lock) {
-            Map<String, CountDownLatch> map = location.config().get(COUNTDOWN_LATCH_MAP);
-            if (map == null) { map = MutableMap.of(); }
+            Map<String, CountDownLatch> map = MutableMap.copyOf(location.config().get(COUNTDOWN_LATCH_MAP));
 
             if (!map.containsKey(scopeUid)) {
                 map.put(scopeUid, new CountDownLatch(1));
             }
             CountDownLatch latch = map.get(scopeUid);
-            location.config().set(COUNTDOWN_LATCH_MAP, map);
+
+            location.config().set(COUNTDOWN_LATCH_MAP, ImmutableMap.copyOf(map));
+            location.getManagementContext().getRebindManager().forcePersistNow(false, null);
 
             return latch;
         }
@@ -343,21 +344,25 @@ public class SoftLayerSameVlanLocationCustomizer extends BasicJcloudsLocationCus
      */
     protected void removeCountDownLatch(JcloudsLocation location, String scopeUid) {
         synchronized (lock) {
-            Map<String, CountDownLatch> map = location.config().get(COUNTDOWN_LATCH_MAP);
-            if (map != null) {
-                map.remove(scopeUid);
-            }
+            Map<String, CountDownLatch> map = MutableMap.copyOf(location.config().get(COUNTDOWN_LATCH_MAP));
+
+            map.remove(scopeUid);
+
+            location.config().set(COUNTDOWN_LATCH_MAP, ImmutableMap.copyOf(map));
+            location.getManagementContext().getRebindManager().forcePersistNow(false, null);
         }
     }
 
     /** Return the public VLAN number for a scope. */
     protected Integer lookupPublicVlanId(JcloudsLocation location, String scopeUid) {
-        synchronized (SoftLayerSameVlanLocationCustomizer.class) {
-            Map<String, Integer> map = location.config().get(PUBLIC_VLAN_ID_MAP);
-            if (map == null) {
-                map = MutableMap.of();
-                location.config().set(PUBLIC_VLAN_ID_MAP, map);
+        synchronized (lock) {
+            Map<String, Integer> map = MutableMap.copyOf(location.config().get(PUBLIC_VLAN_ID_MAP));
+
+            if (map.isEmpty()) {
+                location.config().set(PUBLIC_VLAN_ID_MAP, ImmutableMap.copyOf(map));
+                location.getManagementContext().getRebindManager().forcePersistNow(false, null);
             }
+
             return map.get(scopeUid);
         }
     }
@@ -365,22 +370,25 @@ public class SoftLayerSameVlanLocationCustomizer extends BasicJcloudsLocationCus
     /** Save the public VLAN number for a scope. */
     protected void savePublicVlanId(JcloudsLocation location, String scopeUid, Integer publicVlanId) {
         synchronized (lock) {
-            Map<String, Integer> map = location.config().get(PUBLIC_VLAN_ID_MAP);
-            if (map == null) { map = MutableMap.of(); }
+            Map<String, Integer> map = MutableMap.copyOf(location.config().get(PUBLIC_VLAN_ID_MAP));
 
             map.put(scopeUid, publicVlanId);
-            location.config().set(PUBLIC_VLAN_ID_MAP, map);
+
+            location.config().set(PUBLIC_VLAN_ID_MAP, ImmutableMap.copyOf(map));
+            location.getManagementContext().getRebindManager().forcePersistNow(false, null);
         }
     }
 
     /** Return the private VLAN number for a scope. */
     protected Integer lookupPrivateVlanId(JcloudsLocation location, String scopeUid) {
         synchronized (lock) {
-            Map<String, Integer> map = location.config().get(PRIVATE_VLAN_ID_MAP);
-            if (map == null) {
-                map = MutableMap.of();
-                location.config().set(PRIVATE_VLAN_ID_MAP, map);
+            Map<String, Integer> map = MutableMap.copyOf(location.config().get(PRIVATE_VLAN_ID_MAP));
+
+            if (map.isEmpty()) {
+                location.config().set(PRIVATE_VLAN_ID_MAP, ImmutableMap.copyOf(map));
+                location.getManagementContext().getRebindManager().forcePersistNow(false, null);
             }
+
             return map.get(scopeUid);
         }
     }
@@ -388,11 +396,12 @@ public class SoftLayerSameVlanLocationCustomizer extends BasicJcloudsLocationCus
     /** Save the private VLAN number for a scope. */
     protected void savePrivateVlanId(JcloudsLocation location, String scopeUid, Integer privateVlanId) {
         synchronized (lock) {
-            Map<String, Integer> map = location.config().get(PRIVATE_VLAN_ID_MAP);
-            if (map == null) { map = MutableMap.of(); }
+            Map<String, Integer> map = MutableMap.copyOf(location.config().get(PRIVATE_VLAN_ID_MAP));
 
             map.put(scopeUid, privateVlanId);
-            location.config().set(PRIVATE_VLAN_ID_MAP, map);
+
+            location.config().set(PRIVATE_VLAN_ID_MAP, ImmutableMap.copyOf(map));
+            location.getManagementContext().getRebindManager().forcePersistNow(false, null);
         }
     }
 }
--- a/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/softlayer/SoftLayerSameVlanLocationCustomizer.java
+++ b/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/softlayer/SoftLayerSameVlanLocationCustomizer.java
@@ -1,0 +1,398 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.brooklyn.location.jclouds.softlayer;
+
+import java.util.Map;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+
+import javax.annotation.concurrent.ThreadSafe;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.common.base.Preconditions;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.reflect.TypeToken;
+import com.google.common.util.concurrent.Uninterruptibles;
+
+import org.jclouds.compute.ComputeService;
+import org.jclouds.compute.options.TemplateOptions;
+import org.jclouds.softlayer.SoftLayerApi;
+import org.jclouds.softlayer.compute.options.SoftLayerTemplateOptions;
+import org.jclouds.softlayer.domain.VirtualGuest;
+import org.jclouds.softlayer.features.VirtualGuestApi;
+import org.jclouds.softlayer.reference.SoftLayerConstants;
+
+import org.apache.brooklyn.api.entity.Entity;
+import org.apache.brooklyn.api.sensor.AttributeSensor;
+import org.apache.brooklyn.config.ConfigKey;
+import org.apache.brooklyn.core.config.ConfigKeys;
+import org.apache.brooklyn.core.location.LocationConfigKeys;
+import org.apache.brooklyn.core.sensor.Sensors;
+import org.apache.brooklyn.location.jclouds.BasicJcloudsLocationCustomizer;
+import org.apache.brooklyn.location.jclouds.JcloudsLocation;
+import org.apache.brooklyn.location.jclouds.JcloudsLocationCustomizer;
+import org.apache.brooklyn.location.jclouds.JcloudsMachineLocation;
+import org.apache.brooklyn.location.jclouds.JcloudsSshMachineLocation;
+import org.apache.brooklyn.util.collections.MutableMap;
+import org.apache.brooklyn.util.core.flags.SetFromFlag;
+import org.apache.brooklyn.util.core.task.Tasks;
+import org.apache.brooklyn.util.text.Strings;
+import org.apache.brooklyn.util.time.Duration;
+
+/**
+ * Customizes {@link JcloudsSshMachineLocation machines} in SoftLayer to use
+ * the same VLANs across an application, or other named scope
+ * <p>
+ * Define the scope by setting the {@link #SCOPE_UID scopeUid} ({@code softlayer.vlan.scopeUid})
+ * option. Set {@link #SCOPE_TIMEOUT scopeTimeout} ({@code softlayer.vlan.timeout}) to change
+ * the length of time the customizer will wait for VLAN information; normally 15 minutes.
+ * <p>
+ * The VLAN IDs and latches are stored as {@link ConfigKey configuration} on
+ * the {@link JcloudsLocation location} provisioning the VMs, in a map keyed
+ * on the scope.
+ */
+@ThreadSafe
+public class SoftLayerSameVlanLocationCustomizer extends BasicJcloudsLocationCustomizer {
+
+    private static final Logger LOG = LoggerFactory.getLogger(SoftLayerSameVlanLocationCustomizer.class);
+
+    @SetFromFlag("scopeUid")
+    public static final ConfigKey<String> SCOPE_UID = ConfigKeys.newStringConfigKey(
+            "softlayer.vlan.scopeUid",
+            "The unique identifier for a Softlayer location scope that will have VMs created in the same VLAN");
+
+    @SetFromFlag("scopeTimeout")
+    public static final ConfigKey<Duration> SCOPE_TIMEOUT = ConfigKeys.newDurationConfigKey(
+            "softlayer.vlan.timeout",
+            "The length of time to wait for a Softlayer VLAN ID",
+            Duration.minutes(15));
+
+    public static final ConfigKey<Map<String, CountDownLatch>> COUNTDOWN_LATCH_MAP = ConfigKeys.newConfigKey(
+            new TypeToken<Map<String, CountDownLatch>>() { },
+            "softLayerSameVlanLocationCustomizer.map.latches",
+            "A mapping from scope identifiers to CountDownLatches; used to synchronize threads");
+    public static final ConfigKey<Map<String, Integer>> PUBLIC_VLAN_ID_MAP = ConfigKeys.newConfigKey(
+            new TypeToken<Map<String, Integer>>() { },
+            "softLayerSameVlanLocationCustomizer.map.publicVlanIds",
+            "A mapping from scope identifiers to public VLAN numbers");
+    public static final ConfigKey<Map<String, Integer>> PRIVATE_VLAN_ID_MAP = ConfigKeys.newConfigKey(
+            new TypeToken<Map<String, Integer>>() { },
+            "softLayerSameVlanLocationCustomizer.map.privateVlanIds",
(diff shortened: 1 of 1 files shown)


grkvlt:
@ahgittin this is the first cut of the SoftLayer VLAN customizer for review

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/softlayer/SoftLayerSameVlanLocationCustomizer.java
```diff
@@ -0,0 +1,192 @@
...
+            LOG.warn(message);
+            throw new IllegalArgumentException(message);
+        }
+
+        Semaphore semaphore = null;
+        String scopeUid = getScopeUid(location);
+
+        synchronized (SoftLayerSameVlanLocationCustomizer.class) {
+            semaphore = semaphoreCache.getIfPresent(scopeUid);
+            if (semaphore == null) {
```
ahgittin: couldn't this be `null` if we're rebinding?  should probably share semaphore acquisition method with the `customize(..., TemplateOptions)`.  (also would it make sense to put that method first?)

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/softlayer/SoftLayerSameVlanLocationCustomizer.java
```diff
@@ -0,0 +1,192 @@
...
+ */
+public class SoftLayerSameVlanLocationCustomizer extends BasicJcloudsLocationCustomizer {
+
+    private static final Logger LOG = LoggerFactory.getLogger(SoftLayerSameVlanLocationCustomizer.class);
+
+    public static final ConfigKey<String> SCOPE_UID = ConfigKeys.newStringConfigKey("softlayer.vlan.scopeUid",
+            "The unique identifier for a Softlayer location scope that will have VMs created in the same VLAN");
+
+    public static final ConfigKey<Duration> SCOPE_TIMEOUT = ConfigKeys.newDurationConfigKey("softlayer.vlan.timeout",
+            "The length of time to wait for a Softlayer VLAN ID", Duration.minutes(10));
```
ahgittin: this feels like not long enough.  SL VM provisioning is often fast but not infrequently can take 25m or more.  suggest 1 hour?

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/softlayer/SoftLayerSameVlanLocationCustomizer.java
```diff
@@ -0,0 +1,192 @@
...
+        Semaphore semaphore = null;
+        String scopeUid = getScopeUid(location);
+
+        synchronized (SoftLayerSameVlanLocationCustomizer.class) {
+            semaphore = semaphoreCache.getIfPresent(scopeUid);
+            if (semaphore == null) {
+                // We are first, so create a new semaphore and return
+                LOG.debug("Creating semaphore for scope: {}", scopeUid);
+                semaphore = new Semaphore(1);
+                semaphoreCache.put(scopeUid, semaphore);
```
ahgittin: shouldn't this also acquire the semaphore?

ahgittin:
the failing jenkins is unrelated (it's because we moved the json snippets in testing and should go away when the maven cache updates)

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/softlayer/SoftLayerSameVlanLocationCustomizer.java
```diff
@@ -0,0 +1,192 @@
...
+            }
+        }
+
+        Integer publicVlanId = publicVlanIdCache.getIfPresent(scopeUid);
+        Integer privateVlanId = privateVlanIdCache.getIfPresent(scopeUid);
+        if (privateVlanId != null || publicVlanId != null) {
+            LOG.debug("SoftLayer VLANs already configured for application: {}", scopeUid);
+            return;
+        }
+
```
ahgittin: add comment

    // no vlan info yet known, we are probably the vm and we should set the vlan info for others to then learn about

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/softlayer/SoftLayerSameVlanLocationCustomizer.java
```diff
@@ -0,0 +1,192 @@
...
+                "primaryNetworkComponent.networkVlan;" +
+                "primaryBackendNetworkComponent;" +
+                "primaryBackendNetworkComponent.networkVlan");
+        publicVlanId = guest.getPrimaryNetworkComponent().getNetworkVlan().getVlanNumber();
+        privateVlanId = guest.getPrimaryBackendNetworkComponent().getNetworkVlan().getVlanNumber();
+
+        // Save the VLAN ids and release the semaphore
+        publicVlanIdCache.put(scopeUid, publicVlanId);
+        privateVlanIdCache.put(scopeUid, privateVlanId);
+        semaphore.release();
```
ahgittin: if using `semaphore with owners` we could check whether we are the thread which acquired the semaphore and log a warning if not, for good measure.

i think `release` whether or not we were the creator thread, because we now have a VM others can use.

however if the semaphore already has a permit available then don't release (eg if we're rebinding).

ahgittin:
good start, but i think the acquisition logic is too brittle.

if a machine fails or is cancelled or it somehow doesn't go through the `customize(..., Location)` method the lock is kept forever.  i'd suggest:

* in the `synchronized` block, before acquiring the semaphore, check whether anyone has owned it for more than 30m, if so, forcibly eject them (we have a class `SemaphoreWithOwners` which could be extended to include timestamp of when owning threads were issued it to achieve this)

* if it's easy to check for creation failure we should `release` if we were the owner (or simply if there are no permits available)

ahgittin:
static caches are bad!  :)  the right impl i think would be to search through known locations to look for locations with scope and store the vlan id's on that.  that would also allow us to resize to 0 then back out, and we'll get a new vlan id.  however that may be hard.

could we note that preferred implementation as a comment in the code?

however a more important problem is that the static cache prevents the vlan id's to be persisted.  if we restart brooklyn then scale out we will acquire a new vlan id with this code.  :(  i think that does need to be fixed.

it could be fixed by switching away from the cache.  however a simpler fix could be that the first thread which acquires the semaphore before exiting (without setting a vlan id) first does a query of machines in SL.  if there are some already there we just take their vlan id info and cache that.  it won't be perfect because vlan id *could* change if we rebooted brooklyn and the list of vm's from SL came back in a different order but i think we could live with that.  also it means that we are able to provision in parallel much more often, if there are any machines in SL, and machines in different apps will also go on to the same vlan (not needed but nice).

ahgittin:
also can you update the `more-clouds.md` doc to describe this for softlayer?

ahgittin:
completed.  good start but some things need addressed i think.

grkvlt:
@rdowner Still WIP but candidate for inclusion in the 0.9.0 release

ahgittin:
+1 will try to review in time for 090

grkvlt:
@ahgittin updates to documentation have been made in https://github.com/apache/brooklyn-docs/pull/31

grkvlt:
@ahgittin I think this is ready to merge now

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/softlayer/SoftLayerSameVlanLocationCustomizer.java
```diff
@@ -0,0 +1,398 @@
...
+            }
+        }
+
+        try {
+            // Check if VLAN IDs have been set for this scope
+            LOG.debug("Looking up saved VLAN details {}", scopeUid);
+            Integer publicVlanId = lookupPublicVlanId(location, scopeUid);
+            Integer privateVlanId = lookupPrivateVlanId(location, scopeUid);
+            if (privateVlanId != null || publicVlanId != null) {
+                LOG.debug("SoftLayer VLANs private {} and public {} already configured for scope: {}",
```
ahgittin: should we warn here if the looked up ID is different to those on the machine?

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/softlayer/SoftLayerSameVlanLocationCustomizer.java
```diff
@@ -0,0 +1,398 @@
...
+        }
+
+        // Try and acquire the latch, or time out
+        Duration timeout = getTimeout(location);
+        Tasks.setBlockingDetails("Waiting for VLAN details");
+        try {
+            LOG.debug("Waiting for VLAN details for scope: {}", scopeUid);
+            if (!Uninterruptibles.awaitUninterruptibly(latch, timeout.toMilliseconds(), TimeUnit.MILLISECONDS)) {
+                // Release the latch and throw an exception to prevent others blocking forever
+                latch.countDown();
```
ahgittin: confused here -- imagine provisioning threads A B C and D started, in that order.  i think then:

* A keeps the latch for longer than timeout
* B detects this, clears the latches, and throws
* C picks up the latch
* D is very soon going to timeout and repeat B's behaviour

Feels like B,C, and D need symmetry, and their timers reset if there is a new master, and A should be the thread to fail, rather than B (who here has done nothing wrong apart from being the whistleblower).

I was thinking `SemaphoreWithOwner` because then A can check whether the owner has been stolen from him and fail, B can claim ownership, and C & D can check whether the owner has changed and reset their timer appropriately.  (I think it's okay if C & D wait for the initial timeout to expire before detecting and resetting, btw; it's quite a bit more work to get them to reset their timers as soon as B steals ownership and not necessary IMO.)

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/softlayer/SoftLayerSameVlanLocationCustomizer.java
```diff
@@ -0,0 +1,398 @@
...
+    public static final ConfigKey<Duration> SCOPE_TIMEOUT = ConfigKeys.newDurationConfigKey(
+            "softlayer.vlan.timeout",
+            "The length of time to wait for a Softlayer VLAN ID",
+            Duration.minutes(15));
+
+    public static final ConfigKey<Map<String, CountDownLatch>> COUNTDOWN_LATCH_MAP = ConfigKeys.newConfigKey(
+            new TypeToken<Map<String, CountDownLatch>>() { },
+            "softLayerSameVlanLocationCustomizer.map.latches",
+            "A mapping from scope identifiers to CountDownLatches; used to synchronize threads",
+            MutableMap.<String, CountDownLatch>of());
```
ahgittin: this -- and all config keys -- do not need to return defaults.  the code tests for `null` so that's what i would expect.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/softlayer/SoftLayerSameVlanLocationCustomizer.java
```diff
@@ -0,0 +1,398 @@
...
+    public static final ConfigKey<Map<String, Integer>> PUBLIC_VLAN_ID_MAP = ConfigKeys.newConfigKey(
+            new TypeToken<Map<String, Integer>>() { },
+            "softLayerSameVlanLocationCustomizer.map.publicVlanIds",
+            "A mapping from scope identifiers to public VLAN numbers",
+            MutableMap.<String, Integer>of());
+    public static final ConfigKey<Map<String, Integer>> PRIVATE_VLAN_ID_MAP = ConfigKeys.newConfigKey(
```
ahgittin: i think it would simplify things a lot to define an object with

    transient CountdownLatch latch;
    Integer publicVlanId;
    Integer privateVlanId;

and then a single map capturing this, and then lookup/save this object rather than the ID's and latches separately.  this also allows us to:

* make the countdown transient which i think is needed for when locations are persisted
* combine the various lookup/save methods for vlan id's

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/softlayer/SoftLayerSameVlanLocationCustomizer.java
```diff
@@ -0,0 +1,398 @@
...
+            }
+        } finally {
+            Tasks.resetBlockingDetails();
+        }
+
+        // Looking up saved VLAN details
+        LOG.debug("Looking up saved VLAN details {}", scopeUid);
+        publicVlanId = lookupPublicVlanId(location, scopeUid);
+        privateVlanId = lookupPrivateVlanId(location, scopeUid);
+        if (privateVlanId == null && publicVlanId == null) {
```
ahgittin: i think we need both to be non-null, ie `||` instead of `&&` here -- or if we have `CombinedLatchVlanRecord` (as per comment on unifed map) it has a method `boolean vlansSet()` which tests that both are non-null (and similarly a `setVlans(publicVlan, privateVlan)` which also requires both to be non-null)

ahgittin:
A few more comments I'm afraid @grkvlt .  The latching logic still looks dubious and needs addressing, same for the non-transience of the latch.  The other comments are cosmetic but probably worth doing given changes are needed.

grkvlt:
@ahgittin I found the latches had better semantics than semaphores; will look into your other comments...

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/softlayer/SoftLayerSameVlanLocationCustomizer.java
grkvlt: Once the latch has counted down, that's it. It's acting like a one-time barrier, once it is zero, all other `await()` calls return true immediately. The semantics for semaphores were different, and I didn't like them. So, in your ABCD case:

1. A creates the latch and starts to do work
2. BCD all wait on it
3. A takes too long to finish
4. B times out and calls `countDown()` setting the latch to zero
5. C and D both drop out of their `await()` calls and discover there's no VLANS and fail
6. (Optional) A eventually finishes or crashes and the `finally` block executes, also calling `countDown()` on the latch, which has no effect

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/softlayer/SoftLayerSameVlanLocationCustomizer.java
grkvlt: True. Defensive coding?

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/softlayer/SoftLayerSameVlanLocationCustomizer.java
grkvlt: Dislike creating these data holder objects, but I see what you mean. It wouldn't affect the logic. Also, I understood that a `CountDownLatch` could be persisted no problem?

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/softlayer/SoftLayerSameVlanLocationCustomizer.java
grkvlt: It checks if both *are* null here, so not sure what you mean?

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/softlayer/SoftLayerSameVlanLocationCustomizer.java
grkvlt: Or:

1. A creates the latch and starts to do work
2. BCD all wait on it
3. A crashes and the `finally` block executes, calling `countDown()` on the latch, setting the latch to zero
4. B, C and D all drop out of their `await()` calls and discover there's no VLANS and fail

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/softlayer/SoftLayerSameVlanLocationCustomizer.java
ahgittin: if either is null, you allow it to proceed.  i don't think that's right.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/softlayer/SoftLayerSameVlanLocationCustomizer.java
ahgittin: My read is still that if A takes too long an B detects the timeout:

* B fails -- `throw` on line 246 -- causing the entity start to fail
* C and D return without setting any vlan info -- line 258 -- and then proceed to provision, and then the first of them will proceed to set these as the official info (line 184), whereas the other will quietly exit without even noticing it's possibly on a different subnet (line 163)

So A is still off and running, B has failed, and C and D may be on separate subnets.  However subsequent entities will correctly pick up the vlan's assigned to C.

Would a simple fix be to throw on line 258 (so C and D fail) but also for B to remove the latch in the map (so that a subsequent run call it E will generate a new latch and repeat the process; if in the meantime A succeeds then we'll nicely get its VLAN info; we live with the mess if A succeeds while E is provisioning, as long as we have a warning re E on line 161).

HTH

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/softlayer/SoftLayerSameVlanLocationCustomizer.java
aledsage: The default value for a config key should *never* be mutable! If it is, then someone can get the default value and modify the object. Someone else who then uses the same config key on a different entity/customizer will get the modified object as their default.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/softlayer/SoftLayerSameVlanLocationCustomizer.java
aledsage: Though xstream manages to serialize and deserialize the `CountDownLatch`, it leads to ugly looking serialized state. It's nice if our persisted state uses simpler datatypes wherever possible.

We discussed a similar change in Clocker's `DockerHost` where we removed a persisted attribute of `CountDownLatch`s, keeping them transient.

I haven't read this PR enough to say whether persisting the CountDownLatch seems necessary yet.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/softlayer/SoftLayerSameVlanLocationCustomizer.java
grkvlt: Ah, yes - removing the latch from the map will work. OK, I'll make that change.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/softlayer/SoftLayerSameVlanLocationCustomizer.java
grkvlt: Removed the defaults.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/softlayer/SoftLayerSameVlanLocationCustomizer.java
```diff
@@ -0,0 +1,385 @@
...
+    public static final AttributeSensor<Integer> PUBLIC_VLAN_ID = Sensors.newIntegerSensor(
+            "softLayer.vlan.publicId", "The public VLAN ID for this entity");
+    public static final AttributeSensor<Integer> PRIVATE_VLAN_ID = Sensors.newIntegerSensor(
+            "softLayer.vlan.privateId", "The private VLAN ID for this entity");
+
+    /* Flags passed in on object creation. */
+    private final Map<String, ?> flags;
+
+    /* Lock object for synchronization. */
+    private static final transient Object lock = new Object[0];
```
aledsage: Needs a comment justifying why it is `static`.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/softlayer/SoftLayerSameVlanLocationCustomizer.java
grkvlt: Did you mean the VLAN ID checks in the _other_ `customize()` method, which did have an `||` but is now also an `&&`

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/softlayer/SoftLayerSameVlanLocationCustomizer.java
```diff
@@ -0,0 +1,385 @@
...
+
+            // Save the VLAN ids
+            LOG.debug("Saving VLAN details private {} and public {} for scope: {}",
+                    new Object[] { privateVlanId, publicVlanId, scopeUid });
+            savePublicVlanId(location, scopeUid, publicVlanId);
+            savePrivateVlanId(location, scopeUid, privateVlanId);
+            saveVlanDetails(machine, scopeUid, privateVlanId, publicVlanId);
+        } finally {
+            // Release the latch
+            latch.countDown();
```
aledsage: The problem with persisting latches like this is that nothing explicitly triggers the object to be written to the persisted state again. We are not triggering persistence, so it won't be written to the object store (best case, we'll get persisted some time later when the location that this is associated with next gets persisted).

grkvlt: So, how do we best deal with this sort of situation? Can I trigger a save of the location state somehow?

grkvlt:
```Java
location.getManagementContext().getRebindManager().forcePersistNow(false, null);
```

What about doing this after mutating the config?

ahgittin: FWIW I think no point in persisting latches.  If A has completed it is unnecessary as a new latch will detect the shared vlan info.  If A has not completed it will fail/cancel and so on a restart a new latch is the same as a persisted latch with count=1.

ahgittin: Probably no harm though.

We will need to revisit items like this when we federate of course.  (That was another reason for suggesting SemaphoreWithOwner as we own that class, we could swap it for a federated backing store and this would work.  But we can address that when we do proper federation, unless of course the improved SL networking arrives first in which case this becomes redundant!)

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/softlayer/SoftLayerSameVlanLocationCustomizer.java
ahgittin: I mean right here line 267, if private is null but public is set this check returns false and we set `null` as the `primaryBackend`.  I'm not sure if that will work.  But not a big deal as I expect we'll always get either both non-null or both null.  Not sure what we should do if SL (or a user) sets one non-null and one null.

Maybe worth a comment to that effect...

ahgittin:
the `removeCountDownLatch` needs to be done by B not C/D (else D might be removing a latch created by E)

ahgittin:
otherwise changes look good.  (though they were hard to see, i guess you squashed things?  :(  )

will address and merge

nice work!

ahgittin:
merged, with two changes:

* cc9be55e2dce419539e6ab24db59dad370dc249c fix where latch is removed from map (and add comments, better logging, etc)
* f66d1ad4418cdf7ec4af373f3964e6cfb180a891 reorder so template options method is first, and refactor persistence

grkvlt:
Thanks @ahgittin - I also added 6727aa6209fdc8e91bd1e309f7efefd7eb1a12d6 to fix one of the checks.

grkvlt:
@ahgittin @aledsage What do you think about storing the latches in a `static` map, rather than persisting, and making the code resilient in the face of rebind causing the map to be empty? Should be better, I think, but this works so lets leave it for now before we try and perfect it?

ahgittin:
thanks for the fix @grkvlt.  re latches yes my preference would be a static cache which is not persisted, which is weak (ie can drop stuff) if the latch count is zeroed, and which looks through all locations on initialization -- but agree it's a low priority.  nice work here.

PR #57, opened 2016-03-11
Remove Incubator DISCLAIMER file

rdowner:


aledsage:
Searching for disclaimer files (`find . -name "DISCLAIMER*"`), I also see `./utils/jmx/jmxmp-ssl-agent/src/main/license/DISCLAIMER.shaded` (and its usage in `utils/jmx/jmxmp-ssl-agent/pom.xml`). Shall we deal with that in a separate PR?

Merging now.

PR #58, opened 2016-03-11
Fix SoftwareProcessPersisterInMemorySizeIntegrationTest

neykov:
Seen a failure where 153397 bytes were written, increase maximum. The test is not deterministic.


neykov:
retest this please

ahgittin:
failure unrelated, recorded in new non-det test issue https://issues.apache.org/jira/browse/BROOKLYN-239 /cc @aledsage @sjcorbett 

PR here looks good, merging

PR #61, opened 2016-03-11
BrooklynNode: support initial catalog.bom conf

aledsage:


software/base/src/main/java/org/apache/brooklyn/entity/brooklynnode/BrooklynNodeSshDriver.java
```diff
@@ -224,8 +231,18 @@ public void customize() {
...
         }
 
+        // Upload a local-catalog.bom if required
+        if (brooklynCatalogInitialBomContents != null || brooklynCatalogInitialBomUri != null) {
+            uploadFileContents(brooklynCatalogInitialBomContents, brooklynCatalogInitialBomUri, brooklynCatalogInitialBomRemotePath);
+        }
+
         // Override the ~/.brooklyn/catalog.xml if required
         if (brooklynCatalogContents != null || brooklynCatalogUri != null) {
+            log.warn("Deprecated use of config"
```
neykov: Add `Deprecated since x.x.x`. Looking at the config keys it would be `0.7.0`, but since that's not user visible might make sense to set it to 0.9.0.

neykov:
LGTM

software/base/src/main/java/org/apache/brooklyn/entity/brooklynnode/BrooklynNodeSshDriver.java
aledsage: I'll say since 0.7.0, because that is when we deprecated the field (not ideal, I agree). Don't want to have this code sitting around for longer than we need by saying it was only recently deprecated (even though we've only now added the log.warn message to tell the user about it directly).

PR #62, opened 2016-03-14
does not create a localhost named location automatically

ahgittin:
the location-add-wizard which @tbouron is working on will make it easy for users to do this, and there is no reason `localhost` should be added specially (and good reasons why it shouldn't as it is not representative and it is error-prone!)

i suggest this be tested and reviewed as part of @tbouron's WIP, and merged only at the same time as his.  (if this lands in time for 090 great but not needed.  probably confusing for a user to merge this PR without the wizard.)

we should also update the release notes with something like the following explaining the compatibility break:

```
6. The named location `localhost` is no longer automatically added by default on a fresh Brooklyn install.
Instead UI users are directed to a location wizard where they can configure their targets, including localhost. 
If you require `localhost` to be available on boot, define it as a named location in `brooklyn.properties`
or the default catalog. (The property `brooklyn.location.name.localhost=localhost` is usually sufficient.)
If you want a location named `localhost` afterwards, simply add it to your local catalog using the wizard.
```

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/location/BasicLocationRegistry.java
+++ b/core/src/main/java/org/apache/brooklyn/core/location/BasicLocationRegistry.java
@@ -47,6 +47,7 @@ import org.apache.brooklyn.core.location.internal.LocationInternal;
 import org.apache.brooklyn.core.mgmt.internal.LocalLocationManager;
 import org.apache.brooklyn.core.mgmt.internal.ManagementContextInternal;
 import org.apache.brooklyn.core.typereg.RegisteredTypePredicates;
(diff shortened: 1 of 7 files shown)


neykov:
LGTM

ahgittin:
build failure unrelated, due to wrong path in `karaf/init/` /cc @neykov, pushed a fix to master branch and merged here

https://github.com/apache/brooklyn-server/commit/1465b040bd9fcce4a95ea4fc927a8584de18dc05

ahgittin:
BTW please do not merge until location wizard ui lands

PR #67, opened 2016-03-17
Fix SshMachineLocation cleanup

aledsage:
Previously if a temporary location was created (i.e. using 
CREATE_UNMANAGED=true) then the cleanup task would execute immediately
and would close the ssh-connections-pool (potentially while the other 
thread tried to execute something).


grkvlt:
@aledsage LGTM :frog:

ahgittin:
+1 i've merged this in to #70 which is related, so this issue will close when that is merged

PR #68, opened 2016-03-17
multi-sensor-aggregator: respect the EntityFilter

aledsage:
This is useful for apps that want to override the 
ComputeServiceIndicatorsFromChildrenAndMembers. For example, in 
clocker the app’s health depends on just two of the children (rather 
than all/one or quorum).


grkvlt:
@aledsage this looks good, the test failure is spurious, can be merged

PR #69, opened 2016-03-17
[OSGi] Generate random password if no security provider configured

neykov:
Replicate classical launcher behaviour. Note that when accessing from localhost username will still be requested, but will accept any provider.

### DIFF
--- a/karaf/init/pom.xml
+++ b/karaf/init/pom.xml
@@ -44,6 +44,11 @@
             <artifactId>brooklyn-launcher-common</artifactId>
             <version>${project.version}</version>
         </dependency>
+        <dependency>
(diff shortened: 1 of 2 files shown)


neykov:
With this and #41 merged should change `BrooklynUserWithRandomPasswordSecurityProvider` to always ask for a password. Otherwise the browser will ask for password while the provider will authenticate any user when coming from localhost.

ahgittin:
@neykov should you / will you do the change you mention here or in a new PR ?

neykov:
Better do it in another PR, after sending an email to dev@a.o for the change. It will affect most users.
Also needs updating how BrooklynNode connects to self.
Or should we change behaviour just for Karaf?

PR #70, opened 2016-03-17
fix location leak - use LocationSpec wherever possible

ahgittin:
This improves how locations are handled internally, switching the resolvers to return `LocationSpec` rather than `Location` instances which may or may not be managed.

This allows us to fix a leak if catalog items include a location block (and includes a test for that, /cc @bostko ).  It also simplifies a number of things internally.  There is quite a bit more we'd like to do at some point (/cc @aledsage):

* have locations come entirely from the `TypeRegistry` (ie the catalog), doing away with `LocationDefinition` and `LocationRegistry`
* treat `Location` instances *as* `Entity` instances

However those are bigger changes, and this PR addresses the immediate pain points and makes the code quite a bit tidier.

There are a huge number of tests changed so to facilitate review the main important commits are:

* https://github.com/apache/brooklyn-server/commit/2876a9e5884332c38cdd2f631d0d221de2de35c9 pass location *specs* to entity spec (preferred over location instances)
* https://github.com/apache/brooklyn-server/commit/f755c25d2e66fa93a4a5f9f29ab8edc8d1c5be48 resolvers create location *specs* instead of location instances -- this breaks Resolver compatibility, mostly pretty easy to fix
* https://github.com/apache/brooklyn-server/commit/a3d74701fc37ae9b3daae15865510dfb1b94e327 special flag where a spec needs to do something special to create the instance
  (beta, used only for PortForwardManager)
* https://github.com/apache/brooklyn-server/commit/a75e8f7d1b02a8e094f631a82e1cc6ddc4340fec add catalog info to the object returned via the legacy `/v1/locations` endpoint to facilitate switching to `/v1/catalog/locations` in future, and notes

Builds on #62.

Note also that this breaks backwards compatibility if someone *implements* `LocationResolver`.  A note should be added in the release notes when this is merged.

/cc @grkvlt does clocker provide a `LocationResolver`?  if so it will need changed to return a `LocationSpec` not a `Location` instance.  if it is not a straightforward change ping me and I can help; this PR adds a "special constructor" facility which might apply if a `LocationSpec` doesn't work there

### DIFF
--- a/api/src/main/java/org/apache/brooklyn/api/location/LocationRegistry.java
+++ b/api/src/main/java/org/apache/brooklyn/api/location/LocationRegistry.java
@@ -24,16 +24,16 @@ import java.util.NoSuchElementException;
 
 import javax.annotation.Nullable;
 
+import org.apache.brooklyn.api.entity.Entity;
+import org.apache.brooklyn.api.entity.EntitySpec;
+import org.apache.brooklyn.api.mgmt.LocationManager;
 import org.apache.brooklyn.util.guava.Maybe;
-
-import com.google.common.annotations.Beta;
 
 /**
  * The registry of the sorts of locations that brooklyn knows about. Given a
  * {@LocationDefinition} or a {@link String} representation of a spec, this can
  * be used to create a {@link Location} instance.
  */
-@SuppressWarnings("rawtypes")
 public interface LocationRegistry {
 
     /** map of ID (possibly randomly generated) to the definition (spec, name, id, and props; 
@@ -59,24 +59,32 @@ public interface LocationRegistry {
      * <p>
      * The manage parameter is {@link Boolean} so that null can be used to say rely on anything in the flags.
      * 
-     * @since 0.7.0, but beta and likely to change as the semantics of this class are tuned */
-    @Beta
+     * @since 0.7.0, but beta and likely to change as the semantics of this class are tuned 
+     * @deprecated since 0.9.0 use {@link #getLocationSpec(String, Map)} or {@link #getLocationManaged(String, Map)} */
+    @SuppressWarnings("rawtypes")
+    @Deprecated
     public Maybe<Location> resolve(LocationDefinition ld, Boolean manage, Map locationFlags);
     
     /** As {@link #resolve(LocationDefinition, Boolean, Map), with the location managed, and no additional flags,
-     * unwrapping the result (throwing if not resolvable) */
+     * unwrapping the result (throwing if not resolvable) 
+     * @deprecated since 0.9.0 use {@link #getLocationSpec(LocationDefinition, Map)} and then manage it as needed*/
+    @Deprecated
     public Location resolve(LocationDefinition l);
 
     /** Returns a location created from the given spec, which might correspond to a definition, or created on-the-fly.
      * Optional flags can be passed through to underlying the location. 
-     * @since 0.7.0, but beta and likely to change as the semantics of this class are tuned */
-    @Beta
+     * @since 0.7.0, but beta and likely to change as the semantics of this class are tuned 
+     * @deprecated since 0.9.0 use {@link #getLocationSpec(String, Map)} or {@link #getLocationManaged(String, Map)} */
+    @SuppressWarnings("rawtypes")
+    @Deprecated
     public Maybe<Location> resolve(String spec, Boolean manage, Map locationFlags);
-    
+
     /** See {@link #resolve(String, Boolean, Map)}; asks for the location to be managed, and supplies no additional flags,
-     * and unwraps the result (throwing if the spec cannot be resolve) */
+     * and unwraps the result (throwing if the spec cannot be resolve).
+     * @deprecated since 0.9.0 use {@link #getLocationSpec(String)} or {@link #getLocationManaged(String)} */
+    @Deprecated
     public Location resolve(String spec);
-    
+        
     /** Returns true/false depending whether spec seems like a valid location,
      * that is it has a chance of being resolved (depending on the spec) but NOT guaranteed,
      * as it is not passed to the spec;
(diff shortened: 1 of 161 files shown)


ahgittin:
Having looked at https://github.com/apache/incubator-brooklyn/pull/1124 I think @bostko is right, this does not address a leak where a `Location` is created and passed to an `Entity` as a `start` argument.

Feels like we should do further work to accept `LocationSpec` objects -- or strings etc -- as arguments to `start(...)`, and record explicitly whether an entity is creating the locations (eg using new relationships support -- if that is persisted?). /cc @bostko @aledsage @neykov

ahgittin:
(That can be a separate PR.)

ahgittin:
looks like generated code has changed; will push update shortly.

it should be reviewable for the most part without that

ahgittin:
fixed, ready for review

very good if this, and the ui feature it enables https://github.com/apache/brooklyn-ui/issues/18 can make it into 090 /cc @rdowner

ahgittin:
new unrelated failure, not very enlightening logs:

```
2016-03-17 19:52:16,292 INFO  Starting RestMockSimpleEntity{id=fh3xdiFS} on machine SshMachineLocation[localhost:null@jenkins-ubuntu-1404-4gb-560/10.176.64.103:22(id=xOhYUjcS)]
2016-03-17 19:52:26,329 INFO  TESTNG FAILED: "Surefire test" - org.apache.brooklyn.rest.resources.UsageResourceTest.testGetApplicationUsage() finished in 10239 ms
java.lang.AssertionError: task ftMVXJTj not finished expected [true] but found [false]
	at org.apache.brooklyn.rest.resources.UsageResourceTest.waitForTask(UsageResourceTest.java:351)
	at org.apache.brooklyn.rest.resources.UsageResourceTest.createApp(UsageResourceTest.java:309)
	at org.apache.brooklyn.rest.resources.UsageResourceTest.testGetApplicationUsage(UsageResourceTest.java:161)
```

core/src/main/java/org/apache/brooklyn/core/objs/proxy/SpecialBrooklynObjectConstructor.java
```diff
@@ -0,0 +1,80 @@
...
+ *     adding them would be nice to make Locations more like Entities)
+ * <li>Require the appropriate {@link Entity} or {@link Location} type, on a case-by-case basis,
+ *     where the lifecycle is as normal, and the logic for initializing and delegating is done 
+ *     in the init.  (Possibly this is better than the current implementation!?  It is certainly
+ *     simpler than allowing multiple links to existing items, and would be feasible in many
+ *     cases, e.g. a `same-machine-as` type (assuming we made SshMachineEntity an interface);
+ *     but it would not allow choosing an entity type smartly based on location.)
+ */
+@Beta
+public interface SpecialBrooklynObjectConstructor {
```
neykov: Don't like `Special` in the name. This is essentially a factory for the BrooklynObject instances - could just call it  `BrooklynObjectFactory` (or some derivate including spec in the name). Same for the key.
Also deserves an entry in the spec builder, similar to `.impl`.

core/src/main/java/org/apache/brooklyn/core/mgmt/internal/LocationManagerInternal.java
```diff
@@ -19,10 +19,18 @@
...
+import org.apache.brooklyn.api.location.LocationSpec;
 import org.apache.brooklyn.api.mgmt.LocationManager;
+import org.apache.brooklyn.config.ConfigKey;
+import org.apache.brooklyn.core.config.ConfigKeys;
 
 public interface LocationManagerInternal extends LocationManager, BrooklynObjectManagerInternal<Location> {
 
+    /** Indicates that a call to {@link #createLocation(LocationSpec)} should create a location instance not
+     * linked to the management context. This allows clients to peek at the result but not to use it. */
+    public static final ConfigKey<Boolean> CREATE_UNMANAGED = ConfigKeys.newBooleanConfigKey("brooklyn.internal.location.createUnmanaged",
```
neykov: Do we still need this, after introducing the spec resolver?

core/src/main/java/org/apache/brooklyn/core/objs/proxy/SpecialBrooklynObjectConstructor.java
```diff
@@ -0,0 +1,80 @@
...
+    
+    /** Method to implement to create the given type from the given spec. */
+    public <T> T create(ManagementContext mgmt, Class<T> type, AbstractBrooklynObjectSpec<?,?> spec);
+    
+    public static class Config {
+        @Beta @SuppressWarnings("serial")
+        public static ConfigKey<Class<? extends SpecialBrooklynObjectConstructor>> SPECIAL_CONSTRUCTOR =
```
neykov: If this is to be used from java code only this would better be an object instead of the class. But no reason why not pass it in yaml, so can stay as is.

api/src/main/java/org/apache/brooklyn/api/location/LocationRegistry.java
```diff
@@ -86,25 +94,62 @@
...
+    public Maybe<LocationSpec<? extends Location>> getLocationSpec(LocationDefinition ld);
+    /** As {@link #getLocationSpec(String,Map)} where the caller has a {@link LocationDefinition}. */
+    public Maybe<LocationSpec<? extends Location>> getLocationSpec(LocationDefinition ld, Map<?,?> locationFlags);
     
+    /** A combination of {@link #getLocationSpec(String)} then {@link LocationManager#createLocation(LocationSpec)},
+     * mainly for use in tests or specialised situations where a managed location is needed directly.
+     * The caller is responsible for ensuring that the resulting {@link Location} 
+     * is cleaned up, ie removed from management via {@link LocationManager#unmanage(Location)} directly or linking it to 
+     * an {@link Entity} or another {@link Location} which will unmanage it. */
+    public Location getLocationManaged(String spec);
```
neykov: Personal preference to not mixing responsibilities. Now that this API is dealing with specs better leave it to utility classes to do the delegation to `LocationManager`. Same for following methods.

neykov:
Great set of changes. Good to merge.

api/src/main/java/org/apache/brooklyn/api/location/LocationRegistry.java
ahgittin: I thought so too at first but it is otherwise an ugly syntax to create location from spec string (get spec object then pass to location manager) and hundreds of places it should sit.  I think longer term this method should sit on the `TypeRegistry` as a convenience so defer that sweeping change.

core/src/main/java/org/apache/brooklyn/core/objs/proxy/SpecialBrooklynObjectConstructor.java
ahgittin: Yep - when we better understand how this should be used I agree.  For now I thought something lightweight, and beta, is better.

core/src/main/java/org/apache/brooklyn/core/mgmt/internal/LocationManagerInternal.java
ahgittin: Yes we do in a few places, and probably will always, because some info is only available on the Location, not on the LocationSpec, and callers may want to know that.  (I suppose those places could unmanage it explicitly but why the overhead.)

ahgittin:
great comments - thanks.  will merge as soon as https://github.com/apache/brooklyn-ui/pull/17 lands.

api/src/main/java/org/apache/brooklyn/api/location/LocationRegistry.java
```diff
@@ -86,25 +94,62 @@
...
+     * @deprecated since 0.9.0 use {@link #getListOfLocationsManaged(Object)} */
+    @Deprecated
     public List<Location> resolve(Iterable<?> spec);
     
     /** Takes a string, interpreted as a comma-separated (or JSON style, when you need internal double quotes or commas) list;
      * or a list, passed to {@link #resolve(Iterable)}; or null/empty (empty list),
-     * and returns a list of resolved (created and managed) locations */
+     * and returns a list of resolved (created and managed) locations 
+     * @deprecated since 0.9.0 use {@link #getListOfLocationsManaged(Object)} */
+    @Deprecated
```
bostko: Nice.

PR #71, opened 2016-03-17
Disable tests on Windows

neykov:
Windows doesn't support the expected file attributes, disable tests.

because of https://builds.apache.org/view/Brooklyn/job/brooklyn-master-windows/10


sjcorbett:
Looks good.

PR #75, opened 2016-03-22
Add Windows script for enabling CredSSP

bostko:


logging/logback-includes/src/main/resources/brooklyn/logback-logger-excludes.xml
```diff
@@ -42,6 +42,8 @@
         <appender-ref ref="FILE" />
     </logger>
 
+    <logger name="jclouds.wire" level="INFO"/>
```
aledsage: Is this deliberate? Why?

bostko: Currently the debug log is flooded with too jclouds rest calls. I feel like this shouldn't be part of brooklyn.debug.log

aledsage:
LGTM - merging.

PR #76, opened 2016-03-23
BasicSpecParameter: fix compatibility of persisted state

aledsage:
Support old-style persisted state, by implementing readResolve()


aledsage:
@ahgittin @neykov could you take a look please?

neykov:
How does `readResolve` work, is it called from somewhere?
Still think we need a way to bundle transformers with our code for cases like this.

neykov:
Now is see it's used by `xstream`. That's a neat fix. Worth mentioning it in the comments, otherwise good to merge.

aledsage:
Thanks @neykov - added comment saying "Automatically called by xstream (which is used under the covers by XmlMementoSerializer)."

Merging now.

PR #78, opened 2016-03-23
Skip to next bundle on predicate failure.

geomacy:
Without this change, if the predicate fails, the code flow just picks the next predicate to check,
when really we want to skip to checking the next bundle.

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/util/core/osgi/Osgis.java
+++ b/core/src/main/java/org/apache/brooklyn/util/core/osgi/Osgis.java
@@ -184,9 +184,7 @@ public class Osgis {
             for (Bundle b: framework.getBundleContext().getBundles()) {
                 if (symbolicName!=null && !symbolicName.equals(b.getSymbolicName())) continue;
                 if (version!=null && !Version.parseVersion(version).equals(b.getVersion())) continue;
-                for (Predicate<? super Bundle> predicate: predicates) {
-                    if (!predicate.apply(b)) continue;
-                }
+                if (!Predicates.and(predicates).apply(b)) continue;
 
                 // check url last, because if it isn't mandatory we should only clear if we find a url
                 // for which the other items also match


core/src/main/java/org/apache/brooklyn/util/core/osgi/Osgis.java
```diff
@@ -181,11 +181,11 @@ public BundleFinder requiringFromUrl(String url) {
         public List<Bundle> findAll() {
             boolean urlMatched = false;
             List<Bundle> result = MutableList.of();
-            for (Bundle b: framework.getBundleContext().getBundles()) {
+            allBundles: for (Bundle b: framework.getBundleContext().getBundles()) {
                 if (symbolicName!=null && !symbolicName.equals(b.getSymbolicName())) continue;
                 if (version!=null && !Version.parseVersion(version).equals(b.getVersion())) continue;
                 for (Predicate<? super Bundle> predicate: predicates) {
-                    if (!predicate.apply(b)) continue;
+                    if (!predicate.apply(b)) continue allBundles;
```
neykov: Could `Predicates.and` be used instead of the nested loop? Alternatively can you extract it in a separate method to avoid dealing with labels.

ahgittin: +1 i don't like labels here, and in any case i think the inner loop as svet suggests can be replaced by one line

    if (!Predicates.and(predicates).apply(b)) continue;

geomacy: sounds like a fix to me!  will add that soon.

geomacy:
Have rebased and pushed the change suggested.

neykov:
Kicked a new build due to the unrelated failure.
```
Results :

Tests run: 0, Failures: 0, Errors: 0, Skipped: 0

[WARNING] Failed to notify spy hudson.maven.Maven3Builder$JenkinsEventSpy: Invalid object ID 34 iota=49
[WARNING] Failed to notify spy hudson.maven.Maven3Builder$JenkinsEventSpy: Invalid object ID 34 iota=49
[WARNING] Failed to notify spy hudson.maven.Maven3Builder$JenkinsEventSpy: Invalid object ID 27 iota=49
```

geomacy:
Did this not complete?  The build seems to have finished successfully.

geomacy:
Trying a recycle

geomacy:
Reopening

PR #79, opened 2016-03-24
[OSGi] Improve REST API behaviour in Karaf

neykov:
* Use the correct json serializer
  * Make version files accessible
  * Fix collecting version info from bundles - array should be mutable


ahgittin:
looks good

PR #81, opened 2016-03-25
Winrm4j upgrade

aledsage:
Also see the commit "softwareProcess.restart: even on error, set expected=running", and the details in that commit description.


bostko:
Nice changes and cleanups!

PR #82, opened 2016-03-25
Fix DynamicMultiGroup rebind

aledsage:
The `isRebindComplete()` is really hacky, but not sure how better to do this - @ahgittin any advice?

I'd like to merge this PR asap to fix the problem, and then improve on the impl in subsequent PRs. Unless you think the code changes are dangerous!

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/entity/group/DynamicMultiGroupImpl.java
+++ b/core/src/main/java/org/apache/brooklyn/entity/group/DynamicMultiGroupImpl.java
@@ -28,11 +28,15 @@ import javax.annotation.Nullable;
 import org.apache.brooklyn.api.entity.Entity;
 import org.apache.brooklyn.api.entity.EntitySpec;
 import org.apache.brooklyn.api.entity.Group;
+import org.apache.brooklyn.api.mgmt.ha.ManagementNodeState;
 import org.apache.brooklyn.api.sensor.AttributeSensor;
 import org.apache.brooklyn.core.entity.Entities;
 import org.apache.brooklyn.feed.function.FunctionFeed;
 import org.apache.brooklyn.feed.function.FunctionPollConfig;
 import org.apache.brooklyn.util.collections.MutableMap;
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/apache/brooklyn/entity/group/DynamicMultiGroupImpl.java
```diff
@@ -119,7 +125,36 @@ public void rebind() {
...
+                    LOG.debug("Connecting scanner for {}", DynamicMultiGroupImpl.this);
+                    connectScanner();
+                }
+                private boolean isRebindComplete() {
+                    // TODO Want to determine if finished rebinding (either success or fail is fine).
+                    // But not a clean way to do this that works for both unit tests and live server?!
+                    //  * In RebindTestFixtureWithApp tests, mgmt.getHighAvailabilityManager().getNodeState()
+                    //    always returns INITIALIZING.
+                    //  * The rebind metrics is a hack, and feels very risky for HOT_STANDBY nodes that 
+                    //    may have executed the rebind code multiple times.
```
neykov: `!mgmt().getRebindManager().isAwaitingInitialRebind()`should do the job.
Tests might need an additional `((LocalManagementContext)getManagementContext()).noteStartupComplete();`
See `HaHotCheckTest` for details.

neykov:
LGMT

core/src/main/java/org/apache/brooklyn/entity/group/DynamicMultiGroupImpl.java
aledsage: Ah - the `noteStartupComplete()` is what is missing I think. All the sensible looking methods returned "initialising" in the unit tests. Longer term, it would be good to ensure the test code is more like the production code path!

I'll merge as-is. We can look at this again next week. Thanks for the fast review @neykov !

johnmccabe:
LGTM - checked with a clocker running Mesos with multiple managed tasks and an unmanaged tasks, all rebound successfully

PR #84, opened 2016-03-26
WinRM Parametrize number of retries of network failures

bostko:
For some networks retries could be useful if user is able to configure high number of tries for connection failures.
Requires publishing new version of winrm4j.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -151,7 +151,7 @@
         <jax-rs-api.version>2.0.1</jax-rs-api.version>
         <maxmind.version>0.8.1</maxmind.version>
         <jna.version>4.0.0</jna.version>
-        <winrm4j.version>0.3.4</winrm4j.version>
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -151,7 +151,7 @@
         <jax-rs-api.version>2.0.1</jax-rs-api.version>
         <maxmind.version>0.8.1</maxmind.version>
         <jna.version>4.0.0</jna.version>
-        <winrm4j.version>0.3.4</winrm4j.version>
+        <winrm4j.version>0.3.5</winrm4j.version>
```
aledsage: winrm4j 0.3.5 isn't released yet, I believe - next time can you please make that *very* clear in the pull request description, in case anyone merges it without checking!

aledsage:
winrm4j 0.3.5 is now available; merging.

PR #85, opened 2016-03-26
CLI args: support catalogAdd with multiple URIs

aledsage:


server-cli/src/test/java/org/apache/brooklyn/cli/CliTest.java
```diff
@@ -59,6 +60,8 @@
 import org.apache.brooklyn.core.entity.factory.ApplicationBuilder;
 import org.apache.brooklyn.core.entity.trait.Startable;
 import org.apache.brooklyn.core.location.SimulatedLocation;
+import org.apache.brooklyn.core.mgmt.internal.LocalManagementContext;
+import org.apache.brooklyn.core.mgmt.internal.LocalManagementContextRegistry;
```
johnmccabe: Did you miss this out from the commit?

aledsage: Yes, doh! Have added it and pushed again.

johnmccabe:
Built and run Brooklyn and Clocker and was able to start with 5 catalogs loading as expected, lovely stuff, thanks @aledsage.

Will pull updated commit just to sanity check the unit tests now

johnmccabe:
LGTM checked catalogInitial with catalogAdd, working as expected.

johnmccabe:
Also checked the some persistance behaviours.
- Restarting with passing an unchanged bom is ok
- Retarting with a newer or older version than persisted is ok, the new/old versions get added and all expected versions are visible in the catalog
- Restarting with  an existing version whose contents have changed (ie change the name) puts the server in an error state - expected
- Restarting above errored system with the name reverted is ok, starts successfully
- Alternatively, restarting the errored system with --catalogForce sees it start successfully and the version you've altered gets added to the catalog overriding the persisted version (all other versions are still present)

aledsage:
Thanks @johnmccabe , including for the through testing of restart scenarios! Merging now.

PR #86, opened 2016-03-28
Windows enable credssp script - on exception Write-Error

bostko:


rest/rest-resources/src/main/java/org/apache/brooklyn/rest/resources/CloudResource.java
```diff
@@ -0,0 +1,7 @@
+package org.apache.brooklyn.rest.resources;
+
+/**
+ * Created by valentin on 28/03/16.
+ */
+public class CloudResource {
```
neykov: These couple of files don't seem related to the PR.

neykov:
Other than that looks good.

aledsage:
Thanks @bostko - merging now.

PR #88, opened 2016-03-28
JcloudsSshMachineLocation: avoid NPE on rebind

aledsage:


aledsage:
@neykov can you take a look at this (trivial change) please? I hit it when testing rebind to a customer's historic state.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsSshMachineLocation.java
```diff
@@ -162,8 +162,12 @@ public void init() {
...
     public void rebind() {
         super.rebind();
-        ComputeServiceContext context = jcloudsParent.getComputeService().getContext();
-        runScriptFactory = context.utils().injector().getInstance(RunScriptOnNode.Factory.class);
+        
+        if (jcloudsParent != null) {
+            // can be null on rebind, if location has been "orphaned" somehow
+            ComputeServiceContext context = jcloudsParent.getComputeService().getContext();
+            runScriptFactory = context.utils().injector().getInstance(RunScriptOnNode.Factory.class);
+        }
```
neykov: Won't this cause NPE down the line when trying to execute scripts?

aledsage: @neykov Yes, it would - if the location were ever used. I'll add a log message in the else block.

The current behaviour is that Brooklyn fails to rebind because of the NPE, unless you configure it to ignore errors. It happened for some persisted state that had an "orphaned" location in its persisted state.

The ideal would be that locations are entities, and that the location can thus be marked as "on-fire". But that's much longer term.

neykov: Isn't the rebind failing a good thing in this case - it shows that something is wrong and there will be runtime errors if not addressed.
Can we address the problem in another way? For example:
  * why are parent locations not serialized - is this a bug?
  * serialize parent locations even if not managd

aledsage:
@neykov do you think this is good to merge?

neykov:
Good to merge, but still interested in what you think about my last comment.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsSshMachineLocation.java
aledsage: The problem is the bug happened a long time ago, to persist these locations (or more likely not delete them when the parent was deleted). I'm hoping that's fixed but it might not be.

PR #89, opened 2016-03-28
Windows @ GCE: exec startup script to enable WinRM

aledsage:
Longer term, we should look at how better to configure a startup script. Different clouds within jclouds seem to use userMetadata in different ways, making our code fiddly.

Note this change isn't enough to get Windows@GCE working. We also need to write some code for generating + retrieving a Windows password. See https://cloud.google.com/compute/docs/instances/automate-pw-generation

If testing, the following location config is also important for users, to avoid jclouds waiting for port 22 to be reachable (and thus failing): `templateOptions: { autoCreateKeyPair: false }`

### DIFF
--- a/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
+++ b/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
@@ -1536,10 +1536,45 @@ public class JcloudsLocation extends AbstractCloudMachineProvisioningLocation im
         }
         TemplateOptions options = template.getOptions();
 
+        // For windows, we need a startup-script to be executed that will enable winrm access.
+        // If there is already conflicting userMetadata, then don't replace it (and just warn).
+        // TODO this injection is hacky and (currently) cloud specific.
         boolean windows = isWindows(template, config);
         if (windows) {
-            if (!(config.containsKey(JcloudsLocationConfig.USER_METADATA_STRING) || config.containsKey(JcloudsLocationConfig.USER_METADATA_MAP))) {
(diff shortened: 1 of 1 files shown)


aledsage:
@andreaturli @bostko @neykov can you review please? Would like this available tomorrow (Tuesday) morning if possible!

neykov:
LGTM

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -1536,10 +1536,45 @@ public Template buildTemplate(ComputeService computeService, ConfigBag config) {
...
+                    Map<String, String> metadataMapReplacement = MutableMap.of(startupScriptKey, initScript);
+                    config.put(USER_METADATA_MAP, metadataMapReplacement);
+                    LOG.debug("Adding startup-script to enable WinRM for Windows VM on "+provider);
+                }
+            } else {
+                // For AWS and vCloudDirector, we just set user_metadata_string.
+                // For Azure-classic, there is no capability to execute a startup script.
+                boolean userMetadataString = config.containsKey(JcloudsLocationConfig.USER_METADATA_STRING);
+                boolean userMetadataMap = config.containsKey(JcloudsLocationConfig.USER_METADATA_MAP);
+                if (!(userMetadataString || userMetadataMap)) {
```
bostko: Isn't it better to name it `hasMetadataString` and `hasMetadataMap`

aledsage: Yes, it would have been (I left the names as-is). Feel free to rename that in a new PR.

PR #90, opened 2016-03-29
reduce number of events and log

aledsage:


core/src/main/java/org/apache/brooklyn/core/mgmt/internal/LocalSubscriptionManager.java
```diff
@@ -143,6 +143,9 @@ public void run() {
                         if (s.eventFilter!=null && !s.eventFilter.apply(event))
                             return;
                         try {
+                            int count = s.eventCount.incrementAndGet();
+                            if (count > 0 && count % 1000 == 0) LOG.debug("{} events for subscriber {}", count, s);
```
ahgittin: will always be > 0 due to line above

ahgittin:
LGTM - logging is nice

PR #91, opened 2016-03-30
BYON and MultiLocation resolvers: use LocationSpec for children

aledsage:


ahgittin:
nice work -- merging, to master and 090 branch.  @rdowner ready for release after this i think.

PR #92, opened 2016-03-30
Fully initialize JAAS before launching jsgui (in tests)

neykov:


ahgittin:
helps, but doesn't solve it entirely.  new PR coming with more fixes, and capturing outstanding issues.

PR #93, opened 2016-03-30
a few more OSGi fixes

ahgittin:
doesn't solve anything, but opening this to track outstanding problems

### DIFF
--- a/karaf/features/src/main/feature/feature.xml
+++ b/karaf/features/src/main/feature/feature.xml
@@ -27,6 +27,12 @@
 
     <!-- TODO how to add central.maven.org repo? -->
     <repository>http://central.maven.org/maven2/org/apache/jclouds/karaf/jclouds-karaf/1.9.2/jclouds-karaf-1.9.2-features.xml</repository>
+<!--
+TODO the above link works for loading it, at compile time, as in the file below is then installed into target/assembly/system/
+BUT the assembly's variant of *this* feature file keeps the link above which FAILS if the server is not online.
+it should contain the entry below *instead of* the entry above:
+    <repository>mvn:org.apache.jclouds.karaf/jclouds-karaf/1.9.2/xml/features</repository>
+-->
     <repository>mvn:org.apache.cxf.karaf/apache-cxf/${cxf.version}/xml/features</repository>
 
     <feature name="swagger" version="${swagger.version}" description="Swagger Annotations+Core+JAXRS+Models">
--- a/karaf/apache-brooklyn/src/main/filtered-resources/etc/branding.properties
+++ b/karaf/apache-brooklyn/src/main/filtered-resources/etc/branding.properties
@@ -26,7 +26,7 @@ welcome = \
 \u001B[33m |_.__/|_|  \\___/ \\___/|_|\\_\\_|\\__, |_| |_|  \u001B[0m\n\
 \u001B[33m                               |___/              \u001B[0m\n\
 \u001B[33m                                             \u001B[0m\n\
-\u001B[33m http://brooklyn.incubator.apache.org \u001B[0m\n\
+\u001B[33m http://brooklyn.apache.org \u001B[0m\n\
 \u001B[33m (version ${project.version})\u001B[0m\n\
 \u001B[33m\u001B[0m\n\
 \u001B[33mHit '\u001B[1m<tab>\u001B[0m' for a list of available commands\u001B[0m\n\
--- a/rest/rest-server/src/main/webapp/WEB-INF/web.xml
+++ b/rest/rest-server/src/main/webapp/WEB-INF/web.xml
@@ -38,16 +38,8 @@
         <url-pattern>/*</url-pattern>
     </filter-mapping>
 
-    <!-- Brooklyn REST is usually run as a filter so static content can be placed in a webapp
-         to which this is added; to run as a servlet directly, replace the filter tags 
-         below (after the comment) with the servlet tags (commented out immediately below),
-         (and do the same for the matching tags at the bottom)
-        <servlet>
-            <servlet-name>Brooklyn REST API v1 Servlet</servlet-name>
-            <servlet-class>com.sun.jersey.spi.container.servlet.ServletContainer</servlet-class>
-     -->
     <servlet>
-        <servlet-name>Brooklyn REST API v1 Filter</servlet-name>
+        <servlet-name>Brooklyn REST API v1 Servlet</servlet-name>
         <servlet-class>org.apache.cxf.jaxrs.servlet.CXFNonSpringJaxrsServlet</servlet-class>
 
         <init-param>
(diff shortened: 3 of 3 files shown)


karaf/features/src/main/feature/feature.xml
```diff
@@ -27,6 +27,12 @@
 
     <!-- TODO how to add central.maven.org repo? -->
     <repository>http://central.maven.org/maven2/org/apache/jclouds/karaf/jclouds-karaf/1.9.2/jclouds-karaf-1.9.2-features.xml</repository>
+<!--
+TODO the above link works for loading it, at compile time, as in the file below is then installed into target/assembly/system/
+BUT the assembly's variant of *this* feature file keeps the link above which FAILS if the server is not online.
+it should contain the entry below *instead of* the entry above:
+    <repository>mvn:org.apache.jclouds.karaf/jclouds-karaf/1.9.2/xml/features</repository>
```
ahgittin: @neykov @m4rkmckenna @cmoh any ideas how to solve this problem?

ahgittin:
Two problems seen so far with OSGi:
* the jclouds issue (commented above)
* if brooklyn is configured to allow *anyone* to log in, the LoginModule doesn't seem to work; it looks like `LoginModule` is only activated if basic auth is suppled by the HTTP.  so if an auth-less request comes in, the server returns 401 unauthorized without even calling to Brooklyn code.  (tried w the `BrooklynJavascriptGuiLauncher` but i think the problem might also exist w karaf startup, ie anything using CXF / LoginModule.)  we should probably hook in at the `SecurityHandler` level, or as a custom filter?

Need someone who knows more about OSGi to look at these! /cc @CMoH @neykov 

I'm also having trouble seeing the `catalog.bom` files load but I know my way around that better so can investigate there.

karaf/features/src/main/feature/feature.xml
m4rkmckenna: The repository tag needs to contain a mvn URL (GAV format)
Few options 
1. Add maven central to `org.ops4j.pax.url.mvn.repositories.cfg` **NOTE** This will require internet access @ runtime so karaf can resolve it
2. Ensure that all jclouds dependencies are packaged in the karaf system directory @ compile (so internet access is not required)

I can't verify the above at the minute as my workspace is in a mess :smile:

ahgittin:
Re auth -- I also notice that in karaf mode I can access static content without auth when auth is required.  I don't understand why as the login module is specified in the `web.xml` but this probably shouldn't be allowed.

ahgittin:
another topic -- what's involved to apply our slf4j log levels?  i see reams of logging on every http request.

m4rkmckenna:
@ahgittin RE logging from karaf console `log:set <level> <log>` eg `log:set WARN org.apache.brooklyn`

Or edit the logging config file under karf/etc can't remember the name of it off the top of my head

ahgittin:
@m4rkmckenna wondering can we re-use the existing `logback.xml` items, and what changes will we need (eg suppress the `INFO` on every web request)

m4rkmckenna:
@ahgittin We would need to port the existing logback into https://github.com/apache/brooklyn-server/blob/master/karaf/apache-brooklyn/src/main/resources/etc/org.ops4j.pax.logging.cfg

Should just be simple

neykov:
* re jclouds repo - thanks @m4rkmckenna, that's exactly the info I was missing at the time
* re "if brooklyn is configured to allow anyone to log in" - the LoginModule is only ever accessed to authenticate a pair of credentials, the rest is done in other layers. See if https://github.com/apache/brooklyn-server/pull/94 addresses your problems.
* re " I also notice that in karaf mode I can access static content without auth" - can you confirm this, could have already been logged in another tab.

neykov:
btw checking out if we can re-use logback's config - should be possible to replace the log4j configuration used by default.

neykov:
RE logging - we can replace the backend logger from log4j to logback by:
  * override `etc/startup.properties` replacing `pax-logging-service` for `pax-logging-logback`
  * in `org.ops4j.pax.logging.cfg` replace with `org.ops4j.pax.logging.logback.config.file=${karaf.base}/etc/logback.xml`
  * add `logback.xml` to dist
  * add `pax-logging-logback` as a dependency in the pom so it gets included in the dist

BUT then we need to put any of the included files from `logback.xml` in the etc folder. Jars containing overrides will be ignored. Might be possible to have the defaults (i.e. logback-includes) in a bundle by making it an extension to pax-logging, but needs to be confirmed.

SO do we really need it? Why not go with log4j other than familiarity with logback.

ahgittin:
@m4rkmckenna @CMoH what are your views on logging?

the existing logback config has a few nice features including rollover configuration and good categories for jclouds and nicely formatted messages and custom filenames -- and integration so you get good logging during tests and running in the IDE (without which the IDE can be painful) -- so if we can support it that's definitely my preference, but only if it doesn't feel too odd.

otoh i don't want to waste time fighting karaf so if log4j is the de facto standard in osgi or logback completely alien then we should switch.  but the existence of `pax-logging-logback` suggests it might not be a completely strange thing to do.

@neykov regarding the point that the current `logback.xml` in the dist references e.g. `logging/some-project/logback-partial.xml` that makes me wonder how osgi logging is meant to work:

* does a master `logback.xml` define categories applying to every bundle (eg configured in `o.o.p.logging.logback.config.file`)?

* if this references other files what is the search path?  (an extension bundle is a great idea, if it works.  if so it doesn't feel too difficult.  alternately we could use maven to copy the items from that bundle and put them in the right structure in the dist; that might be preferable in any case.)

* can we define additional logging config on a per-bundle basis?  (i'm not sure we'd do that even if we could, but maybe it would show us a cleaner way.)

running in the IDE w nice logging is quite important i think -- debug gets painful otherwise -- so that should be a consideration.

neykov:
> does a master logback.xml define categories applying to every bundle
It's dist-wide.

> what is the search path?
It's using the `pax-logging` class loader and delegating to standard logback code for parsing the file (see [here](https://github.com/ops4j/org.ops4j.pax.logging/blob/master/pax-logging-logback/src/main/java/org/ops4j/pax/logging/logback/internal/PaxLoggingServiceImpl.java#L317)). An extension bundle could work, depending on how it resolves relative paths (initial path is disk-based absolute path)

> can we define additional logging config on a per-bundle basis
Don't think so.

m4rkmckenna:
@ahgittin Re [Auth / Unauthenticated Access to static content] 
I couldn't re-produce ... Tested using a chrome incognito tab and was prompted for auth when trying to access http://localhost:8081/assets/css/brooklyn.css

ahgittin:
@m4rkmckenna k maybe something was funny on my system.  did you try curl without auth?  browsers do funny things sometimes, i never trust em.

karaf/features/src/main/feature/feature.xml
neykov: Fixed in https://github.com/apache/brooklyn-server/pull/98

PR #94, opened 2016-03-31
Fix noConsoleSecurity option

neykov:
Also skip security by default for tests


neykov:
btw forgot to mention that this fixes the master build tests

rest/rest-server-jersey/src/test/java/org/apache/brooklyn/rest/BrooklynRestApiLauncherTest.java
```diff
@@ -36,17 +36,17 @@
 
 public class BrooklynRestApiLauncherTest extends BrooklynRestApiLauncherTestFixture {
 
-    @Test
+    @Test(enabled = false) // to be deleted
```
sjcorbett: Can these tests just be deleted?

neykov: Yes, but doesn't really matter because already deleted in https://github.com/apache/brooklyn-server/pull/95.

sjcorbett:
@neykov Thanks.

PR #95, opened 2016-03-31
Remove jersey compatibility

neykov:
My favourite kind of PRs -  0 additions and 9,808 deletions.

Now that 0.9.0 is branched we can remove the jersey compatibility.


neykov:
Build failures fixed in #94. Once merged I expect some conflicts here, so will fix and trigger a rebuild.

ahgittin:
delicious!  however is the passwordless login mentioned in #92 fixed?  not strictly a pre-req for this as we've already switched to cxf as the default but it is a high priority.

neykov:
yes, completely fixed in #94

sjcorbett:
@ahgittin @neykov Are you ready for this to be merged?

neykov:
Good to go. Thanks!

PR #97, opened 2016-03-31
Fix persisting localhost

aledsage:
When running `org.apache.brooklyn.qa.load.LoadTest`, I hit exceptions about serializing localhost machine location. I didn't try rebinding, but I presume the localhost machine's location object would not have been deserialised correctly (or not been available at all).

This PR fixes the two exceptions I hit.


ahgittin:
yeah the `mutexSupport` is static so not sure why we go to such great lengths to pass it around; LGTM

PR #98, opened 2016-04-03
Fix build on Windows

neykov:
Build fails, trying to resolve the http url as a file system path.


karaf/features/src/main/feature/feature.xml
```diff
@@ -26,7 +26,7 @@
     <repository>mvn:org.apache.karaf.features/spring/${karaf.version}/xml/features</repository>
 
     <!-- TODO how to add central.maven.org repo? -->
```
bostko: Is this comment still needed?

neykov: Thanks for pointing it out, removed.

geomacy:
👍 looks good to me

neykov:
Thanks @bostko, @geomacy. Merging.

PR #100, opened 2016-04-05
Support removal of catalog entries from bundle BOM when the bundle is stopped.

geomacy:
This change follows on from https://github.com/apache/brooklyn-server/pull/80.

With this change any catalog members that were added by scanning the bundle's
catalog.bom when the bundle started will be removed upon stopping the bundle.

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/catalog/internal/CatalogBomScanner.java
+++ b/core/src/main/java/org/apache/brooklyn/core/catalog/internal/CatalogBomScanner.java
@@ -25,14 +25,11 @@ import org.apache.brooklyn.api.catalog.BrooklynCatalog;
 import org.apache.brooklyn.api.catalog.CatalogItem;
 import org.apache.brooklyn.api.mgmt.ManagementContext;
 import org.apache.brooklyn.util.collections.MutableList;
-import org.apache.brooklyn.util.exceptions.CompoundRuntimeException;
 import org.apache.brooklyn.util.exceptions.Exceptions;
 import org.apache.brooklyn.util.stream.Streams;
 import org.apache.brooklyn.util.yaml.Yamls;
 import org.osgi.framework.Bundle;
-import org.osgi.framework.BundleContext;
 import org.osgi.framework.BundleEvent;
-import org.osgi.framework.FrameworkUtil;
 import org.osgi.framework.ServiceReference;
 import org.osgi.util.tracker.BundleTracker;
 import org.slf4j.Logger;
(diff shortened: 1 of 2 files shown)


geomacy:
Note this also addresses some of the comments from the review of https://github.com/apache/brooklyn-server/pull/80, namely:

https://github.com/apache/brooklyn-server/pull/80#discussion_r57437307
https://github.com/apache/brooklyn-server/pull/80#discussion_r57437307

The others remain pending.

core/src/main/java/org/apache/brooklyn/core/catalog/internal/CatalogBomScanner.java
```diff
@@ -100,54 +103,82 @@ public ManagementContext getManagementContext() {
...
+                    LOG.warn("Caught {} unloading {} {} from catalog", new String [] {
+                        e.getMessage(), item.getSymbolicName(), item.getVersion()
+                    });
+                    exceptions.add(e);
+                }
+            }
 
-            return bundle.getBundleId();
+            if (0 < exceptions.size()) {
+                throw new CompoundRuntimeException(
```
neykov: What happens when an exception is thrown here?
Shouldn't block bundle unloading because of failure to delete the catalog items. Could be removed by the user already.

geomacy: Hm, I guess I have been assuming that this will just result in a message being logged by Karaf as opposed to it causing a failure - I will have to test that out.

neykov: If it's just for the message, then better ignore it as it's already been logged? Otherwise the exception logs will get duplicated.

neykov:
Looks good, but needs discussion on the bundle loading comment above.

core/src/main/java/org/apache/brooklyn/core/catalog/internal/CatalogBomScanner.java
```diff
@@ -100,54 +103,75 @@ public ManagementContext getManagementContext() {
...
+            }
+            LOG.debug("Unloading catalog BOM entries from {} {} {}", bundleIds(bundle));
+            final BrooklynCatalog catalog = getManagementContext().getCatalog();
+            for (CatalogItem<?, ?> item : items) {
+                LOG.debug("Unloading {} {} from catalog", item.getSymbolicName(), item.getVersion());
+
+                try {
+                    catalog.deleteCatalogItem(item.getSymbolicName(), item.getVersion());
+                } catch (Exception e) {
+                    // Gobble exception to avoid possibility of causing problems stopping bundle.
```
neykov: add `Exceptions.propagateIfFatal(e)` for the case of `InterruptedException`

geomacy: done in 1844634

neykov:
Thanks @geomacy. Can you (at least partially) squash and I'll merge it.

geomacy:
hi @neykov that's squashed with previous commit and pushed in 47421cb

PR #101, opened 2016-04-05
Don't force binding to 127.0.0.1 when using AnyoneSecurityProvider

neykov:
Keeps backwards compatibility - using AnyoneSecurityProvider wouldn't force brooklyn server to bind to loopback only, but then changed to being treated same as noSecurityProvider. Revert to initial behaviour.

/cc @johnmccabe


johnmccabe:
@neykov I'm seeing it bind to all interfaces as expected now, but we still see the loopback logged to the console. I can't recall seeing it log that its bound to all interfaces in the past but it might be nice to show that for correctness.

neykov:
@johnmccabe It's meant as a simple copy&paste aid (or click for terminals that support it). So better to use a real address.

PR #102, opened 2016-04-05
pass brooklynConfig when making call to getLocationSpec

johnmccabe:
Fixes [BROOKLYN-246](https://issues.apache.org/jira/browse/BROOKLYN-246)

Was failing for byon locations where the host wasn't specified in the type.

For example the byon catalog location as shown in `testByonLocationHostsInConfig` had been failing due to `extractConfig` in the `ByonLocationResolver` throwing an `IllegalArgumentException` as no hosts were found.


johnmccabe:
fyi @aledsage @rdowner @neykov

aledsage:
LGTM - very nice when it's a one-line fix plus tests! Merging.

PR #103, opened 2016-04-05
Fix BROOKLYN-249

aledsage:
See more detailed descriptions in each commit message.

### DIFF
--- a/camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/JcloudsRebindWithExternalConfigTest.java
+++ b/camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/JcloudsRebindWithExternalConfigTest.java
@@ -1,0 +1,139 @@
+package org.apache.brooklyn.camp.brooklyn;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import java.io.File;
+import java.util.Map;
+
+import org.apache.brooklyn.api.entity.Application;
+import org.apache.brooklyn.api.entity.Entity;
+import org.apache.brooklyn.api.entity.EntitySpec;
+import org.apache.brooklyn.camp.brooklyn.spi.creation.CampTypePlanTransformer;
+import org.apache.brooklyn.core.config.external.InPlaceExternalConfigSupplier;
+import org.apache.brooklyn.core.entity.trait.Startable;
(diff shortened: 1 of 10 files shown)


core/src/test/java/org/apache/brooklyn/location/byon/ByonLocationResolverRebindTest.java
```diff
@@ -0,0 +1,69 @@
...
+import org.apache.brooklyn.api.location.LocationSpec;
+import org.apache.brooklyn.api.location.MachineLocation;
+import org.apache.brooklyn.api.location.MachineProvisioningLocation;
+import org.apache.brooklyn.core.mgmt.rebind.RebindTestFixtureWithApp;
+import org.apache.brooklyn.location.ssh.SshMachineLocation;
+import org.testng.annotations.Test;
+
+import com.google.common.collect.ImmutableMap;
+
+public class ByonLocationResolverRebindTest extends RebindTestFixtureWithApp {
```
aledsage: This should have been included in https://github.com/apache/brooklyn-server/pull/91, but was accidentally missed out!

Not quite appropriate to include it in this commit, but no big deal in my opinion.

neykov:
Looks good. Are the external properties in locations already covered by tests (don't see related changes in last commit)?

aledsage:
Existing tests include `ExternalConfigYamlTest.testExternalisedLocationConfigReferencedFromYaml` and `testExternalisedLocationConfigSetViaProvisioningPropertiesReferencedFromYaml`.

We could do with more tests for other specific types of location. For example, I worry that a BYON won't correctly resolve external config.

camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/JcloudsRebindWithExternalConfigTest.java
```diff
@@ -0,0 +1,139 @@
+package org.apache.brooklyn.camp.brooklyn;
```
sjcorbett: This is missing the ASF license header.

sjcorbett: I'll just push a fix to master.

PR #105, opened 2016-04-07
Fixed command for installing python via apt-get and yum by using sudo

jongyoul:
When I use a "Template 2: Bash Web Server" with non-root user and password-less sudo, I've got a error because I don't have any python installed. We should use sudo command for installing python.


aledsage:
@jongyoul thanks for fixing!

I presume you have already signed the apache Contributor License Agreement, when working on the Apache Zeppelin project? Therefore happy to merge.

jongyoul:
@aledsage I've already signed CLA. Thanks for merging it. I'm trying to find the reason why CI fails.

PR #106, opened 2016-04-07
Add catalog.bom application white/blacklists.

geomacy:
Note this PR depends on https://github.com/apache/brooklyn-server/pull/100 and should be merged after it.

It is desirable to have a config time mechanism to control what
bundles are allowed to present applications in the catalog and
what aren't.

This change adds a white/blacklist mechanism to the catalog.bom
bundle scanning. Each list takes the form of a CSV whitelist of
regexes to match against bundle symbolic ids.  A bundle will be
allowed to add an application (template) to the catalog if its
bundle id matches one of the regexes on the whitelist, and also
does not match one of the regexes on the blacklist.

The configuration keys and default values for the lists are:

brooklyn.catalog.osgi.application.whitelist=.*
brooklyn.catalog.osgi.application.blacklist=

(i.e. by default accept all bundles).

these values are specified in the file org.apache.brooklyn.core.catalog.bomscanner.cfg

Note that two 'default'  example applications are supplied in the bom
for brooklyn-software-base.

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/catalog/internal/CatalogBomScanner.java
+++ b/core/src/main/java/org/apache/brooklyn/core/catalog/internal/CatalogBomScanner.java
@@ -91,7 +91,7 @@ public class CatalogBomScanner {
 
     public void setWhiteList(String whiteListText) {
         LOG.debug("Setting whiteList to ", whiteListText);
-        this.whiteList = Strings.parseCsv(whiteListText, ",");
+        this.whiteList = Strings.parseCsv(whiteListText);
     }
 
     public List<String> getBlackList() {
@@ -104,7 +104,7 @@ public class CatalogBomScanner {
 
     public void setBlackList(String blackListText) {
         LOG.debug("Setting blackList to ", blackListText);
-        this.blackList = Strings.parseCsv(blackListText, ",");
+        this.blackList = Strings.parseCsv(blackListText);
     }
 
     public class CatalogPopulator extends BundleTracker<Iterable<? extends CatalogItem<?, ?>>> {
@@ -170,7 +170,7 @@ public class CatalogBomScanner {
             } catch (Exception e) {
                 Exceptions.propagateIfFatal(e);
                 LOG.warn(Strings.join(new String[] {
-                    "Failed to remove", item.getSymbolicName(), item.getVersion(), "{} {} from catalog"
+                    "Failed to remove", item.getSymbolicName(), item.getVersion(), "from catalog"
                 }, " "), e);
             }
         }
@@ -200,7 +200,9 @@ public class CatalogBomScanner {
             return catalogItems;
         }
 
-        private Iterable<? extends CatalogItem<?, ?>> removeAnyApplications(Iterable<? extends CatalogItem<?, ?>> catalogItems) {
+        private Iterable<? extends CatalogItem<?, ?>> removeAnyApplications(
+            Iterable<? extends CatalogItem<?, ?>> catalogItems) {
+
             List<CatalogItem<?, ?>> result = MutableList.of();
 
             for (CatalogItem<?, ?> item: catalogItems) {
--- a/core/src/main/java/org/apache/brooklyn/core/catalog/internal/CatalogBomScanner.java
+++ b/core/src/main/java/org/apache/brooklyn/core/catalog/internal/CatalogBomScanner.java
@@ -21,12 +21,12 @@ package org.apache.brooklyn.core.catalog.internal;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Iterables;
-import org.apache.brooklyn.api.catalog.BrooklynCatalog;
 import org.apache.brooklyn.api.catalog.CatalogItem;
 import org.apache.brooklyn.api.mgmt.ManagementContext;
 import org.apache.brooklyn.util.collections.MutableList;
 import org.apache.brooklyn.util.exceptions.Exceptions;
 import org.apache.brooklyn.util.stream.Streams;
+import org.apache.brooklyn.util.text.Strings;
 import org.apache.brooklyn.util.yaml.Yamls;
 import org.osgi.framework.Bundle;
 import org.osgi.framework.BundleEvent;
@@ -34,6 +34,7 @@ import org.osgi.framework.ServiceReference;
 import org.osgi.util.tracker.BundleTracker;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.yaml.snakeyaml.DumperOptions;
 import org.yaml.snakeyaml.Yaml;
 
 import java.io.IOException;
@@ -42,17 +43,26 @@ import java.net.URL;
 import java.util.List;
 import java.util.Map;
 
+import static org.apache.brooklyn.api.catalog.CatalogItem.CatalogItemType.TEMPLATE;
+
 public class CatalogBomScanner {
+
+    private final String ACCEPT_ALL_BY_DEFAULT = ".*";
 
     private static final Logger LOG = LoggerFactory.getLogger(CatalogBomScanner.class);
     private static final String CATALOG_BOM_URL = "catalog.bom";
     private static final String BROOKLYN_CATALOG = "brooklyn.catalog";
     private static final String BROOKLYN_LIBRARIES = "brooklyn.libraries";
 
+    private List<String> whiteList = ImmutableList.of(ACCEPT_ALL_BY_DEFAULT);
+    private List<String> blackList = ImmutableList.of();
+
     private CatalogPopulator catalogTracker;
 
     public void bind(ServiceReference<ManagementContext> managementContext) throws Exception {
-        LOG.debug("Binding management context");
+        LOG.debug("Binding management context with whiteList [{}] and blacklist [{}]",
+            Strings.join(getWhiteList(), "; "),
+            Strings.join(getBlackList(), "; "));
         catalogTracker = new CatalogPopulator(managementContext);
     }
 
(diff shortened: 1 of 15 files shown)


core/src/main/java/org/apache/brooklyn/core/catalog/internal/CatalogBomScanner.java
```diff
@@ -74,6 +80,17 @@ public void unbind(ServiceReference<ManagementContext> managementContext) throws
 
     public class CatalogPopulator extends BundleTracker<Iterable<? extends CatalogItem<?, ?>>> {
 
+        private final String BROOKLYN_PACKAGE_REGEX = "org.apache.brooklyn.*";
+
+        private final ConfigKey<List<String>> WHITELIST =
+            ConfigKeys.newConfigKey(new TypeToken<List<String>>(){}, "brooklyn.catalog.osgi.application.whitelist",
+                "CSV Whitelist of patterns to match bundle symbolic id if bundle is to be permitted to add " +
+                    "applications (templates) to the catalog", ImmutableList.of(BROOKLYN_PACKAGE_REGEX));
```
neykov: If I am evaluating Brooklyn and load my own bundle in the container I'd expect that its `catalog.bom` is picked up out of the box. I think Brooklyn should accept all bundles, letting downstream distributions set their own policy.

geomacy: That's a very good point. Will update.

core/src/main/java/org/apache/brooklyn/core/catalog/internal/CatalogBomScanner.java
```diff
@@ -74,6 +80,17 @@ public void unbind(ServiceReference<ManagementContext> managementContext) throws
 
     public class CatalogPopulator extends BundleTracker<Iterable<? extends CatalogItem<?, ?>>> {
 
+        private final String BROOKLYN_PACKAGE_REGEX = "org.apache.brooklyn.*";
+
+        private final ConfigKey<List<String>> WHITELIST =
+            ConfigKeys.newConfigKey(new TypeToken<List<String>>(){}, "brooklyn.catalog.osgi.application.whitelist",
+                "CSV Whitelist of patterns to match bundle symbolic id if bundle is to be permitted to add " +
```
neykov: Mention that regexes are accepted.

core/src/main/java/org/apache/brooklyn/core/catalog/internal/CatalogBomScanner.java
```diff
@@ -74,6 +80,17 @@ public void unbind(ServiceReference<ManagementContext> managementContext) throws
...
+                    "applications (templates) to the catalog", ImmutableList.of(BROOKLYN_PACKAGE_REGEX));
+        private final ConfigKey<List<String>> BLACKLIST =
+            ConfigKeys.newConfigKey(new TypeToken<List<String>>(){}, "brooklyn.catalog.osgi.application.blacklist",
+                "CSV Blacklist of patterns to match bundle symbolic id to prevent selected whitelisted bundles " +
+                    "adding applications (templates) to the catalog", ImmutableList.<String>of());
+
```
neykov: Since this is a configuration set at dist build time do you think it will be better if set in [`org.apache.brooklyn.osgilauncher.cfg`](https://github.com/apache/brooklyn-server/blob/master/karaf/init/src/main/resources/OSGI-INF/blueprint/blueprint.xml#L33)? 

`brooklyn.properties` is reserved for users, so they will have to add this key themselves.

core/src/main/java/org/apache/brooklyn/core/catalog/internal/CatalogBomScanner.java
```diff
@@ -137,9 +153,19 @@ public void removedBundle(Bundle bundle, BundleEvent bundleEvent, Iterable<? ext
             }
         }
 
+        private void removeFromCatalog(CatalogItem<?, ?> item) {
+            try {
+                getManagementContext().getCatalog().deleteCatalogItem(item.getSymbolicName(), item.getVersion());
+            } catch (Exception e) {
+                LOG.warn("Failed to remove {} {} from catalog as {}", new String[] {
```
neykov: `propagateIfFatal` got lost in the rebase. Better insert it as the first expression after the catch.

core/src/main/java/org/apache/brooklyn/core/catalog/internal/CatalogBomScanner.java
```diff
@@ -121,12 +138,11 @@ public void removedBundle(Bundle bundle, BundleEvent bundleEvent, Iterable<? ext
...
             }
             LOG.debug("Unloading catalog BOM entries from {} {} {}", bundleIds(bundle));
-            final BrooklynCatalog catalog = getManagementContext().getCatalog();
             for (CatalogItem<?, ?> item : items) {
                 LOG.debug("Unloading {} {} from catalog", item.getSymbolicName(), item.getVersion());
 
                 try {
-                    catalog.deleteCatalogItem(item.getSymbolicName(), item.getVersion());
+                    removeFromCatalog(item);
                 } catch (Exception e) {
```
neykov: try-catch not needed any more.

core/src/main/java/org/apache/brooklyn/core/catalog/internal/CatalogBomScanner.java
```diff
@@ -137,9 +153,19 @@ public void removedBundle(Bundle bundle, BundleEvent bundleEvent, Iterable<? ext
             }
         }
 
+        private void removeFromCatalog(CatalogItem<?, ?> item) {
+            try {
+                getManagementContext().getCatalog().deleteCatalogItem(item.getSymbolicName(), item.getVersion());
+            } catch (Exception e) {
+                LOG.warn("Failed to remove {} {} from catalog as {}", new String[] {
+                    item.getSymbolicName(), item.getVersion(), e.getMessage()
```
neykov: Why not pass the exception as a throwable argument to log, instead of including the message only. Could mask an underlying problem (for example NPE will have an empty message).

core/src/main/java/org/apache/brooklyn/core/catalog/internal/CatalogBomScanner.java
```diff
@@ -155,9 +181,42 @@ public void removedBundle(Bundle bundle, BundleEvent bundleEvent, Iterable<? ext
...
+            final List<String> white = managementContext.getConfig().getConfig(WHITELIST);
+            final List<String> black = managementContext.getConfig().getConfig(BLACKLIST);
+
+            return on(bundle, white) && !on(bundle, black);
+        }
+
+        private boolean on(Bundle bundle, List<String> list) {
+            for (String candidate : list) {
+                final String symbolicName = bundle.getSymbolicName();
+                if (symbolicName.matches(candidate)) {
```
neykov: Do we need to trim the individual values? Or does that already happen in the coercion step? What if there's just only one item?

core/src/main/java/org/apache/brooklyn/core/catalog/internal/CatalogBomScanner.java
```diff
@@ -155,9 +181,42 @@ public void removedBundle(Bundle bundle, BundleEvent bundleEvent, Iterable<? ext
...
         }
 
+        private void removeAnyApplications(Iterable<? extends CatalogItem<?, ?>> catalogItems) {
+
+            final Iterator<? extends CatalogItem<?, ?>> cit = catalogItems.iterator();
+            while (cit.hasNext()) {
+                final CatalogItem<?, ?> item = cit.next();
+                if (TEMPLATE.equals(item.getCatalogItemType())) {
+                    removeFromCatalog(item);
+                    cit.remove();
```
neykov: (minor) No need for an iterator & removing the items from `catalogItems`. `removeFromCatalog` does all that's needed.

core/src/main/java/org/apache/brooklyn/core/catalog/internal/CatalogBomScanner.java
```diff
@@ -168,11 +227,18 @@ private String addLibraryDetails(Bundle bundle, String bomText) {
                     LOG.warn("Unexpected syntax for {} (expected Map), ignoring", BROOKLYN_CATALOG);
                 }
             }
-            final String updatedBom = new Yaml().dump(bom);
+            final String updatedBom = backToYaml(bom);
```
neykov: Something that I didn't consider before when discussing this approach is that parsing and re-serializing the yaml will change its formatting. Not a problem for Brooklyn internals, but will be surprising to users when viewing the plan in the catalog. Not to be addressed in this PR, but we should keep it in mind.

core/src/main/java/org/apache/brooklyn/core/catalog/internal/CatalogBomScanner.java
geomacy: I guess it would be handy enough to add them in that file and then it could be referred to in the documentation (will do documentation as a separate PR).

core/src/main/java/org/apache/brooklyn/core/catalog/internal/CatalogBomScanner.java
geomacy: This method modifies its argument, which is ultimately used as a return value for addingBundle, which becomes the set of items that is tracked by the BundleTracker, so if I remove something from the catalog in here, I also need to remove it from the iterable.  This probably this isn't clear enough from the code as written, so I shall change this to return a new Iterable rather than modify the arg.

core/src/main/java/org/apache/brooklyn/core/catalog/internal/CatalogBomScanner.java
geomacy: I'll trim the value, it's good to be explicit.  It should work fine with just one item.

core/src/main/java/org/apache/brooklyn/core/catalog/internal/CatalogBomScanner.java
geomacy: That is true, but the formatting added here will hopefully be good enough looking to be acceptable.  Certainly worth keeping in mind.

geomacy:
Code review comments done and commits squashed.

geomacy:
added license header to cfg file and re-squashed

utils/common/src/main/java/org/apache/brooklyn/util/text/Strings.java
```diff
@@ -916,4 +916,17 @@ public static String toUniqueString(Object x, int optionalMax) {
...
+
+    /**
+     * Parses a string as comma separated values and returns a list of the entries with whitespace removed.
+     * @param csv
+     * @param separators
+     */
+    public static List<String> parseCsv(String csv, String separators) {
+        List<String> result = MutableList.of();
+        for (String value: csv.split(separators)) {
+            result.add(value.trim());
```
neykov: Suggest adding some tests for this method. `.split` will return non-empty array for the empty string I believe

core/src/main/java/org/apache/brooklyn/core/catalog/internal/CatalogBomScanner.java
```diff
@@ -121,25 +157,27 @@ public void removedBundle(Bundle bundle, BundleEvent bundleEvent, Iterable<? ext
...
+            }
+        }
+
+        private void removeFromCatalog(CatalogItem<?, ?> item) {
+            try {
+                getManagementContext().getCatalog().deleteCatalogItem(item.getSymbolicName(), item.getVersion());
+            } catch (Exception e) {
+                Exceptions.propagateIfFatal(e);
+                LOG.warn(Strings.join(new String[] {
+                    "Failed to remove", item.getSymbolicName(), item.getVersion(), "{} {} from catalog"
```
neykov: Remove `{} {}`

neykov:
Looks great, a couple of minor points worth addressing.

PR #107, opened 2016-04-07
[OSGi] Load properties from config admin

neykov:
Load properties from:
  * global brooklyn.properties
  * local properties
  * config admin's brooklyn PID

It's a one-way integration where changes in config admin will cause reloading the properties in brooklyn. Setting properties programmatically in brooklyn will not update config admin.

--

Since we are relying on config admin to read the configuration (which could be coming from multiple places) there's no check on the security attributes of the source.

### DIFF
--- a/karaf/init/src/main/java/org/apache/brooklyn/launcher/osgi/OsgiLauncher.java
+++ b/karaf/init/src/main/java/org/apache/brooklyn/launcher/osgi/OsgiLauncher.java
@@ -15,6 +15,9 @@
  */
 package org.apache.brooklyn.launcher.osgi;
 
+import java.io.IOException;
+import java.util.Map;
+
 import javax.annotation.Nullable;
 
 import org.apache.brooklyn.api.mgmt.ManagementContext;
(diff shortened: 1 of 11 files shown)


geomacy:
👍  I've read through the code and don't have any particular comments to make, I also pulled and built the branch and confirmed that brooklyn.cfg is read and properties from it are used to configure Brooklyn and are dynamically updatable via the config:properties-set command in Karaf console.  

In its present form this will read properties, separately, from the global and local files, and from etc/brooklyn.cfg, and you can only see the values from brooklyn.cfg in the karaf console.  It might be interesting to discuss whether all values could be brought into Configuration Admin and managed from there?

neykov:
Thanks @geomacy. My thinking was that we should point `etc/brooklyn.cfg` to point to `~/.brooklyn/brooklyn.properties` instead (or maybe in addition to) - not done in these changes, just a matter of configuration.

geomacy:
Something worth thinking about I agree, but as far as this PR goes I think it's fine as is.

karaf/init/src/main/java/org/apache/brooklyn/launcher/osgi/OsgiLauncher.java
```diff
@@ -23,36 +26,78 @@
...
 
 /**
  * Initializer for brooklyn-core when running in an OSGi environment.
- *
- * Temporarily here; should be totally contained in blueprint beans' init-methods.
  */
 public class OsgiLauncher extends BasicLauncher<OsgiLauncher> {
 
     private static final Logger LOG = LoggerFactory.getLogger(OsgiLauncher.class);
+    public static final String BROOKLYN_CONFIG_PID = "brooklyn";
```
m4rkmckenna: IMO the PID should be `org.apacha.brooklyn`

neykov: I was wondering about the same. I'd like the file on disk to be called `brooklyn.cfg` making it a much more friendlier to autocomplete and shorter to type. Can we redirect the pid to a custom file name?

geomacy: I don't think we need to feel strictly constrained to using package style naming, 'org.apache.brooklyn'.  It's certainly appropriate if the configuration is for a module, where the package style is needed to distinguish itself from others that might have the same basename, e.g. x.y.z.core.security.cfg versus x.y.z.software.base.security.cfg. In this case however we're configuring the whole product, rather than a module. I think we can be a bit flexible here for the convenience of the simpler file name.

geomacy: I also think it doesn't really matter too much either way.

PR #108, opened 2016-04-07
New PerUserEntitlementManager keyword: user

sjcorbett:
`PerUserEntitlementManager` transforms the "user" keyword into an entitlement to everything but the most privileged actions (everything but `ROOT` and `SEE_ALL_SERVER_INFO`). Practically this means a user can do everything but: use `ScriptApi` to run any Groovy, stop the server, access management context configuration via `ServerApi`, reload Brooklyn properties and export persistence data.

I wasn't sure about the choice of "user" for the keyword. I wanted to indicate that it was a step down from "root". Other ideas people gave me include "manager" and "administrator", neither of which are hugely satisfactory to me. Suggestions welcome.

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/mgmt/entitlement/Entitlements.java
+++ b/core/src/main/java/org/apache/brooklyn/core/mgmt/entitlement/Entitlements.java
@@ -329,6 +329,7 @@ public class Entitlements {
         return FineGrainedEntitlements.anyOf(
             FineGrainedEntitlements.allowing(SEE_ENTITY),
             FineGrainedEntitlements.allowing(SEE_ACTIVITY_STREAMS),
+            FineGrainedEntitlements.allowing(SEE_CATALOG_ITEM),
             FineGrainedEntitlements.seeNonSecretSensors(),
             FineGrainedEntitlements.seeNonSecretConfig()
         );
--- a/core/src/main/java/org/apache/brooklyn/core/mgmt/entitlement/Entitlements.java
+++ b/core/src/main/java/org/apache/brooklyn/core/mgmt/entitlement/Entitlements.java
@@ -21,6 +21,7 @@ package org.apache.brooklyn.core.mgmt.entitlement;
 import java.util.Arrays;
 import java.util.List;
 
+import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 
 import org.apache.brooklyn.api.entity.Entity;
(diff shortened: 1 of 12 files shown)


core/src/main/java/org/apache/brooklyn/core/mgmt/entitlement/Entitlements.java
```diff
@@ -398,15 +420,17 @@ public static void clearEntitlementContext() {
...
     
     // ----------------- initialization ----------------
 
     public final static String ENTITLEMENTS_CONFIG_PREFIX = "brooklyn.entitlements";
-    
-    public static ConfigKey<String> GLOBAL_ENTITLEMENT_MANAGER = ConfigKeys.newStringConfigKey(ENTITLEMENTS_CONFIG_PREFIX+".global", 
+
+    public static final ConfigKey<String> GLOBAL_ENTITLEMENT_MANAGER = ConfigKeys.newStringConfigKey(ENTITLEMENTS_CONFIG_PREFIX + ".global",
         "Class for entitlements in effect globally; "
         + "short names 'minimal', 'readonly', or 'root' are permitted here, with the default 'root' giving full access to all declared users; "
```
andreaturli: shall we add `user` or the description doesn't need to be exhaustive?

sjcorbett: Good point

core/src/main/java/org/apache/brooklyn/core/mgmt/entitlement/PerUserEntitlementManager.java
```diff
@@ -64,7 +64,7 @@ private static EntitlementManager load(BrooklynProperties properties, String typ
     public PerUserEntitlementManager(BrooklynProperties properties) {
```
andreaturli: I've seen this manager replaces the deprecated `PerUserEntitlementManagerWithDefault` - shall we remove that one now that we are almost at 0.10.0-SNAPSHOT? 
not sure what's the rule of thumb for that :)

andreaturli:
code looks good to me and tests are green.

I'd like to see some additional doc for this feature, as far as we have only [entitlements.md](https://github.com/apache/brooklyn-docs/blob/master/guide/java/entitlements.md) where we describe the available `EntitlementManagers` but of course in a subsequent PR.

+1

sjcorbett:
@andreaturli Thanks.

@aledsage @ahgittin Any thoughts on terminology?

core/src/main/java/org/apache/brooklyn/core/mgmt/entitlement/PerUserEntitlementManager.java
sjcorbett: I will leave this for a future change.

sjcorbett:
@andreaturli Thanks for your comments. Look like no others are forthcoming so will merge this.

PR #111, opened 2016-04-11
pull displayName from flags in LocationTransformer if present - fixes BROOKLYN-250

johnmccabe:
Untyped config from the YAML is being pushed into flags rather than config, so rather than expose flags in the locations API (which we're moving away from to the catalog API), I'm now checking for the presence of a displayName parameter in flags and setting it in the LocationTransformer.

Tests added.


sjcorbett:
The test failure is unrelated but interesting - 

```
org.apache.brooklyn.entity.stock.BasicStartableTest.testTransitionsThroughLifecycles (from TestSuite)

Expected: eventually IsEqualTo([starting, running, stopping, stopped]); got most recently: [starting, running, stopping, stopping, stopped] (waited 1s 4ms 696us 7ns, checked 99)
Stacktrace

java.lang.AssertionError: Expected: eventually IsEqualTo([starting, running, stopping, stopped]); got most recently: [starting, running, stopping, stopping, stopped] (waited 1s 4ms 696us 7ns, checked 99)
```

@aledsage Have you seen duplicate states before?

tbouron:
Tested, LGTM

neykov:
LGTM

PR #112, opened 2016-04-12
Configurable removal strategies

nakomis:


nakomis:
Depended on by https://github.com/apache/brooklyn-library/pull/31 to be merged at the same time

sjcorbett:
Test failure is valid:
```
java.lang.IllegalArgumentException: Cannot coerce or set org.apache.brooklyn.entity.group.DynamicClusterTest$22@2bb2743b to dynamiccluster.configurable.removalStrategy[ConfigKey:org.apache.brooklyn.entity.group.RemovalStrategy]
	at org.apache.brooklyn.core.config.internal.AbstractConfigMapImpl.coerceConfigVal(AbstractConfigMapImpl.java:83) ~[classes/:na]
Caused by: org.apache.brooklyn.util.core.flags.ClassCoercionException: Cannot coerce type null to org.apache.brooklyn.entity.group.RemovalStrategy (org.apache.brooklyn.entity.group.DynamicClusterTest$22@2bb2743b): no adapter known
	at org.apache.brooklyn.util.core.flags.TypeCoercions.coerce(TypeCoercions.java:285) ~[classes/:na]
```

core/src/main/java/org/apache/brooklyn/entity/group/DynamicCluster.java
```diff
@@ -148,6 +148,11 @@
     ConfigKey<EntityFactory> FACTORY = ConfigKeys.newConfigKey(
             EntityFactory.class, "dynamiccluster.factory", "factory for creating new cluster members", null);
 
+
+    @SetFromFlag("removalStrategy")
+    ConfigKey<RemovalStrategy> CONFIGURABLE_REMOVAL_STRATEGY = ConfigKeys.newConfigKey(RemovalStrategy.class, "dynamiccluster.configurable.removalStrategy");
+
+    @Deprecated /** @deprecated since 0.9.0; use {@link CONFIGURABLE_REMOVAL_STRATEGY instead} */
```
sjcorbett: We are up to 0.10.0 now.

sjcorbett: `{@link CONFIGURABLE_REMOVAL_STRATEGY instead}` should be `{@link CONFIGURABLE_REMOVAL_STRATEGY} instead`.

core/src/main/java/org/apache/brooklyn/entity/group/DynamicClusterImpl.java
```diff
@@ -259,17 +268,39 @@ protected void initEnrichers() {
...
     protected Function<Collection<Entity>, Entity> getRemovalStrategy() {
         Function<Collection<Entity>, Entity> result = getConfig(REMOVAL_STRATEGY);
         return (result != null) ? result : defaultRemovalStrategy;
     }
 
+    protected RemovalStrategy getConfigurableRemovalStrategy() {
+        RemovalStrategy result = getConfig(CONFIGURABLE_REMOVAL_STRATEGY);
+        // Fallback to the deprecated REMOVAL_STRATEGY for backwards-compatibility
+        if (result == null) {
+            result = new RemovalStrategy() {
```
sjcorbett: Make this a named inner class. New anonymous classes risk pain with persistence later.

core/src/main/java/org/apache/brooklyn/entity/group/DynamicClusterImpl.java
```diff
@@ -151,12 +151,14 @@ public ZoneFailureDetector apply(final String input) {
      */
     protected final Object mutex = new Object[0];
 
-    private static final Function<Collection<Entity>, Entity> defaultRemovalStrategy = new Function<Collection<Entity>, Entity>() {
-        @Override public Entity apply(Collection<Entity> contenders) {
```
sjcorbett: Again on persistence. @aledsage is deleting this anonymous inner class safe? (seems a little silly to have to ask!)

core/src/main/java/org/apache/brooklyn/entity/group/DynamicCluster.java
```diff
@@ -148,6 +148,11 @@
     ConfigKey<EntityFactory> FACTORY = ConfigKeys.newConfigKey(
             EntityFactory.class, "dynamiccluster.factory", "factory for creating new cluster members", null);
 
+
+    @SetFromFlag("removalStrategy")
```
sjcorbett: `@SetFromFlag("removalStrategy")` is on both this and the deprecated key. Which one wins?

sjcorbett:
Thinking about this a bit more closely, you've deprecated a `ConfigKey<Function<Collection<Entity>, Entity>>` in favour of a `ConfigKey<RemovalStrategy>`, but the `RemovalStrategy` class implements `Function<Collection<Entity>, Optional<Entity>`. You could avoid all of the deprecation hassle by leaving the config key the same and altering `RemovalStrategy` to implement `Function<Collection<Entity>, Entity>`. The result is identical, no?

nakomis:
Hadn't twigged that I wouldn't need to change the config key. Much nice without all the deprecation. Have close https://github.com/apache/brooklyn-library/pull/31 as it's no longer needed

sjcorbett:
Looks much cleaner. I'm wary that turning `defaultRemovalStrategy` from an anonymous inner class to a named inner class will be problematic for anyone with persisted state (because the `$1`, `$2` etc. numbering will change). Given that it's not imperative for this feature are you ok to revert that particular portion?

core/src/main/java/org/apache/brooklyn/entity/group/DynamicClusterImpl.java
neykov: Rule of thumb is that anything that goes into a config or sensor (or is nested in such an object) will get persisted. Looking at the code this is not the case, but can never know what an extending class might be doing, so better be on the safe side.

core/src/main/java/org/apache/brooklyn/entity/group/FirstFromRemovalStrategy.java
```diff
@@ -0,0 +1,53 @@
...
+
+public class FirstFromRemovalStrategy extends RemovalStrategy {
+
+    public static final ConfigKey<List<RemovalStrategy>> STRATEGIES = ConfigKeys.newConfigKey(new TypeToken<List<RemovalStrategy>>() {}, "firstfrom.strategies",
+            "An ordered list of removal strategies to be used to determine which entity to remove");
+
+    @Nullable
+    @Override
+    public Entity apply(@Nullable Collection<Entity> input) {
+        List<RemovalStrategy> strategies = config().get(STRATEGIES);
```
neykov: From YAML standpoint would be better if these are set on the cluster entity. Could inject the cluster through a method in `RemovalStrategy`, or call a static util method (hiding a threadlocal) (yikes I know :) ).

core/src/main/java/org/apache/brooklyn/entity/group/SensorMatchingRemovalStrategy.java
```diff
@@ -0,0 +1,50 @@
...
+import org.apache.brooklyn.api.sensor.AttributeSensor;
+import org.apache.brooklyn.config.ConfigKey;
+import org.apache.brooklyn.core.config.ConfigKeys;
+
+import com.google.common.reflect.TypeToken;
+
+public class SensorMatchingRemovalStrategy<T> extends RemovalStrategy {
+    public static final ConfigKey<AttributeSensor> SENSOR = ConfigKeys.newConfigKey(new TypeToken<AttributeSensor>() {}, "sensor.matching.sensor");
+    // Would be nice to use ConfigKey<T>, but TypeToken<T> cannot be instantiated at runtime
+    public static final ConfigKey<Object> DESIRED_VALUE = ConfigKeys.newConfigKey(new TypeToken<Object>() {}, "sensor.matching.value");
```
neykov: No need for a `TypeToken` here and above, could use class references directly.

neykov:
LGTM, could make it more YAML friendly.

core/src/main/java/org/apache/brooklyn/entity/group/DynamicClusterImpl.java
aledsage: Agreed, let's be paranoid. Mark it as deprecated, with a comment saying why we have not deleted it.

There are two ways changing your anonymous inner class can break persistence: one is if someone has persisted it; two is if deleted or moving the class changes the classnames of other anonymous inner classes (there are 5 of them in DynamicClusterImpl)!

core/src/main/java/org/apache/brooklyn/entity/group/SensorMatchingRemovalStrategy.java
aledsage: @neykov scarily, changing the use of TypeToken will renumber all the anonymous inner classes below it! If anyone has persisted those, it will cause problems.

We really do need to get to be stricter by default, that we won't accept persisting anonymous inner classes!

aledsage: Ah, it's a brand new class - in which case yes, change to just `AttributeSensor.class` and `Object.class`.

core/src/main/java/org/apache/brooklyn/entity/group/FirstFromRemovalStrategy.java
aledsage: Following up on @neykov's point - from a yaml perspective, how would one use this currently? What is the advantage of having `RemovalStrategy extends BasicConfigurableObject`? Are you expecting people to use (in yaml) the `$brooklyn:object` syntax to instantiate the removal strategy? (if so, they can use `brooklyn.config` within the `$brooklyn:object:`). I still don't see the benefit of that over a pojo.

nakomis: The benefit of using `BasicConfigurableObject` is that the configuration is the configuration (i.e. setting of ConfigKeys) is managed automatically by the `ConfigurationSupport`, rather than implementors having to read the values directly from the properties passed to the constructor

nakomis:
PR Comments addressed. @neykov would be interesting to hear your thoughts for making it more YAML friendly - perhaps for a future update?

neykov:
@nakomis Didn't realise this can be used with $broolyn:object. That's enough YAML friendly for me :).

PR #113, opened 2016-04-12
Fix NPE in forcePersistNow

grkvlt:
Fixes issue where `forcePeristNow()` fails due to `persistenceRealChangeListener` being null


neykov:
Looks like there's a similar problem with `forcePersistNow` for the `full` case - should we fix the method in a more general way?

grkvlt:
@neykov good catch, I'll do that...

grkvlt:
Now logs when `forcePersistNow()` is called and no persistence mechanism is set.

aledsage:
Cherry-picking into 0.9.x branch.

PR #114, opened 2016-04-13
Driver copyResources walks the filesystem 

sjcorbett:
So blueprint writers can replace
```yaml
  brooklyn.config:
    runtimeTemplates:
      /path/to/source1: dest1
      /path/to/source2: dest2
      ...
      /path/to/sourceN: destN
```
With
```yaml
  brooklyn.config:
    runtimeTemplates:
      /path/to: dest
```
One minor difference: in the first example the two files end up in `runDir/` but in the second in `runDir/dest/`. If using the latter syntax I think this would be expected.

Each file is copied in a task so there's better visibility of what is in progress at any given point.

### DIFF
--- a/launcher/src/main/java/org/apache/brooklyn/launcher/BrooklynLauncher.java
+++ b/launcher/src/main/java/org/apache/brooklyn/launcher/BrooklynLauncher.java
@@ -384,7 +384,7 @@ public class BrooklynLauncher extends BasicLauncher<BrooklynLauncher> {
         if (getPersistMode() != PersistMode.DISABLED) {
             try {
                 Stopwatch stopwatch = Stopwatch.createStarted();
-                if (managementContext.getHighAvailabilityManager().getPersister() != null) {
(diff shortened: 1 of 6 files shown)


Graeme-Miller:
lgtm

launcher/src/main/java/org/apache/brooklyn/launcher/BrooklynLauncher.java
```diff
@@ -384,7 +384,7 @@ public void terminate() {
         if (getPersistMode() != PersistMode.DISABLED) {
             try {
                 Stopwatch stopwatch = Stopwatch.createStarted();
-                if (managementContext.getHighAvailabilityManager().getPersister() != null) {
+                if (managementContext.getHighAvailabilityManager() != null && managementContext.getHighAvailabilityManager().getPersister() != null) {
```
neykov: Do you think we should do the same checks in `AbstractManagementContext.terminate()` in case startup failed?

neykov:
Very useful. Merging.

PR #115, opened 2016-04-16
BROOKLYN-257: allow nulling imageId in jcloudsLocation.obtain()

aledsage:
Fixes https://issues.apache.org/jira/browse/BROOKLYN-257.

To fix this, I was tempted to instead change `JcloudsLocation.buildTemplate` so that it ignored all null values, changing the code below:

```
        for (Map.Entry<ConfigKey<?>, CustomizeTemplateBuilder> entry : SUPPORTED_TEMPLATE_BUILDER_PROPERTIES.entrySet()) {
            ConfigKey<?> name = entry.getKey();
            CustomizeTemplateBuilder code = entry.getValue();
            if (config.containsKey(name))
                code.apply(templateBuilder, config, config.get(name));
        }
```

to:

```
        for (Map.Entry<ConfigKey<?>, CustomizeTemplateBuilder> entry : SUPPORTED_TEMPLATE_BUILDER_PROPERTIES.entrySet()) {
            ConfigKey<?> name = entry.getKey();
            CustomizeTemplateBuilder code = entry.getValue();
            if (config.containsKey(name) && config.get(name) != null)
                code.apply(templateBuilder, config, config.get(name));
        }
```

But is that risky? Do we rely for any config keys on `null` meaning something other than "use the jclouds default"?

### DIFF
--- a/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
+++ b/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
@@ -1146,11 +1146,15 @@ public class JcloudsLocation extends AbstractCloudMachineProvisioningLocation im
                     }})
             .put(HARDWARE_ID, new CustomizeTemplateBuilder() {
                     public void apply(TemplateBuilder tb, ConfigBag props, Object v) {
-                        tb.hardwareId(((CharSequence)v).toString());
+                        if (v != null) {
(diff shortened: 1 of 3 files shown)


neykov:
+1 to doing this in `buildTemplate`. Looking at the `CustomizeTemplateBuilder`s they either will break with null values or ignore them if present.
Merging as is, it's useful on it's own. Further changes can be done in a separate commit.

aledsage:
Cherry-picking this into 0.9.x branch.

PR #116, opened 2016-04-20
Fix JcloudsLocation.resumeMachine when password supplied

aledsage:


locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -2133,11 +2133,17 @@ protected NodeMetadata findNodeOrThrow(ConfigBag config) {
...
+        String pkd = osCredentials.getPrivateKeyData();
+        String password = osCredentials.getPassword();
         if (Strings.isNonBlank(pkd)) {
             LoginCredentials expectedCredentials = LoginCredentials.fromCredentials(new Credentials(user, pkd));
             //override credentials
             node = NodeMetadataBuilder.fromNodeMetadata(node).credentials(expectedCredentials).build();
+        } else if (Strings.isNonBlank(password)) {
+            LoginCredentials expectedCredentials = LoginCredentials.fromCredentials(new Credentials(user, password));
+            //override credentials
+            node = NodeMetadataBuilder.fromNodeMetadata(node).credentials(expectedCredentials).build();
```
andreaturli: maybe you could move this out of the 'else' (and of the 'if' as well)

andreaturli:
@aledsage, code looks sensible just a minor comment. do we have a test that capture this behavior?

andreaturli:
jenkins failure doesn't seem related to this PR
`Build timed out (after 40 minutes). Marking the build as aborted.`

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
aledsage: @andreaturli not sure I follow. The if-else means that it preserves previous behaviour when there is privateKeyData. If someone supplies both privateKeyData and a password, then the privateKeyData takes precedence. If I moved this out of the else/if, then it would overwrite the credentials set at lines 2138-2140.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -2133,11 +2133,17 @@ protected NodeMetadata findNodeOrThrow(ConfigBag config) {
...
         }
 
         return node;
```
andreaturli: @aledsage it was a minor suggestion. I guess, we can just return the node:
```
return NodeMetadataBuilder.fromNodeMetadata(node).credentials(expectedCredentials).build();
```
and remove lines 2145-2146 and 2141-2142, maybe.
if we pulled it out from the if-else and you can

aledsage: I'm going to leave as-is because there is an implicit third case (where neither if block matches). In that case, we don't override the node's credentials (so use whatever the compute-service gave us). I suspect that will not have credentials at all.

aledsage:
Thanks @andreaturli - merging now.

PR #117, opened 2016-04-20
Fix conversion of unsigned bytes to IP addresses

grkvlt:
IP addresses with octets greater than 127 should be parsed as unsigned bytes, to prevent them being displayed with negative numbers.


andreaturli:
+1

PR #119, opened 2016-04-20
Add method HttpFeed.Builder#credentials(Credentials)

bostko:
Useful when we want to use a custom implementations of Credentials like a UsernamePasswordCredentials like implementation with updatable password.

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/feed/http/HttpFeed.java
+++ b/core/src/main/java/org/apache/brooklyn/feed/http/HttpFeed.java
@@ -197,6 +197,10 @@ public class HttpFeed extends AbstractFeed {
             this.credentials = new UsernamePasswordCredentials(username, password);
             return this;
         }
+        public Builder credentials(Credentials credentials) {
+            this.credentials = credentials;
+            return this;
+        }
         public Builder credentialsIfNotNull(String username, String password) {
             if (username != null && password != null) {
                 this.credentials = new UsernamePasswordCredentials(username, password);


core/src/main/java/org/apache/brooklyn/feed/http/HttpFeed.java
```diff
@@ -197,6 +197,10 @@ public Builder credentials(String username, String password) {
             this.credentials = new UsernamePasswordCredentials(username, password);
             return this;
         }
+        public Builder credentials(Credentials credentials) {
```
aledsage: Not sure about this. Nothing in the HttpFeed's public API talks about `org.apache.http.auth`. If we add this public method, then `org.apache.http.auth.Credentials` is exposed (rather than it being an implementation detail).

Mind you, some of our other "public" classes (e.g. `HttpTool`) exposes `org.apache.http.*` in its API.

What is the use-case for adding this?

@rdowner what are your thoughts - should we expose this in the API?

ahgittin:
@bostko i agree w @aledsage, it's nicer if we can avoid putting `o.a.h.a.Credentials` in the api:  what is the use case?  maybe a use-case specific method would be better.

aledsage:
@bostko any comments on this? Do you agree that we shouldn't expose `o.a.h.a.Credentials` in the API? What is the use-case?

bostko:
An example is when you want to provide the HttpFeed with a mutable UsernameAndPasswordCredentials. When the credentials are changed you should it is useful to be able to alter the credentials object from code who is aware of the credentials change. Example https://github.com/brooklyncentral/brooklyn-ambari/pull/101/files

aledsage:
@bostko Ah, thanks for explaining the use-case. Could we instead fix that with an overloaded method:

        public Builder credentials(Supplier<String> username, Supplier<String> password) { ...

Under-the-covers, that could either instantiate a mutable `UsernameAndPasswordCredentials`.

We'd need to be careful with a mutable credentials object. I'm not sure how well `HttpClient` will cope with being instantiated with mutable credentials. Hopefully it would be ok; I've never tried. (we create the `HttpClient` in `HttpFeed.preStart()`, and then re-use it for each poll.

core/src/main/java/org/apache/brooklyn/feed/http/HttpFeed.java
neykov: Agree that we should avoid it if possible. Depending on the underlying implementation being mutable and changing it **after** being passed (that's how I understand the change) doesn't sound like a good idea.

neykov:
Bump @bostko.

bostko:
@neykov what are your thoughts on this?

neykov:
Agree with @aledsage. Do we really need to marry our API to http client? A mutable credentials object depends on implementation details.
Why need this, what's the driver?

bostko:
Closing it until it is really needed.

PR #120, opened 2016-04-21
Add reading of default.catalog.bom.

geomacy:


karaf/init/src/main/java/org/apache/brooklyn/launcher/osgi/OsgiLauncher.java
```diff
@@ -38,6 +39,7 @@
 public class OsgiLauncher extends BasicLauncher<OsgiLauncher> {
 
     private static final Logger LOG = LoggerFactory.getLogger(OsgiLauncher.class);
+    private static final String DEFAULT_CATALOG_BOM = "file:conf/brooklyn/default.catalog.bom";
```
neykov: Will this work if we start karaf from a non-home folder?

geomacy: It works fine, Karaf must be able to work out its Karaf root on its own.

neykov: What do you think about moving the file to the `etc` folder?

neykov:
Looks great, would like to discuss the bom location.

karaf/init/src/main/java/org/apache/brooklyn/launcher/osgi/OsgiLauncher.java
geomacy: I did think about that and wasn't sure.  In the end I plumped for `conf` in order to keep the file location the same as for the classic launcher, but didn't feel very strongly that this was the clear choice.  If you think it's probably better to 'go the Karaf way' at the expense of a little backward compatibility then I'll change this.  Karaf is already quite different from classic in many ways.

geomacy:
Changed location of file to etc, squashed and re-pushed.

neykov:
Thanks @geomacy, merging.

PR #121, opened 2016-04-21
Fix loading yaml CAMP specs from bundles classpath

neykov:
Also fix the build path of helper test bundles, so they end up in the correct folder.


sjcorbett:
Looks good.

PR #122, opened 2016-04-21
Test case for resources used by vanilla software processes

sjcorbett:
@neykov Here's the test case for resources used by `VanillaSoftwareProcesses`. 

@rdowner The blueprint in this pull request would be a good entry in an FAQ. Would be useful to talk to you about how to get it there.


neykov:
Thanks @sjcorbett, merging.

PR #125, opened 2016-04-25
[BROOKLYN-343] A test to illustrate a failing case where cross-referencing catalog items fails

neykov:
On the other hand removing `services` works just fine.

The problem is that the camp parser doesn't see items which are in the same catalog and already parsed. Only already added items are visible. The catalog parsers does a poor-man's validation on the top level items only, we need to replace that with the normal CAMP parser workflow, but able to see items in the current catalog preceding the current item.


aledsage:
@neykov if this is still an issue, then I suggest we create a jira for it, update this PR to include a link to that jira, mark the test as `@Test(groups={"Broken"})`, and merge it.

If it is already fixed, then we should check if there is already a test for it and if not then merge this PR.

Thoughts?

neykov:
I believe it's still failing. I'll mark it as `Broken` and create a Jira to track it.

geomacy:
Might be worth linking the new JIRA to https://cloudsoft.jira.com/browse/AMP-982; I don't mean it's the same issue but it appears similar in area and might be worth taking into consideration when doing this one.

neykov:
Marked it as Broken. Merging.

PR #126, opened 2016-04-26
Fix JmxAttributeSensor

googlielmo:
- fix bug where period was wrongly taken from entity config
 - fix add feed to entity 
 - add unit test


aledsage:
LGTM; merging.

PR #127, opened 2016-04-27
Add configuration option to SSH sensor to trim output

grkvlt:
Removes surrounding whitespace which makes further transformation simpler.

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/sensor/ssh/SshCommandSensor.java
+++ b/core/src/main/java/org/apache/brooklyn/core/sensor/ssh/SshCommandSensor.java
@@ -20,6 +20,15 @@ package org.apache.brooklyn.core.sensor.ssh;
 
 import java.util.Map;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.common.annotations.Beta;
+import com.google.common.base.Function;
+import com.google.common.base.Functions;
+import com.google.common.base.Preconditions;
+import com.google.common.base.Supplier;
+
 import org.apache.brooklyn.api.entity.EntityInitializer;
 import org.apache.brooklyn.api.entity.EntityLocal;
 import org.apache.brooklyn.config.ConfigKey;
@@ -35,14 +44,6 @@ import org.apache.brooklyn.util.core.config.ConfigBag;
 import org.apache.brooklyn.util.core.flags.TypeCoercions;
 import org.apache.brooklyn.util.os.Os;
 import org.apache.brooklyn.util.text.Strings;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
-import com.google.common.annotations.Beta;
-import com.google.common.base.Function;
-import com.google.common.base.Functions;
-import com.google.common.base.Preconditions;
-import com.google.common.base.Supplier;
 
 /** 
  * Configurable {@link EntityInitializer} which adds an SSH sensor feed running the <code>command</code> supplied
@@ -69,7 +70,7 @@ public final class SshCommandSensor<T> extends AddSensor<T> {
 
         // TODO create a supplier for the command string to support attribute embedding
         command = Preconditions.checkNotNull(params.get(SENSOR_COMMAND), "command");
-        
+
         executionDir = params.get(SENSOR_EXECUTION_DIR);
     }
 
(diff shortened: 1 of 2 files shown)


grkvlt:
I have left the default for the new `TRIM` option as `false` to keep existing behaviour. Should be pretty safe to merge. @aledsage WDYT?

core/src/main/java/org/apache/brooklyn/core/sensor/ssh/SshCommandSensor.java
```diff
@@ -104,7 +107,7 @@ public String get() {
                 .onSuccess(Functions.compose(new Function<String, T>() {
                         @Override
                         public T apply(String input) {
-                            return TypeCoercions.coerce(input, getType(type));
+                            return TypeCoercions.coerce(trim ? input.trim() : input, getType(type));
```
aledsage: Defend against null?

core/src/main/java/org/apache/brooklyn/feed/ssh/SshFeed.java
```diff
@@ -28,6 +28,19 @@
 import java.util.concurrent.Callable;
 import java.util.concurrent.TimeUnit;
 
+import org.slf4j.Logger;
```
aledsage: Would prefer us not reordering imports (most of our code is in the default order that eclipse gives - but this change is not). But I can live with it.

aledsage:
LGTM; only a couple of very minor comments.

neykov:
Wouldn't a better fix be to always trim on coercing strings (i.e. in the TypeCoercions code)? At the very least trimming should be enabled by default.

core/src/main/java/org/apache/brooklyn/feed/ssh/SshFeed.java
kiuby88: @sjcorbett send me a [interesting link]( http://stackoverflow.com/a/17194980/41195) to add Eclipse's import style to IntelliJ IDE. I hope it will be useful for you.

johnmccabe:
LGTM, I tend to agree with @neykov that it should be on by default.

ahgittin:
Agree with @neykov `TypeCoercions` should ignore whitespace in most cases.

That would render this PR unnecessary additional complexity.  Suggest closing this and fixing `TypeCoercions`.

grkvlt:
@neykov I'll change this to trim by default, and add null checks as @aledsage suggests

grkvlt:
@ahgittin missed your comment - will look into `TypeCoercions` instead - are you thinking just add a `.trim()` when coercont to `java.lang.String` then?

grkvlt:
@ahgittin @neykov finally had a look at this, and turns out that in `TypeCoercions` we already call `value.trim()` in the `stringToPrimitive()` method. It's really only a problem when we want the output of the SSH command to _remain_ as string sensor data, that may be getting processed elsewhere, and in that case we just want to remove the final `\n` that is getting added. We _could_ have added something [here](https://github.com/apache/brooklyn-server/blob/master/core/src/main/java/org/apache/brooklyn/util/core/flags/TypeCoercions.java#L175) where we would check if `targetType instanceof String` but not convinced now that this is the right solution. But, since the issue only occurs in SSH command based sensors, I have just added a call to `Strings.trimEnd()` which is null-safe, and removed the configuration option for trimming entirely.

core/src/main/java/org/apache/brooklyn/feed/ssh/SshFeed.java
```diff
@@ -267,13 +261,13 @@ public SshPollValue call() throws Exception {
...
         SshMachineLocation machine = getMachine();
-        Boolean execAsCommand = getConfig(EXEC_AS_COMMAND);
+        Boolean execAsCommand = config().get(EXEC_AS_COMMAND);
         if (log.isTraceEnabled()) log.trace("Ssh polling for {}, executing {} with env {}", new Object[] {machine, command, env});
         ByteArrayOutputStream stdout = new ByteArrayOutputStream();
         ByteArrayOutputStream stderr = new ByteArrayOutputStream();
 
         int exitStatus;
-        ConfigBag flags = ConfigBag.newInstance()
+        ConfigBag flags = ConfigBag.newInstanceExtending(config().getBag())
```
aledsage: @grkvlt this change sneaked in without comment ;-)
It looks reasonable (though the config on the feed will normally not be low-level ssh config, I'd guess). But could you comment on it to say why, and to confirm it is deliberate?

aledsage:
@grkvlt I agree the trim looks fine here. Happy to merge, once you've confirmed about the change for `ConfigBag.newInstanceExtending(config().getBag())`.

andreaturli:
please merge it as it looks very useful!

core/src/main/java/org/apache/brooklyn/feed/ssh/SshFeed.java
grkvlt: This change is because we are adding some extra config overrides, and it seemed to be the correct method to use based on the JavaDoc.

grkvlt:
I will merge this, including `ConfigBag#newInstanceExtending(ConfigBag)` change.

PR #128, opened 2016-04-28
jclouds.ImageChooser configurable

aledsage:
This makes a custom-supplied `JcloudsLocationConfig.IMAGE_CHOOSER` customizable: if it implements `ConfigAwareChooser` then `cloneFor(ConfigBag)` will be called. This follows the same pattern as `ComputeServiceAwareChooser` (both inside `BrooklynImageChooser`).

This also adds `BrooklynImageChooser.imageChooserFromOrderings(Iterable<? extends Ordering<? super Image>>)`, so that a downstream image chooser can supply multiple orderings (the first taking priority, and subsequent ones to break any ties).

Strictly speaking, this is a break to semantic versioning to put this in the 0.9.x branch. However, the ImageChooser is quite niche. (Frankly, we should have marked `BrooklynImageChooser` as `@Beta` until it had been used in anger in more places.) The change is backwards compatible. It's just that if someone used this in a 0.9.1 and then switched back to 0.9.0 then they `ConfigAwareChooser` wouldn't be there so it wouldn't compile. On balance, I'm fine with us making that sacrifice.

The use case is...

On VMware vCloudDirector there is a customer preference for VM images that are local to the vDC (virtual datacenter) over those in other vDCs and public images. They are using imageNameRegex so jclouds can return us the corresponding image in multiple vDCs and in the public catalog. To choose the local image in a Brooklyn image chooser, we need to know where the VM is going to be provisioned. We therefore need to know the config being supplied to the `jcloudsLocation.obtain()` method *inside* an image chooser.

### DIFF
--- a/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/BrooklynImageChooser.java
+++ b/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/BrooklynImageChooser.java
@@ -18,12 +18,15 @@
  */
 package org.apache.brooklyn.location.jclouds;
 
+import static com.google.common.base.Preconditions.checkNotNull;
+
 import java.util.Arrays;
 import java.util.Comparator;
 import java.util.Iterator;
 import java.util.List;
 
 import org.apache.brooklyn.util.collections.MutableList;
+import org.apache.brooklyn.util.core.config.ConfigBag;
 import org.jclouds.compute.ComputeService;
 import org.jclouds.compute.domain.Image;
 import org.jclouds.compute.domain.OperatingSystem;
@@ -33,7 +36,11 @@ import org.slf4j.LoggerFactory;
 
 import com.google.common.annotations.Beta;
 import com.google.common.base.Function;
+import com.google.common.base.Predicates;
 import com.google.common.collect.ComparisonChain;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Iterables;
+import com.google.common.collect.Lists;
 import com.google.common.collect.Ordering;
 import com.google.common.math.DoubleMath;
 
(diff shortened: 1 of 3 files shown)


locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/BrooklynImageChooser.java
```diff
@@ -292,27 +325,63 @@ public int compare(Image left, Image right) {
...
+            } else {
+                return this;
+            }
+        }        
+        @Override
+        public ImageChooserFromOrdering cloneFor(ConfigBag config) {
+            if (Iterables.tryFind(orderings, Predicates.instanceOf(ConfigAwareChooser.class)).isPresent()) {
+                List<Ordering<? super Image>> clonedOrderings = Lists.newArrayList();
+                for (Ordering<? super Image> ordering : orderings) {
+                    if (ordering instanceof ConfigAwareChooser) {
```
Graeme-Miller: Slightly concerned about the use of instanceof here, but I can see we are using the same approach as the ComputeServiceAwareChooser

Graeme-Miller:
Looks good to me, nothing wrong witht he additions as they are following the pattern already set down by the ComputeServiceImageChooser.

However, I do feel like this class could use a refactor at some point in the future, to split out the classes/interfaces, and possibly to remove the instanceof checks. Don't let that hold up this PR though

ahgittin:
Makes sense to backport this to `0.9.x`.  I don't view that as a violation of semver because for semver `a.b.c` when `a=0` special semantics apply.  We've been stricter than semver requires in this case, taking `b` is a major version and `c` a minor version.  In this strict case `0.9.1` is not a point release but a minor release, meaning new features can be added so long as it doesn't break backwards compatibility.

aledsage:
Thanks all - will merge. The PR failed because it can't find a 0.9.1-SNAPSHOT dependency. Not sure what is up with the build (and why it needs something to pre-exist?!).

I'll close-and-reopen another PR against master (like I should have done first), and then cherry-pick it into 0.9.x.

aledsage:
Replaced by https://github.com/apache/brooklyn-server/pull/129 (which is against master). Closing this, and will cherry-pick to 0.9.x afterwards.

PR #131, opened 2016-04-29
Add Karaf equivalents of cloud commands.

geomacy:
Refactors the CloudExplorer class into a core CloudExplorerSupport class
that does the work without knowing about any command line functionality,
plus a smaller CloudExplorer that does the command line handling and
invokes the CloudExplorerSupport to do the work.

Adds Karaf commands with the same names that use Karaf's command line
support to invoke the CloudExplorerSupport.

Adds the Karaf commands bundle as a boot feature so the commands are
readily available on startup.

### DIFF
--- a/karaf/commands/src/main/java/org/apache/brooklyn/karaf/commands/cloud/explorer/BlobCommand.java
+++ b/karaf/commands/src/main/java/org/apache/brooklyn/karaf/commands/cloud/explorer/BlobCommand.java
@@ -16,29 +16,30 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-package org.apache.brooklyn.karaf.commands;
+package org.apache.brooklyn.karaf.commands.cloud.explorer;
 
-import org.apache.karaf.shell.api.action.Action;
-import org.apache.karaf.shell.api.action.Argument;
+import org.apache.brooklyn.launcher.command.support.CloudExplorerSupport.GetBlob;
 import org.apache.karaf.shell.api.action.Command;
 import org.apache.karaf.shell.api.action.Option;
 import org.apache.karaf.shell.api.action.lifecycle.Service;
 
-@Command(scope = "brooklyn", name = "catalog", description = "Manage the local brooklyn catalog")
+import static org.apache.brooklyn.karaf.commands.cloud.explorer.AbstractCloudExplorerCommand.CLOUD_EXPLORER_SCOPE;
+
+@Command(scope = CLOUD_EXPLORER_SCOPE, name = GetBlob.NAME, description = GetBlob.DESCRIPTION)
 @Service
-public class Catalog implements Action {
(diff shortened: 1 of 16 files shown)


karaf/commands/src/main/java/org/apache/brooklyn/karaf/commands/cloud/explorer/BlobCommand.java
```diff
@@ -0,0 +1,45 @@
...
+    @Option(name = GetBlob.CONTAINER_ARGUMENT_NAME, description = GetBlob.CONTAINER_ARGUMENT_DESC, required = true)
+    private String container;
+
+    @Option(name = GetBlob.BLOB_ARGUMENT_NAME, description = GetBlob.BLOB_ARGUMENT_DESC, required = true)
+    private String blob;
+
+    @Override
+    public Object execute() throws Exception {
+
+        final GetBlob GetBlob =
```
neykov: Preference to lowerCamelCase for variables here and in the other commands.

launcher-common/pom.xml
```diff
@@ -74,6 +74,11 @@
             <version>${project.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.apache.brooklyn</groupId>
+            <artifactId>brooklyn-locations-jclouds</artifactId>
+            <version>0.10.0-SNAPSHOT</version>
```
neykov: Use `${project.version}` instead.

neykov:
LGTM

geomacy:
Thanks @neykov, will do a quick update to fix those comments soon.

geomacy:
hi @neykov comments addressed, thanks.

launcher-common/src/main/java/org/apache/brooklyn/launcher/command/support/CloudExplorerSupport.java
```diff
@@ -0,0 +1,437 @@
...
+            stdout.println("\tidentity: "+loc.getIdentity());
+            if (loc.getEndpoint() != null) stdout.println("\tendpoint: "+loc.getEndpoint());
+            if (loc.getRegion() != null) stdout.println("\tregion: "+loc.getRegion());
+
+            try {
+                doCall(loc, "\t");
+            } finally {
+                stdout.println("}");
+            }
+        }
```
m4rkmckenna: This could be simplified by using `org.apache.karaf.shell.support.table.ShellTable`

Unless original formatting is required to remain the same

geomacy: I've added a TODO note to consider doing this, it's a good idea.  I don't think it needs done as part of this PR.

m4rkmckenna:
LGTM :: Tested against openstack

As mentioned above the output could be made prettier by using `ShellTable` but this shouldn't hold up this PR

PR #133, opened 2016-05-04
Fix clocker pattern: reload-properties listener

aledsage:
See https://github.com/brooklyncentral/clocker/issues/290

This PR is testing how to fix that in Brooklyn, and to ensure that we continue to support that use-case in clocker.

Note the second commit (update for checkNotNull on rebind) is just to bring this code up-to-date with how Clocker actually does it. It is the first commit that is actually interesting here.

@grkvlt please review.


ahgittin:
this is tests only, demoing and asserting behaviour that downstream projects rely on.  merging.

PR #134, opened 2016-05-06
Make entity ids lowercase

grkvlt:
Needed to allow use of entity and application ids as part of DNS names, which
are case-insensitive. Extension to 10 characters gives equivalent amount of
randomness, so collision probability is unchanged.

### DIFF
--- a/utils/common/src/main/java/org/apache/brooklyn/util/text/Identifiers.java
+++ b/utils/common/src/main/java/org/apache/brooklyn/util/text/Identifiers.java
@@ -91,21 +91,29 @@ public class Identifiers {
      * <b>NOTE</b> This version is 30-50% faster than the old double-based one,
      * which computed a random every 3 turns -- takes about 600 ns to do id
      * of len 10, compared to 10000 ns for old version [on 1.6ghz machine]
+     * <p>
+     * <b>TODO</b> The integer value passed to {@link Randonm#nextInt(int)}
+     * will overflow if the length of the character sets passed in is more
+     * than 128. It is possible to mitigate this by truncating the strings,
+     * or calculating the maximum number of characters per invocation of
+     * {@code nextInt()} by taking the logarithm of {@link Integer#MAX_INT}
+     * using the length of the character set as the base. Currently this
+     * method is private to prevent overly long arguments.
      */
-    public static String makeRandomId(int l, String validStartChars, String validNonStartChars) {
+    private static String makeRandomId(int l, String validStartChars, String validNonStartChars) {
         if (l <= 0) return "";
         char[] id = new char[l];
         int s = validStartChars.length();
         int n = validNonStartChars.length();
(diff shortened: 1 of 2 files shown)


grkvlt:
Here are the relevant combinations for before:
```
62^8  =   218 340 105 584 895
```
And possibilities considered for after:
```
36^8  =     2 821 109 907 456
36^9  =   101 559 956 668 416
36^10 = 3 656 158 440 062 976
```

neykov:
LGTM, but note that `makeRandomPassword` has a "special" behaviour in that it will pick initial characters from all pools and only after that merge them and select randomly regardless of the pool. Might be better to extend `makeRandomId` to accept character pools, otherwise all IDs will always follow the format <letter><number><letter or number>{8}

grkvlt:
@neykov I like your suggestion of overloading `makeRandomId()` so I will implement that instead

grkvlt:
@neykov I have updated the `Identifiers#makeRandomId(int)` to have the overload you suggested

utils/common/src/main/java/org/apache/brooklyn/util/text/Identifiers.java
```diff
@@ -87,29 +87,34 @@
...
+     * which computed a random every 3 turns -- takes about 600 ns to do id
+     * of len 10, compared to 10000 ns for old version [on 1.6ghz machine]
+     */
+    public static String makeRandomId(int l, String validStartChars, String validNonStartChars) {
+        if (l <= 0) return "";
         char[] id = new char[l];
-        int d = random.nextInt( (26+26) * (26+26+10) * (26+26+10) * (26+26+10) * (26+26+10));
+        int s = validStartChars.length();
+        int n = validNonStartChars.length();
+        int d = random.nextInt(s * n * n * n * n);
```
neykov: This will overflow for very low `n` (~73).

grkvlt: @neykov Good catch, suppose we could change to a long, or reduce to re-randomising after less characters...

grkvlt: I'm going to use groups of 4 characters, and will limit the character set strings to a length of 128 maximum. Do you think it should error on too many characters, or silently truncate to the first 128?

neykov: Failing on too many characters is better I think
One more option - if input is too long switch to a slower implementation, but one which can handle the input correctly.
For this case though suggest making the parametrized `makeRandomId` private, adding the proper javadoc warnings and adding another method for lower case IDs. It requires the least effort and is good enough for now.

grkvlt:
@neykov I added a note on overflow and made the method private

neykov:
Thanks @grkvlt, merging.

PR #135, opened 2016-05-09
Remove feature brookly-software-cm

duncangrant:


geomacy:
👍 LGTM.  This was causing problems with a build, when tests were failing upon building with a clean Maven repository.  Taking the cm feature out allows the server tests to succeed. At some stage it would be good to produce separate feature.xml for features defined in brooklyn-library, and have these incorporated into a final feature set by brooklyn-dist, but this PR is fine for now.

neykov:
+1 for the quick fix. A proper solution would be to move `karaf/apacke-brooklyn` to `brooklyn-dist`.

PR #136, opened 2016-05-11
Fix/required open ports

kiuby88:
This PR tries to fix the issue described by issue [BROOKLYN-263](  https://issues.apache.org/jira/browse/BROOKLYN-263).

`BrooklynNode` requires `brooklynnode.webconsole.httpPort` config key as a open port. However, the config was not being considered as a open port, because `InboundPortsUtils#getRequiredOpenPorts` just consideres either config keys whose name matches with `*.port` or config keys that are based on `PortRanges` and that are not configured (using default values).

This PR adds to `InboundPortsUtils#getRequiredOpenPorts` the capability to take in account explicit configured config keys that are based on `PortRanges`.

### DIFF
--- a/software/base/src/main/java/org/apache/brooklyn/entity/software/base/InboundPortsUtils.java
+++ b/software/base/src/main/java/org/apache/brooklyn/entity/software/base/InboundPortsUtils.java
@@ -27,13 +27,13 @@ import org.apache.brooklyn.api.entity.Entity;
 import org.apache.brooklyn.api.location.PortRange;
 import org.apache.brooklyn.config.ConfigKey;
 import org.apache.brooklyn.core.entity.EntityInternal;
+import org.apache.brooklyn.util.collections.MutableMap;
 import org.apache.brooklyn.util.collections.MutableSet;
 import org.apache.brooklyn.util.core.flags.TypeCoercions;
 import org.apache.brooklyn.util.guava.Maybe;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Sets;
 import com.google.common.reflect.TypeToken;
@@ -82,7 +82,7 @@ public class InboundPortsUtils {
             Set<ConfigKey<?>> configKeys = Sets.newHashSet(extraConfigKeys);
             configKeys.addAll(entity.getEntityType().getConfigKeys());
             // Also add dynamically added config keys
-            Map<ConfigKey<?>, ?> configuredConfigKeys = ImmutableMap.copyOf(((EntityInternal) entity).config().getBag().getAllConfigAsConfigKeyMap());
+            Map<ConfigKey<?>, ?> configuredConfigKeys = MutableMap.copyOf(((EntityInternal) entity).config().getBag().getAllConfigAsConfigKeyMap());
             configKeys.addAll(configuredConfigKeys.keySet());
 
             if (portRegex == null) {
--- a/software/base/src/main/java/org/apache/brooklyn/entity/software/base/InboundPortsUtils.java
+++ b/software/base/src/main/java/org/apache/brooklyn/entity/software/base/InboundPortsUtils.java
@@ -18,9 +18,10 @@
  */
 package org.apache.brooklyn.entity.software.base;
 
-import com.google.common.collect.ImmutableSet;
-import com.google.common.collect.Sets;
-import com.google.common.reflect.TypeToken;
+import java.util.Collection;
+import java.util.Map;
+import java.util.Set;
+import java.util.regex.Pattern;
 
 import org.apache.brooklyn.api.entity.Entity;
 import org.apache.brooklyn.api.location.PortRange;
@@ -32,9 +33,10 @@ import org.apache.brooklyn.util.guava.Maybe;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.util.Collection;
-import java.util.Set;
-import java.util.regex.Pattern;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Sets;
+import com.google.common.reflect.TypeToken;
 
 public class InboundPortsUtils {
     private static final Logger log = LoggerFactory.getLogger(InboundPortsUtils.class);
@@ -44,10 +46,11 @@ public class InboundPortsUtils {
      * If {@code portsAutoInfer} is {@code true} then
      * return the first value for each {@link org.apache.brooklyn.core.sensor.PortAttributeSensorAndConfigKey}
      * config key {@link PortRange} plus any ports defined with a config key matching the provided regex.
-     * @param entity the Entity
+     *
+     * @param entity         the Entity
      * @param portsAutoInfer if {@code true} then also return the first value for each {@link org.apache.brooklyn.core.sensor.PortAttributeSensorAndConfigKey}
-     * config key {@link PortRange} plus any ports defined with a config keys matching the provided regex
-     * @param portRegex the regex to match config keys that define inbound ports
+     *                       config key {@link PortRange} plus any ports defined with a config keys matching the provided regex
+     * @param portRegex      the regex to match config keys that define inbound ports
      * @return a collection of port numbers
      */
     public static Collection<Integer> getRequiredOpenPorts(Entity entity, Boolean portsAutoInfer, String portRegex) {
@@ -60,11 +63,12 @@ public class InboundPortsUtils {
      * return the first value for each {@link org.apache.brooklyn.core.sensor.PortAttributeSensorAndConfigKey}
      * config key {@link PortRange} plus any ports defined with a config key matching the provided regex.
      * This method also accepts an extra set of config keys in addition to those that are defined in the EntityType of the entity itself.
-     * @param entity the Entity
+     *
+     * @param entity          the Entity
      * @param extraConfigKeys extra set of config key to inspect for inbound ports
-     * @param portsAutoInfer if {@code true} then return the first value for each {@link org.apache.brooklyn.core.sensor.PortAttributeSensorAndConfigKey}
-     * config key {@link PortRange} plus any ports defined with a config keys matching the provided regex
-     * @param portRegex the regex to match config keys that define inbound ports
+     * @param portsAutoInfer  if {@code true} then return the first value for each {@link org.apache.brooklyn.core.sensor.PortAttributeSensorAndConfigKey}
+     *                        config key {@link PortRange} plus any ports defined with a config keys matching the provided regex
+     * @param portRegex       the regex to match config keys that define inbound ports
      * @return a collection of port numbers
      */
     public static Collection<Integer> getRequiredOpenPorts(Entity entity, Set<ConfigKey<?>> extraConfigKeys, Boolean portsAutoInfer, String portRegex) {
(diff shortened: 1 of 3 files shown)


kiuby88:
There is another different fix. Config key names should be review following the patter `*.port`.

software/base/src/main/java/org/apache/brooklyn/entity/software/base/InboundPortsUtils.java
```diff
@@ -98,4 +107,21 @@
...
+    /**
+     * Checks if a configkey can be managed as a config key {@link PortRange}. The method checks if
+     * the config type or the entity config key value are assignable from the {@link PortRange}.
+     * Moreover the method retrieves true if the config key name matches with pattern.
+     *
+     * @param configuredConfigKeys the configured entity config keys
+     * @param configKey            the configkey that could be assigned as a {@link org.apache.brooklyn.api.location.PortRange}
+     * @param portsPattern         pattern to recognized Config key {@link org.apache.brooklyn.api.location.PortRange} of the name   @return if the config key is recognized as a config key {@link PortRange}
+     */
+    private static boolean isAssignableFromPortConfigKey(Map<ConfigKey<?>, ?> configuredConfigKeys, ConfigKey<?> configKey, Pattern portsPattern) {
```
googlielmo: I like this.
Just one thing: The method name only expresses half of the check (the "assignable from" part), not the regex match. I'd recommend keeping it simple and reinstating the regex check out of this method's responsibility (i.e., in the `if` above).

software/base/src/main/java/org/apache/brooklyn/entity/software/base/InboundPortsUtils.java
```diff
@@ -98,4 +107,21 @@
...
+        Object a = configuredConfigKeys.get(configKey);
+        return PortRange.class.isAssignableFrom(configKey.getType())
+                || (portsPattern.matcher(configKey.getName()).matches())
+                || (a != null) && (a instanceof PortRange);
```
googlielmo: Null check not actually needed.

kiuby88:
@googlielmo thanks a lot for your advice.
Squashing?

software/base/src/main/java/org/apache/brooklyn/entity/software/base/InboundPortsUtils.java
```diff
@@ -77,12 +82,17 @@
             Set<ConfigKey<?>> configKeys = Sets.newHashSet(extraConfigKeys);
             configKeys.addAll(entity.getEntityType().getConfigKeys());
             // Also add dynamically added config keys
-            configKeys.addAll(((EntityInternal)entity).config().getBag().getAllConfigAsConfigKeyMap().keySet());
+            Map<ConfigKey<?>, ?> configuredConfigKeys = ImmutableMap.copyOf(((EntityInternal) entity).config().getBag().getAllConfigAsConfigKeyMap());
+            configKeys.addAll(configuredConfigKeys.keySet());
```
ahgittin: why create the intermediate `Immutable`?  is there a risk if a value in the may is null that it will now throw NPE?

software/base/src/main/java/org/apache/brooklyn/entity/software/base/InboundPortsUtils.java
```diff
@@ -77,12 +82,17 @@
...
+
+            if (portRegex == null) {
+                portRegex = ".*\\.port"; // defaults to legacy regex if not specified
+            }
```
ahgittin: please avoid whitespace changes, import reorgs (I know you know this now!), and preference style changes like the above (if it's an egregiously bad style change it)

those things all make PR's harder to review

ahgittin:
minor comment about `ImmutableMap` needs addressing, then good to merge.

if you've not already done so can you also update `brooklyn-docs/yaml/custom-entities.md` section on "Port Inferencing" to include a mention of this behaviour you've added here?

software/base/src/main/java/org/apache/brooklyn/entity/software/base/InboundPortsUtils.java
kiuby88: First, I started to use directly the `((EntityInternal) entity).config().getBag().getAllConfigAsConfigKeyMap()` but I found an exception when this tests was executed, because the entity was accessed by another thread (I found and `InterruptedException` what threw a NPE. 
Then, I decided to use some intermediate element, I used an immutable because the Map should not be modified. Does it make sense?

software/base/src/main/java/org/apache/brooklyn/entity/software/base/InboundPortsUtils.java
kiuby88: Ok, let me roll back the aforementioned changes

kiuby88:
Thanks for your reply, @ahgittin. Let me update the documentation in this PR

software/base/src/main/java/org/apache/brooklyn/entity/software/base/InboundPortsUtils.java
kiuby88: @ahgittin, does it make sense for you?

googlielmo:
@kiuby88 have all comments been addressed?
If so, can someone please merge this? @ahgittin

kiuby88:
@googlielmo I think I have addressed all. But I think @ahgittin would like to take a look about the `ImmutableMap` question comment.

software/base/src/main/java/org/apache/brooklyn/entity/software/base/InboundPortsUtils.java
aledsage: @kiuby88 I didn't follow your description of the problem: "but I found an exception when this tests was executed, because the entity was accessed by another thread (I found and InterruptedException what threw a NPE. " 

Was it an `InterruptedException`? Those really are special cases - just let your method terminate, propagating the exception if you are interrupted. Or was it a NullPointerException, or perhaps a ConcurrentModificationException? The call to `getAllConfigAsConfigKeyMap()` will give you a copy. So you should be safe to use that directly, without taking another copy.

---
In terms of the low-level comment about use of `ImmutableMap`...
 
Alex is right: we need to be careful about where we use `ImmutableMap` because it will throw a `NullPointerException` if any key or value is null. In our config, a user might explicitly have set a config value to null (e.g. to override the default). Therefore we should not try to copy `getAllConfigAsConfigKeyMap()` as an immutable map.

I do really love guava's `ImmutableMap`. I think it's a smart move to fail on null in many many cases. Unfortunately, because of the way nulls are used in various parts of Brooklyn (in particular config and sensors), we can't do it here.

You could change this to `MutableMap.copyOf(...)`. I agree that you don't want to modify the copy, but you are also not sharing it so it's ok if you leave it mutable. (we do not have a `ImmutableMapAllowingNulls`, so let's just make do with `MutableMap`.

software/base/src/test/java/org/apache/brooklyn/entity/brooklynnode/BrooklynNodeTest.java
```diff
@@ -120,18 +132,58 @@ public void run() {
...
+        //EntityTestUtils.assertAttributeEqualsEventually(brooklynNode, BrooklynNode.SERVICE_UP, true);
+        EntityAsserts.assertAttributeEqualsEventually(brooklynNode, BrooklynNode.SERVICE_UP, true);
+        assertTrue(brooklynNode.getRequiredOpenPorts().contains(8082));
+
+        brooklynNode.stop();
+        EntityAsserts.assertAttributeEquals(brooklynNode, BrooklynNode.SERVICE_UP, false);
+    }
+
+    @ImplementedBy(BrooklynNodeWitOpenedPortsImpl.class)
+    public interface BrooklynNodeWitOpenedPorts extends BrooklynNode {
```
aledsage: Typo: "WithOpened" instead of "WitOpened"

software/base/src/test/java/org/apache/brooklyn/entity/brooklynnode/BrooklynNodeTest.java
```diff
@@ -120,18 +132,58 @@ public void run() {
...
+                        .configure(BrooklynNode.ON_EXISTING_PROPERTIES_FILE, BrooklynNode.ExistingFileBehaviour.DO_NOT_USE)
+                        .configure(BrooklynNode.NO_WEB_CONSOLE_AUTHENTICATION, true)
+                        .configure(BrooklynNode.HTTP_PORT, PortRanges.fromString("8082+")));
+
+        app.start(ImmutableList.of(loc));
+        assertNotNull(app);
+
+        log.info("started " + app + " containing " + brooklynNode + " for " + JavaClassNames.niceClassAndMethod());
+
+        //EntityTestUtils.assertAttributeEqualsEventually(brooklynNode, BrooklynNode.SERVICE_UP, true);
```
aledsage: Why commented out? Delete this line, as it is replaced by the line below it I presume.

General note: never leave commented out code, unless there is a very clear comment with it that says what it is for and under what circumstances one would want to uncomment it.

software/base/src/test/java/org/apache/brooklyn/entity/software/base/InboundPortsUtilsTest.java
```diff
@@ -50,4 +46,23 @@ public void testGetRequiredOpenPortsGetsDynamicallyAddedKeys() {
...
+        PortAttributeSensorAndConfigKey newTestConfigKeyPort2 = ConfigKeys.newPortSensorAndConfigKey("new.test.config.port.string.second", "port");
+
+        ConfigKey<String> newTestConfigKeyString = ConfigKeys.newStringConfigKey("new.test.config.key.string");
+        entity.config().set(newTestConfigKeyPort, PortRanges.fromString("8888+"));
+        entity.config().set(newTestConfigKeyPort2, PortRanges.fromInteger(9999));
+        entity.config().set(newTestConfigKeyString, "foo.bar");
+
+        Collection<Integer> dynamicRequiredOpenPorts = InboundPortsUtils.getRequiredOpenPorts(entity, ImmutableSet.<ConfigKey<?>>of(), true, null);
+        Assert.assertTrue(dynamicRequiredOpenPorts.contains(8888));
+        Assert.assertTrue(dynamicRequiredOpenPorts.contains(9999));
```
aledsage: It feels like your code handles an additional case, which is not tested: setting config (e.g of type `ConfigKey<Object>`) where the value is a `PortRanges.fromInteger(2222)`. Your new code will include that, even though the type and name of the config key didn't give us a clue about it.

If that is the case, then worth including it in the test case.

(In contrast, your existing test cases use config keys of type PortAttributeSensorAndConfigKey, so we can tell from the key that it is a port (without looking at the value)).

software/base/src/main/java/org/apache/brooklyn/entity/software/base/InboundPortsUtils.java
kiuby88: @aledsage, thanks a lot for your explanation. 
I see the problem about `ImmutableMap` now and I understand the difference between both of them, so I will use `MutableMap`.
About the exception, the method terminated propagating a `NullPointerException` and I saw a `NullPointerException` in the console, but I did not understand why a `NullPointerException` was thrown. Then, I found an `InterruptedException` in the stack. I do not know if it makes sense for you, but this `InterruptedException` helped me to understand the problem and then I decided to save a copy of the information that was being accessed.

software/base/src/test/java/org/apache/brooklyn/entity/brooklynnode/BrooklynNodeTest.java
kiuby88: Oops! I forgort remove this line. Thanks.

software/base/src/test/java/org/apache/brooklyn/entity/software/base/InboundPortsUtilsTest.java
kiuby88: @aledsage thanks for the deep review. I have addressing all of your comments.

kiuby88:
@aledsage thanks for the deep review. I have addressing all of your comments.

aledsage:
Thanks @kiuby88 - confirmed all tests pass for me locally. Merging now.

PR #137, opened 2016-05-12
FOR REVIEW: easier access for entity config by name (not ConfigKey)

alasdairhodge:
* Inspired by several sightings of client code downcasting `Entity` to `EntityInternal` just for the purpose of accessing the raw config map with simple string keys.
* Ensures that appropriate type coersions are performed on the returned result, including resolution of `DeferredSupplier` values.
* IMO, "ad-hod" config (i.e. named entries in the `brooklyn.config` clause that do not correspond to explicit `ConfigKey` declarations) is extremely useful, especially for customising the behaviour of stock entities without subclassing.

Comments welcome!

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/util/core/flags/TypeCoercions.java
+++ b/core/src/main/java/org/apache/brooklyn/util/core/flags/TypeCoercions.java
@@ -56,6 +56,7 @@ import org.apache.brooklyn.util.JavaGroovyEquivalents;
 import org.apache.brooklyn.util.collections.MutableSet;
 import org.apache.brooklyn.util.collections.QuorumCheck;
 import org.apache.brooklyn.util.collections.QuorumCheck.QuorumChecks;
+import org.apache.brooklyn.util.core.task.DeferredSupplier;
 import org.apache.brooklyn.util.core.task.Tasks;
 import org.apache.brooklyn.util.exceptions.Exceptions;
 import org.apache.brooklyn.util.guava.Maybe;
@@ -138,6 +139,9 @@ public class TypeCoercions {
     @SuppressWarnings({ "unchecked" })
     public static <T> T coerce(Object value, TypeToken<T> targetTypeToken) {
         if (value==null) return null;
+        if (value instanceof DeferredSupplier<?>) {
+            value = ((DeferredSupplier<?>) value).get();
+        }
         Class<? super T> targetType = targetTypeToken.getRawType();
 
         //recursive coercion of parameterized collections and map entries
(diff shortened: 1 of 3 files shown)


kiuby88:
@alasdairhodge I have compiled this PR using my mac OS X 10.9.5. I do not know why Jenkins is not happy

bostko:
It seems the issue is unrelated to this PR. @neykov could you check the test failure?

core/src/main/java/org/apache/brooklyn/util/core/flags/TypeCoercions.java
```diff
@@ -138,6 +139,9 @@ private TypeCoercions() {}
     @SuppressWarnings({ "unchecked" })
     public static <T> T coerce(Object value, TypeToken<T> targetTypeToken) {
         if (value==null) return null;
+        if (value instanceof DeferredSupplier<?>) {
+            value = ((DeferredSupplier<?>) value).get();
```
aledsage: I probably agree with this, but I worry about the implications. The `deferredSupplier.get()` call might be very slow. Are we always doing this in a context where we accept that slowness (e.g. never called by `getNonBlocking()` without first checking that it's not a deferredSupplier).

@alasdairhodge can you rebase this against master - we've fixed quite a few non-deterministic test failures, so hopefully the jenkins build will be better behaved this time.

aledsage:
@alasdairhodge can you rebase this please so we can see if all the tests pass?

I'm still not sure about us calling `deferredSupplier.get()` inside coerce. I have a couple of big worries. First that it might have unanticipated side-effects for other code (e.g. code that is supposed to be non-blocking might now block). And second that it is only doing a small part of what `ValueResolver` does (e.g. handle `DeferredSupplier`, `Task`, `Future`, and deep inspection to resolve values inside maps, lists, etc).

The second point means that it might work in certain use-cases and then break in other use-cases. If we want the value to really be resolved, then perhaps we should go through the same code path as the existing `config().get(key)`, but with the type being passed in rather than being obtained from the type defined in the `ConfigKey`.

core/src/main/java/org/apache/brooklyn/util/core/flags/TypeCoercions.java
alasdairhodge: @aledsage: Now that the coersion methods have been moved into `brooklyn-utils-common`, we can't implement this check as-is without adding a new dependency. `DeferredSupplier` is arguably also a "common utility" so could be moved, but that ovbiously has implications beyong this PR. Let's review.

alasdairhodge: @aledsage: Revisiting this, `DeferredSupplier` simply restates the contract of (Guava) `Supplier`. While superficially redundant, it may offer a (tenuous) distinction between "immediately resolvable" and "deferred-ly resolvable" quantities. Seems weak though, and potentially *very* confusing for users.

PR #139, opened 2016-05-16
BROOKLYN-267: test for brooklyn.parameters inheritance

aledsage:


camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/catalog/CatalogYamlEntityTest.java
```diff
@@ -804,6 +808,52 @@ public void testHardcodedCatalog() throws Exception {
...
+                "services: \n" +
+                "  - type: "+id+":"+version;
+        Entity app = createAndStartApplication(yaml);
+        Entity child = Iterables.getOnlyElement(app.getChildren());
+        ConfigKey<?> configKey = app.getEntityType().getConfigKey("test.myconf");
+        assertNotNull(configKey);
+        assertEquals(app.config().get(configKey), "myval");
+        assertEquals(app.config().get(ConfigKeys.newStringConfigKey("myconf2.from.root")), "myval");
+        assertEquals(child.config().get(ConfigKeys.newStringConfigKey("myconf3.from.root")), "myval");
+        assertEquals(app.config().get(ConfigKeys.newStringConfigKey("myconf2")), "myval");
```
geomacy: Do you want to do any tests on myconf3 in here?

geomacy:
Nice test 👍.  Would it be valuable to add variants that mutate the config value, e.g. if the app overrides the value from the root, and then demonstrating the child 'inheriting' the mutated value?

camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/catalog/CatalogYamlEntityTest.java
aledsage: The assertion for `myconf3` is commented out, because it fails. When we address BROOKLYN-267, then we can uncomment that.

aledsage:
@geomacy thanks for the review. Agree we can add more tests, including for overriding config etc. But I won't do that just now.

aledsage:
The build error looks unrelated:

```
[ERROR] Failed to execute goal on project brooklyn-rest-api: Could not resolve dependencies for project org.apache.brooklyn:brooklyn-rest-api:jar:0.10.0-SNAPSHOT: Could not find artifact org.apache.brooklyn:brooklyn-jsgui:jar:tests:0.10.0-SNAPSHOT in Nexus (http://repository.apache.org/snapshots) -> [Help 1]
```

I'm guessing that is down to some strange build environment thing. Maybe something else was messing with the maven local cache for brooklyn-rest-api while this was being compiled? Or is our build order and dependencies messed up, with brooklyn-rest-api depending on brooklyn-jsgui for tests?!

Merging this PR now anyway.

PR #140, opened 2016-05-17
Fix BROOKLYN-268: DynamicCluster on-fire when initially empty

aledsage:
Fixes `DynamicClusterTest.testServiceUpAfterStartingWithNoMembers`.

It feels like there should be a more general fix to this. But given that the DynamicCluster code is explicitly setting service.isUp to true, then it feels ok to also explicitly say that we know there are no service-not-up-indicators.

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/entity/group/DynamicClusterImpl.java
+++ b/core/src/main/java/org/apache/brooklyn/entity/group/DynamicClusterImpl.java
@@ -259,6 +259,7 @@ public class DynamicClusterImpl extends AbstractGroupImpl implements DynamicClus
                 Preconditions.checkNotNull(getConfig(INITIAL_SIZE), "Cluster initial size overridden to be null. Must be set explicitly.")==0) {
             // if initial size is 0 then override up check to allow zero if empty
             config().set(UP_QUORUM_CHECK, QuorumChecks.atLeastOneUnlessEmpty());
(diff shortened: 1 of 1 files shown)


kiuby88:
Hi @aledsage 
I have tried to reproduce the issue using OS X 10.9.5, Ubuntu 14.04.3 LTS and 12.04.05 LTS all of then with maven 3.3.9 and jdk 1.7 but I was not lucky, I cannot reproduce the error. However, I find the issue that you mentioned in [Jenkins' report](https://builds.apache.org/job/brooklyn-server-pull-requests/393/) of #141. I am a little confused because we should find the same error in the last PRs, for example in #142 where `DynamicClusterImpl` does not include `sensors().set(ServiceStateLogic.SERVICE_NOT_UP_INDICATORS, MutableMap.<String, Object>of());`. WDYT? May the issue be in a different part?

In any case I have also tested this PR in the aforementioned OO.SS and it passes all the tests. So, it could be merged.

aledsage:
@kiuby88 unfortunately the test seems to be non-deterministic (without this fix). That is presumably because there is a race between the enricher setting the service.isUp to null, and DynamicClusterImpl explicitly setting the service.isUp to true.

Thanks for reviewing!

@ahgittin any comments on this (and the subtleties of the `ServiceStateLogic`)? Or is this good to merge?

kiuby88:
I see. Thanks for the explanation, @aledsage.

aledsage:
Merging this now, to fix the failing test. @ahgittin if you have any subsequent comments then please do let me know.

PR #141, opened 2016-05-17
Transformer enricher: support triggerSensors

aledsage:
Some uses of the transformer want to be triggered by multiple sensor
events (e.g. when they will combine two sensor values). This change
allows us to subscribe to multiple sensors, and to trigger the
evaulation when any of those change.

The sourceSensor is now optional. You can use it in the same way as
previously, or you can use just triggerSensors, or you can combine the
two. If combined, then the triggerSensors will cause us to look up the
latest value of the sourceSensor and synthesise an event as though
the source sensor had just published.

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/enricher/stock/AbstractTransformer.java
+++ b/core/src/main/java/org/apache/brooklyn/enricher/stock/AbstractTransformer.java
@@ -18,6 +18,8 @@
  */
 package org.apache.brooklyn.enricher.stock;
 
+import java.util.List;
+
 import org.apache.brooklyn.api.entity.Entity;
 import org.apache.brooklyn.api.entity.EntityLocal;
 import org.apache.brooklyn.api.sensor.AttributeSensor;
@@ -27,12 +29,17 @@ import org.apache.brooklyn.api.sensor.SensorEventListener;
 import org.apache.brooklyn.config.ConfigKey;
 import org.apache.brooklyn.core.config.ConfigKeys;
 import org.apache.brooklyn.core.enricher.AbstractEnricher;
+import org.apache.brooklyn.core.mgmt.BrooklynTaskTags;
 import org.apache.brooklyn.core.sensor.BasicSensorEvent;
+import org.apache.brooklyn.core.sensor.Sensors;
+import org.apache.brooklyn.util.collections.MutableMap;
+import org.apache.brooklyn.util.core.task.Tasks;
 import org.apache.brooklyn.util.javalang.JavaClassNames;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.google.common.base.Function;
+import com.google.common.collect.ImmutableList;
 import com.google.common.reflect.TypeToken;
 
 @SuppressWarnings("serial")
@@ -40,12 +47,18 @@ public abstract class AbstractTransformer<T,U> extends AbstractEnricher implemen
 
     private static final Logger LOG = LoggerFactory.getLogger(AbstractTransformer.class);
 
-    public static ConfigKey<Entity> PRODUCER = ConfigKeys.newConfigKey(Entity.class, "enricher.producer");
+    public static final ConfigKey<Entity> PRODUCER = ConfigKeys.newConfigKey(Entity.class, "enricher.producer");
 
-    public static ConfigKey<Sensor<?>> SOURCE_SENSOR = ConfigKeys.newConfigKey(new TypeToken<Sensor<?>>() {}, "enricher.sourceSensor");
+    public static final ConfigKey<Sensor<?>> SOURCE_SENSOR = ConfigKeys.newConfigKey(new TypeToken<Sensor<?>>() {}, "enricher.sourceSensor");
 
-    public static ConfigKey<Sensor<?>> TARGET_SENSOR = ConfigKeys.newConfigKey(new TypeToken<Sensor<?>>() {}, "enricher.targetSensor");
+    public static final ConfigKey<Sensor<?>> TARGET_SENSOR = ConfigKeys.newConfigKey(new TypeToken<Sensor<?>>() {}, "enricher.targetSensor");
     
+    public static final ConfigKey<List<? extends Sensor<?>>> TRIGGER_SENSORS = ConfigKeys.newConfigKey(
+            new TypeToken<List<? extends Sensor<?>>>() {}, 
+            "enricher.triggerSensors",
+            "Sensors that will trigger re-evaluation",
+            ImmutableList.<Sensor<?>>of());
+
     protected Entity producer;
     protected Sensor<T> sourceSensor;
     protected Sensor<U> targetSensor;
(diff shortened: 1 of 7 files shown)


core/src/main/java/org/apache/brooklyn/enricher/stock/AbstractTransformer.java
```diff
@@ -59,27 +72,55 @@ public void setEntity(EntityLocal entity) {
         super.setEntity(entity);
 
         this.producer = getConfig(PRODUCER) == null ? entity: getConfig(PRODUCER);
-        this.sourceSensor = (Sensor<T>) getRequiredConfig(SOURCE_SENSOR);
+        this.sourceSensor = (Sensor<T>) getConfig(SOURCE_SENSOR);
         Sensor<?> targetSensorSpecified = getConfig(TARGET_SENSOR);
+        List<? extends Sensor<?>> triggerSensorsSpecified = getConfig(TRIGGER_SENSORS);
+        List<? extends Sensor<?>> triggerSensors = triggerSensorsSpecified != null ? triggerSensorsSpecified : ImmutableList.<Sensor<?>>of();
```
neykov: `triggerSensorsSpecified` could be the empty array, but not likely.

neykov: doesn't really matter, ignore.

core/src/main/java/org/apache/brooklyn/enricher/stock/AbstractTransformer.java
```diff
@@ -59,27 +72,55 @@ public void setEntity(EntityLocal entity) {
...
+                        AbstractTransformer.this.onEvent(new BasicSensorEvent(sourceSensor, producer, value, event.getTimestamp()));
+                    } else {
+                        // Dangerous casting, but will assume that the transform doesn't care - otherwise 
+                        // it would have declared a sourceSensor!
+                        AbstractTransformer.this.onEvent((SensorEvent<T>)event);
+                    }
+                }
+            };
+            for (Object sensor : triggerSensors) {
+                if (sensor instanceof String) {
```
neykov: `triggerSensors` is of type `List<? extends Sensor<?>>` so can't be string

core/src/main/java/org/apache/brooklyn/enricher/stock/AbstractTransformer.java
```diff
@@ -59,27 +72,55 @@ public void setEntity(EntityLocal entity) {
...
+        List<? extends Sensor<?>> triggerSensorsSpecified = getConfig(TRIGGER_SENSORS);
+        List<? extends Sensor<?>> triggerSensors = triggerSensorsSpecified != null ? triggerSensorsSpecified : ImmutableList.<Sensor<?>>of();
         this.targetSensor = targetSensorSpecified!=null ? (Sensor<U>) targetSensorSpecified : (Sensor<U>) this.sourceSensor;
-        if (producer.equals(entity) && targetSensorSpecified==null) {
+        if (targetSensor == null) {
+            throw new IllegalArgumentException("Enricher "+JavaClassNames.simpleClassName(this)+" has no "+TARGET_SENSOR.getName()+", and it cannot be inferred as "+SOURCE_SENSOR.getName()+" is also not set");
+        }
+        if (sourceSensor == null && triggerSensors.isEmpty()) {
+            throw new IllegalArgumentException("Enricher "+JavaClassNames.simpleClassName(this)+" has no "+SOURCE_SENSOR.getName()+" and no "+TRIGGER_SENSORS.getName());
+        }
```
neykov: Does it make sense to use `enricher.triggerSensors` with anything else than `targetValue`? Probably not, and as you comment would be dangerous to allow arbitrary transformers due to casting issues.

core/src/main/java/org/apache/brooklyn/enricher/stock/AbstractTransformer.java
```diff
@@ -59,27 +72,55 @@ public void setEntity(EntityLocal entity) {
...
+            SensorEventListener<Object> triggerListener = new SensorEventListener<Object>() {
+                @Override public void onEvent(SensorEvent<Object> event) {
+                    if (sourceSensor != null) {
+                        // Simulate an event, as though our sourceSensor changed
+                        Object value = producer.getAttribute((AttributeSensor<?>)sourceSensor);
+                        AbstractTransformer.this.onEvent(new BasicSensorEvent(sourceSensor, producer, value, event.getTimestamp()));
+                    } else {
+                        // Dangerous casting, but will assume that the transform doesn't care - otherwise 
+                        // it would have declared a sourceSensor!
+                        AbstractTransformer.this.onEvent((SensorEvent<T>)event);
```
neykov: Personal preference - I'd consider passing null values here for sensor & value instead. If `sourceSensor` is null it would make more sense to fail with a NPE for the case where the transformer expects a specific sensor.

neykov:
Minor comments only, mostly for discussion. Can be merged as is.

core/src/main/java/org/apache/brooklyn/enricher/stock/AbstractTransformer.java
aledsage: You could use triggerSensors in combination with sourceSensor, as an additional way of it recalculating a value from the sourceSensor. Not quite sure when you'd want to do that but it should work!

core/src/main/java/org/apache/brooklyn/enricher/stock/AbstractTransformer.java
aledsage: See `TransformingEnricherTest.testTriggeringSensorNamesResolvedFromStrings` for a test that it works with strings passed in. Given that it's yaml, then a user could easily pass in such a String value. I'm not convinced that our `TypeCoercions` would kick-in for the values within the triggerSensors' `List`, to automatically convert form String to sensor.

core/src/main/java/org/apache/brooklyn/enricher/stock/AbstractTransformer.java
aledsage: Agreed. I'll change that to pass null, and check that the tests pass.

aledsage:
Test failure is unrelated; will deal with that in a separate pull request. Closing and re-opening to kick off jenkins again.

core/src/main/java/org/apache/brooklyn/enricher/stock/AbstractTransformer.java
neykov: That's pretty nasty casting there :). Why not make `TRIGGER_SENSORS` typed as list of objects - you are using it as such anyway.

I was actually surprised that coercion doesn't happen for this case - I see it as a bug/shortcoming of Brooklyn. I checked the code path and it is [coercing to `Object.class` on purpose](https://github.com/apache/brooklyn-server/blob/d03f254ba19472e3fb5e0a7260cb07679991684a/core/src/main/java/org/apache/brooklyn/core/config/BasicConfigKey.java#L292). Any idea why that is?

core/src/main/java/org/apache/brooklyn/enricher/stock/AbstractTransformer.java
```diff
@@ -59,27 +72,55 @@ public void setEntity(EntityLocal entity) {
...
+            SensorEventListener<Object> triggerListener = new SensorEventListener<Object>() {
+                @Override public void onEvent(SensorEvent<Object> event) {
+                    if (sourceSensor != null) {
+                        // Simulate an event, as though our sourceSensor changed
+                        Object value = producer.getAttribute((AttributeSensor<?>)sourceSensor);
+                        AbstractTransformer.this.onEvent(new BasicSensorEvent(sourceSensor, producer, value, event.getTimestamp()));
+                    } else {
+                        // Dangerous casting, but will assume that the transform doesn't care - otherwise 
+                        // it would have declared a sourceSensor!
+                        AbstractTransformer.this.onEvent(null);
```
neykov: Comment is no longer applicable.

grkvlt:
@aledsage If you do `git commit --amend && git push --force branch` then GitHub will rebuild without the noise of needing to close and re-open the PR

aledsage:
Thanks @neykov - I've updated comment and will now merge.

PR #142, opened 2016-05-17
Adds TestEndpointReachable (for yaml test-cases)

aledsage:


test-framework/src/main/java/org/apache/brooklyn/test/framework/TestEndpointReachableImpl.java
```diff
@@ -0,0 +1,180 @@
...
+            throw new IllegalArgumentException(String.format("The entity [%s] with endpoint [%s] has no port", getClass().getName(), endpoint));
+        }
+        return result;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public void stop() {
+        ServiceStateLogic.setExpectedState(this, Lifecycle.STOPPING);
```
neykov: Should be `STOPPED`

test-framework/src/test/java/org/apache/brooklyn/test/framework/TestEndpointReachableTest.java
```diff
@@ -0,0 +1,222 @@
...
+    }
+
+    protected ServerSocket openServerPort() {
+        int startPort = 58767;
+        int endPort = 60000;
+        int port = startPort;
+        InetAddress localAddress = Networking.getLocalHost();
+        do {
+            try {
+                ServerSocket ss = new ServerSocket(port, 1024, localAddress);
```
neykov: Passing port `0` will let the OS allocate an available port - no need for the retry logic.

neykov:
Looks good, nice test coverage. What do you think about adding an option to reverse the test - successful if can't connect (for example by adding `assertFails: true`). It's as useful doing the reverse check.
Current implementation of `Networking.isReachable` might take a while to give up if the port is not reachable.

test-framework/src/test/java/org/apache/brooklyn/test/framework/TestEndpointReachableTest.java
aledsage: Ah, doh: you're absolutely right!

aledsage:
I also wondered about the inverse. Agree it's worth adding. For consistency with the other TestCase types, I'll go for YAML syntax like that below (where it defaults to `reachable: true`):

```
- type: org.apache.brooklyn.test.framework.TestEndpointReachable
  name: Endpoint reachable
  brooklyn.config:
    targetId: entity-under-test
    endpointSensor: datastore.url
    assertions:
      reachable: false
```

I'm not going to worry about `Networking.isReachable` being slow to fail just now.

neykov:
+1 for proposed syntax

neykov:
Good to merge, once Jenkins is happy. 
Current failure: `[ERROR] TestEndpointReachableImpl.java[47:1] (imports) IllegalImport: Import from illegal package - com.google.api.client.repackaged.com.google.common.base.Objects.`

aledsage:
Thanks @neykov - merging now.

PR #143, opened 2016-05-18
Adds class renames for downstream Clocker project

nakomis:


aledsage:
It feels like this should be an override in downstream projects, rather than in Brooklyn itself. Do you want to close this?

As discussed with @nakomis offline, a longer-term change we want to make is... instead of a single `deserializingClassRenames.properties`, we'd like to pick up multiple files so that downstream projects can be additive.

We could have a '/brooklyn/deserializingClassRenames' resource folder. For any `.properties` file inside that folder (how ever it was added to the classpath), we'd load it and include it in the renames.

We could also do multiple calls to rename for a given class, until the name became stable (or an infinite loop was detected!). Then we can have separate files for 0.7.0->0.8.0, 0.8.0->0.9.0.

Not sure how that would working with OSGi with Karaf @neykov ? What different configuration mechanism would we use, instead of searching the classpath?

aledsage:
@nakomis did we conclude that this was the responsibility of downstream projects, and that working on https://github.com/apache/brooklyn-server/pull/144 was the right way to go?

If so, can you please close this PR.

aledsage:
@nakomis do you want to close this PR, while we discuss https://github.com/apache/brooklyn-server/pull/144?

PR #145, opened 2016-05-18
Suppress exceptions in the isRunning check for Windows

bostko:
Some Windows Blueprints require restart.
That's why for the isRunning check,
I catched WinRM IO Exceptions which could happen because of the restart.

### DIFF
--- a/software/base/src/main/java/org/apache/brooklyn/entity/software/base/VanillaWindowsProcessWinRmDriver.java
+++ b/software/base/src/main/java/org/apache/brooklyn/entity/software/base/VanillaWindowsProcessWinRmDriver.java
@@ -22,10 +22,14 @@ import org.apache.brooklyn.api.entity.EntityLocal;
 import org.apache.brooklyn.core.entity.Attributes;
 import org.apache.brooklyn.location.winrm.WinRmMachineLocation;
 import org.apache.brooklyn.util.core.internal.winrm.WinRmTool;
+import org.apache.brooklyn.util.exceptions.Exceptions;
 import org.apache.brooklyn.util.net.UserAndHostAndPort;
 import org.apache.brooklyn.util.text.Strings;
+import org.apache.cxf.interceptor.Fault;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
+import javax.xml.ws.WebServiceException;
 
 public class VanillaWindowsProcessWinRmDriver extends AbstractSoftwareProcessWinRmDriver implements VanillaWindowsProcessDriver {
     private static final Logger LOG = LoggerFactory.getLogger(VanillaWindowsProcessWinRmDriver.class);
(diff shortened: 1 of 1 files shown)


software/base/src/main/java/org/apache/brooklyn/entity/software/base/VanillaWindowsProcessWinRmDriver.java
```diff
@@ -80,9 +80,15 @@ public void launch() {
...
-        int exitCode = executeCommandInTask(
-                getEntity().getConfig(VanillaWindowsProcess.CHECK_RUNNING_COMMAND),
-                getEntity().getConfig(VanillaWindowsProcess.CHECK_RUNNING_POWERSHELL_COMMAND), "is-running-command");
+        int exitCode = 1;
+        try {
+            exitCode = executeCommandInTask(
+                    getEntity().getConfig(VanillaWindowsProcess.CHECK_RUNNING_COMMAND),
+                    getEntity().getConfig(VanillaWindowsProcess.CHECK_RUNNING_POWERSHELL_COMMAND), "is-running-command");
+        } catch (javax.xml.ws.WebServiceException/*Wraps Soap exceptions*/|org.apache.cxf.interceptor.Fault/*Wraps IO exceptions*/ e) {
+            LOG.warn(getEntity() + " isRunning check failed. Executing WinRM command failed.", e);
```
aledsage: It's worth including a comment with the reason here (as you did in the description of the PR). For example:

    // If machine is restarting, then will get WinRM IOExceptions - don't propagate such exceptions

aledsage:
Looks good. One minor comment worth adding to the code, then good to merge.

nakomis:
A future improvement would be for `AbstractSoftwareProcessWinRmDriver.rebootAndWait` to set a sensor to indicate that the machine is rebooting, which could be checked in the catch block, but I don't think that's worth holding up this PR. Other than that, LGTM

software/base/src/main/java/org/apache/brooklyn/entity/software/base/VanillaWindowsProcessWinRmDriver.java
```diff
@@ -80,9 +84,22 @@ public void launch() {
 
     @Override
     public boolean isRunning() {
-        int exitCode = executeCommandInTask(
-                getEntity().getConfig(VanillaWindowsProcess.CHECK_RUNNING_COMMAND),
-                getEntity().getConfig(VanillaWindowsProcess.CHECK_RUNNING_POWERSHELL_COMMAND), "is-running-command");
+        int exitCode = 1;
```
aledsage: Very minor (for next time). I wouldn't bother giving this an initial value. It will always get a value inside the try-block, and the catch block will always either return or throw. So the initial value of `1` will always get overwritten. Better to be explicit about that by not giving it any initial value.

software/base/src/main/java/org/apache/brooklyn/entity/software/base/VanillaWindowsProcessWinRmDriver.java
```diff
@@ -80,9 +84,22 @@ public void launch() {
...
+        try {
+            exitCode = executeCommandInTask(
+                    getEntity().getConfig(VanillaWindowsProcess.CHECK_RUNNING_COMMAND),
+                    getEntity().getConfig(VanillaWindowsProcess.CHECK_RUNNING_POWERSHELL_COMMAND), "is-running-command");
+        } catch (Exception e) {
+            // If machine is restarting, then will get WinRM IOExceptions - don't propagate such exceptions
+            Throwable interestingCause = Exceptions.getFirstThrowableOfType(e, WebServiceException.class) != null ?
+                    Exceptions.getFirstThrowableOfType(e, WebServiceException.class/*Wraps Soap exceptions*/) : Exceptions.getFirstThrowableOfType(e, Fault.class/*Wraps IO exceptions*/);
+            if (interestingCause != null) {
+                LOG.warn(getEntity() + " isRunning check failed. Executing WinRM command failed.", e);
```
aledsage: I wonder how many times / how often we might get this exception logged if the VM is rebooting for several minutes.

aledsage:
A couple of minor thoughts, but merging now.

PR #146, opened 2016-05-20
Fix BROOKLYN-271: Avoiding object hashcode in entities' toString.

kiuby88:
Fixing [BROOKLYN-271](https://issues.apache.org/jira/browse/BROOKLYN-271)
Avoiding object references in catalog ConfigKey cards, for example:

captura de pantalla 2016-05-19 a la s 17 46 04

Last picture shows that default value of `softwareProcess.lifecycleTasks` config key does not contain the object hashcode `@XXXXX`. Then, objects `toString` methods are modified to avoid the hashcode.

### DIFF
--- a/api/src/main/java/org/apache/brooklyn/api/entity/EntitySpec.java
+++ b/api/src/main/java/org/apache/brooklyn/api/entity/EntitySpec.java
@@ -432,5 +432,10 @@ public class EntitySpec<T extends Entity> extends AbstractBrooklynObjectSpec<T,E
     private void checkMutable() {
         if (immutable) throw new IllegalStateException("Cannot modify immutable entity spec "+this);
     }
-    
+
+    @Override
+    public String toString(){
+        return getClass().getName();
+    }
+
 }
--- a/api/src/main/java/org/apache/brooklyn/api/entity/EntitySpec.java
+++ b/api/src/main/java/org/apache/brooklyn/api/entity/EntitySpec.java
@@ -433,9 +433,4 @@ public class EntitySpec<T extends Entity> extends AbstractBrooklynObjectSpec<T,E
         if (immutable) throw new IllegalStateException("Cannot modify immutable entity spec "+this);
     }
 
-    @Override
-    public String toString(){
-        return getClass().getName();
-    }
-
 }
--- a/core/src/main/java/org/apache/brooklyn/entity/group/zoneaware/ProportionalZoneFailureDetector.java
+++ b/core/src/main/java/org/apache/brooklyn/entity/group/zoneaware/ProportionalZoneFailureDetector.java
@@ -21,6 +21,7 @@ package org.apache.brooklyn.entity.group.zoneaware;
 import org.apache.brooklyn.api.location.Location;
 import org.apache.brooklyn.util.time.Duration;
 
+import com.google.common.base.Objects;
 import com.google.common.base.Ticker;
 
 public class ProportionalZoneFailureDetector extends AbstractZoneFailureDetector {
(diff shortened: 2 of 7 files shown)


kiuby88:
@aledsage I think it should be better than https://github.com/apache/brooklyn-docs/pull/68

aledsage:
Thanks @kiuby88 - agreed this will improve the auto-generated entity docs (as per BROOKLYN-271).

In general, I like the toString to give more specifics about the object where possible. That is possible for some of these, and not for others. I'll comment on individual methods where I think we could include more.

core/src/main/java/org/apache/brooklyn/entity/group/zoneaware/ProportionalZoneFailureDetector.java
```diff
@@ -56,4 +56,10 @@ protected boolean doHasFailed(Location loc, ZoneHistory zoneHistory) {
             return numDatapoints >= minDatapoints && proportionFailure >= maxProportionFailures;
         }
     }
+
+    @Override
+    public String toString(){
+        return getClass().getName();
```
aledsage: I'd go for the code below (using `com.google.common.base.Objects`):

    @Override
    public String toString() {
        return Objects.toStringHelper(this)
                .add("minDatapoints", minDatapoints)
                .add("timeToConsider",  timeToConsider)
                .add("maxProportionFailures", maxProportionFailures)
        .toString();
    }

api/src/main/java/org/apache/brooklyn/api/entity/EntitySpec.java
```diff
@@ -432,5 +432,10 @@ public Entity getParent() {
     private void checkMutable() {
         if (immutable) throw new IllegalStateException("Cannot modify immutable entity spec "+this);
     }
-    
+
+    @Override
+    public String toString(){
+        return getClass().getName();
```
aledsage: Interestingly, there is already a `toString`, inherited from `AbstractBrooklynObjectSpec`. That will include the entity type, and will append `+"@"+Integer.toHexString(System.identityHashCode(this)`!

I can see why we might want the identityHashCode (so that when looking at the logs we can tell if something is using the same instance). But on-balance, I think it's fine to not include it. Therefore, instead of this impl I suggest you change `AbstractBrooklynObjectSpec` to be:

    @Override
    public String toString() {
        return Objects.toStringHelper(this).omitNullValues()
                .add("type", type)
                .add("displayName", displayName)
                .toString();
    }

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/BrooklynImageChooser.java
```diff
@@ -371,7 +371,12 @@ public ImageChooserFromOrdering cloneFor(ConfigBag config) {
             } else {
                 return this;
             }
-        }        
+        }
+
+        @Override
+        public String toString(){
+            return getClass().getName();
+        }
```
kiuby88: @aledsage, this tries to avoid the following entry in the catalog  for `JcloudsLocation`.
captura de pantalla 2016-05-31 a la s 10 24 10

I think this impl is enough, maybe [ordering](https://github.com/apache/brooklyn-server/blob/master/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/BrooklynImageChooser.java#L330)?

aledsage:
LGTM, thanks @kiuby88 - merging. I incorrectly marked BROOKLYN-271 as resolved before this PR (and https://github.com/apache/brooklyn-library/pull/38) were merged - sorry about that, and thanks for spotting!

PR #148, opened 2016-05-20
Test and fix locations not being unmanaged

neykov:
Only `MachineLocation` locations created by `SoftwareProcess` are unmanaged. Currently there's no owner for `LocationResolver` locations so there's no one to unmanage them.

The changes will mark locations created as part of bringing up a spec as owner by the entity where they are attached. When the entity is unmanaged the locations will be unmanaged as well.

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/entity/AbstractEntity.java
+++ b/core/src/main/java/org/apache/brooklyn/core/entity/AbstractEntity.java
@@ -72,6 +72,8 @@ import org.apache.brooklyn.core.internal.storage.BrooklynStorage;
 import org.apache.brooklyn.core.internal.storage.Reference;
 import org.apache.brooklyn.core.internal.storage.impl.BasicReference;
 import org.apache.brooklyn.core.location.Locations;
+import org.apache.brooklyn.core.mgmt.BrooklynTags;
+import org.apache.brooklyn.core.mgmt.BrooklynTags.NamedStringTag;
 import org.apache.brooklyn.core.mgmt.internal.EffectorUtils;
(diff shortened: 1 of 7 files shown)


aledsage:
This looks good, but I think the `log.warn` is too extreme:

For example, if you deploy this simple blueprint:

    location: localhost
    services:
    - type: org.apache.brooklyn.entity.group.DynamicCluster
      brooklyn.config:
        cluster.initial.size: 1
        memberSpec:
          $brooklyn:entitySpec:

Then it logs:

    2016-05-27 10:02:17,140 WARN  Adding location LocalhostMachineProvisioningLocation{id=mB4Tjwyt, name=localhost} to entity EmptySoftwareProcessImpl{id=iVSKZf22} which is already owner by another entity xUAFEbnM. Locations which are owned by a specific entity should not be shared with other entities as they will be unmanaged together with their owner, regardless of other references to them.

This is because the `MachineProvisionigLocation` is passed by the app to the child SoftwareProcess. But it is already owned by the app. The DynamicCluster passes the location via `entity.start(location)`.

Note that we don't get this log message if we were just using a simple app + entity (rather than a dynamically resized `DynamicCluster`). This is because the location would be passed differently to the child entity: in `MachineLifecycleEffectorTasks.getLocation`, it finds the location from its ancestor rather than calling `addLocations()`.

aledsage:
I'll take a look at changing the logging.

aledsage:
Superseded by https://github.com/apache/brooklyn-server/pull/160, which changes the logging and also fixed the failing tests `BrooklynMementoPersisterInMemoryTest.testDeleteAndLoadMemento`.

PR #150, opened 2016-05-20
Ability to perform installation on a Windows Blueprint which changes Computer Name and requires reboot

bostko:
- update the `install.reboot.required` action to use computername when running the install command and the reboot command
- set small and reasonable default value for connection retries of winrm4j client

### DIFF
--- a/software/base/src/main/java/org/apache/brooklyn/entity/software/base/AbstractSoftwareProcessWinRmDriver.java
+++ b/software/base/src/main/java/org/apache/brooklyn/entity/software/base/AbstractSoftwareProcessWinRmDriver.java
@@ -29,6 +29,7 @@ import java.util.Map;
 import java.util.concurrent.Callable;
 import java.util.concurrent.TimeUnit;
 
+import org.apache.cxf.interceptor.Fault;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -55,6 +56,8 @@ import org.apache.brooklyn.util.repeat.Repeater;
 import org.apache.brooklyn.util.stream.Streams;
 import org.apache.brooklyn.util.text.Strings;
 import org.apache.brooklyn.util.time.Duration;
+
+import javax.xml.ws.WebServiceException;
 
 public abstract class AbstractSoftwareProcessWinRmDriver extends AbstractSoftwareProcessDriver implements NativeWindowsScriptRunner {
     private static final Logger LOG = LoggerFactory.getLogger(AbstractSoftwareProcessWinRmDriver.class);
@@ -161,7 +164,12 @@ public abstract class AbstractSoftwareProcessWinRmDriver extends AbstractSoftwar
     }
 
     protected int executeCommandInTask(String command, String psCommand, String phase) {
+        return executeCommandInTask(command, psCommand, phase, null);
+    }
+
+    protected int executeCommandInTask(String command, String psCommand, String phase, String ntDomain) {
         return newScript(phase)
+                .setNtDomain(ntDomain)
                 .setCommand(command)
                 .setPsCommand(psCommand)
                 .failOnNonZeroResultCode()
@@ -278,10 +286,16 @@ public abstract class AbstractSoftwareProcessWinRmDriver extends AbstractSoftwar
         }
 
         WinRmToolResponse response;
+
+        ImmutableMap.Builder winrmProps = ImmutableMap.builder();
+        if (flags.get(WinRmTool.COMPUTER_NAME) != null) {
+            winrmProps.put(WinRmTool.COMPUTER_NAME, flags.get(WinRmTool.COMPUTER_NAME));
+        }
+
         if (Strings.isBlank(regularCommand)) {
-            response = getLocation().executePsScript(ImmutableList.of(powerShellCommand));
+            response = getLocation().executePsScript(winrmProps.build(), ImmutableList.of(powerShellCommand));
         } else {
-            response = getLocation().executeCommand(ImmutableList.of(regularCommand));
+            response = getLocation().executeCommand(winrmProps.build(), ImmutableList.of(regularCommand));
         }
 
         if (currentTask != null) {
@@ -321,14 +335,35 @@ public abstract class AbstractSoftwareProcessWinRmDriver extends AbstractSoftwar
     }
 
     public void rebootAndWait() {
+        rebootAndWait(null);
+    }
+
+    public void rebootAndWait(String hostname) {
         try {
-            executePsScriptNoRetry(ImmutableList.of("Restart-Computer -Force"));
+            if (hostname != null) {
+                getLocation().executePsScript(ImmutableMap.of(WinRmTool.COMPUTER_NAME, hostname), ImmutableList.of("Restart-Computer -Force"));
+            } else {
+                getLocation().executePsScript(ImmutableList.of("Restart-Computer -Force"));
+            }
         } catch (Exception e) {
-            // Restarting the computer will cause the command to fail; ignore the exception and continue
-            Exceptions.propagateIfFatal(e);
+            Throwable interestingCause = findExceptionCausedByWindowsRestart(e);
+            if (interestingCause != null) {
+                LOG.debug("Restarting... exception while closing winrm session from the restart command {}", getEntity(), e);
+            } else {
+                throw e;
+            }
         }
+
         waitForWinRmStatus(false, entity.getConfig(VanillaWindowsProcess.REBOOT_BEGUN_TIMEOUT));
         waitForWinRmStatus(true, entity.getConfig(VanillaWindowsProcess.REBOOT_COMPLETED_TIMEOUT)).getWithError();
+    }
+
+    /**
+     * If machine is restarting, then will get WinRM IOExceptions
+     */
+    protected Throwable findExceptionCausedByWindowsRestart(Exception e) {
+        return Exceptions.getFirstThrowableOfType(e, WebServiceException.class) != null ?
+                Exceptions.getFirstThrowableOfType(e, WebServiceException.class/*Wraps Soap exceptions*/) : Exceptions.getFirstThrowableOfType(e, Fault.class/*Wraps IO exceptions*/);
     }
 
     private String getDirectory(String fileName) {
--- a/software/base/src/main/java/org/apache/brooklyn/entity/software/base/VanillaWindowsProcess.java
+++ b/software/base/src/main/java/org/apache/brooklyn/entity/software/base/VanillaWindowsProcess.java
@@ -20,6 +20,7 @@ package org.apache.brooklyn.entity.software.base;
 
 import java.util.Collection;
 
+import com.google.common.annotations.Beta;
 import org.apache.brooklyn.api.catalog.Catalog;
 import org.apache.brooklyn.api.catalog.CatalogConfig;
 import org.apache.brooklyn.api.entity.ImplementedBy;
@@ -96,9 +97,12 @@ public interface VanillaWindowsProcess extends AbstractVanillaProcess {
 
     ConfigKey<Boolean> PRE_INSTALL_REBOOT_REQUIRED = ConfigKeys.newBooleanConfigKey("pre.install.reboot.required",
             "indicates that a reboot should be performed after the pre-install command is run", false);
-    
+
+    @Beta
     ConfigKey<Boolean> INSTALL_REBOOT_REQUIRED = ConfigKeys.newBooleanConfigKey("install.reboot.required",
-            "indicates that a reboot should be performed after the install command is run", false);
+            "indicates that a reboot should be performed after the install command is run." +
+            "When running the install command and the reboot command this parameter adds computername when authenticating.",
+            false);
     
     ConfigKey<Boolean> CUSTOMIZE_REBOOT_REQUIRED = ConfigKeys.newBooleanConfigKey("customize.reboot.required",
             "indicates that a reboot should be performed after the customize command is run", false);
--- a/software/base/src/main/java/org/apache/brooklyn/entity/software/base/VanillaWindowsProcessWinRmDriver.java
+++ b/software/base/src/main/java/org/apache/brooklyn/entity/software/base/VanillaWindowsProcessWinRmDriver.java
@@ -20,6 +20,7 @@ package org.apache.brooklyn.entity.software.base;
 
 import org.apache.brooklyn.api.entity.EntityLocal;
 import org.apache.brooklyn.core.entity.Attributes;
+import org.apache.brooklyn.entity.software.base.lifecycle.WinRmExecuteHelper;
 import org.apache.brooklyn.location.winrm.WinRmMachineLocation;
 import org.apache.brooklyn.util.core.internal.winrm.WinRmTool;
 import org.apache.brooklyn.util.exceptions.Exceptions;
@@ -49,15 +50,27 @@ public class VanillaWindowsProcessWinRmDriver extends AbstractSoftwareProcessWin
 
     @Override
     public void install() {
+        // TODO: At some point in the future, this should probably be refactored to get the name of the machine in WinRmMachineLocation and set it as the hostname sensor
+        String hostname = null;
+        if (entity.getConfig(VanillaWindowsProcess.INSTALL_REBOOT_REQUIRED)) {
+            WinRmExecuteHelper checkHostnameTask = newScript("Checking hostname")
+                    .setCommand("hostname")
+                    .failOnNonZeroResultCode()
+                    .gatherOutput();
+            checkHostnameTask.execute();
+            hostname = Strings.trimEnd(checkHostnameTask.getResultStdout());
+        }
+
         // TODO: Follow install path of VanillaSoftwareProcessSshDriver
         if(Strings.isNonBlank(getEntity().getConfig(VanillaWindowsProcess.INSTALL_COMMAND)) || Strings.isNonBlank(getEntity().getConfig(VanillaWindowsProcess.INSTALL_POWERSHELL_COMMAND))) {
             executeCommandInTask(
                     getEntity().getConfig(VanillaWindowsProcess.INSTALL_COMMAND),
                     getEntity().getConfig(VanillaWindowsProcess.INSTALL_POWERSHELL_COMMAND),
-                    "install-command");
+                    "install-command",
+                    hostname);
         }
         if (entity.getConfig(VanillaWindowsProcess.INSTALL_REBOOT_REQUIRED)) {
-            rebootAndWait();
+            rebootAndWait(hostname);
         }
     }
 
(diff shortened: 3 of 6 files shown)


software/base/src/main/java/org/apache/brooklyn/entity/software/base/VanillaWindowsProcessWinRmDriver.java
```diff
@@ -76,10 +77,13 @@ public void customize() {
 
     @Override
     public void launch() {
-        executeCommandInTask(
```
bostko: Counter to the launch command, what do you think about making install command obligatory?

software/winrm/src/main/java/org/apache/brooklyn/util/core/internal/winrm/WinRmTool.java
```diff
@@ -46,7 +46,7 @@
     ConfigKey<String> PROP_HOST = newStringConfigKey("host", "Host to connect to (required)", null);
     ConfigKey<Integer> PROP_PORT = ConfigKeys.newIntegerConfigKey("port", "WinRM port to use when connecting to the remote machine");
     ConfigKey<Boolean> USE_HTTPS_WINRM = ConfigKeys.newBooleanConfigKey("winrm.useHttps", "The parameter tells the machine sensors whether the winrm port is over https. If the parameter is true then 5986 will be used as a winrm port.", false);
-    ConfigKey<Integer> RETRIES_OF_NETWORK_FAILURES = ConfigKeys.newIntegerConfigKey("retriesOfNetworkFailures", "The parameter sets the number of retries for connection failures. If you use high value, consider taking care for the machine's network.");
+    ConfigKey<Integer> RETRIES_OF_NETWORK_FAILURES = ConfigKeys.newIntegerConfigKey("retriesOfNetworkFailures", "The parameter sets the number of retries for connection failures. If you use high value, consider taking care for the machine's network.", 4);
```
aledsage: Remind me: are we careful where we use this retry for non-idempotent commands (e.g. in case the command was received and partially executed before we got the network failure, which is presumably indistinguishable from the command never been started)? In such as case, do we still retry or do we refuse to retry?

software/base/src/main/java/org/apache/brooklyn/entity/software/base/VanillaWindowsProcess.java
```diff
@@ -76,6 +76,10 @@
     @CatalogConfig(label = "Stop Command", priority=1)
     ConfigKey<String> STOP_COMMAND = VanillaSoftwareProcess.STOP_COMMAND;
 
+    @CatalogConfig(label = "Suppress failures occurred during installation. When using this parameter, Apache Brooklyn relies that your blueprint will report failures on the check running phase.")
+    ConfigKey<Boolean> SUPPRESS_INSTALL_FAILURE = ConfigKeys.newBooleanConfigKey("install.windows.suppressFailure", "Suppress any installation failure. It is a common case to restart windows together with the installation process." +
```
aledsage: I think we should indicate what the default value is (`false` I presume), rather than leaving it null.

software/base/src/main/java/org/apache/brooklyn/entity/software/base/VanillaWindowsProcessWinRmDriver.java
```diff
@@ -54,7 +54,8 @@ public void install() {
             executeCommandInTask(
                     getEntity().getConfig(VanillaWindowsProcess.INSTALL_COMMAND),
                     getEntity().getConfig(VanillaWindowsProcess.INSTALL_POWERSHELL_COMMAND),
-                    "install-command");
+                    "install-command",
+                    getEntity().getConfig(VanillaWindowsProcess.SUPPRESS_INSTALL_FAILURE));
```
aledsage: The `getConfig` call can return null. We should do `Boolean.TRUE.equals(getEntity().getConfig(VanillaWindowsProcess.SUPPRESS_INSTALL_FAILURE))` to avoid the risk of a `NullPointerException`.

software/base/src/main/java/org/apache/brooklyn/entity/software/base/AbstractSoftwareProcessWinRmDriver.java
```diff
@@ -161,12 +161,16 @@ public WinRmMachineLocation getMachine() {
...
+    }
+
+    protected int executeCommandInTask(String command, String psCommand, String phase, boolean suppressFailures) {
         return newScript(phase)
                 .setCommand(command)
                 .setPsCommand(psCommand)
                 .failOnNonZeroResultCode()
                 .gatherOutput()
-                .execute();
+                .execute(suppressFailures);
```
aledsage: Why is `suppressFailures` special (i.e. passing it as a parameter to `execute()`, rather than calling `.suppressFailures(suppressFailures)` to configure the script before calling a no-arg `execute()`?

software/base/src/main/java/org/apache/brooklyn/entity/software/base/lifecycle/WinRmExecuteHelper.java
```diff
@@ -82,16 +82,29 @@ public WinRmExecuteHelper setPsCommand(String psCommand) {
      * _and_ block until completion, throwing on error */
     @Beta
     public Task<Integer> queue() {
-        return DynamicTasks.queueIfPossible(newTask()).orSubmitAndBlock().getTask();
+        return queue(false);
+    }
+
+    private Task<Integer> queue(boolean suppressFailures) {
```
aledsage: Again, why is `suppressFailures` a parameter on `queue()`, rather than config you can set on the `WinRmExecuteHelper`? I'd favour adding a method `suppressFailures(boolean)`.

software/base/src/main/java/org/apache/brooklyn/entity/software/base/VanillaWindowsProcessWinRmDriver.java
aledsage: Why make `install` obligatory?

bostko: To clarify my suggestion is only for Windows.
From the experience we have with windows, I see only blueprints with install command and no blueprints with launch command.

software/winrm/src/main/java/org/apache/brooklyn/util/core/internal/winrm/WinRmTool.java
bostko: It is indistinguishable. Restarting from the install command will be retried 4 times and then it will be catched on the install phase if `install.windows.suppressFailure` is used.

bostko:
@aledsage thank you! I addressed your comments.

software/base/src/main/java/org/apache/brooklyn/entity/software/base/VanillaWindowsProcess.java
```diff
@@ -76,6 +76,10 @@
...
+            " In such case, it's hard to tell when the install command finished and entered into restart and you may want to support connectivity or other failures.", false);
```
aledsage: Put the `, false` on a new line. Otherwise someone reading through the code will miss it (unless they scroll).

software/base/src/main/java/org/apache/brooklyn/entity/software/base/VanillaWindowsProcessWinRmDriver.java
aledsage: Let's look at that in a separate PR. Making both optional (neither obligatory) would be fine with me.

bostko: Is it ok for launch to be non-obligatory like it is now in the PR?

bostko:
@aledsage @nakomis is it ok to merge?

aledsage:
Speaking to @bostko off-line, it seems that things are more complicated when it comes to a windows script kicking off a reboot. The reboot might not happen immediately.

The install script that triggers the reboot might return, and then we'll start executing the subsequent scripts. But in the background, the windows VM is doing stuff and is about to reboot. So it's a race as to which script we will be executing when the reboot actually happens. If it's not the install script, then the configuration option added in this PR will not be enough. The provisioning of the windows blueprint might then fail at the post-intall step (or uploading of runtime-files, or customize or launch, etc).

If it is a subsequent step that fails, we really would need to retry it. But it might well not be idempotent! 

It would be good to have a more holistic approach to fix this (and/or a consistent approach applied to all the steps).

Perhaps we need a "reboot expected" configuration option. This would be similar to `install.reboot.required`, but would not trigger the reboot. Instead it would wait for something that looks like a reboot (i.e it going offline and then back online), before continuing with the subsequent tasks.

@nakomis what are your thoughts?

bostko:
Thank you for the comment @aledsage !
I think we should streamline adding parameters like the one you said and also `waitWindowsToStart` parameter.

nakomis:
@bostko Can you give some more information about why `-NoRebootOnCompletion` in conjunction with `INSTALL_REBOOT_REQUIRED` isn't appropriate here? On the face of it, it seems much cleaner than retrying failing (and potentially non-idempotent) installation scripts

bostko:
@nakomis you are right that `INSTALL_REBOOT_REQUIRED` can be used for the install command to complete. I think it is fare enough solution.
However for Active Directory blueprint `-NoRebootOnCompletion` won't work because installing Active Directory is messing up the windows authentication and triggering a reboot is not possible. I tried launching manually a command against a machine that had installed Active Directory with `-NoRebootOnCompletion`  and it wasn't possible to authenticate against it with using passing the new domain name in the user.

nakomis:
@bostko I've been trying to reproduce the issue manually, by provisioning a windows VM, installing ADDS remotely using pywinrm with `-NoRebootOnCompletion`, then reconnecting to reboot, then reconnecting using domain credentials. It all worked successfully

I think it's going to be worth spending some time looking into this - the `-NoRebootOnCompletion` and `INSTALL_REBOOT_REQUIRED` pattern seems much cleaner, even if we need to jump through some hoops to get the authentication working correctly

Here's my rough notes from my testing:

Created a VM:

```
services:
- type: org.apache.brooklyn.entity.software.base.VanillaWindowsProcess
  brooklyn.config:
    launch.command: echo true
    checkRunning.command: echo true
location: AWS Oregon Win
```

RDPd to the VM, and ran:

```
winrm set winrm/config/service/auth @{Basic="True"}
```

Then created `C:\ads.ps1`

```
Add-windowsfeature AD-Domain-Services
Import-Module ADDSDeployment
$pwd = "thepassword$!123Password"
$securepwd = ConvertTo-SecureString $pwd -AsPlainText -Force
$domainName = "mydomain.com"
$domainNetbiosName = "mydomain"
Install-ADDSForest -DomainName $domainName -DomainNetbiosName $domainNetbiosName -NoRebootOnCompletion -SafeModeAdministratorPassword: $securepwd -Force: $true
```

pywinrm to the VM and ran:

```
s = winrm.Session('52.39.83.40', auth=('Administrator', 'password'))
r = s.run_ps("C:\\ads.ps1")
r.std_out
```

Tried connecting again (success):
```
s = winrm.Session('52.39.83.40', auth=('Administrator', 'password'))
r = s.run_ps("ls")
r.std_out
```

Rebooted (success):
```
s = winrm.Session('52.39.83.40', auth=('Administrator', 'password'))
r = s.run_ps("Restart-Computer -Force")
r.std_out
```

Tried various connections... (all successful)
```
s = winrm.Session('52.39.83.40', auth=('Administrator', 'password'))
r = s.run_ps("ls")
r.std_out
```

```
s = winrm.Session('52.39.83.40', auth=('Administrator@WIN-H2SOO1DO89C', 'password'))
r = s.run_ps("ls")
r.std_out
```

```
s = winrm.Session('52.39.83.40', auth=('Administrator@mydomain', 'password'))
r = s.run_ps("ls")
r.std_out
```

bostko:
@nakomis I amended the PR.
To summarize, the special think for the Blueprint I tested was that:
1. By default the install command restarts the VM on its own. A restart command executed through winrm may report non-zero code status or connection exception.
2. It changes the domain settings of the Windows Machine. By default Apache Brooklyn uses username and password to authenticate via NTLM against the Windows VM. When such operation is performed the winrm connection hangs and you cannot continue execute commands.

So to resolve the issues above I made changes to my blueprint to not restart automatically and restart by setting the `install.reboot.required` config.
To make sure that the install command won't hang and the reboot command can be triggered,
for these two commands when `install.reboot.required` is used I added the computer name when authenticating.
Basically adding a Computer name to the authentication should be safe and tends to be the better way for authentication, but because it is tricky to get this name and it is not still tested I made the change only for when `install.reboot.required` is used.
I tested my changes against different clouds with Basic and NTLM authentication and all works.
The clouds which I tested against were:
- AWS EC2
- Azure
- Vcloud Director

software/base/src/main/java/org/apache/brooklyn/entity/software/base/AbstractSoftwareProcessWinRmDriver.java
```diff
@@ -321,12 +332,22 @@ public int copyTo(InputStream source, String destination) {
...
+    public void rebootAndWait(String hostname) {
         try {
-            executePsScriptNoRetry(ImmutableList.of("Restart-Computer -Force"));
+            if (hostname != null) {
+                getLocation().executePsScript(ImmutableMap.of(WinRmTool.COMPUTER_NAME, hostname), ImmutableList.of("Restart-Computer -Force"));
+            } else {
+                getLocation().executePsScript(ImmutableList.of("Restart-Computer -Force"));
+            }
         } catch (Exception e) {
             // Restarting the computer will cause the command to fail; ignore the exception and continue
```
nakomis: This comment is no longer true, and the LOG message below should be changed to something like `LOG.warn("Exception when restarting machine: ", e);`

software/base/src/main/java/org/apache/brooklyn/entity/software/base/VanillaWindowsProcessWinRmDriver.java
```diff
@@ -49,15 +50,28 @@ public void start() {
 
     @Override
     public void install() {
+        String hostname = null;
```
nakomis: At some point in the future, this should probably be refactored to get the name of the machine in `WinRmMachineLocation` and set it as the hostname sensor. Please add a TODO note here suggesting the same

nakomis:
A couple of minor comments, but other than that, LGTM. We should probably refactor the away the `winrm.computerName` config key, and instead set the `hostname` sensor correctly during (or immediately after) machine provisioning, but that can happen in a future PR

Have successfully tested a deployment of Active Directory Services to a Windows VM on AWS Oregon

nakomis:
LGTM

software/base/src/main/java/org/apache/brooklyn/entity/software/base/AbstractSoftwareProcessWinRmDriver.java
```diff
@@ -321,12 +332,21 @@ public int copyTo(InputStream source, String destination) {
...
         try {
-            executePsScriptNoRetry(ImmutableList.of("Restart-Computer -Force"));
+            if (hostname != null) {
+                getLocation().executePsScript(ImmutableMap.of(WinRmTool.COMPUTER_NAME, hostname), ImmutableList.of("Restart-Computer -Force"));
+            } else {
+                getLocation().executePsScript(ImmutableList.of("Restart-Computer -Force"));
+            }
         } catch (Exception e) {
-            // Restarting the computer will cause the command to fail; ignore the exception and continue
+            LOG.warn("Exception when restarting machine: ", e);
```
aledsage: I'd prefer this exception message to indicate that it is an expected exception. If someone looks in the logs and sees this problem, it will look like there was a problem restarting.

Should we also log at debug? Can we narrow down the kind of exception(s) we expect? (if we're not 100% certain, that's absolutely fine to just catch Exception).

software/base/src/main/java/org/apache/brooklyn/entity/software/base/VanillaWindowsProcessWinRmDriver.java
```diff
@@ -49,15 +50,29 @@ public void start() {
...
+        if (entity.getConfig(VanillaWindowsProcess.INSTALL_REBOOT_REQUIRED)) {
+            WinRmExecuteHelper checkHostnameTask = newScript("Checking hostname")
+                    .setCommand("hostname")
+                    .failOnNonZeroResultCode()
+                    .gatherOutput();
+            if (checkHostnameTask.execute() != 0) {
+                LOG.warn("Cannot determine hostname for entity {} machine {}", getEntity(), getLocation());
+            }
+            hostname = Strings.trimEnd(checkHostnameTask.getResultStdout());
```
aledsage: Should we really be extracting hostname from stdout if the command failed to execute (i.e. if we did the above log statement)? Presumably we'll get either an empty string or some error message. What are the implications of subsequently passing that to `rebootAndWait(hostname)`?

software/base/src/main/java/org/apache/brooklyn/entity/software/base/AbstractSoftwareProcessWinRmDriver.java
nakomis: My understanding was that this exception was now *not* expected, @bostko is that the case?

software/base/src/main/java/org/apache/brooklyn/entity/software/base/VanillaWindowsProcessWinRmDriver.java
bostko: @aledsage I find it as the most reasonable way to get the actual hostname.
The `LOG.warn` statement which was there is not valid.
`.failOnNonZeroResultCode()` predicate will be used from the WinRmExecuteHelper to throw an exception if it is not valid.
I expect that this command should always succeed when winrm-ing to a working windows machine.

nakomis:
The latest changes looks sensible. Is there any (external) documentation saying what exceptions we may see during `rebootAndWait`, or are the exceptions used (`WebServiceException` and `Fault`) from experimentation?

bostko:
@nakomis `WebServiceException` and `Fault` are exceptions thrown from the SOAP client we use for WinRM.
I will see if I can add some documentation.

bostko:
Testing now the changes.

bostko:
@nakomis @aledsage I tested the current changes. Do you have other suggestions?

aledsage:
Thanks all - merging.

software/base/src/main/java/org/apache/brooklyn/entity/software/base/VanillaWindowsProcessWinRmDriver.java
```diff
@@ -49,15 +50,27 @@ public void start() {
 
     @Override
     public void install() {
+        // TODO: At some point in the future, this should probably be refactored to get the name of the machine in WinRmMachineLocation and set it as the hostname sensor
+        String hostname = null;
+        if (entity.getConfig(VanillaWindowsProcess.INSTALL_REBOOT_REQUIRED)) {
+            WinRmExecuteHelper checkHostnameTask = newScript("Checking hostname")
```
aledsage: @bostko @nakomis why do we retrieve this *before* executing the install command?

My understanding is that we added the hostname lookup to handle "Ability to perform installation on a Windows Blueprint which changes Computer Name and requires reboot." What will change the Computer Name? Do we expect it to be the "install.command"? If so, surely we should retrieve the hostname *after* executing the install command?

bostko: Actually  AD blueprint doesn't change the Computer Name, it is changing the `%UserDnsDomain%` and `%USERDOMAIN%` which breaks the default authentication where we supply only Windows username without any computer name.

software/winrm/src/main/java/org/apache/brooklyn/util/core/internal/winrm/WinRmTool.java
aledsage: For the record, it looks like this used to default to 16 in winrm4j, if no value was passed in by Brooklyn: https://github.com/cloudsoft/winrm4j/blob/0.3.5/client/src/main/java/io/cloudsoft/winrm4j/client/WinRmClient.java#L581

PR #151, opened 2016-05-23
Merge Map values in Location config

bostko:
Merge location configuration and application's provisioning.properties.
This PR will modify the behaviour of existing blueprints, I believe to better.
When deploying a blueprint with provisioning.properties and templateOptions configuration for example with will again prefer the provisioning.properties but it will also take any old values in templateOptions.

### DIFF
--- a/camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/EmptySoftwareProcessYamlTest.java
+++ b/camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/EmptySoftwareProcessYamlTest.java
@@ -21,13 +21,17 @@ package org.apache.brooklyn.camp.brooklyn;
 import java.util.Iterator;
 import java.util.Map;
 
+import com.google.common.collect.ImmutableMap;
 import org.apache.brooklyn.api.entity.Entity;
 import org.apache.brooklyn.api.location.Location;
+import org.apache.brooklyn.config.ConfigKey;
+import org.apache.brooklyn.core.config.ConfigKeys;
 import org.apache.brooklyn.core.entity.Attributes;
 import org.apache.brooklyn.core.entity.BrooklynConfigKeys;
 import org.apache.brooklyn.core.entity.Entities;
 import org.apache.brooklyn.core.entity.EntityAsserts;
 import org.apache.brooklyn.entity.software.base.EmptySoftwareProcess;
+import org.apache.brooklyn.location.jclouds.JcloudsLocationConfig;
 import org.apache.brooklyn.location.ssh.SshMachineLocation;
 import org.apache.brooklyn.util.collections.Jsonya;
 import org.slf4j.Logger;
@@ -57,6 +61,36 @@ public class EmptySoftwareProcessYamlTest extends AbstractYamlTest {
         EmptySoftwareProcess entity = (EmptySoftwareProcess) app.getChildren().iterator().next();
         Map<String, Object> pp = entity.getConfig(EmptySoftwareProcess.PROVISIONING_PROPERTIES);
         Assert.assertEquals(pp.get("minRam"), 16384);
+    }
+
+    public void testProvisioningPropertiesOverrideValues() throws Exception {
+        ConfigKey<Map> testConfig = ConfigKeys.newConfigKey(Map.class, "someOtherLocationOption");
+        Entity app = createAndStartApplication(
+                "location:",
+                "  localhost:",
+                "    templateOptions:",
+                "      testKey: testValue",
+                "    someOtherLocationOption:",
+                "      testKeyX1: testValueX1",
+                "services:",
+                "- type: "+EmptySoftwareProcess.class.getName(),
+                "  provisioning.properties:",
+                "    minRam: 16384",
+                "    templateOptions:",
+                "      overrideLoginUser: testUser",
+                "    someOtherLocationOption:",
+                "      testKeyX2: testValueX2");
+        waitForApplicationTasks(app);
+
+        log.info("App started:");
+        Entities.dumpInfo(app);
+
+        EmptySoftwareProcess entity = (EmptySoftwareProcess) app.getChildren().iterator().next();
+        Map<String, Object> pp = entity.getConfig(EmptySoftwareProcess.PROVISIONING_PROPERTIES);
+        Assert.assertEquals(pp.get("minRam"), 16384);
+        Map<String, Object> templateOptions = entity.getLocations().iterator().next().getConfig(JcloudsLocationConfig.TEMPLATE_OPTIONS);
+        Assert.assertEquals(templateOptions, ImmutableMap.of("overrideLoginUser", "testUser", "testKey", "testValue"));
+        Assert.assertEquals(entity.getLocations().iterator().next().getConfig(testConfig), ImmutableMap.of("testKeyX2", "testValueX2"));
     }
 
     @Test(groups="Integration")
--- a/core/src/main/java/org/apache/brooklyn/core/location/AbstractLocation.java
+++ b/core/src/main/java/org/apache/brooklyn/core/location/AbstractLocation.java
@@ -459,7 +459,7 @@ public abstract class AbstractLocation extends AbstractBrooklynObject implements
 
         @Override
         public void addToLocalBag(Map<String, ?> vals) {
-            configBag.putAll(vals);
+            configBag = ConfigBag.newInstanceExtending(configBag, vals);
         }
 
         @Override
--- a/core/src/main/java/org/apache/brooklyn/location/byon/FixedListMachineProvisioningLocation.java
+++ b/core/src/main/java/org/apache/brooklyn/location/byon/FixedListMachineProvisioningLocation.java
@@ -39,6 +39,7 @@ import org.apache.brooklyn.config.ConfigKey;
 import org.apache.brooklyn.core.config.ConfigKeys;
 import org.apache.brooklyn.core.location.AbstractLocation;
 import org.apache.brooklyn.core.location.cloud.CloudLocationConfig;
+import org.apache.brooklyn.core.objs.AbstractConfigurationSupportInternal;
 import org.apache.brooklyn.location.ssh.SshMachineLocation;
 import org.apache.brooklyn.util.collections.CollectionFunctionals;
 import org.apache.brooklyn.util.collections.MutableMap;
@@ -279,7 +280,7 @@ implements MachineProvisioningLocation<T>, Closeable {
     public T obtain(Map<?,?> flags) throws NoMachinesAvailableException {
         T machine;
         T desiredMachine = (T) flags.get("desiredMachine");
-        ConfigBag allflags = ConfigBag.newInstanceExtending(config().getBag()).putAll(flags);
+        ConfigBag allflags = ConfigBag.newInstanceExtending(config().getBag(), flags);
         Function<Iterable<? extends MachineLocation>, MachineLocation> chooser = allflags.get(MACHINE_CHOOSER);
         
         synchronized (lock) {
--- a/core/src/main/java/org/apache/brooklyn/util/core/config/ConfigBag.java
+++ b/core/src/main/java/org/apache/brooklyn/util/core/config/ConfigBag.java
@@ -29,8 +29,8 @@ import javax.annotation.Nonnull;
 import org.apache.brooklyn.config.ConfigKey;
 import org.apache.brooklyn.config.ConfigKey.HasConfigKey;
 import org.apache.brooklyn.core.config.ConfigKeys;
+import org.apache.brooklyn.util.collections.CollectionHelpers;
 import org.apache.brooklyn.util.collections.MutableMap;
-import org.apache.brooklyn.util.core.config.ConfigBag;
 import org.apache.brooklyn.util.core.flags.TypeCoercions;
 import org.apache.brooklyn.util.guava.Maybe;
 import org.apache.brooklyn.util.javalang.JavaClassNames;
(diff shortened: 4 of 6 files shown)


camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/EmptySoftwareProcessYamlTest.java
```diff
@@ -60,6 +62,31 @@ public void testProvisioningProperties() throws Exception {
...
+        waitForApplicationTasks(app);
+
+        log.info("App started:");
+        Entities.dumpInfo(app);
+
+        EmptySoftwareProcess entity = (EmptySoftwareProcess) app.getChildren().iterator().next();
+        Map<String, Object> pp = entity.getConfig(EmptySoftwareProcess.PROVISIONING_PROPERTIES);
+        Assert.assertEquals(pp.get("minRam"), 16384);
+        Map<String, Object> templateOptions = entity.getLocations().iterator().next().getConfig(JcloudsLocationConfig.TEMPLATE_OPTIONS);
+        Assert.assertEquals(templateOptions, ImmutableMap.of("overrideLoginUser", "test", "test", "test"));
```
nakomis: A minor point (shouldn't hold up this PR), but `testKey`, `testValue`, and `testUser` would be clearer here

bostko:
Is the naming OK. Do you think I named properly the util class?

core/src/main/java/org/apache/brooklyn/location/byon/FixedListMachineProvisioningLocation.java
```diff
@@ -345,7 +346,7 @@ protected void updateMachineConfig(T machine, Map<?, ?> flags) {
             // For backwards compatibility, where peristed state did not have this.
             origConfigs = Maps.newLinkedHashMap();
         }
-        Map<String, Object> strFlags = ConfigBag.newInstance(flags).getAllConfig();
+        Map<String, Object> strFlags = ConfigBag.newInstanceExtending(((AbstractConfigurationSupportInternal)machine.config()).getBag(), flags).getAllConfig();
```
nakomis: Does this need to be applied to the other implementations of `MachineProvisioningLocation.obtain()`?

Will this effect the behaviour of existing blueprints?

core/src/main/java/org/apache/brooklyn/util/core/config/ConfigBag.java
```diff
@@ -126,7 +127,11 @@ public void markUsed(String key) {
     /** As {@link #newInstanceExtending(ConfigBag)} but also putting the supplied values. */
```
nakomis: This is now merging, not putting existing values

core/src/main/java/org/apache/brooklyn/location/byon/FixedListMachineProvisioningLocation.java
bostko: It is already used in `JcloudsLocation.obtain()`

utils/common/src/main/java/org/apache/brooklyn/util/collections/CollectionHelpers.java
```diff
@@ -0,0 +1,55 @@
...
+ */
+package org.apache.brooklyn.util.collections;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Map;
+
+/**
+ * @author valentin
```
nakomis: We typically avoid author javadocs

camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/EmptySoftwareProcessYamlTest.java
nakomis: The name of the util class is ok, but line 86 of `EmptySoftwareProcessYamlTest` (the line I commented on) would be clearer if it read `Assert.assertEquals(templateOptions, ImmutableMap.of("overrideLoginUser", "testUser", "testKey", "testValue"));`

core/src/main/java/org/apache/brooklyn/location/byon/FixedListMachineProvisioningLocation.java
nakomis: Ok, didn't see that. Is this changing `FixedListMachineProvisioningLocation` to work in the same way as `JcloudsLocation`? If so, it should use `ResolvingConfigBag` for consistency with `JcloudsLocation`

utils/common/src/main/java/org/apache/brooklyn/util/collections/CollectionHelpers.java
```diff
@@ -0,0 +1,52 @@
...
+        result.putAll(defaultEntries);
+        for (Map.Entry<String, Object> newEntry: overrideWith.entrySet()) {
+            if (!defaultEntries.containsKey(newEntry.getKey())) {
+                result.put(newEntry.getKey(), newEntry.getValue());
+            } else if (newEntry.getValue() instanceof Map) {
+                Map<Object, Object> mergedValue = MutableMap.of();
+                if (defaultEntries.get(newEntry.getKey()) instanceof Map) {
+                    mergedValue.putAll((Map) defaultEntries.get(newEntry.getKey()));
+                } else {
+                    log.warn("defaultEntries {} is expected to be Map but it is {}", newEntry.getKey(), defaultEntries);
```
neykov: For incompatible types better keep non-map behaviour of overriding the key.

utils/common/src/main/java/org/apache/brooklyn/util/collections/CollectionHelpers.java
```diff
@@ -0,0 +1,52 @@
...
+                }
+                for (Map.Entry<Object, Object> innerValue : ((Map<Object,Object>) newEntry.getValue()).entrySet()) {
+                    mergedValue.put(innerValue.getKey(), innerValue.getValue());
```
neykov: Why not `putAll(newEntry.getValue())`?

utils/common/src/main/java/org/apache/brooklyn/util/collections/CollectionHelpers.java
```diff
@@ -0,0 +1,52 @@
...
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Map;
+
+public class CollectionHelpers {
+    private static final Logger log = LoggerFactory.getLogger(CollectionHelpers.class);
+
+    public static Map<String, Object> mergeNestedMaps(Map<String, Object> defaultEntries, Map<String, Object> overrideWith) {
```
neykov: Not clear from the name that it's a shallow merge.

core/src/main/java/org/apache/brooklyn/util/core/config/ConfigBag.java
```diff
@@ -126,7 +127,11 @@ public void markUsed(String key) {
...
     @Beta
     public static ConfigBag newInstanceExtending(final ConfigBag configBag, Map<?,?> optionalAdditionalValues) {
-        return newInstanceExtending(configBag).putAll(optionalAdditionalValues);
+        if (optionalAdditionalValues != null) {
+            return newInstance(CollectionHelpers.mergeNestedMaps(configBag.getAllConfig(), (Map<String, Object>) optionalAdditionalValues));
+        } else {
+            return newInstanceExtending(configBag);
+        }
```
neykov: I think we shouldn't do this, too intrusive. Better make the merging behaviour specific to `provisioning.properties`.

utils/common/src/main/java/org/apache/brooklyn/util/collections/CollectionHelpers.java
bostko: it is again the non-map behavior. Such values will be overridden by the code bellow. This was the case for all the values previously.

core/src/main/java/org/apache/brooklyn/util/core/config/ConfigBag.java
neykov: Or even more constrained to `templateOptions`.

neykov:
I like the idea of the PR, but feel that it should be more constrained in scope. Either having this behaviour only for `prosivioning.properties` or even more to `templateOptions`

camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/EmptySoftwareProcessYamlTest.java
```diff
@@ -60,6 +62,31 @@ public void testProvisioningProperties() throws Exception {
     }
 
     @Test(groups="Integration")
+    public void testProvisioningPropertiesOverrideValues() throws Exception {
+        Entity app = createAndStartApplication(
+                "location:",
+                "  localhost:",
+                "    templateOptions:",
+                "      testKey: testValue",
```
aledsage: I'd include in the test an assertion about what takes precedence. (and you can turn off all attempts to ssh to the machine). For example, have:

    location:
      useJcloudsSshInit: false
      waitForSshable: false
      minRam: 1G
      minCores: 1
      templateOptions:
        locKey1: locVal1
        locKey2: locVal2
    services:
    - type: EmptySoftwareProcess
      brooklyn.config:
        onbox.base.dir.skipResolution: true
        sshMonitoring.enabled: false
        provisioning.properties:
          minRam: 2G
          templateOptions:
            locKey1: overriddenLocVal1
            entKey1: entVal1

core/src/main/java/org/apache/brooklyn/core/location/AbstractLocation.java
```diff
@@ -459,7 +459,7 @@ public ConfigBag getLocalBag() {
 
         @Override
         public void addToLocalBag(Map<String, ?> vals) {
-            configBag.putAll(vals);
+            configBag = ConfigBag.newInstanceExtending(configBag, vals);
```
aledsage: I don't understand what this change is fixing. Can you explain the reasoning please?

core/src/main/java/org/apache/brooklyn/location/byon/FixedListMachineProvisioningLocation.java
```diff
@@ -279,7 +280,7 @@ public T obtain() throws NoMachinesAvailableException {
     public T obtain(Map<?,?> flags) throws NoMachinesAvailableException {
         T machine;
         T desiredMachine = (T) flags.get("desiredMachine");
-        ConfigBag allflags = ConfigBag.newInstanceExtending(config().getBag()).putAll(flags);
+        ConfigBag allflags = ConfigBag.newInstanceExtending(config().getBag(), flags);
```
aledsage: This change looks like it makes no difference (now that you've backed out your changes to the `newInstanceExtending()` method). I'm fine with leaving this change in though.

core/src/main/java/org/apache/brooklyn/util/core/config/ConfigBag.java
```diff
@@ -126,7 +126,11 @@ public void markUsed(String key) {
     /** As {@link #newInstanceExtending(ConfigBag)} but also putting the supplied values. */
     @Beta
     public static ConfigBag newInstanceExtending(final ConfigBag configBag, Map<?,?> optionalAdditionalValues) {
-        return newInstanceExtending(configBag).putAll(optionalAdditionalValues);
```
aledsage: As per @neykov 's comments, do we really want this change?

    I think we shouldn't do this, too intrusive. Better make the merging behaviour specific to provisioning.properties.
    Or even more constrained to templateOptions.
    I like the idea of the PR, but feel that it should be more constrained in scope. Either having this behaviour only for prosivioning.properties or even more to templateOptions

It feels high risk. It's unclear what impact it will have on other users (e.g. someone writing a YAML entity that extends another entity, overriding a config key; for that particular example, we need to discuss it on dev@brooklyn a lot more first).

utils/common/src/main/java/org/apache/brooklyn/util/collections/CollectionHelpers.java
```diff
@@ -0,0 +1,53 @@
...
+
+import java.util.Map;
+
+public class CollectionHelpers {
+    private static final Logger log = LoggerFactory.getLogger(CollectionHelpers.class);
+
+    /**
+     * Does a shallow merge of two maps.
+     */
+    public static Map<String, Object> mergeMapsAndTheirInnerMapValues(Map<String, Object> defaultEntries, Map<String, Object> overrideWith) {
```
aledsage: This is the sort of method that it's really easy to write great test coverage for. We should do that, including where default entry is of different type than overriding entry etc.

It's unclear what a user would expect, when merging maps. Why stop at one level down? What about merging or overriding lists?

aledsage:
I agree with @neykov : "I like the idea of the PR, but feel that it should be more constrained in scope. Either having this behaviour only for `prosivioning.properties` or even more to `templateOptions`".

If doing this across the board, we'd need to think *very* carefully about its implications for entities extending entities, write lots of test cases, discuss the desired behaviour on dev@brooklyn, etc.

Right now, I'm not sure what the consequences are of changing `ConfigBag.newInstanceExtending`, which is called in a lot of places.

core/src/main/java/org/apache/brooklyn/core/location/AbstractLocation.java
bostko: It merges the values that the location has in config with vals flags come from provisioning.properties

bostko: The exact answer here is, that the properties are set twice in `FixedListMachineProvisioningLocation#obtain`
1. when allflags is constructed and used in `provisionMore` call
2. after that the configuration of the object is updated for a second time in `updateMachineConfig` call

ahgittin:
+1 absolutely must be more constrained -- taking to ML

aledsage:
This has now been discussed as part of the dev@brooklyn email thread "[PROPOSAL] merging config keys".

I'm working on a general solution for that, so please DO NOT MERGE this PR.

aledsage:
This has been superseded by https://github.com/apache/brooklyn-server/pull/173.

@bostko can you please confirm that PR #173 does what you want?

PR #152, opened 2016-05-23
Resolve AddSensor and AddEffector config

sjcorbett:
Resolves useful parameters of implementations of the `AddEffector` and `AddSensor` entity initializers, meaning that their values can be sourced from DslComponents.

Affects:
* SshCommandEffector's command and executionDir
* SshCommandSensor's command and executionDir
* CreatePasswordSensor's passwordLength and acceptableChars
* JmxAttributeSensor's objectName, attribute and defaultValue

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/sensor/password/CreatePasswordSensor.java
+++ b/core/src/main/java/org/apache/brooklyn/core/sensor/password/CreatePasswordSensor.java
@@ -24,6 +24,7 @@ import org.apache.brooklyn.api.entity.EntityLocal;
 import org.apache.brooklyn.config.ConfigKey;
 import org.apache.brooklyn.core.config.ConfigKeys;
 import org.apache.brooklyn.core.effector.AddSensor;
+import org.apache.brooklyn.core.entity.EntityInitializers;
 import org.apache.brooklyn.util.core.config.ConfigBag;
 import org.apache.brooklyn.util.text.Identifiers;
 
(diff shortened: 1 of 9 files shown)


core/src/main/java/org/apache/brooklyn/core/sensor/password/CreatePasswordSensor.java
```diff
@@ -33,23 +34,21 @@
...
-        passwordLength = params.get(PASSWORD_LENGTH);
-        acceptableChars = params.get(ACCEPTABLE_CHARS);
     }
 
     @Override
     public void apply(EntityLocal entity) {
         super.apply(entity);
 
+        Integer passwordLength = EntityInitializers.resolve(params, PASSWORD_LENGTH);
+        String acceptableChars = EntityInitializers.resolve(params, ACCEPTABLE_CHARS);
```
neykov: `apply` is called before managing any of the entities in the spec hierarchy (i.e. a camp plan) so this can't be resolved at this point. It would just block app creation. Could you add some tests to prove me wrong :). 
`StaticSensor` does the right thing by pushing all of it in tasks.

grkvlt:
@sjcorbett One thing you should probably do here is move `AddSensor` to a different package, perhaps `org.apache.brooklyn.core.sensor` instead? Also, would there be any benefit gained by making the initializers be `EntityAdjunct` objects, supporting configuration properly etc.

core/src/main/java/org/apache/brooklyn/core/sensor/password/CreatePasswordSensor.java
sjcorbett: You're right, this was lazy of me.

sjcorbett:
@grkvlt @neykov Thanks for your comments. This whole approach is flawed because it doesn't survive rebind. Entity initialisers have been written to fire once, which is unfortunate because it's counterintuitive when used alongside everything else in a blueprint. I will explore the suggestion of using `EntityAdjuncts`.

PR #153, opened 2016-05-23
[WIP] Add support for JSR-223 scripting of effectors

grkvlt:
This is an attempt to add scripting language support to Brooklyn blueprints, so that effectors can be added in various languages directly. JSR-223 supports JavaScript and I have also added the Python 2.7.0 plugin, but many other languages are available.

See the `script-effector-example.yaml` file for an example of how these scripting languages can be used to implement effectors.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -150,11 +150,6 @@
             <groupId>org.apache.commons</groupId>
             <artifactId>commons-lang3</artifactId>
         </dependency>
-        <dependency>
-            <groupId>org.python</groupId>
-            <artifactId>jython</artifactId>
-            <version>2.7.0</version>
-        </dependency>
 
         <dependency>
             <groupId>org.testng</groupId>
@@ -198,6 +193,12 @@
         <dependency>
             <groupId>com.beust</groupId>
             <artifactId>jcommander</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.python</groupId>
+            <artifactId>jython</artifactId>
+            <version>2.7.0</version>
             <scope>test</scope>
         </dependency>
         <dependency>
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -150,6 +150,11 @@
             <groupId>org.apache.commons</groupId>
             <artifactId>commons-lang3</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.python</groupId>
+            <artifactId>jython</artifactId>
+            <version>2.7.0</version>
+        </dependency>
 
         <dependency>
             <groupId>org.testng</groupId>
--- a/core/src/test/java/org/apache/brooklyn/core/effector/script/ScriptEffectorTest.java
+++ b/core/src/test/java/org/apache/brooklyn/core/effector/script/ScriptEffectorTest.java
@@ -1,0 +1,68 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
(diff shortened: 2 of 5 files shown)


grkvlt:
@ahgittin I'd be interested in your thoughts about this idea... Still needs documentation, more examples and tests

core/pom.xml
```diff
@@ -150,6 +150,11 @@
             <groupId>org.apache.commons</groupId>
             <artifactId>commons-lang3</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.python</groupId>
+            <artifactId>jython</artifactId>
+            <version>2.7.0</version>
```
johnmccabe: Is it possible/feasible to install additional python modules for use with jython? Does the cPython ecosystem place nicely?

grkvlt: I believe that as long as the archives are on the CLASSPATH this is possible, and executing `import whatever` will work, and have seen examples, but I have not tested this.

core/src/test/java/org/apache/brooklyn/core/effector/script/ScriptEffectorTest.java
```diff
@@ -0,0 +1,68 @@
...
+                .addInitializer(new ScriptEffector(ConfigBag.newInstance(ImmutableMap.of(
+                        AddEffector.EFFECTOR_NAME, "javaScriptEffector",
+                        ScriptEffector.EFFECTOR_SCRIPT_LANGUAGE, "js",
+                        ScriptEffector.EFFECTOR_SCRIPT_RETURN_VAR, "o",
+                        ScriptEffector.EFFECTOR_SCRIPT_RETURN_TYPE, String.class,
+                        ScriptEffector.EFFECTOR_SCRIPT_CONTENT, "var o; o = \"jsval\";")))));
+        Maybe<Effector<?>> javaScriptEffector = entity.getEntityType().getEffectorByName("javaScriptEffector");
+        Assert.assertTrue(javaScriptEffector.isPresentAndNonNull(), "The JavaScript effector does not exist");
+        Object result = Entities.invokeEffector(entity, entity, javaScriptEffector.get()).getUnchecked();
+        Assert.assertTrue(result instanceof String, "Returned value is of type String");
```
duncangrant: Should be "Returned value is **not** of type string"?

grkvlt:
@sjcorbett Once #152 is merged I will update this to take advantage of the same features

core/pom.xml
aledsage: For a while, we bundled jython (via winrm4j, until we re-wrote that as pure Java). The jython libraries pulled in about 20 or 30MB of dependencies. It would certainly be nice if folk can add that to the classpath (or OSGi bundles) themselves, but I don't think we want to ship it with Brooklyn.

neykov: Agree that jython is better left as an optional addon - not worth the weight it brings in. Plus we already bundle groovy and javascript is provided by the JRE so there are alternatives to choose from.

neykov:
+1 for the idea - very useful in yaml context. I've been wanting to see this for a while.
What I think needs discussion is the security implications this brings. Currently as long as you can add (Java) catalog items you can gain full access to the Brooklyn instance so only privileged users should be allowed to do it. For application deployments "regular" users can only use existing catalog items which won't let them tinker with Brooklyn internals. Having this in the stock distribution will change that.
Take for example the Cloud Foundry Service Bridge. Any CF user is able to spin up an application in Brooklyn.

grkvlt:
@neykov that's a reasonable point - this exposes `managementContext` as a script engine attribute binding, which lets you do anything. However it's the same set of capabilities as if you had a catalog entry with a `brooklyn.libraries` entry pointing to a custom Jar file defining your entity and its effectors, isn't it?

core/pom.xml
grkvlt: I'll add it as a test-only dependency instead, to validate that we can call additional JSR-223 languages if required, but document that we only support JavaScript (as either Rhino in 7.x or Nashorn in 8.x) by default,

grkvlt:
I talked with @neykov about this, and we agreed that a sandbox would be required, and also that it would be difficult. I have been thinking about how to implement it, and one mechanism might be to inject a `brooklyn` object that allows only method calls to the equivalent of the `$brooklyn:` functions in the DSL. I'm still thinking about this, and will post to the Brooklyn dev list when I have something more concrete.

neykov:
Looks like it's possible to secure a script engine so it can run any 3d party scripts, but it's language spefic, there's nothing in JSR-223 that we can rely on.
See what it would take to secure a Rhino engine in http://codeutopia.net/blog/2009/01/02/sandboxing-rhino-in-java/.

grkvlt:
Closing this, as much of the desired functionality can be obtained through use of the meta effectors in #605

PR #156, opened 2016-05-25
BROOKLYN-272: disable more failing non-deterministic tests

aledsage:


aledsage:
Only affects tests, disabling those that failed in apache jenkins. Merging now.

PR #164, opened 2016-05-27
Update ConfigMapTest

aledsage:
* Updates the config tests (in particular `ConfigMapTest` was still calling entity constructors).
* Merges `ConfigMapTest` and `EntityConfigTest`, because there was no clear difference for what the tests were covering - so risked duplication if people didn't look in both.
* Deletes EntityConfigMapUsageLegacyTest, which tested the old (defunct) way of creating an entity via its constructor - everything (except some old tests) should now use EntitySpec.


aledsage:
Jenkins test failure was unrelated - I investigated it in https://github.com/apache/brooklyn-server/pull/166, but still not sure what caused it to fail :-(
Rebased against master, to cause jenkins to build again.

PR #170, opened 2016-05-31
Don't inherit "brooklyn.wrapper_app" key in children

neykov:


camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/ApplicationsYamlTest.java
```diff
@@ -232,6 +232,25 @@ public void testDoesNotWrapCatalogItemAndUsesCatalogServiceName() throws Excepti
...
+                "  item:",
+                "    type: " + BasicEntity.class.getName());
+        Entity app = createAndStartApplication(
+                "services:",
+                "- type: simple:" + TEST_VERSION);
+        Entity entity = Iterables.getOnlyElement(app.getChildren());
+        assertNull(entity.getConfig(EntityManagementUtils.WRAPPER_APP_MARKER));
+        // Note that "brooklyn.wrapper_app" will still make it into 
+        //   ((EntityInternal) entity).config().getBag().getAllConfigAsConfigKeyMap();
+        // so the UI will still show the marker as inherited by the entity.
```
aledsage: It's annoying that the UI will still show it. I guess that's because `getBag()` is also asking the parent for its bag? I'd have thought `.inheritance(ConfigInheritance.NONE)` would be enough to prevent that, but sounds like I would be wrong!

aledsage:
LGTM - merging.

camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/ApplicationsYamlTest.java
neykov: `getBag()` is doing (through `EntityConfigMap`):
```
        return ConfigBag.newInstanceCopying(localConfigBag)
                .putAll(ownConfig)
                .putIfAbsent(inheritedConfig)
                .putIfAbsent(inheritedConfigBag)
                .seal();
```

The problem is that `wrapper_app` in `inheritedConfig` has a generic `ConfigKey`, not the instance from `EntityManagementUtils.WRAPPER_APP_MARKER` so it doesn't have the inheritance value set.

PR #171, opened 2016-06-01
Use login port for ssh

duncangrant:
Currently we assume port 22 for ssh unless a
portforwarder has been configured.  This uses the getLoginPort
from nodemetadata to determine the ssh port.


grkvlt:
Looks good to me, I would like to do some testing on this before I merge it, though.

aledsage:
LGTM - I saw this in action when pairing with @duncangrant so will merge now.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -2896,7 +2896,7 @@ protected LoginCredentials waitForSshable(final ComputeService computeService, f
         String user = (users.size() == 1) ? Iterables.getOnlyElement(users) : "{" + Joiner.on(",").join(users) + "}";
         String vmIp = hostAndPortOverride.isPresent() ? hostAndPortOverride.get().getHostText() : getFirstReachableAddress(node, setup);
         if (vmIp==null) LOG.warn("Unable to extract IP for "+node+" ("+setup.getDescription()+"): subsequent connection attempt will likely fail");
-        int vmPort = hostAndPortOverride.isPresent() ? hostAndPortOverride.get().getPortOrDefault(22) : 22;
+        int vmPort = hostAndPortOverride.isPresent() ? hostAndPortOverride.get().getPortOrDefault(22) : getLoginPortOrDefault(node, 22);
```
neykov: Shouldn't `22` in `getPortOrDefault(22)` be replaced as well?

PR #173, opened 2016-06-01
BROOKLYN-286: merge config keys map values

aledsage:


aledsage:
Apologies that this is such a huge pull request :-)
But many of the lines are tests, or refactoring tests.
See individual commits for details of the changes.

bostko:
I think it worths putting Implement ConfigKey.typeInheritance changes in a separate PR.
The jenkins check seems relevant.

aledsage:
I have broken a few of these commits out into their own PRs, to allow parts of it to be reviewed + merged faster. Please comment on those smaller PRs - we can address comments there, and then rebase this larger PR.

* https://github.com/apache/brooklyn-server/pull/174
* https://github.com/apache/brooklyn-server/pull/175
* https://github.com/apache/brooklyn-server/pull/176

neykov:
Jenkins failure is related: `Too many files with unapproved license: 1 See RAT report`. I've merged the smaller PRs so if you could also rebase.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -3308,4 +3314,20 @@ public PersistenceObjectStore newPersistenceObjectStore(String container) {
...
     }
 
+    // TODO Duplicate of EntityConfigMap.deepMerge
+    private <T> Maybe<?> deepMerge(Maybe<? extends T> val1, Maybe<? extends T> val2, ConfigKey<?> keyForLogging) {
+        if (val2.isAbsent() || val2.isNull()) {
+            return val1;
+        } else if (val1.isAbsent()) {
+            return val2;
+        } else if (val1.isNull()) {
+            return val1; // an explicit null means an override; don't merge
```
neykov: Can't set `null` values in yaml, because we are using `ImmutableXXX` all over the place. Worth confirming with a test. If that's really the case we should think about other ways of overriding/fixing support for null values.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -3308,4 +3314,20 @@ public PersistenceObjectStore newPersistenceObjectStore(String container) {
...
+        } else if (val1.get() instanceof Map && val2.get() instanceof Map) {
+            return Maybe.of(CollectionMerger.builder().build().merge((Map<?,?>)val1.get(), (Map<?,?>)val2.get()));
```
neykov: I'd expect a shallow merge here. That is being able to override the values of maps in `templateOptions`. But the arguments could go either way. Doesn't really matter in the and as all maps in `templateOptions` are of simple types.

aledsage:
Now that those sub-PRs are merged (thanks @neykov ), I've rebased against master and fixed the missing license headers.

aledsage:
I've also extracted the commit "JcloudsLocation: handle config null values" to its own PR so that it can be reviewed + merged in isolation: https://github.com/apache/brooklyn-server/pull/178

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
neykov: Didn't notice you've already addressed this in a commit, ignore.

utils/common/src/main/java/org/apache/brooklyn/util/collections/CollectionMerger.java
```diff
@@ -0,0 +1,234 @@
...
+import java.util.Map;
+import java.util.Set;
+
+import org.apache.brooklyn.util.guava.Maybe;
+
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Iterables;
+import com.google.common.collect.Sets;
+
+public class CollectionMerger {
```
aledsage: This should be marked `@Beta`.

utils/common/src/main/java/org/apache/brooklyn/config/ConfigInheritance.java
```diff
@@ -20,31 +20,60 @@
...
 
+import org.apache.brooklyn.util.text.Strings;
+
 import com.google.common.annotations.Beta;
 
 @SuppressWarnings("serial")
 public abstract class ConfigInheritance implements Serializable {
 
-    public static final ConfigInheritance ALWAYS = new Always();
+    public enum InheritanceMode {
```
aledsage: This should be marked `@Beta`.

aledsage: I don't like the use of `ConfigInheritance` and `InheritanceMode`. It feels like we should just have one thing, particularly given the existing implementations of `ConfigInheritance` are so simple. But `ConfigInheritance` is not `@Beta` so I'm being conservative about changing that.

Any suggestions for a cleaner way?

neykov: Agree they represent the same thing currently. But on the other hand `ConfigInheritance` allows us to have context-dependent behaviour. We didn't need it so far, but doesn't hurt leaving it "just in case".

core/src/main/java/org/apache/brooklyn/core/entity/internal/EntityConfigMap.java
```diff
@@ -111,47 +110,107 @@ public EntityConfigMap(AbstractEntity entity, Map<ConfigKey<?>, Object> storage)
...
+    
+    private <T> Maybe<?> deepMerge(Maybe<? extends T> val1, Maybe<? extends T> val2, ConfigKey<?> keyForLogging) {
+        if (val2.isAbsent() || val2.isNull()) {
+            return val1;
+        } else if (val1.isAbsent()) {
+            return val2;
+        } else if (val1.isNull()) {
+            return val1; // an explicit null means an override; don't merge
+        } else if (val1.get() instanceof Map && val2.get() instanceof Map) {
+            return Maybe.of(CollectionMerger.builder().build().merge((Map<?,?>)val1.get(), (Map<?,?>)val2.get()));
```
neykov: In the general case a shallow merge would be the expected behaviour for me. Deep merge makes sense only for `provisioning.properties` (where in practice it's of depth 2 and only because of the special treatment of `templateOptions`). wdyt about having two inheritance modes - DEEP_MERGE, SHALLOW_MERGE? Or at least rename MERGE to DEEP_MERGE for now?

camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/ConfigInheritanceYamlTest.java
```diff
@@ -300,14 +338,212 @@ public void testOverridesParentConfig() throws Exception {
...
+                    "      mykey: myval",
+                    "  brooklyn.children:",
+                    "  - type: entity-with-keys",
+                    "    brooklyn.config:",
+                    "      map.type-merged:",
+                    "        mykey2: myval2",
+                    "      map.type-always:",
+                    "        mykey2: myval2",
+                    "      map.type-never:",
+                    "        mykey2: myval2");
```
neykov: How would inheritance work if there are three levels involved? What if they don't have `brooklyn.parameters` declared at each level?

camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/creation/BrooklynComponentTemplateResolver.java
```diff
@@ -257,37 +261,63 @@ private void configureEntityConfig(EntitySpec<?> spec, Set<String> encounteredRe
...
+                continue;
+            }
+            ConfigKey<?> key = entry.getValue();
+            InheritanceMode mode = getInheritanceMode(key, spec);
+            if (mode == InheritanceMode.NONE) {
+                spec.removeConfig(key);
+                spec.removeFlag(key.getName());
+            }
+        }
+
```
neykov: Should do analogous merging for locations?

camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/ConfigInheritanceYamlTest.java
aledsage: If one doesn't declare `brooklyn.parameters` (or you don't include `inheritance.parant` or `inheritance.type`), then it defaults to "always" - i.e. inherit the value, but override it (without merge) if there is a value in the sub-type.

aledsage: If 3 levels within a map, then if merging it will do a "deep merge", as per `CollectionMerger`. I agree it's not clear what a user would expect. I suspect some users would think it will be shallow, and other users will think deep (depending on the particularly values in the config map, presumably).

Or do you mean 3 levels of entity (e.g. A extends B, and C extends B)? Then the rules apply at each level: B will merge or override the values of A, and then C will merge or override those values of B (based on the config key definition).

camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/creation/BrooklynComponentTemplateResolver.java
aledsage: Agree we should - there's a disabled test that demonstrates it doesn't work for locations! Do you think it's ok to do that in a separate PR? Do you think we *must* do it for locations before the next release if this PR is merged (i.e. it's very wrong if it works for entities but not locations)?

camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/ConfigInheritanceYamlTest.java
neykov: I meant 3 levels of entity. Say a root `env`, a cluster `env` and per-node `env`. For example:
- root : env{APP_NAME="XXX"}
-+ group1
--+ cluster1 env{CLUSTER_NAME="XXX"}
---+ entity env{ENTITY_ID="XXX"}

Then we need to declare env in `brooklyn.parameters` at each level (even those that don't have a value for it - group1), so the entity will inherit all the way to the root, correct?

core/src/main/java/org/apache/brooklyn/core/entity/internal/EntityConfigMap.java
aledsage: @neykov I kind of agree, but am not sure. The examples I can think of are all shallow except for templateOptions. So doing a deep merge will have no impact (because the values of `shell.env` etc will be shallow). Therefore keeping the default as deep feels like we have a consistent rule that makes it work for templateOptions, and hopefully doesn't impact the other uses.

I worry slightly about deep for `templateOptions` (versus an explicit depth=2). If a value within templateOptions is a map, then it will be deep-merged rather than overriding. Most of the values are not, but there are some - e.g. jclouds `TemplateOptions.userMetadata(Map<String, String> userMetadata)`. What would someone expect? Would they expect to merge additional userMetadata or to override the value?

There is also potentially surprising behaviour for lists + sets within templateOptions. These values are merged to concatenate the sets/lists. For example, `TemplateOptions.networks(Iterable<String> networks)` will be merged.

I think for now we just want a sensible default that can be easily explained. And we can worry later about the edge cases where the default doesn't match what someone wants, and it's tricky (or even impossible) for them to avoid that without changing their super-type location-definition.

Another possibility is that we have a parameterised `MERGE` that says the depth. But that seems overly elaborate. Same for saying whether lists/sets should be merged or not.

So in conclusion:
* should we just rename it to DEEP_MERGE?
* should we change the defaults so that it doesn't merge (i.e. concatenate) lists/sets deep within a map when doing a merge?

utils/common/src/main/java/org/apache/brooklyn/config/ConfigInheritance.java
aledsage: Agreed. If we went back in time to review the very first pull request, then I'd change it to be simpler (rather than context-dependent behaviour). But now that we have the code, let's keep it "just in case".

core/src/main/java/org/apache/brooklyn/core/entity/internal/EntityConfigMap.java
neykov: * +1 for `DEEP_MERGE` - better make it explicit
* I didn't realize lists are concatenated - default behaviour in `CollectionMerger` is not to concatenate them and don't see where this is explicitly enabled?

neykov: You convinced me that `DEEP_MERGE` makes sense for current needs.

camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/creation/BrooklynComponentTemplateResolver.java
neykov: It's definitely not wrong, can be done in a future PR. It's just not consistent, so if we start advertising this functionality (and we will since it already changes existing behaviour) we better not surprise users.

neykov:
Great changes @aledsage. Looking forward to using them in blueprints.

core/src/main/java/org/apache/brooklyn/core/entity/internal/EntityConfigMap.java
aledsage: You're right about lists not being merged by default - `CollectionMergerTest.testMergeMapsDefaultsToOverridingSubLists` already tests that.

camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/ConfigInheritanceYamlTest.java
aledsage: That's a really good point! I've confirmed that behaviour in a test. In your example, we'd get ENTITY_ID and CLUSTER_NAME, but not APP_NAME.

That feels very hard to fix. We'd have to change the way `EntityConfigMap.inheritedConfig` is populated and stored. Currently the entity is given a copy of its parent's config (rather than looking it up on the parent each time).

In fact, I'm not sure that declaring `brooklyn.parameters` at each level would work around it! The `BasicConfigurationSupport.refreshInheritedConfig` calls `entity.getAllConfig()`. That just does `Map result; result.putAll(inheritedConfig); result.putAll(ownConfig);` so it ignores the config key's settings for inheritance.

I'm tempted to leave that as a documented limitation, with the failing test marked as WIP+Broken. Does that sound sensible to you @neykov ?

neykov: Yes, agree.

PR #174, opened 2016-06-01
pr173/rename confMapThing.obj

aledsage:

The previous name caused problems because there is also a
“confMapThing” of type MapConfigKey. That has special behaviour,
where it looks up any config with that prefix - so it picked up
any config defined against confMapThing.obj as well.

### DIFF
--- a/core/src/test/java/org/apache/brooklyn/core/entity/EntityConfigTest.java
+++ b/core/src/test/java/org/apache/brooklyn/core/entity/EntityConfigTest.java
@@ -220,21 +220,21 @@ public class EntityConfigTest extends BrooklynAppUnitTestSupport {
                     }})
                 .build();
         TestEntity entity = mgmt.getEntityManager().createEntity(EntitySpec.create(TestEntity.class)
-                .configure(TestEntity.CONF_MAP_THING_OBJECT, ImmutableMap.<String, Object>of("mysub", task))
+                .configure(TestEntity.CONF_MAP_OBJ_THING, ImmutableMap.<String, Object>of("mysub", task))
                 .configure(TestEntity.CONF_NAME, task));
         
         // Will initially return absent, because task is not done
-        assertTrue(entity.config().getNonBlocking(TestEntity.CONF_MAP_THING_OBJECT).isAbsent());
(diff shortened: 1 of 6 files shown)


neykov:
LGTM, but tasks shouldn't be cancelled while trying to get their result. I think this needs further investigation.

core/src/test/java/org/apache/brooklyn/core/entity/EntityConfigTest.java
```diff
@@ -208,8 +210,18 @@ public void testGetConfigMapWithExplicitMap() throws Exception {
...
+    // TODO This now fails because the task has been cancelled, in entity.config().get().
+    // But it used to pass (e.g. with commit 56fcc1632ea4f5ac7f4136a7e04fabf501337540).
+    // It failed after the rename of CONF_MAP_THING_OBJ to CONF_MAP_OBJ_THING, which 
+    // suggests there was an underlying problem that was masked by the unfortunate naming
+    // of the previous "test.confMapThing.obj".
+    //
+    // Presumably an earlier call to task.get() timed out, causing it to cancel the task?
+    // I (Aled) question whether we want to support passing a task (rather than a 
+    // DeferredSupplier or TaskFactory, for example). Our EntitySpec.configure is overloaded
+    // to take a Task, but that feels wrong!?
```
neykov: Passing Tasks as config keys makes sense for me (same as passing a future). Creating a task to compute some value and passing it to an entity is a common thing in Brooklyn.

aledsage: @neykov you're right - we support passing any of Task, Future and DeferredSupplier. I agree we should continue to support those, and should fix whatever is cancelling this task. I think it's probably to do with the implementation of `MapConfigKey`, given the test only broke after the rename (which meant that "obj" couldn't be mistaken for a sub-key of the other MapConfigKey called "test.confMapThing".

I also see that `MapConfigKey.applyEntryValueToMap` does an instanceof check for `Supplier`, but that `ValueResolver` does not - that seems wrong.

PR #175, opened 2016-06-01
Pr173/yaml config tests

aledsage:

Builds on #174 (the config rename is required), so please review + merge that first.

This just tests the existing functionality. Note there are a couple of disabled tests, which demonstrate existing limitations.

### DIFF
--- a/camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/ConfigInheritanceYamlTest.java
+++ b/camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/ConfigInheritanceYamlTest.java
@@ -1,0 +1,316 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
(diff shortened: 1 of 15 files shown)


camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/ConfigInheritanceYamlTest.java
```diff
@@ -0,0 +1,316 @@
...
+    }
+    
+    /**
+     * TODO This hangs because the attributeWhenReady self-reference is resolved against the entity
+     * looking up the config value (i.e. the child). Therefore it waits for the TestEntity to have
+     * a value for that sensor, but this never happens. The way to avoid this is to explicitly set
+     * the component that the attributeWhenReady should apply to (e.g. see {@link #testInheritsParentConfigTask()}.
+     * 
+     * Do we want to just exclude this test? Or do we want to "fix" it? Which entity should the
+     * attributeWhenReady apply to?
```
neykov: +1 for fixing it. We have a similar problem with resource references where they will be resolved against the libraries of the catalog item using the resource, not the one that's declaring it. Need to tag config with the place where it was declared.

neykov:
LGTM

PR #176, opened 2016-06-01
pr173: Refactor stubbing of jclouds for tests

aledsage:

Reason for refactoring is:
1. share the one impl of the stubbed ComputeServiceRegistry (previously we had two)
2. allow that to be re-used by other tests in PR #173, that location's templateOptions config values are correctly merged.


neykov:
LGTM

PR #177, opened 2016-06-01
Adds PublicNetworkFaceEnricher

aledsage:


core/src/main/java/org/apache/brooklyn/core/location/access/PublicNetworkFaceEnricher.java
```diff
@@ -0,0 +1,373 @@
...
+        try {
+            new URI(sensorVal.toString());
+            return true;
+        } catch (URISyntaxException e) {
+            return false;
+        }
+    }
+
+    protected boolean isPort(Object sensorVal) {
+        return (sensorVal instanceof Integer || sensorVal instanceof Long || sensorVal instanceof Short);
```
grkvlt: Not convinced we'd ever use `Long` as a type for a port sensor. And `Short` doesn't even have the full range - why not add `Character` if we're doing this, since it's the only type that accurately represents a port number, being a 16 bit unsigned value from 0-65535?

ahgittin: i'd check that `toString.trim()` matches regex `[0-9]+`, just in case the port is a string number.  that's unambiguously not a URL so we're fine.

core/src/main/java/org/apache/brooklyn/core/location/access/PublicNetworkFaceEnricher.java
```diff
@@ -0,0 +1,373 @@
...
+ *            sensor: main.uri
+ * - type: org.apache.brooklyn.entity.proxy.nginx.NginxController
+ *   brooklyn.config:
+ *     member.sensor.hostandport: $brooklyn:sensor("main.uri.mapped.public")
+ *     serverPool: cluster
+ * }
+ * </pre>
+ */
+@Beta
+public class PublicNetworkFaceEnricher extends AbstractEnricher {
```
grkvlt: Why not `JoinPublicNetworkEnricher`?

core/src/main/java/org/apache/brooklyn/core/location/access/PublicNetworkFaceEnricher.java
```diff
@@ -0,0 +1,373 @@
...
+    public static ConfigKey<Collection<? extends AttributeSensor<?>>> SENSORS = ConfigKeys.newConfigKey(
+            new TypeToken<Collection<? extends AttributeSensor<?>>>() {}, 
+            "sensors",
+            "The multiple sensors whose mapped values are to be re-published (with suffix \"mapped.public\"); "
+                    + "either 'sensor' or 'sensors' should be specified");
+
+    public static final ConfigKey<PortForwardManager> PORT_FORWARD_MANAGER = ConfigKeys.newConfigKey(
+            PortForwardManager.class, 
+            "portForwardManager",
+            "The PortForwardManager storing the port-mappings; if null, the global instance will be used");
```
neykov: Aren't PFM's set on the entity? What if we use this enricher in clocker context.

core/src/main/java/org/apache/brooklyn/core/location/access/PublicNetworkFaceEnricher.java
```diff
@@ -0,0 +1,373 @@
...
+            subscriptions().subscribe(entity, sensor, new SensorEventListener<Object>() {
+                @Override public void onEvent(SensorEvent<Object> event) {
+                    LOG.debug("{} attempting transformations, triggered by sensor-event {}->{}, to {}", 
+                            new Object[] {PublicNetworkFaceEnricher.this, event.getSensor().getName(), event.getValue(), entity});
+                    tryTransform((AttributeSensor<?>)event.getSensor());
+                }});
+        }
+
+        tryTransformAll();
+    }
```
neykov: Is the method called on rebind? If not need to re-subscribe on rebind.

core/src/main/java/org/apache/brooklyn/core/location/access/PublicNetworkFaceEnricher.java
```diff
@@ -0,0 +1,373 @@
...
+
+    protected Maybe<MachineLocation> getMachine() {
+        return Machines.findUniqueMachineLocation(entity.getLocations());
+    }
+    
+    protected PortForwardManager getPortForwardManager() {
+        PortForwardManager portForwardManager = config().get(PORT_FORWARD_MANAGER);
+        if (portForwardManager == null) {
+            portForwardManager = (PortForwardManager) getManagementContext().getLocationRegistry().getLocationManaged("portForwardManager(scope=global)");
```
neykov: `portForwardManager(scope=global)` is used at so many places now, that it deserves to be promoted to a constant.

core/src/main/java/org/apache/brooklyn/core/location/access/PublicNetworkFaceEnricher.java
```diff
@@ -0,0 +1,373 @@
...
+            return Maybe.absent();
+        }
+    }
+
+    protected boolean isUri(Object sensorVal) {
+        if (sensorVal instanceof URI || sensorVal instanceof URL) {
+            return true;
+        }
+        try {
+            new URI(sensorVal.toString());
```
neykov: This will match most strings (as relative URIs). Suggest checking for an existing prefix as well.

core/src/main/java/org/apache/brooklyn/core/location/access/PublicNetworkFaceEnricher.java
```diff
@@ -0,0 +1,373 @@
...
+    protected Maybe<String> transformUri(Entity source, MachineLocation machine, String sensorVal) {
+        URI uri = URI.create(sensorVal);
+        int port = uri.getPort();
+        if (port == -1 && "http".equalsIgnoreCase(uri.getScheme())) port = 80;
+        if (port == -1 && "https".equalsIgnoreCase(uri.getScheme())) port = 443;
+
+        if (port != -1) {
+            HostAndPort publicTarget = getPortForwardManager().lookup(machine, port);
+            if (publicTarget == null) {
+                // TODO What if publicTarget is still null, but will be set soon? We're not subscribed to changes in the PortForwardManager!
```
neykov: I thought that's what `onAssociationCreated` is about.

core/src/main/java/org/apache/brooklyn/core/location/access/PublicNetworkFaceEnricher.java
```diff
@@ -0,0 +1,373 @@
...
+                // TODO Should we return null or sensorVal? In this method we always return sensorVal;
+                //      but in HostAndPortTransformingEnricher we always return null!
+                LOG.trace("network-facing enricher not transforming {} URI {}, because no port-mapping for {}", new Object[] {source, sensorVal, machine});
+                return Maybe.absent();
+            }
+            URI result;
+            try {
+                result = new URI(uri.getScheme(), uri.getUserInfo(), publicTarget.getHostText(), publicTarget.getPort(), uri.getPath(), uri.getQuery(), uri.getFragment());
```
neykov: For the case where no explicit port is defined in source url - It's quite possible that the port mapping has the same public port as the internal one. Suggest comparing and passing -1 for the port if that's the case so that the URL doesn't contain default ports.

core/src/main/java/org/apache/brooklyn/core/location/access/PublicNetworkFaceEnricher.java
aledsage: The enricher doesn't do anything other than publish sensors - it therefore doesn't "join" the public network. The idea behind "face" is that it's setting the sensors for its public-face (a bit like its public facade, but it's not a "facade" in gang-of-4-patterns).

core/src/main/java/org/apache/brooklyn/core/location/access/PublicNetworkFaceEnricher.java
grkvlt: +1 suggested this in #181 as well

core/src/main/java/org/apache/brooklyn/core/location/access/PublicNetworkFaceEnricher.java
grkvlt: @neykov remember we aren't doing this on every sensor, only the ones we have configured.

grkvlt: @aledsage I'd like to have some logic that determines the type based on the sensor name suffix, i.e. `_thing_.url` and `_thing_.url` and _then_ checks whether the value is appropriate for the name. And, if the name doesn't match any of the suffixes we just use the type of the value like this.

core/src/main/java/org/apache/brooklyn/core/location/access/PublicNetworkFaceEnricher.java
aledsage: @neykov I'm not particularly concerned about its use in Clocker 1.x, if the pattern is not a natural one elsewhere.

Looking at the entity's portForwardManager config would be possible, but we don't do that for other enrichers so I don't think we should do it here.

Do you think I should reuse the key `BrooklynAccessUtils.PORT_FORWARDING_MANAGER` (which is named "brooklyn.portforwarding.manager")?

I even wondered about not making it configurable, so the enricher would always use the default global one!

Thoughts?

core/src/main/java/org/apache/brooklyn/core/location/access/PublicNetworkFaceEnricher.java
aledsage: Yes, it's called on rebind. I'll add a test for it though.

core/src/main/java/org/apache/brooklyn/core/location/access/PublicNetworkFaceEnricher.java
aledsage: I'm updating this to explicitly check if `publicTarget.hasPort()`. When constructing this URI, we should therefore never do it without an explicit port.

aledsage:
@neykov @grkvlt @ahgittin I've incorporated those comments - can you take another look please?

core/src/main/java/org/apache/brooklyn/core/network/OnPublicNetworkEnricher.java
```diff
@@ -0,0 +1,496 @@
...
+ * {@code
+ * services:
+ * - type: cluster
+ *   id: cluster
+ *   brooklyn.config:
+ *    memberSpec:
+ *      $brooklyn:entitySpec:
+ *        type: org.apache.brooklyn.entity.webapp.tomcat.TomcatServer
+ *        brooklyn.enrichers:
+ *        - type: org.apache.brooklyn.core.location.access.PublicNetworkFaceEnricher
```
neykov: Update class name in the example

core/src/main/java/org/apache/brooklyn/core/location/access/PublicNetworkFaceEnricher.java
neykov: Looking at `BrooklynAccessUtils.getBrooklynAccessibleAddress` it's looking at the entity for a PFM. How would things work with a non-default PFM - does it need to be set both at the entity and the enricher?
Are we deprecating this method in light of the "Working with multiple networks proposal? 

Looking at the big picture - the port forwarding happens on the location, using the ports of the entity. Doesn't this mean that a PFM is per location (but for practical reasons entity)?

I'm trying to figure out how using a non-default PFM would look like - haven't used one before.

I like the idea of always using the default one - once we have a better understanding of using non-default PFMs and a concrete use case then we can extend the enricher for it.

core/src/main/java/org/apache/brooklyn/core/network/OnPublicNetworkEnricher.java
```diff
@@ -0,0 +1,496 @@
...
+            if (lowerInput.endsWith("uri")) {
+                return input + ".mapped." + network;
+            } else if (lowerInput.endsWith("url")) {
+                return input + ".mapped." + network;
+            } else if (lowerInput.endsWith("endpoint")) {
+                return input + ".mapped." + network;
+            } else if (lowerInput.endsWith("port")) {
+                String prefix = input.substring(0, input.length() - "port".length());
+                if (prefix.endsWith(".")) prefix = prefix.substring(0, prefix.length() - 1);
+                return prefix + ".endpoint.mapped." + network;
```
neykov: Can you explain what's the logic here? Why remove the port suffix?
`redis.httpPort` -> `redis.http.endpoint.mapped.public`
`redis.port` -> `redis.endpoint.mapped.public`
Isn't endpoint reserved for url-like pairs of host + port?

core/src/main/java/org/apache/brooklyn/core/network/OnPublicNetworkEnricher.java
```diff
@@ -0,0 +1,496 @@
...
+        if (machine.isAbsent()) {
+            return;
+        }
+        tryTransform(machine.get(), sensor);
+    }
+    
+    protected void tryTransform(MachineLocation machine, AttributeSensor<?> sensor) {
+        Object sensorVal = entity.sensors().get(sensor);
+        if (sensorVal == null) {
+            return;
```
neykov: On stop all IP/hostname sensors are removed. The port forwards are also lost presumably? Is it better to null the mapped sensors as well?

core/src/main/java/org/apache/brooklyn/core/network/OnPublicNetworkEnricher.java
```diff
@@ -0,0 +1,496 @@
...
+        return portForwardManager;
+    }
+
+    protected void checkConfig() {
+        AttributeSensor<?> sensor = getConfig(SENSOR);
+        Collection<? extends AttributeSensor<?>> sensors = getConfig(SENSORS);
+        Maybe<Object> rawMapMatching = config().getRaw(MAP_MATCHING);
+        String mapMatching = config().get(MAP_MATCHING);
+        
+        if (sensor != null && sensors != null && sensors.isEmpty()) {
```
neykov: Should be `!sensors.isEmpty()`

neykov:
Finished review. Not sure how this gets used with non-default PFMs, otherwise looks good.

core/src/main/java/org/apache/brooklyn/core/location/access/PublicNetworkFaceEnricher.java
aledsage: Yes, I think we will be deprecating this method (i.e. the non-default PFM, or maybe even the PFM) in light of the "Working with multipe networks" proposal.

A PFM just needs to be scoped such that there will not be clashes in the "publicIpId". That field came about from the early days of doing cloudstack integration. It's not really appropriate in many cases. In advanced networking use-cases, we've created a PFM for the "subnet" (which generally meant for the app, or a group of apps that we were carefully including as children of a single SubnetTier entity.

For now, I think we should assume "always using the default". I'm tempted to remove that as a config key from the enricher. But given it's a config key on other things, then it seemed sensible to do the same here. So I'll just leave it as-is for now. We can deprecate config/use of the PFM across the entire code-base if we decide to do that.

core/src/main/java/org/apache/brooklyn/core/network/OnPublicNetworkEnricher.java
aledsage: The value set for the new sensor will always be a host:port, when mapping a port sensor. So "redis.port" will be an int, where as "redis.endpoint.mapped.public" will be a string ip:port.

I wondered about leaving it as "redis.port.mapped.public" (which would be a simpler rule to explain), but felt that including "endpoint" would make it clearer that this is not just a port value.

core/src/main/java/org/apache/brooklyn/core/network/OnPublicNetworkEnricher.java
aledsage: I wondered about that. It depends why the port-mapping was removed from the PFM. If it's because the entity is stopping, then the enricher shouldn't bother trying to keep up with that. But if it's because the the mappings are more dynamic, then it would make sense to remove it.

My reasoning was that we don't delete the main.uri sensor for a Tomcat entity when the tomcat process stops, so we can similarly leave these sensors as-is.

But like I said, if things are a lot more dynamic for DNAT rules added/removed on-the-fly then we should revisit.

core/src/main/java/org/apache/brooklyn/core/network/OnPublicNetworkEnricher.java
aledsage: Good spot! Have added tests as well.

aledsage:
I've incorporated comments; merging now.

We can discuss non-default PFMs as part of the Working with multiple networks" proposal.

core/src/main/java/org/apache/brooklyn/core/network/OnPublicNetworkEnricher.java
neykov: Makes sense with that explanation in mind. But then mapping urls seems wrong. The workflow should be port -> mapped.endpoint.public -> main.uri. Mapping urls means that one of them will always be wrong (either the source or the mapped one).

ahgittin: I prefer the simpler `.port` model as per the feature proposal doc.  Reasoning:
* it *is* a port, on the address associated with the indicated network
* simplicity and consistency --- `XXX` always becomes `XXX.mapped.NETWORK`
* there are times when accessing in other entities that we will need the mapped host and mapped port individually, not the mapped endpoint (eg `ssh -p`)

PR #180, opened 2016-06-03
prevent OOME on big streams and better logging for that case

ahgittin:


core/src/main/java/org/apache/brooklyn/core/mgmt/BrooklynTaskTags.java
```diff
@@ -190,9 +190,16 @@ protected WrappedStream(String streamType, ByteArrayOutputStream stream) {
         public Integer getStreamSize() {
             return streamSize.get();
         }
-        // there is a stream api so don't return everything unless explicitly requested!
+        // there is a stream call on Activity REST api which accesses streamContent.get() directly;
+        // so when serializing the tag, abbreviate things
         @JsonProperty("streamContents")
         public String getStreamContentsAbbreviated() {
+            // TODO would be nice to just get the first 80 chars but that's a refactoring
+            // which might affect persistence.  if stream is very large (100MB+) then we sometimes
```
aledsage: Agree we should be cautious about breaking persistence, so am happy to merge as-is. But would be really nice if we refactor in a safe way (e.g. keeping the old field, but changing the constructor args and adding a new field), so we can write something that gets the truncated data. (And in the future it could return it paginated, but not sure if that would be at all sensible?!).

aledsage:
LGTM; merging.

core/src/main/java/org/apache/brooklyn/core/mgmt/BrooklynTaskTags.java
ahgittin: +1

PR #182, opened 2016-06-04
Change to Catalog API so that policy config is correctly returned

m4rkmckenna:


rest/rest-resources/src/main/java/org/apache/brooklyn/rest/transform/CatalogTransformer.java
```diff
@@ -131,7 +132,11 @@ public static CatalogItemSummary catalogItemSummary(BrooklynRestResourceUtils b,
     }
 
     public static CatalogPolicySummary catalogPolicySummary(BrooklynRestResourceUtils b, CatalogItem<? extends Policy,PolicySpec<?>> item, UriBuilder ub) {
-        Set<PolicyConfigSummary> config = ImmutableSet.of();
+        final Set<PolicyConfigSummary> config = new HashSet<>();
+        final PolicySpec<?> spec = (PolicySpec<?>) b.getCatalog().createSpec((CatalogItem) item);
```
neykov: wrap in a `try-catch`, could throw on invalid catalog item (for example some dependency no longer exists)

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/transform/EntityTransformer.java
```diff
@@ -122,6 +123,12 @@ public static EntityConfigSummary entityConfigSummary(ConfigKey<?> config, Strin
         Map<String, URI> mapOfLinks =  links==null ? null : ImmutableMap.copyOf(links);
         return new EntityConfigSummary(config, label, priority, mapOfLinks);
     }
+
+    public static PolicyConfigSummary policyConfigSummary(ConfigKey<?> config, String label, Double priority, Map<String, URI> links) {
+        Map<String, URI> mapOfLinks =  links==null ? null : ImmutableMap.copyOf(links);
```
neykov: Isn't this already handled in the constructor below?

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/transform/CatalogTransformer.java
```diff
@@ -131,7 +132,11 @@ public static CatalogItemSummary catalogItemSummary(BrooklynRestResourceUtils b,
...
+        for (final SpecParameter<?> input : spec.getParameters()){
```
neykov: Parameters are already ordered and you lose the ordering with the `HashSet`. Suggest to use `LinkedHashSet`. (for the case of = priority).
The API doesn't actually guarantee ordering (it's a `Set`), but Javascript will keep it anyway. Decided to keep it like this (for the entity case) and not introduce backward compatibility issues until needed.

m4rkmckenna:
@neykov comments addressed

neykov:
LGTM, merging.

PR #185, opened 2016-06-07
Hide sensitive parameters from effectors

bostko:
For review.
@ygy can you check it?

### DIFF
--- a/api/src/main/java/org/apache/brooklyn/api/effector/ParameterType.java
+++ b/api/src/main/java/org/apache/brooklyn/api/effector/ParameterType.java
@@ -27,6 +27,8 @@ import javax.management.MBeanParameterInfo;
  *
  * @see Effector
  */
+// TODO rename this to ParameterInfo rather than type. A ParameterType object is used for each parameter not for each type of parameter.
+// check MethodEffector.AnnotationsOnMethod.toParameterType
 public interface ParameterType<T> extends Serializable {
     
     public String getName();
--- a/core/src/main/java/org/apache/brooklyn/core/effector/BasicParameterType.java
+++ b/core/src/main/java/org/apache/brooklyn/core/effector/BasicParameterType.java
@@ -75,7 +75,7 @@ public class BasicParameterType<T> implements ParameterType<T> {
 
     @Override
     public Class<T> getParameterClass() { return type; }
-    
+
     @Override
     public String getParameterClassName() { return type.getCanonicalName(); }
 
--- a/core/src/test/java/org/apache/brooklyn/core/mgmt/internal/TestEntityWithEffectors.java
+++ b/core/src/test/java/org/apache/brooklyn/core/mgmt/internal/TestEntityWithEffectors.java
@@ -1,0 +1,30 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.brooklyn.core.mgmt.internal;
+
+import org.apache.brooklyn.api.entity.ImplementedBy;
+import org.apache.brooklyn.core.annotation.Effector;
+import org.apache.brooklyn.core.annotation.EffectorParam;
+import org.apache.brooklyn.core.test.entity.TestEntity;
+
+@ImplementedBy(TestEntityWithEffectorsImpl.class)
+public interface TestEntityWithEffectors extends TestEntity {
+    @Effector(description = "Reset User pass effetor")
+    Void resetUserPassword(@EffectorParam(name = "newPassword") String param1);
+}
--- a/core/src/test/java/org/apache/brooklyn/core/mgmt/internal/TestEntityWithEffectors.java
+++ b/core/src/test/java/org/apache/brooklyn/core/mgmt/internal/TestEntityWithEffectors.java
@@ -25,6 +25,11 @@ import org.apache.brooklyn.core.test.entity.TestEntity;
 
 @ImplementedBy(TestEntityWithEffectorsImpl.class)
 public interface TestEntityWithEffectors extends TestEntity {
-    @Effector(description = "Reset User pass effetor")
-    Void resetUserPassword(@EffectorParam(name = "newPassword") String param1);
+    @Effector(description = "Reset password")
+    Void resetPassword(@EffectorParam(name = "newPassword") String param1, @EffectorParam(name = "secretPin") Integer secretPin);
+
+    @Effector(description = "Reset User and password effector")
+    Void invokeUserAndPassword(@EffectorParam(name = "user") String user,
+                               @EffectorParam(name = "newPassword", defaultValue = "Test") String newPassword,
+                               @EffectorParam(name = "paramDefault", defaultValue = "DefaultValue") String paramDefault);
 }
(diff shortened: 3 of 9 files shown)


bostko:
After when you have the code from this branch and brooklyn-ui you can use the blueprint bellow to check how it looks in the UI
```yaml
location: localhost
services:
- type: org.apache.brooklyn.giulauncher.effectors.TestEntityWithEffectors
```

rest/rest-api/src/main/java/org/apache/brooklyn/rest/domain/EffectorSummary.java
```diff
@@ -61,6 +64,10 @@ public String getType() {
             return type;
         }
 
+        public boolean isSensitive() {
```
bostko: What do you think about naming it `shouldSanitize`?

core/src/main/java/org/apache/brooklyn/core/effector/BasicParameterType.java
```diff
@@ -44,6 +45,7 @@ public BasicParameterType(Map<?, ?> arguments) {
         if (arguments.containsKey("type")) type = (Class<T>) arguments.get("type");
         if (arguments.containsKey("description")) description = (String) arguments.get("description");
         if (arguments.containsKey("defaultValue")) defaultValue = (T) arguments.get("defaultValue");
+        if (arguments.containsKey("sensitive")) defaultValue = (T) arguments.get("sensitive");
```
sjcorbett: Mistakenly overwrites `defaultValue`.

sjcorbett: Is it possible for the value in the map to be null? If yes then use `Boolean.TRUE.equals(args.get(..))`.

core/src/test/java/org/apache/brooklyn/core/mgmt/internal/TestEntityWithEffectors.java
```diff
@@ -0,0 +1,34 @@
...
+package org.apache.brooklyn.core.mgmt.internal;
+
+import org.apache.brooklyn.api.entity.ImplementedBy;
+import org.apache.brooklyn.core.annotation.Effector;
+import org.apache.brooklyn.core.annotation.EffectorParam;
+import org.apache.brooklyn.core.test.entity.TestEntity;
+
+@ImplementedBy(TestEntityWithEffectorsImpl.class)
+public interface TestEntityWithEffectors extends TestEntity {
+    @Effector(description = "Reset User pass effetor")
```
sjcorbett: typo: `effetor`

core/src/test/java/org/apache/brooklyn/core/mgmt/internal/TestEntityWithEffectorsImpl.java
```diff
@@ -0,0 +1,42 @@
...
+ */
+package org.apache.brooklyn.core.mgmt.internal;
+
+import org.apache.brooklyn.core.annotation.EffectorParam;
+import org.apache.brooklyn.core.test.entity.TestEntityImpl;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class TestEntityWithEffectorsImpl extends TestEntityImpl implements TestEntityWithEffectors {
+    private static final Logger log = LoggerFactory.getLogger(EffectorUtils.class);
```
sjcorbett: EffectorUtils?

launcher/src/test/java/org/apache/brooklyn/giulauncher/effectors/TestEntityWithEffectors.java
```diff
@@ -0,0 +1,33 @@
...
+package org.apache.brooklyn.giulauncher.effectors;
+
+import org.apache.brooklyn.api.entity.ImplementedBy;
+import org.apache.brooklyn.core.annotation.Effector;
+import org.apache.brooklyn.core.annotation.EffectorParam;
+import org.apache.brooklyn.core.test.entity.TestEntity;
+
+@ImplementedBy(TestEntityWithEffectorsImpl.class)
+public interface TestEntityWithEffectors extends TestEntity {
+    @Effector(description = "Reset User pass effetor")
```
sjcorbett: `effetor` here and below again.

launcher/src/test/java/org/apache/brooklyn/giulauncher/effectors/TestEntityWithEffectorsImpl.java
```diff
@@ -0,0 +1,43 @@
...
+package org.apache.brooklyn.giulauncher.effectors;
+
+import org.apache.brooklyn.core.annotation.EffectorParam;
+import org.apache.brooklyn.core.mgmt.internal.EffectorUtils;
+import org.apache.brooklyn.core.test.entity.TestEntityImpl;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class TestEntityWithEffectorsImpl extends TestEntityImpl implements TestEntityWithEffectors {
+    private static final Logger log = LoggerFactory.getLogger(EffectorUtils.class);
```
sjcorbett: Wrong class again.

launcher/src/test/java/org/apache/brooklyn/giulauncher/effectors/TestEntityWithEffectors.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+package org.apache.brooklyn.giulauncher.effectors;
+
+import org.apache.brooklyn.api.entity.ImplementedBy;
+import org.apache.brooklyn.core.annotation.Effector;
+import org.apache.brooklyn.core.annotation.EffectorParam;
+import org.apache.brooklyn.core.test.entity.TestEntity;
+
+@ImplementedBy(TestEntityWithEffectorsImpl.class)
+public interface TestEntityWithEffectors extends TestEntity {
```
aledsage: Why do we define this again in `launcher`? 

The `launcher/pom.xml` already has a dependency on core's tests, so we should just be able to use them without duplicating:

        <dependency>
            <artifactId>brooklyn-core</artifactId>
            <scope>test</scope>
        </dependency>

api/src/main/java/org/apache/brooklyn/api/effector/ParameterType.java
```diff
@@ -45,4 +47,6 @@
      * @return The default value for this parameter, if not supplied during an effector call.
      */
     public T getDefaultValue();
+
+    public boolean isSensitive();
```
aledsage: Given your question about naming `shouldSanitize` or `isSensitive`, let's mark this as `@Beta`.

aledsage: I really worry about adding this - do we actually respect it properly? In many places we just have a `ConfigBag` without the `ParameterType` definition, so we can only infer from the name whether it looks like a password/credential.

If we include `isSensitive` here, then it will set a false expectation of what we'll do with it.

Should we just rely on `sanitize`, and get rid of `isSensitive`?

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/transform/EffectorTransformer.java
```diff
@@ -83,7 +84,8 @@ public static EffectorSummary effectorSummaryForCatalog(Effector<?> effector) {
                 .context(entity).timeout(ValueResolver.REAL_QUICK_WAIT).getMaybe();
             return new ParameterSummary(parameterType.getName(), parameterType.getParameterClassName(), 
                 parameterType.getDescription(), 
-                WebResourceUtils.getValueForDisplay(defaultValue.orNull(), true, false));
+                WebResourceUtils.getValueForDisplay(defaultValue.orNull(), true, false),
+                Sanitizer.IS_SECRET_PREDICATE.apply(parameterType.getName()));
```
aledsage: It looks like you added `parameterType.isSensitive`, but then don't use it here (and instead rely on `Sanitizer`). As I said in earlier comment, I think that's fine - in fact probably best to delete `isSensitive` if we're not going to use it everywhere.

core/src/main/java/org/apache/brooklyn/core/effector/BasicParameterType.java
```diff
@@ -32,6 +32,7 @@
     private Class<T> type;
     private String description;
     private Boolean hasDefaultValue = null;
+    private boolean shouldSanitize = false;
```
aledsage: Why include `shouldSanitize` here? It's private and thus never used.

bostko: Sorry, you are right. It is needed only in `EffectorSummary.ParameterSummary`
Check my commit now.

aledsage:
Thanks @bostko - merging now.

PR #186, opened 2016-06-07
Rename SimpleShellCommandTest to TestSshCommand

aledsage:


test-framework/src/main/java/org/apache/brooklyn/test/framework/TestSshCommandImpl.java
```diff
@@ -0,0 +1,269 @@
...
+
+        ProcessTaskWrapper<Integer> job = DynamicTasks.queue(etf);
+        job.asTask().blockUntilEnded();
+        return new Result(job);
+    }
+
+
+
+    private IllegalArgumentException illegal(String message, String... messages) {
+        Iterable<String> allmsgs = Iterables.concat(MutableList.of(this.toString() + ":", message), Arrays.asList(messages));
```
geomacy: Just for interest, why change this from the previous implementation that just joined the arrays?

geomacy:
👍  Looks like a good change to me.   Will require matching updates in the brooklyn-docs.

test-framework/src/main/java/org/apache/brooklyn/test/framework/TestSshCommandImpl.java
aledsage: There were previously generics warnings. I *think* what would happen before with `Joiner.on(' ').join(this.toString() + ":", message, messages)` is that it would give you a joined the iterable of `[toString()+":", message, messages]` (i.e. it would have called `messages.toString()` rather than concatenating the contents of that array with the others. You get away with that because the toString is ok. You'd just get some extra square brackets in the result.

aledsage:
`TestSshCommandIntegrationTest.shouldCaptureStdoutAndStderrOfScript ` is a real failure - that should have been marked as `groups="Integration"`.

aledsage:
Test failure is unrelated - `ConfigYamlTest.testDeferredSupplierToAttributeWhenReady`. This is getting ridiculous! I strongly suspect there is an underlying problem in AbstractEntity.initEnrichers that causes service.isUp to be cleared, which can cause *many* tests to fail. But that it's a very unlikely race, so we only see it one in a few thousand test runs.

I'll look at that separately.

For now, merging this PR.

PR #187, opened 2016-06-07
BROOKLYN-291: don’t override manually-set service.isUp

aledsage:


aledsage:
Thanks @ahgittin - merging now.

PR #189, opened 2016-06-08
Let downstream projects pass custom brooklyn properties to Brooklyn

neykov:
The properties passed in `getBrooklynPropertiesSupplier` have less precedence than global and local  properties.


aledsage:
LGTM; merging.

PR #190, opened 2016-06-08
Sanitizer: avoid NPE if sanitize null key

aledsage:


aledsage:
@bostko @nakomis can you review this please? We now hit the NPE after merging #185, because some `ParameterType`s of effectors have null parameter names, it seems!

Merging now.

PR #191, opened 2016-06-09
Add AbstractTransformer.ALLOW_CYCLIC_PUBLISHING

aledsage:
If set to true, then allows the transformer to read and publish on the 
same sensor. This is needed for some use-cases in 
brooklyncentral/advanced-networking.

### DIFF
--- a/core/src/test/java/org/apache/brooklyn/enricher/stock/TransformingEnricherTest.java
+++ b/core/src/test/java/org/apache/brooklyn/enricher/stock/TransformingEnricherTest.java
@@ -24,6 +24,7 @@ import org.apache.brooklyn.api.entity.EntitySpec;
 import org.apache.brooklyn.api.sensor.AttributeSensor;
 import org.apache.brooklyn.api.sensor.EnricherSpec;
 import org.apache.brooklyn.api.sensor.Sensor;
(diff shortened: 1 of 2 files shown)


aledsage:
See https://github.com/brooklyncentral/advanced-networking/pull/99 for an example usage of this.

core/src/test/java/org/apache/brooklyn/enricher/stock/TransformingEnricherTest.java
```diff
@@ -225,17 +226,38 @@ public void testInfersTargetSensorSameAsSource() throws Exception {
...
+                        if (input != null && input < 10) {
+                            return input + 1;
+                        } else {
+                            return Entities.UNCHANGED;
+                        }
+                    }}));
+
+        app.sensors().set(intSensorA, 3);
+
+        EntityAsserts.assertAttributeEqualsEventually(app, intSensorA, 10);
```
nakomis: Follow up with `EntityAsserts.assertAttributeEqualsContinually` to ensure value remains at 10

nakomis:
One comment, but other than that, LGTM

aledsage:
Thanks @nakomis - merging.

PR #192, opened 2016-06-09
Adds OnSubnetNetworkEnricher

aledsage:
Some entities publish their endpoint using the public hostname (e.g. `MySqlNode` publishes `datastore.url` using the sensor `host.name`). If trying to connect to it from within the same subnet, this can fail if the public ip/hostname is not accessible.

This topic is covered in the "Working with Multiple Networks" proposal (see dev@brooklyn email thread). That suggests we should use the subnet address in things like the datastore.url. That will likely break things for some of our users. But while we discuss/finalise that proposal...

This enricher allows one to publish another sensor that definitely uses the subnet address. It will publish the sensor as things like `datastore.url.mapping.subnet`.

This is really a temporary solution (hence marking this as `@Beta`).

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/network/AbstractOnNetworkEnricher.java
+++ b/core/src/main/java/org/apache/brooklyn/core/network/AbstractOnNetworkEnricher.java
@@ -61,23 +61,16 @@ public abstract class AbstractOnNetworkEnricher extends AbstractEnricher {
     private static final Logger LOG = LoggerFactory.getLogger(AbstractOnNetworkEnricher.class);
 
     @SuppressWarnings("serial")
-    public static final ConfigKey<AttributeSensor<?>> SENSOR = ConfigKeys.newConfigKey(
-            new TypeToken<AttributeSensor<?>>() {}, 
-            "sensor",
-            "The sensor whose mapped value is to be re-published (with suffix \"mapped.networkName\"); "
-                    + "either 'sensor' or 'sensors' should be specified");
-
-    @SuppressWarnings("serial")
     public static ConfigKey<Collection<? extends AttributeSensor<?>>> SENSORS = ConfigKeys.newConfigKey(
             new TypeToken<Collection<? extends AttributeSensor<?>>>() {}, 
             "sensors",
             "The multiple sensors whose mapped values are to be re-published (with suffix \"mapped.networkName\"); "
-                    + "if neither 'sensor' or 'sensors' is specified, defaults to 'mapAll'");
+                    + "if 'sensors' is not specified, defaults to 'mapMatching'");
 
     public static ConfigKey<String> MAP_MATCHING = ConfigKeys.newStringConfigKey(
             "mapMatching",
             "Whether to map all, based on a sensor naming convention (re-published with suffix \"mapped.networkName\"); "
-                    + "if neither 'sensor' or 'sensors' is specified, defaults to matchin case-insensitive suffix of "
+                    + "if 'sensors' is not specified, defaults to matching case-insensitive suffix of "
                     + "'port', 'uri', 'url' or 'endpoint' ",
             "(?i).*(port|uri|url|endpoint)");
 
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/apache/brooklyn/core/network/AbstractOnNetworkEnricher.java
```diff
@@ -0,0 +1,407 @@
...
+            Optional<HostAndPort> mappedEndpoint = getMappedEndpoint(source, machine, sensorVal);
+            if (!mappedEndpoint.isPresent()) {
+                LOG.debug("network-facing enricher not transforming {} port {}, because no port-mapping for {}", new Object[] {source, sensorVal, machine});
+                return Maybe.absent();
+            }
+            if (!mappedEndpoint.get().hasPort()) {
+                LOG.debug("network-facing enricher not transforming {} port {}, because no port in target {} for {}", new Object[] {source, sensorVal, mappedEndpoint, machine});
+                return Maybe.absent();
+            }
+            return Maybe.of(mappedEndpoint.get().toString());
```
Graeme-Miller: sensor is a port, but we are returning a HostAndPort here

core/src/main/java/org/apache/brooklyn/core/network/AbstractOnNetworkEnricher.java
```diff
@@ -0,0 +1,407 @@
...
+import com.google.common.net.HostAndPort;
+import com.google.common.reflect.TypeToken;
+
+@Beta
+public abstract class AbstractOnNetworkEnricher extends AbstractEnricher {
+
+    private static final Logger LOG = LoggerFactory.getLogger(AbstractOnNetworkEnricher.class);
+
+    @SuppressWarnings("serial")
+    public static final ConfigKey<AttributeSensor<?>> SENSOR = ConfigKeys.newConfigKey(
```
Graeme-Miller: why have SENSOR and SENORS?
Would it not be simpler to just have SENSORS? If someone only wants one sensor to be mapped they can configure SENSORS with one entry.

Graeme-Miller:
Couple of comments then LGTM

core/src/main/java/org/apache/brooklyn/core/network/AbstractOnNetworkEnricher.java
```diff
@@ -0,0 +1,407 @@
...
+            "sensors",
+            "The multiple sensors whose mapped values are to be re-published (with suffix \"mapped.networkName\"); "
+                    + "if neither 'sensor' or 'sensors' is specified, defaults to 'mapAll'");
+
+    public static ConfigKey<String> MAP_MATCHING = ConfigKeys.newStringConfigKey(
+            "mapMatching",
+            "Whether to map all, based on a sensor naming convention (re-published with suffix \"mapped.networkName\"); "
+                    + "if neither 'sensor' or 'sensors' is specified, defaults to matchin case-insensitive suffix of "
+                    + "'port', 'uri', 'url' or 'endpoint' ",
+            "(?i).*(port|uri|url|endpoint)");
```
geomacy: Maybe worth adding an explicit "." in front of the port|uri|url|endpoint?    Just in case e.g. "port" would unintentionally match a sensor called xyz_export or something?

core/src/main/java/org/apache/brooklyn/core/network/AbstractOnNetworkEnricher.java
```diff
@@ -0,0 +1,407 @@
...
+            if (lowerInput.endsWith("uri")) {
+                return input + ".mapped." + network;
+            } else if (lowerInput.endsWith("url")) {
+                return input + ".mapped." + network;
+            } else if (lowerInput.endsWith("endpoint")) {
+                return input + ".mapped." + network;
+            } else if (lowerInput.endsWith("port")) {
+                String prefix = input.substring(0, input.length() - "port".length());
+                if (prefix.endsWith(".")) prefix = prefix.substring(0, prefix.length() - 1);
+                return prefix + ".endpoint.mapped." + network;
```
geomacy: Why ".endpoint.mapped" here and not ".port.mapped"?

core/src/main/java/org/apache/brooklyn/core/network/AbstractOnNetworkEnricher.java
```diff
@@ -0,0 +1,407 @@
...
+        } catch (URISyntaxException e) {
+            return false;
+        }
+    }
+
+    protected boolean isPort(Object sensorVal) {
+        if (sensorVal instanceof Integer || sensorVal instanceof Long) {
+            return Networking.isPortValid(((Number)sensorVal).intValue());
+        } else if (sensorVal instanceof CharSequence) {
+            return sensorVal.toString().trim().matches("[0-9]+");
```
geomacy: Is it not worth doing the isPortValid check here too? i.e. convert to number and then same as line 258

core/src/main/java/org/apache/brooklyn/core/network/AbstractOnNetworkEnricher.java
geomacy: Although that then won't match xyz.HostAndPort.  Hm. Would it be worth checking for that suffix explicitly?

geomacy:
Couple of minor questions and comments but none that *need* done,  looks good to me.

core/src/main/java/org/apache/brooklyn/core/network/AbstractOnNetworkEnricher.java
aledsage: Correct. The mapped value tells us exactly which IP to use, as well as the port number.

core/src/main/java/org/apache/brooklyn/core/network/AbstractOnNetworkEnricher.java
aledsage: Yes, you're right. I'll delete `SENSOR` before this gets into any official release!

core/src/main/java/org/apache/brooklyn/core/network/AbstractOnNetworkEnricher.java
aledsage: @geomacy yes, you're right. I checked what we're doing elsewhere and it does assume a ".port" suffix:
https://github.com/apache/brooklyn-server/blob/rel/apache-brooklyn-0.9.0/software/base/src/main/java/org/apache/brooklyn/entity/software/base/SoftwareProcess.java#L59-L61
https://github.com/apache/brooklyn-server/blob/rel/apache-brooklyn-0.9.0/software/base/src/main/java/org/apache/brooklyn/entity/software/base/InboundPortsUtils.java#L76-L96

core/src/main/java/org/apache/brooklyn/core/network/AbstractOnNetworkEnricher.java
Graeme-Miller: ok, makes sense are we are using a different name for the mapped sensor

aledsage:
Thanks all - merging now.

core/src/main/java/org/apache/brooklyn/core/network/AbstractOnNetworkEnricher.java
neykov: It does, but that rules out sensors like `httpPort` which are used in the wild. Perhaps we should standardize on the sensor naming instead.

core/src/main/java/org/apache/brooklyn/core/network/AbstractOnNetworkEnricher.java
```diff
@@ -0,0 +1,402 @@
...
+            if (Strings.isBlank(mapMatching)) {
+                throw new IllegalStateException(this+" requires 'sensors' config (when 'mapMatching' is explicitly blank)");
+            }
+        } else if (rawMapMatching.isPresent()) {
+            throw new IllegalStateException(this+" must not have explicit 'mapMatching' and 'sensors' config");
+        }
+    }
+    
+    protected Collection<AttributeSensor<?>> resolveSensorsConfig() {
+        Collection<? extends AttributeSensor<?>> sensors = getConfig(SENSORS);
```
neykov: Could've gone one step further and accept both `String` and `List` after removing `SENSOR`.

PR #193, opened 2016-06-09
fix for issue where $brooklyn:external refs in provisioning.properties are resolved too early

ahgittin:
several more tests, including one which failed before the fix -- and of course the fix itself

PR #194, opened 2016-06-10
Prevents premature resolving of external config

nakomis:
If external config is resolved too early, it can lead to resolved values being persisted (security flaw). Also, changes to the external config supplier are not reflected when the config is subsequently read

### DIFF
--- a/camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/JcloudsLocationExternalConfigYamlTest.java
+++ b/camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/JcloudsLocationExternalConfigYamlTest.java
@@ -1,0 +1,124 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.brooklyn.camp.brooklyn;
+
+import static org.testng.Assert.assertEquals;
+
+import java.io.File;
+import java.io.StringReader;
+
+import org.apache.brooklyn.api.entity.Entity;
+import org.apache.brooklyn.api.location.Location;
+import org.apache.brooklyn.camp.brooklyn.ExternalConfigYamlTest.MyExternalConfigSupplier;
+import org.apache.brooklyn.camp.brooklyn.ExternalConfigYamlTest.MyExternalConfigSupplierWithoutMapArg;
+import org.apache.brooklyn.config.ConfigKey;
+import org.apache.brooklyn.core.config.ConfigKeys;
+import org.apache.brooklyn.core.entity.StartableApplication;
+import org.apache.brooklyn.core.internal.BrooklynProperties;
+import org.apache.brooklyn.core.mgmt.internal.LocalManagementContext;
+import org.apache.brooklyn.core.mgmt.rebind.RebindTestUtils;
+import org.apache.brooklyn.core.objs.BrooklynObjectInternal;
+import org.apache.brooklyn.entity.software.base.EmptySoftwareProcess;
+import org.apache.brooklyn.location.jclouds.JcloudsLocation;
+import org.apache.brooklyn.util.core.internal.ssh.SshTool;
+import org.apache.brooklyn.util.core.task.DeferredSupplier;
+import org.apache.brooklyn.util.guava.Maybe;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+import com.google.common.base.Joiner;
+import com.google.common.collect.Iterables;
+
+// also see ExternalConfigYamlTest
+public class JcloudsLocationExternalConfigYamlTest extends AbstractYamlRebindTest {
+
+    private static final Logger log = LoggerFactory.getLogger(ExternalConfigYamlTest.class);
+
+    private static final ConfigKey<String> MY_CONFIG_KEY = ConfigKeys.newStringConfigKey("my.config.key");
+
+    @Override
+    protected BrooklynProperties createBrooklynProperties() {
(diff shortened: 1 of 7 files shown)


nakomis:
This is in addition to the changes in https://github.com/apache/brooklyn-server/pull/193

camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/JcloudsLocationExternalConfigYamlTest.java
```diff
@@ -0,0 +1,143 @@
...
+
+// also see ExternalConfigYamlTest
+public class JcloudsLocationExternalConfigYamlTest extends AbstractYamlRebindTest {
+
+    private static final Logger log = LoggerFactory.getLogger(ExternalConfigYamlTest.class);
+
+    private static final ConfigKey<String> MY_CONFIG_KEY = ConfigKeys.newStringConfigKey("my.config.key");
+
+    /** @return A started management context */
+    protected LocalManagementContext createOrigManagementContext() {
```
ahgittin: i think in 0.10.0-SNAPSHOT the superclass has a `createBrooklynProperties()` method you could extend instead, removing the `create*ManagementContext(...)` overrides

camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/JcloudsLocationExternalConfigYamlTest.java
```diff
@@ -0,0 +1,143 @@
...
+        Location l = Iterables.getOnlyElement( entity.getLocations() );
+        log.info("Location: "+l);
+        assertEquals(l.config().get(JcloudsLocation.PASSWORD), "myval");
+
+        Maybe<Object> rawConfig = ((BrooklynObjectInternal.ConfigurationSupportInternal)l.config()).getRaw(ConfigKeys.newStringConfigKey("password"));
+        log.info("Raw config password: "+rawConfig);
+        Assert.assertTrue(rawConfig.isPresentAndNonNull());
+        Assert.assertTrue(rawConfig.get() instanceof DeferredSupplier, "Expected deferred raw value; got "+rawConfig.get());
+
+        // these are null as only recognised provisioning properties are transmitted by jclouds
```
ahgittin: this is an ugly inconsistency between how jclouds handles `obtain(flags)` and others (eg localhost, see the test in the other class) which i had never noticed

@nakomis @aledsage @andreaturli @neykov wdyt, should unrecognised flags passed in `obtain(...)` *(a)* be passed through and set on the provisioned machine location, as in localhost, or *(b)* be dropped (optionally with warning or failure), as in jclouds ?

(not necessarily for this PR, this is just where we've noticed it)

core/src/main/java/org/apache/brooklyn/util/core/config/ConfigBag.java
```diff
@@ -367,6 +367,22 @@ public Object getStringKey(String key) {
         }
     }
 
+    /** Puts into this bag the raw value for the given key in the given bag, if it was present. */
+    @Beta
+    public <T> ConfigBag copyRaw(ConfigBag source, ConfigKey<T> key) {
```
ahgittin: this is the significant api addition

it feels right to me but wdyt @aledsage ?

also @nakomis we should probably add a test for this

ahgittin: would `copy(...)` be a better name?  `Raw` makes it feel ugly when actually i think this is quite a nice pattern for dealing with transferring values without resolving them

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -2422,7 +2437,7 @@ protected JcloudsWinRmMachineLocation createWinRmMachineLocation(ComputeService
                     .configure(WinRmMachineLocation.WINRM_CONFIG_PORT, sshHostAndPort.isPresent() ? sshHostAndPort.get().getPort() : node.getLoginPort())
                     .configure("user", getUser(setup))
                     .configure(WinRmMachineLocation.USER, setup.get(USER))
-                    .configure(WinRmMachineLocation.PASSWORD, setup.get(PASSWORD))
+                    .configure(ConfigBag.newInstance().copyRaw(setup, PASSWORD, WinRmMachineLocation.PASSWORD).getAllConfigRaw())
```
ahgittin: this changes the spec configuration to set `password` as a flag instead of a config key, and it feels ugly

as @nakomis noted to me offline the WinRM impl doesn't create a user so this field is probably actually being ignored, and even if not the flag/config change should result in the same WinRM machine location configuration

but if we do start doing this more (which seems likely if we want to prevent resolution) we probably want to introduce something cleaner such as

    configureCopyKey(ConfigBag, ConfigKey<?>)
    configureCopyKeys(ConfigBag, ConfigKey<?>, ConfigKey<?>, ...ConfigKey<?>)
    configureCopyKeyAs(ConfigBag, ConfigKey<?> sourceName, ConfigKey<?> targetName)

core/src/main/java/org/apache/brooklyn/util/core/config/ConfigBag.java
ahgittin: or in line with the suggestion on `configureCopy...` below

    copyKey(ConfigBag, ConfigKey<?>)
    copyKeys(ConfigBag, ConfigKey<?>, ConfigKey<?>, ...ConfigKey<?>)
    copyKeyAs(ConfigBag, ConfigKey<?> sourceName, ConfigKey<?> targetName)

ahgittin:
LGTM, minor comments, including test and naming

let's give it a day for @aledsage @andreaturli @neykov to comment on names/future and see whether to do in this PR but happy to merge as is

neykov:
I'm fine with merging the proposed changes, but I think it's papering over a broken abstraction. We can't hide the fact that we have deferred values in the configuration so better expose it into the API and make use of it.

I propose that longer term we add methods that return something similar to `ListenableFuture` (even for already resolved values). Then use the existing `get` methods as a shorthand when we will make immediate use of the value. More complex scenarios will push the futures around. This will even allow having chains of dependent values so that when one changes all dependents change as well - for example having a live `$brooklyn:formatString` updating as the underlying sensors update.
This still leaves the case of complex objects - we are currently resolving their parts ad-hoc (for example maps). Could we improve our parser so it knows that an object contains futures and wrap the object itself in a future?

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
nakomis: This would require adding a dependency on brooklyn-core to brooklyn-api, which would create a circular dependency

nakomis:
PR comments addressed

ahgittin:
+1 @neykov we could have `coerced()` and `future()` methods with `get()` a shorthand for the former

we still have the question what should `obtain(...)` do with unrecognised flags -- current behaviour is inconsistent.  address that another day.

merging, thanks @nakomis .

PR #195, opened 2016-06-10
Sanitize effector parameters when reporting an exception

bostko:
It is possible to show such parameters on other places.

### DIFF
--- a/core/src/test/java/org/apache/brooklyn/core/mgmt/internal/TestEntityWithEffectors.java
+++ b/core/src/test/java/org/apache/brooklyn/core/mgmt/internal/TestEntityWithEffectors.java
@@ -21,15 +21,79 @@ package org.apache.brooklyn.core.mgmt.internal;
 import org.apache.brooklyn.api.entity.ImplementedBy;
 import org.apache.brooklyn.core.annotation.Effector;
 import org.apache.brooklyn.core.annotation.EffectorParam;
+import org.apache.brooklyn.core.effector.MethodEffector;
 import org.apache.brooklyn.core.test.entity.TestEntity;
 
+/**
+ * Entity for testing that secret effector parameters are:
+ * <ul>
+ *   <li>excluded from the activities view
+ *   <li>not logged
+ *   <li>masked out in the UI
+ * </ul>
+ * Of those, only the first is unit-tested.
+ * 
+ * To test manually...
+ * 
+ * Configure logback to log everything at trace:
+ * <pre>
+ * {@code
+ * <configuration>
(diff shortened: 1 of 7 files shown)


ygy:
LGTM

aledsage:
Whoever merges this, please look at marking https://issues.apache.org/jira/browse/BROOKLYN-282 as resolved.

aledsage:
@bostko I've created https://github.com/bostko/brooklyn-server/pull/1 - please review. If you merge it into your branch, then it should automatically appear in this PR.

For the apache record, that PR does:

    Also refactors EffectorUtils.prepareArgsForEffector (deprecating old
    method), and updates how MethodEffector calls it so we have the 
    parameter names.

    The previous calls to `sanitizeArgs(args)` where args was not of type `Map` would do nothing. 
    Therefore if this was ever called with an array etc, we'd log all of the parameter values.

core/src/test/java/org/apache/brooklyn/core/mgmt/internal/TestEntityWithEffectors.java
```diff
@@ -21,15 +21,77 @@
...
+ * export JAVA_OPTS="-Xms256m -Xmx1g -XX:MaxPermSize=256m -Dlogback.configurationFile=/path/to/logback-trace.xml"
+ * ./bin/brooklyn launch --persist auto --persistenceDir /path/to/persistedState
+ * }
+ * </pre>
+ * 
+ * Deploy the blueprint below:
+ * <pre>
+ * {@code
+ * services:
+ * - type: org.apache.brooklyn.core.mgmt.internal.TestEntityWithEffectors
```
bostko: I had to add 
```xml
    <dependency>
        <artifactId>brooklyn-core</artifactId>
    </dependency>
```
In my maven file to use this entity.
@aledsage do you think we should this note here?

aledsage: The way I built it was to export TestEntityWithEffectors*.java as a jar from my IDE, and add it to ./lib/patch of a pre-installed Brooklyn. That didn't require adding the pom.

Which pom did you add that to? Maybe we should add something more general like "You'll have to ensure this is on the classpath of your brooklyn server, before testing". I'll update my PR accordingly.

bostko:
@aledsage thank you for the changes!

aledsage:
Unfortunately jenkins is not running PR builds. I ran `mvn clean install` manually, and it worked. So merging.

PR #196, opened 2016-06-13
Close streams after usage

neykov:
Fixes https://builds.apache.org/job/brooklyn-master-windows/151/console

### DIFF
--- a/camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/AbstractYamlTest.java
+++ b/camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/AbstractYamlTest.java
@@ -19,7 +19,6 @@
 package org.apache.brooklyn.camp.brooklyn;
 
 import java.io.Reader;
-import java.io.StringReader;
 import java.util.Map;
 import java.util.Set;
 
@@ -112,20 +111,21 @@ public abstract class AbstractYamlTest {
         }
     }
 
-    protected Reader loadYaml(String yamlFileName, String ...extraLines) throws Exception {
+    protected String loadYaml(String yamlFileName, String ...extraLines) throws Exception {
         String input = new ResourceUtils(this).getResourceAsString(yamlFileName).trim();
         StringBuilder builder = new StringBuilder(input);
         for (String l: extraLines)
             builder.append("\n").append(l);
-        return new StringReader(builder.toString());
+        return builder.toString();
     }
     
     protected Entity createAndStartApplication(String... multiLineYaml) throws Exception {
         return createAndStartApplication(joinLines(multiLineYaml));
     }
     
+    /** @deprecated since 0.10.0, use {@link #createAndStartApplication(String)} instead */
     protected Entity createAndStartApplication(Reader input) throws Exception {
-        return createAndStartApplication(Streams.readFullyAndClose(input));
+        return createAndStartApplication(Streams.readFully(input));
     }
 
     protected Entity createAndStartApplication(String input) throws Exception {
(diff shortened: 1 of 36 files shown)


camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/AbstractYamlTest.java
```diff
@@ -125,7 +125,7 @@ protected Entity createAndStartApplication(String... multiLineYaml) throws Excep
     }
     
     protected Entity createAndStartApplication(Reader input) throws Exception {
-        return createAndStartApplication(Streams.readFully(input));
+        return createAndStartApplication(Streams.readFullyAndClose(input));
```
aledsage: Overall, I'm fine with this because it's just in tests. But it feels like the responsibility of the caller to close the Reader that they created, rather than this method.

core/src/main/java/org/apache/brooklyn/core/catalog/internal/CatalogBomScanner.java
```diff
@@ -268,7 +268,7 @@ private void addLibraryDetails(Bundle bundle, Map<String, Object> catalog) {
 
         private String readBom(URL bom) {
             try (final InputStream ins = bom.openStream()) {
-                return Streams.readFullyString(ins);
+                return Streams.readFullyStringAndClose(ins);
```
aledsage: The stream is created in a try-with-resources [1], so no need to close. Is your thinking that if we call `readFullyStringAndClose()` everywhere, then it will be easier to make sure we're closing things everywhere? (e.g. we can search for uses of readFullyString, and not have many usages to review).

[1] https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html

core/src/main/java/org/apache/brooklyn/util/core/crypto/SecureKeys.java
```diff
@@ -78,11 +78,11 @@ public static X509Principal getX509PrincipalWithCommonName(String commonName) {
         return new X509Principal("" + "C=None," + "L=None," + "O=None," + "OU=None," + "CN=" + commonName);
     }
 
-    /** reads RSA or DSA / pem style private key files (viz {@link #toPem(KeyPair)}), extracting also the public key if possible
+    /** reads RSA or DSA / pem style private key files (viz {@link #toPem(KeyPair)}), extracting also the public key if possible. Closes the stream.
      * @throws IllegalStateException on errors, in particular {@link PassphraseProblem} if that is the problem */
     public static KeyPair readPem(InputStream input, final String passphrase) {
         // TODO cache is only for fallback "reader" strategy (2015-01); delete when Parser confirmed working
-        byte[] cache = Streams.readFully(input);
+        byte[] cache = Streams.readFullyAndClose(input);
```
aledsage: Feels like the responsibility of the caller to close the stream, rather than this method.

Looking at things like Guava's `ByteStreams`, they've deprecated all the methods that closed the stream that was passed in (and switched to a completely different pattern that uses `ByteSource`).

That makes me think it is the wrong thing to close the stream we are given.

Are there any good examples of good modern APIs that close the stream that is passed in?

aledsage:
Unconvinced of the change in `SecureKeys`. The rest looks really good.

core/src/main/java/org/apache/brooklyn/util/core/crypto/SecureKeys.java
neykov: Agree in general. It's the way we use readPem that prompted me to do the change - we are not closing the streams in any of the callers. Will have another look on how it can be improved.

core/src/main/java/org/apache/brooklyn/core/catalog/internal/CatalogBomScanner.java
neykov: Haven't noticed that, will revert.
Yes, my thinking was that it doesn't really make sense to keep the stream open after reading it fully. I've updated the javadoc to encourage the usage of xxxAndClose alternatives. Only very specific cases will need to keep it unclosed.

core/src/main/java/org/apache/brooklyn/util/core/crypto/SecureKeys.java
aledsage: @neykov How about overloading `readPem` to take a byte array? It looks like that is the way it is used anyway.

And for the tests that call it, I wonder if we should add `byte[] ResourceUtils.getResourceAsBytes(String url)`. Again it feels like many callers of the existing `getResourceFromUrl` either want it as a string or as bytes.

neykov:
Thanks for the great review @aledsage. Addressed comments.

aledsage:
LGTM; once jenkins has confirmed the PR builds then good to merge (please do feel free to do that yourself @neykov if you get to it first).

PR #197, opened 2016-06-13
Fix JcloudsSshMachineLocation.getSubnetIp

aledsage:
Previously would throw NPE if privateAddresses was empty.


neykov:
LGTM, merging.

PR #198, opened 2016-06-14
Change org.apache.cxf log level to ERROR

aledsage:
The previous change to “SEVERE” doesn’t work, because that is not a
logbook logging level.

 In the stdout of Brooklyn, we were still getting:

    10:55:48,833 |-INFO in ch.qos.logback.classic.joran.action.LoggerAction - Setting level of logger [org.apache.cxf] to DEBUG

With this change, we get:

    10:58:19,047 |-INFO in ch.qos.logback.classic.joran.action.LoggerAction - Setting level of logger [org.apache.cxf] to ERROR


neykov:
Looks good, merging.

PR #199, opened 2016-06-14
BROOKLYN-299: fix LocationUsage mutex

aledsage:
Previously, we sometimes tried to ssh to a VM when it was destroyed to get its metadata. Avoid that!

Also changes the `LocalUsageManager`, to call toMetadataRecord() outside of the mutex.


neykov:
I like the fix. Changes look good. Merging.

PR #203, opened 2016-06-16
Fix CAS operations on AtomicBoolean

grkvlt:
Change to use correct idiom when using `AtomicBoolean` to guard against multiple invocations.

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/internal/BrooklynInitialization.java
+++ b/core/src/main/java/org/apache/brooklyn/core/internal/BrooklynInitialization.java
@@ -61,14 +61,14 @@ public class BrooklynInitialization {
      * 
      */
     
-    public synchronized static void initAll() {
-        if (done.get()) return;
-        initTypeCoercionStandardAdapters();
-        initSecureKeysBouncyCastleProvider();
-        initNetworking();
-        initPortRanges();
-        initLegacyLanguageExtensions();
-        done.set(true);
+    public static void initAll() {
+        if (done.compareAndSet(false, true)) {
+            initTypeCoercionStandardAdapters();
+            initSecureKeysBouncyCastleProvider();
+            initNetworking();
+            initPortRanges();
+            initLegacyLanguageExtensions();
+        }
     }
 
     @SuppressWarnings("deprecation")
--- a/core/src/main/java/org/apache/brooklyn/core/mgmt/persist/FileBasedObjectStore.java
+++ b/core/src/main/java/org/apache/brooklyn/core/mgmt/persist/FileBasedObjectStore.java
@@ -33,18 +33,9 @@ import java.util.Date;
 import java.util.List;
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicBoolean;
 
 import javax.annotation.Nullable;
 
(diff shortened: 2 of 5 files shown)


core/src/main/java/org/apache/brooklyn/core/internal/BrooklynInitialization.java
```diff
@@ -61,21 +61,22 @@ public static void initLegacyLanguageExtensions() {
...
+            initLegacyLanguageExtensions();
+        }
     }
 
     @SuppressWarnings("deprecation")
     public synchronized static void reinitAll() {
-        done.set(false);
-        org.apache.brooklyn.util.core.BrooklynLanguageExtensions.reinit();
-        initAll();
+        if (done.compareAndSet(true, false)) {
```
neykov: This changes the behaviour - won't run if `initAll` is not called before that. Better stick to the original, wdyt?

neykov:
Nice cleanup.

core/src/main/java/org/apache/brooklyn/core/internal/BrooklynInitialization.java
aledsage: +1

core/src/main/java/org/apache/brooklyn/core/mgmt/persist/FileBasedObjectStore.java
```diff
@@ -94,25 +94,19 @@ public File getBaseDir() {
...
-        if (doneFirstContentiousWrite.get())
-            return;
-        synchronized (this) {
-            if (doneFirstContentiousWrite.get())
-                return;
-            try {
-                if (deferredBackupNeeded) {
-                    // defer backup and path creation until first write
-                    // this way if node is standby or auto, the backup is not created superfluously
+        if (doneFirstContentiousWrite.getAndSet(true)) return;
```
aledsage: This also changes the semantics - I haven't dug deep enough to confirm that it's right.

I'm assuming that nothing else is synchronizing on this object (from some other class), but it's very hard to tell - one of the main reasons I hate `synchronized (this)`.

Previously, it only set `doneFirstContentiousWrite` if the call to `backupDirByCopying` was successful. Otherwise, it would try again in the next thread that called `prepareForMasterUse()`.

I think that your behaviour is better @grkvlt .

Thoughts?

@ahgittin - anything to add (I think you wrote this original code)?

core/src/main/java/org/apache/brooklyn/core/internal/BrooklynInitialization.java
grkvlt: Oops, will fix that!

core/src/main/java/org/apache/brooklyn/core/mgmt/persist/FileBasedObjectStore.java
grkvlt: @aledsage you're right that the semantics changed here. in fact, I think all we need is a normal boolean and a `synchronized` critical section, the atomic CAS aspect of the `AtomicBoolean` aren't being used here, since we can be sure only one thread at a time is checking the value inside the critical section. I've refactored to reflect that.

grkvlt:
See also 4388cbe580a57d092b5e99ac3209c2fa9076757f and f554e61fe622dcb3b05197246c11393891b5f67c which were needed because the merge somehow missed those changes.

PR #204, opened 2016-06-16
Move machine metrics to SoftwareProcess entity

grkvlt:
Moves the sensor feed for machine metrics to the SoftwareProcess, controlled by a configuration key with default set to _false_ normally and _true_ for `MachineEntity`. Also contributes static helper methods for feed `PollConfig` creation.

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/entity/AbstractEntity.java
+++ b/core/src/main/java/org/apache/brooklyn/core/entity/AbstractEntity.java
@@ -27,6 +27,22 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
+import org.apache.commons.lang3.builder.EqualsBuilder;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.common.annotations.Beta;
+import com.google.common.base.Function;
+import com.google.common.base.Objects;
+import com.google.common.base.Objects.ToStringHelper;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Iterables;
+import com.google.common.collect.Lists;
+import com.google.common.collect.Maps;
+import com.google.common.collect.Sets;
+
 import org.apache.brooklyn.api.effector.Effector;
 import org.apache.brooklyn.api.entity.Application;
 import org.apache.brooklyn.api.entity.Entity;
@@ -100,21 +116,6 @@ import org.apache.brooklyn.util.core.task.DeferredSupplier;
 import org.apache.brooklyn.util.guava.Maybe;
 import org.apache.brooklyn.util.javalang.Equals;
 import org.apache.brooklyn.util.text.Strings;
-import org.apache.commons.lang3.builder.EqualsBuilder;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
-import com.google.common.annotations.Beta;
-import com.google.common.base.Function;
-import com.google.common.base.Objects;
-import com.google.common.base.Objects.ToStringHelper;
-import com.google.common.collect.ImmutableList;
-import com.google.common.collect.ImmutableMap;
-import com.google.common.collect.ImmutableSet;
-import com.google.common.collect.Iterables;
-import com.google.common.collect.Lists;
-import com.google.common.collect.Maps;
-import com.google.common.collect.Sets;
 
 /**
  * Default {@link Entity} implementation, which should be extended whenever implementing an entity.
@@ -144,7 +145,7 @@ import com.google.common.collect.Sets;
  * The legacy (pre 0.5) mechanism for creating entities is for others to call the constructor directly.
  * This is now deprecated.
  */
-public abstract class AbstractEntity extends AbstractBrooklynObject implements EntityLocal, EntityInternal {
+public abstract class AbstractEntity extends AbstractBrooklynObject implements EntityInternal {
     
     private static final Logger LOG = LoggerFactory.getLogger(AbstractEntity.class);
     
--- a/core/src/main/java/org/apache/brooklyn/core/entity/AbstractEntity.java
+++ b/core/src/main/java/org/apache/brooklyn/core/entity/AbstractEntity.java
@@ -221,7 +221,7 @@ public abstract class AbstractEntity extends AbstractBrooklynObject implements E
 
     private Collection<AbstractPolicy> policiesInternal = Lists.newCopyOnWriteArrayList();
     private Collection<AbstractEnricher> enrichersInternal = Lists.newCopyOnWriteArrayList();
-    Collection<Feed> feeds = Lists.newCopyOnWriteArrayList();
+    private Collection<Feed> feedsInternal = Lists.newCopyOnWriteArrayList();
 
     // FIXME we do not currently support changing parents, but to implement a cluster that can shrink we need to support at least
     // orphaning (i.e. removing ownership). This flag notes if the entity has previously had a parent, and if an attempt is made to
@@ -248,6 +248,8 @@ public abstract class AbstractEntity extends AbstractBrooklynObject implements E
     private final BasicEnricherSupport enrichers = new BasicEnricherSupport();
 
     private final BasicGroupSupport groups = new BasicGroupSupport();
+
+    private final BasicFeedSupport feeds = new BasicFeedSupport();
 
     /**
      * The config values of this entity. Updating this map should be done
@@ -1664,9 +1666,8 @@ public abstract class AbstractEntity extends AbstractBrooklynObject implements E
      * {@link BasicPolicySupport}.
      */
     @Beta
-    // TODO revert to private when config() is reverted to return SensorSupportInternal
     public class BasicPolicySupport implements PolicySupportInternal {
-        
+
         @Override
         public Iterator<Policy> iterator() {
             return asList().iterator();
@@ -1692,11 +1693,11 @@ public abstract class AbstractEntity extends AbstractBrooklynObject implements E
                 LOG.debug("Removing "+old+" when adding "+policy+" to "+AbstractEntity.this);
                 remove(old);
             }
-            
+
             CatalogUtils.setCatalogItemIdOnAddition(AbstractEntity.this, policy);
             policiesInternal.add((AbstractPolicy)policy);
             ((AbstractPolicy)policy).setEntity(AbstractEntity.this);
-            
+
             getManagementSupport().getEntityChangeListener().onPolicyAdded(policy);
             sensors().emit(AbstractEntity.POLICY_ADDED, new PolicyDescriptor(policy));
         }
@@ -1707,19 +1708,19 @@ public abstract class AbstractEntity extends AbstractBrooklynObject implements E
             add(policy);
             return policy;
         }
-        
+
         @Override
         public boolean remove(Policy policy) {
             ((AbstractPolicy)policy).destroy();
             boolean changed = policiesInternal.remove(policy);
-            
+
             if (changed) {
                 getManagementSupport().getEntityChangeListener().onPolicyRemoved(policy);
                 sensors().emit(AbstractEntity.POLICY_REMOVED, new PolicyDescriptor(policy));
             }
             return changed;
         }
-        
+
         @Override
         public boolean removeAllPolicies() {
             boolean changed = false;
@@ -1737,7 +1738,6 @@ public abstract class AbstractEntity extends AbstractBrooklynObject implements E
      * {@link BasicEnricherSupport}.
      */
     @Beta
-    // TODO revert to private when config() is reverted to return SensorSupportInternal
     public class BasicEnricherSupport implements EnricherSupportInternal {
         @Override
         public Iterator<Enricher> iterator() {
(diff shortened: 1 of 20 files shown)


software/base/src/main/java/org/apache/brooklyn/entity/software/base/SoftwareProcessImpl.java
```diff
@@ -85,15 +89,23 @@
  * It exposes sensors for service state (Lifecycle) and status (String), and for host info, log file location.
  */
 public abstract class SoftwareProcessImpl extends AbstractEntity implements SoftwareProcess, DriverDependentEntity {
-    private static final Logger log = LoggerFactory.getLogger(SoftwareProcessImpl.class);
-    
+
+    private static final Logger LOG = LoggerFactory.getLogger(SoftwareProcess.class);
```
neykov: Did you remove the `Impl` suffix on purpose? I think we should keep it - it points to the file where the logging occurs.

software/base/src/main/java/org/apache/brooklyn/entity/software/base/SoftwareProcessImpl.java
```diff
@@ -373,6 +468,8 @@ protected void postRestart() {
      */
     protected void disconnectSensors() {
         connectedSensors = false;
+        boolean retrieveMachineMetrics = config().get(RETRIEVE_MACHINE_METRICS);
+        if (retrieveMachineMetrics) disconnectMachineSensors();
```
neykov: (minor) I'd always call that and depend on the null check inside.

neykov:
LGTM. Will hold off merging so people get the chance to read and respond the dev@ mail.

software/base/src/main/java/org/apache/brooklyn/entity/software/base/SoftwareProcessImpl.java
```diff
@@ -287,6 +299,89 @@ protected void postDriverStart() {
...
+                            .onFailureOrException(Functions.constant(-1d))
+                            .onSuccess(new Function<SshPollValue, Double>() {
+                                @Override
+                                public Double apply(SshPollValue input) {
+                                    String loadAverage = Strings.getFirstWordAfter(input.getStdout(), "load average:").replace(",", "");
+                                    return Double.valueOf(loadAverage);
+                                }
+                            }))
+                    .poll(SshPollConfig.forSensor(CPU_USAGE)
+                            .command("cat /proc/stat")
```
ceeaspb: A problem with the cpu stats is that they are avarage since boot time, not since the last feed poll [ .period(Duration.THIRTY_SECONDS) ]
likely a separate issue but can we change this so it only accounts for usage since the last poll. It would require maintaining the counters as at the last poll time to calculate the difference.

software/base/src/main/java/org/apache/brooklyn/entity/software/base/SoftwareProcessImpl.java
```diff
@@ -287,6 +299,89 @@ protected void postDriverStart() {
...
+                            .onFailureOrException(Functions.constant(-1d))
+                            .onSuccess(new Function<SshPollValue, Double>() {
+                                @Override
+                                public Double apply(SshPollValue input) {
+                                    List<String> cpuData = Splitter.on(" ").omitEmptyStrings().splitToList(Strings.getFirstLine(input.getStdout()));
+                                    Integer system = Integer.parseInt(cpuData.get(1));
+                                    Integer user = Integer.parseInt(cpuData.get(3));
+                                    Integer idle = Integer.parseInt(cpuData.get(4));
+                                    return (double) (system + user) / (double) (system + user + idle);
```
ceeaspb: the divisor total time is missing some fields new in 2.6.3* and later kernels, eg. see 
http://procps.cvs.sourceforge.net/viewvc/procps/procps/vmstat.c
eg.
Div= duse+dsys+didl+diow+dstl;

software/base/src/main/java/org/apache/brooklyn/entity/software/base/SoftwareProcessImpl.java
grkvlt: @ceeaspb I changed this to a command that sums the CPU field of the `ps` output, which should work on almost all Linux machines

grkvlt:
Configure an entity to use the machine metrics as follows:

```YAML
services:
    brooklyn.initializers:
      - type: org.apache.brooklyn.entity.machine.AddMachineMetrics
```

software/base/src/main/java/org/apache/brooklyn/entity/software/base/SoftwareProcessImpl.java
```diff
@@ -85,15 +81,20 @@
  * It exposes sensors for service state (Lifecycle) and status (String), and for host info, log file location.
  */
 public abstract class SoftwareProcessImpl extends AbstractEntity implements SoftwareProcess, DriverDependentEntity {
-    private static final Logger log = LoggerFactory.getLogger(SoftwareProcessImpl.class);
-    
+
+    private static final Logger LOG = LoggerFactory.getLogger(SoftwareProcessImpl.class);
+
+    static {
+        MachineAttributes.init();
```
Graeme-Miller: do we need to call init, if we are already calling it in AddMachineMetrics?

neykov:
Couldn't figure out how feeds get stopped with the enricher? What about starting when rebind is involved? `connectSensors` is called on rebind so that took care of it before.

neykov:
Really happy with how the PR turned out btw, great changes!

software/base/src/main/java/org/apache/brooklyn/entity/software/base/SoftwareProcessImpl.java
ceeaspb: @grkvlt ok thanks
on review, ps averages the cpu since process start time.
It also caps the utilisation at 99% I think even if the %age is per cpu core, I'd have to check that though.

http://procps.cvs.sourceforge.net/viewvc/procps/procps/ps/display.c?revision=1.32&content-type=text%2Fplain
  used_jiffies = buf->utime + buf->stime;
  if(include_dead_children) used_jiffies += (buf->cutime + buf->cstime);
  seconds = seconds_since_boot - buf->start_time / Hertz;
  if(seconds) pcpu = (used_jiffies * 1000ULL / Hertz) / seconds;

grkvlt:
@neykov the feeds are managed by `FeedSupport` in the entity - I think this should stop them, and also handle rebind

grkvlt:
Still needs a bit of testing done with deployed clusters

neykov:
Looks like Brooklyn will handle starting them (on [creation](https://github.com/apache/brooklyn-server/blob/master/core/src/main/java/org/apache/brooklyn/feed/ssh/SshFeed.java#L168) and on [rebind](https://github.com/apache/brooklyn-server/blob/master/core/src/main/java/org/apache/brooklyn/core/mgmt/rebind/BasicEntityRebindSupport.java#L169)). But looks like they are not stopped as part of the entity lifecycle - couldn't find anything in the sources.

grkvlt:
@neykov the feed stopping should be addressed in another PR I think?

neykov:
Not convinced. What happens when SshFeed can't find the machine on a stopped entity? Does it start logging errors or is that silently discarded?
There's a similar problem on creation - `apply` is called just after `init`, so there's no machine available yet.

grkvlt:
@neykov it just doesn't do anything (well, throws an exception and returns `null`) until it can actually SSH

grkvlt:
Tested on various clouds and works well as input to an `AutoScalerPolicy`

grkvlt:
@neykov see last commit, feeds are now stopped when software process stops, using a new task.

core/src/main/java/org/apache/brooklyn/core/entity/AbstractEntity.java
```diff
@@ -1737,7 +1738,6 @@ public boolean removeAllPolicies() {
      * {@link BasicEnricherSupport}.
      */
     @Beta
-    // TODO revert to private when config() is reverted to return SensorSupportInternal
```
neykov: I believe this still needs to be addressed, though the comment has copy&paste error referring to sensors.

software/base/src/main/java/org/apache/brooklyn/entity/machine/MachineEntityImpl.java
```diff
@@ -61,82 +51,17 @@ public void init() {
...
+            AddMachineMetrics.addMachineMetricsEnrichers(this);
+        } else {
+            LOG.warn("Not adding machine metrics feed as no suitable location available on entity");
+        }
     }
 
-    @Override
     public void disconnectSensors() {
-        if (sensorFeed != null) sensorFeed.stop();
+        if (machineMetricsFeed != null) machineMetricsFeed.stop();
```
neykov: Could be removed now that feeds are stopped out of the box.

software/base/src/main/java/org/apache/brooklyn/entity/machine/MachineEntityImpl.java
```diff
@@ -61,82 +51,17 @@ public void init() {
...
-                            public Long apply(SshPollValue input) {
-                                List<String> memoryData = Splitter.on(" ").omitEmptyStrings().splitToList(Strings.getFirstLine(input.getStdout()));
-                                return Long.parseLong(memoryData.get(1));
-                            }
-                        }))
-                .build();
-
+        Maybe<SshMachineLocation> location = Machines.findUniqueMachineLocation(getLocations(), SshMachineLocation.class);
+        if (location.isPresent() && location.get().getOsDetails().isLinux()) {
+            machineMetricsFeed = AddMachineMetrics.createMachineMetricsFeed(this);
```
neykov: This will add the feeds multiple times, on each start. It's a problem with the original code as well. We have checks in place to prevent actually having multiple instances of the same feed, so it's not a problem in practice. Feels like feeds API needs a cleanup.

software/base/src/main/java/org/apache/brooklyn/entity/machine/MachineEntityImpl.java
```diff
@@ -61,82 +51,17 @@ public void init() {
...
+            AddMachineMetrics.addMachineMetricsEnrichers(this);
```
neykov: Move this to the `init` method - don't want enrichers added on each startup.

neykov:
Will merge after moving enricher creation to `init` method.

Stopping feeds in the stop effector feels like the right thing to do. This is the practice we follow in connect/disconnectSensors and having this done out of the box will simplify entity implementations. Still think feeds need some more attention in future PRs, especially around creation. Currently we create the feeds even if it's already added to the entity (and try to handle duplicates gracefully). Rebind will start the feeds without taking into consideration the entity state. Overall feels like feed workflow needs cleanup.

grkvlt:
@neykov good feedback; will address your comments

software/base/src/main/java/org/apache/brooklyn/entity/machine/MachineEntityImpl.java
```diff
@@ -18,126 +18,41 @@
...
-                                return Long.parseLong(memoryData.get(1));
-                            }
-                        }))
-                .build();
-
-    }
+        Maybe<SshMachineLocation> location = Machines.findUniqueMachineLocation(getLocations(), SshMachineLocation.class);
+        if (location.isPresent() && location.get().getOsDetails().isLinux()) {
+            LOG.info("Adding machine metrics feed and enrichers.");
+            AddMachineMetrics.createMachineMetricsFeed(this);
```
neykov: Having the feeds initialized here means that stopping and starting the entity won't start the feeds again. Better move them back to `connectSensors` while keeping `addMachineMetricsEnrichers` here.

software/base/src/main/java/org/apache/brooklyn/entity/software/base/lifecycle/MachineLifecycleEffectorTasks.java
```diff
@@ -709,9 +710,11 @@ protected void doStop(ConfigBag parameters, Callable<StopMachineDetails<Integer>
         DynamicTasks.queue("pre-stop", new PreStopCustomTask());
 
         Maybe<MachineLocation> machine = Machines.findUniqueMachineLocation(entity().getLocations());
-        Task<String> stoppingProcess = null;
+        Task<List<?>> stoppingProcess = null;
         if (canStop(stopProcessMode, entity())) {
-            stoppingProcess = DynamicTasks.queue("stopping (process)", new StopProcessesAtMachineTask());
+            stoppingProcess = Tasks.parallel(
```
Graeme-Miller: Aled and I debugged the hanging test today and narrowed it down to this line. The stoppingProcess task is created, but never started.

The following code can be used instead:
`
stoppingProcess = Tasks.parallel(
                    Tasks.<String>builder().displayName("stopping (process)").body(new StopProcessesAtMachineTask()).build(),
                    Tasks.<String>builder().displayName("stopping (feeds)").body(new StopFeedsAtMachineTask()).build());
            DynamicTasks.queue(stoppingProcess);
`

software/base/src/main/java/org/apache/brooklyn/entity/machine/MachineEntityImpl.java
grkvlt: OK, good point. I didn't want to duplicate the `SshMachineLocation` check and was in a hurry ;(

software/base/src/main/java/org/apache/brooklyn/entity/software/base/lifecycle/MachineLifecycleEffectorTasks.java
grkvlt: Of course! That's easily fixed - will do that.

grkvlt:
@Graeme-Miller I have updated, rebased, and fixed tests; will merge with master now.

PR #206, opened 2016-06-16
Add debug log for onEffectorStarting method

bostko:
- the method is used by proxy calls and exceptions in here could be stripped out


aledsage:
I think there is a more general fix we need, for us not logging the cause of exceptions. I'm going to dig into that a little more. Hopefully we can do something at a higher level that means we don't need this log statement. We also want to avoid logging it multiple times.

aledsage:
@bostko I've opened https://issues.apache.org/jira/browse/BROOKLYN-313 to give lots of detail, and created https://github.com/apache/brooklyn-server/pull/245 to fix it. Can you review that please? If you're happy with the fix there, then can you close this PR?

PR #207, opened 2016-06-16
Ensure detection of failed pre.install.command and related commands.

geomacy:
Currently 'pre.install.command' and related steps do not fail if the command(s) returns a non-zero exit.
This can mean your install will fail but Brooklyn won't detect it (perhaps until some subsequent stage, or not at all).

This change adds detection of the return code and failure if it is non-zero. The 'contract' for the methods is added as a comment.  The AbstractSoftwareProcessSshDriver is changed to detect the failure and throw an exception. The AbstractSoftwareProcessWinRmDriver already does the right thing and doesn't need changed.

While this could break some existing blueprints, I think in such cases it's more likely that it is highlighting a problem that has been missed, rather than causing a problem because someone is explicitly relying on that behaviour.

### DIFF
--- a/software/base/src/main/java/org/apache/brooklyn/entity/software/base/AbstractSoftwareProcessSshDriver.java
+++ b/software/base/src/main/java/org/apache/brooklyn/entity/software/base/AbstractSoftwareProcessSshDriver.java
@@ -27,6 +27,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
+import org.apache.brooklyn.config.ConfigKey;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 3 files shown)


software/base/src/main/java/org/apache/brooklyn/entity/software/base/AbstractSoftwareProcessSshDriver.java
```diff
@@ -285,45 +296,54 @@ public void copyInstallResources() {
...
+        int result = execute(ImmutableList.of(getEntity().getConfig(configKey)), label);
+        if (0 != result) {
+            log.debug("Executing {} failed with return code {}", label, result);
+            throw new IllegalStateException("commands for " + configKey.getName() + " failed with return code " + result);
+        }
+    }
+
     @Override
     public void runPreInstallCommand() {
         if(Strings.isNonBlank(getEntity().getConfig(BrooklynConfigKeys.PRE_INSTALL_COMMAND))) {
```
neykov: Could move all of these checks to the `executeSuccessfully` method, but happy to merge as is.

software/base/src/main/java/org/apache/brooklyn/entity/software/base/AbstractSoftwareProcessSshDriver.java
```diff
@@ -285,45 +296,54 @@ public void copyInstallResources() {
         }
     }
 
+    private void executeSuccessfully(ConfigKey<String> configKey, String label) {
+        log.debug("Executing {} on entity {}", label, entity.getDisplayName());
+        int result = execute(ImmutableList.of(getEntity().getConfig(configKey)), label);
+        if (0 != result) {
+            log.debug("Executing {} failed with return code {}", label, result);
+            throw new IllegalStateException("commands for " + configKey.getName() + " failed with return code " + result);
```
neykov: Don't think `execute` logs the stdin/stdout/stderr on error - could be useful. For the time being a TODO here will do.

neykov:
Looks great - will be very useful to have this merged. A very minor comment, happy to be merged as is.

software/base/src/main/java/org/apache/brooklyn/entity/software/base/AbstractSoftwareProcessSshDriver.java
geomacy: That would be cleaner, let me do that.

software/base/src/main/java/org/apache/brooklyn/entity/software/base/AbstractSoftwareProcessSshDriver.java
geomacy: Will add that.

geomacy:
Changes made, squashed and pushed.

PR #208, opened 2016-06-20
Serialize complex objects as JSON in shell.env.

neykov:
Serialize complex objects to JSON instead of calling toString() when building the shell environment variables for `SoftwareProcess`.

```
brooklyn.catalog:
  id: mapping-entity
  item:
    type: VanillaSoftwareProcess
    env:
      MAPPINGS: $brooklyn:config("mappings")

services:
- type:  mapping-entity
  brooklyn.config:
    mappings:
      key1: val1
      key2: val2
```

Will result in an `MAPPINGS={"key1":"val1","key2":"val2"} ` environment variable.

### DIFF
--- a/camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/EntitiesYamlTest.java
+++ b/camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/EntitiesYamlTest.java
@@ -33,6 +33,7 @@ import org.apache.brooklyn.api.entity.Application;
 import org.apache.brooklyn.api.entity.Entity;
 import org.apache.brooklyn.api.entity.EntitySpec;
 import org.apache.brooklyn.api.location.Location;
+import org.apache.brooklyn.api.location.LocationSpec;
 import org.apache.brooklyn.api.mgmt.Task;
 import org.apache.brooklyn.api.sensor.AttributeSensor;
 import org.apache.brooklyn.camp.brooklyn.spi.dsl.methods.BrooklynDslCommon;
@@ -54,6 +55,8 @@ import org.apache.brooklyn.core.test.entity.TestEntityImpl;
 import org.apache.brooklyn.entity.group.DynamicCluster;
 import org.apache.brooklyn.entity.group.DynamicFabric;
 import org.apache.brooklyn.entity.software.base.SameServerEntity;
+import org.apache.brooklyn.entity.software.base.VanillaSoftwareProcess;
+import org.apache.brooklyn.entity.software.base.SoftwareProcessShellEnvironmentTest.EnvRecordingLocation;
 import org.apache.brooklyn.entity.stock.BasicEntity;
 import org.apache.brooklyn.util.collections.MutableMap;
 import org.apache.brooklyn.util.core.task.Tasks;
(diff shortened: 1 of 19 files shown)


software/base/src/test/java/org/apache/brooklyn/entity/software/base/ShellEnvironmentSerializerTest.java
```diff
@@ -0,0 +1,74 @@
...
+        assertSerialize(3.14, "3.14");
+        assertSerialize(3.140, "3.14");
+        assertSerialize(.140, "0.14");
+        assertSerialize(0.140, "0.14");
+        assertSerialize(Boolean.TRUE, "true");
+        assertSerialize(Boolean.FALSE, "false");
+        assertSerialize(dt, Long.toString(dt.getTime()));
+        assertSerialize(null, null);
+        assertSerialize(ImmutableList.of(str, 3.14, 0.14));
+        assertSerialize(ImmutableMap.of("string", str, "num1", 3.14, "num2", 0.14));
```
grkvlt: Just to be sure, can we have a test with serializing a map of maps?

grkvlt:
@neykov great, this should be very useful

camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/EntitiesYamlTest.java
```diff
@@ -1011,6 +1014,43 @@ public void testEntityTypeAsImpl() throws Exception {
...
+                "  brooklyn.config:",
+                "    map-config:",
+                "      key1: val1",
+                "      key2: $brooklyn:entity(\"self\")",
+                "    list-config:",
+                "    - 4.12",
+                "    - true",
+                "    - $brooklyn:entity(\"self\")",
+                "    shell.env:",
+                "      mapRef: $brooklyn:config(\"map-config\")",
```
geomacy: maybe name this MAP_REF and similarly below, following convention of case for environment vars.

software/base/src/main/java/org/apache/brooklyn/entity/software/base/AbstractSoftwareProcessSshDriver.java
```diff
@@ -344,9 +346,23 @@ public void runPostLaunchCommand() {
...
     public Map<String, String> getShellEnvironment() {
-        return Strings.toStringMap(entity.getConfig(SoftwareProcess.SHELL_ENVIRONMENT), "");
+        Map<String, Object> env = entity.getConfig(SoftwareProcess.SHELL_ENVIRONMENT);
+        if (env == null) {
+            return null;
+        }
+        ShellEnvironmentSerializer envSerializer = new ShellEnvironmentSerializer(((EntityInternal)entity).getManagementContext());
+        Map<String, String> serializedEnv = Maps.newHashMap();
+        for (Entry<String, Object> entry : env.entrySet()) {
+            String key = serializeShellEnv(envSerializer, entry.getKey());
```
geomacy: Is it really necessary to serialise the key?  It might even be worth checking here precisely that the key matches the requirements for names of environment variables?

software/base/src/main/java/org/apache/brooklyn/entity/software/base/ShellEnvironmentSerializer.java
```diff
@@ -0,0 +1,51 @@
...
+    public ShellEnvironmentSerializer(ManagementContext mgmt) {
+        mapper = BrooklynObjectsJsonMapper.newMapper(mgmt);
+    }
+
+    public String serialize(Object value) {
+        if (value == null) return null;
+        if (value instanceof String) return (String)value;
+        try {
+            String str = mapper.writeValueAsString(value);
+            // Avoid dealing with unquoting and unescaping the serialized result is a string
```
geomacy: "as" a string?  Not 100% sure what you mean here.

software/base/src/main/java/org/apache/brooklyn/entity/software/base/ShellEnvironmentSerializer.java
```diff
@@ -0,0 +1,51 @@
...
+            if (isString(str)) {
```
geomacy: Maybe better to call this isQuotedString or some such  - isString(str) looks a bit tautological to me.

software/base/src/main/java/org/apache/brooklyn/entity/software/base/AbstractSoftwareProcessSshDriver.java
neykov: That's to keep backwards compatibility. `Strings.toStringMap` would call `toString` on the keys.
Using invalid keys will fail the task on the bash side. I'd like to avoid validating the key here as depending on where you look you'll find different definitions for a valid name.

software/base/src/main/java/org/apache/brooklyn/entity/software/base/ShellEnvironmentSerializer.java
neykov: If `value` is a string it will get serialized as quoted json string. We don't want that, instead we need the raw string. Instead of stripping the quotes and removing escapes within the string I just take the source `value` and use that.

software/base/src/main/java/org/apache/brooklyn/entity/software/base/ShellEnvironmentSerializer.java
neykov: It refers to the json value within the string. You are right, I think `isJsonString` best matches the intent.

utils/common/src/main/java/org/apache/brooklyn/util/javalang/Reflections.java
```diff
@@ -242,7 +242,13 @@ public Object loadInstance(String classname) throws ReflectionNotFoundException,
 
     /** Invokes a suitable constructor, supporting varargs and primitives */
     public static <T> Optional<T> invokeConstructorWithArgs(Class<T> clazz, Object[] argsArray, boolean setAccessible) {
-        Reflections reflections = new Reflections(clazz.getClassLoader());
+        ClassLoader cl = clazz.getClassLoader();
```
geomacy: Is it worth making this available as another public method in Reflections?

geomacy:
Looks good to me.

geomacy:
One more thought on this - it would be really good to get some additional text in the documentation describing these new capabilities.

aledsage:
Thanks all, merging now.

utils/common/src/main/java/org/apache/brooklyn/util/javalang/Reflections.java
neykov: It's very specific to this method call as we don't really care what the fallback classloader is.

PR #209, opened 2016-06-21
Test relative entities

sjcorbett:
Adds an entity for test cases that need to resolve their target from another entity. For example, to make assertions on the child of a group member.

### DIFF
--- a/camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/methods/DslComponent.java
+++ b/camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/methods/DslComponent.java
@@ -61,14 +61,24 @@ public class DslComponent extends BrooklynDslDeferredSupplier<Entity> {
     private final DslComponent scopeComponent;
     private final Scope scope;
 
+    /**
+     * Resolve componentId in the {@link Scope#GLOBAL} scope.
+     */
     public DslComponent(String componentId) {
         this(Scope.GLOBAL, componentId);
     }
-    
+
+    /**
+     * Resolve componentId in scope relative to the current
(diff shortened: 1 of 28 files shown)


neykov:
Great idea. Good to merge.
Some thoughts I had while reading the code:
* Do we really need to have a spec of the child test? Could set TARGET_ENTITY on self and let that propagate to children through the inheritance chain. This will allow nesting children using `brooklyn.children`.
* `target` and `targetId` are reserved in the framework for the actual target. Could rename the config in the test to `ROOT_ENTITY` or `ANCHOR_ENTITY`. This is kind of related to the above, allowing target inheritance.
* Could pull the functionality up in `TargetableTestComponentImpl`, letting any entity use the scoping.

---

Spending some more time on it I now see that it's all designed to work together with the looping test case, so the above points are mute.

---

One final thing. Instead of configuring `DslComponent.Scope` could directly pass `DslComponent` as the configuration and evaluate that against the target. This would allow for more complex queries like `$brooklyn:ancestor("anchor-id").descendant("scoped-id")`.

aledsage:
@sjcorbett LGTM. Can you take a look at @neykov 's comments and see what you think. Please feel free to merge when you're ready.

sjcorbett:
@neykov Thank you for your comments. I have removed the `testSpec` config key in favour of children, replaced the use of `Scope` with plain `DslComponents` and reworked the use of `target`. The entity now resolves its `anchor` by checking a new `anchor` config key, `target` or `targetId` (in order), then sets its `target` to the resolved entity, overwriting any previous value. The anchor is published as a sensor.

I will leave extending this to all the test entities for future work.

sjcorbett:
I'll merge this now.

neykov:
Thanks @sjcorbett, changes look great.

PR #211, opened 2016-06-21
BROOKLYN-264 Stop app while VM still being provisioned: vm is left

bostko:
- Initial Live Tests

### DIFF
--- a/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
+++ b/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
@@ -63,6 +63,7 @@ import org.apache.brooklyn.config.ConfigKey.HasConfigKey;
 import org.apache.brooklyn.core.BrooklynVersion;
 import org.apache.brooklyn.core.config.ConfigUtils;
 import org.apache.brooklyn.core.config.Sanitizer;
+import org.apache.brooklyn.core.entity.Attributes;
 import org.apache.brooklyn.core.location.AbstractLocation;
 import org.apache.brooklyn.core.location.BasicMachineMetadata;
 import org.apache.brooklyn.core.location.LocationConfigKeys;
@@ -732,7 +733,10 @@ public class JcloudsLocation extends AbstractCloudMachineProvisioningLocation im
                     if (LOG.isDebugEnabled())
                         LOG.debug("NOTE: unused flags passed to obtain VM in "+setup.getDescription()+": "
                                 + Sanitizer.sanitize(setup.getUnusedConfig()));
-                
+                Object callerContext = setup.get(JcloudsLocationConfig.CALLER_CONTEXT);
+                if (callerContext != null && callerContext instanceof Entity) {
+                    ((Entity)callerContext).sensors().set(Attributes.JCLOUDS_PROVISIONING_RUNNING, Attributes.JcloudsProvisioningState.STARTED);
+                }
                 nodes = computeService.createNodesInGroup(groupId, 1, template);
                 provisionTimestamp = Duration.of(provisioningStopwatch);
             } finally {
--- a/software/base/src/main/java/org/apache/brooklyn/entity/software/base/lifecycle/MachineLifecycleEffectorTasks.java
+++ b/software/base/src/main/java/org/apache/brooklyn/entity/software/base/lifecycle/MachineLifecycleEffectorTasks.java
@@ -28,6 +28,8 @@ import java.util.concurrent.Callable;
 
 import javax.annotation.Nullable;
 
+import org.apache.brooklyn.core.server.BrooklynServerConfig;
+import org.apache.brooklyn.util.repeat.Repeater;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -77,10 +79,11 @@ import org.apache.brooklyn.entity.software.base.SoftwareProcess.RestartSoftwareP
 import org.apache.brooklyn.entity.software.base.SoftwareProcess.StopSoftwareParameters;
 import org.apache.brooklyn.entity.software.base.SoftwareProcess.StopSoftwareParameters.StopMode;
 import org.apache.brooklyn.entity.stock.EffectorStartableImpl.StartParameters;
+import org.apache.brooklyn.util.collections.MutableSet;
+
 import org.apache.brooklyn.location.localhost.LocalhostMachineProvisioningLocation;
 import org.apache.brooklyn.location.ssh.SshMachineLocation;
 import org.apache.brooklyn.util.collections.MutableMap;
-import org.apache.brooklyn.util.collections.MutableSet;
 import org.apache.brooklyn.util.core.config.ConfigBag;
 import org.apache.brooklyn.util.core.task.DynamicTasks;
 import org.apache.brooklyn.util.core.task.Tasks;
@@ -441,6 +444,9 @@ public abstract class MachineLifecycleEffectorTasks {
                 }
             }
             entity().addLocations(ImmutableList.of((Location) machine));
+            if (entity().getAttribute(Attributes.JCLOUDS_PROVISIONING_RUNNING) != null) {
+                entity().sensors().remove(Attributes.JCLOUDS_PROVISIONING_RUNNING);
+            }
 
             // elsewhere we rely on (public) hostname being set _after_ subnet_hostname
             // (to prevent the tiny possibility of races resulting in hostname being returned
@@ -709,7 +715,26 @@ public abstract class MachineLifecycleEffectorTasks {
 
         DynamicTasks.queue("pre-stop", new PreStopCustomTask());
 
+        // BROOKLYN-263:
+        // With this change the stop effector will wait for Location to provision so it can stop it.
+        // It will not retry if non-jclouds location is used in the entity.
         Maybe<MachineLocation> machine = Machines.findUniqueMachineLocation(entity().getLocations());
+        if (Attributes.JcloudsProvisioningState.STARTED.equals(entity().sensors().get(Attributes.JCLOUDS_PROVISIONING_RUNNING))
+                && machine.isAbsent()) {
+            entity().sensors().set(Attributes.JCLOUDS_PROVISIONING_RUNNING, null);
+            Repeater.create("Wait for a machine to appear")
+                    .until(new Callable<Boolean>() {
+                        @Override
+                        public Boolean call() throws Exception {
+                            return Machines.findUniqueMachineLocation(entity().getLocations()).isPresent();
+                        }
+                    })
+                    .every(Duration.FIVE_SECONDS)
+                    .limitTimeTo(entity().getManagementContext().getConfig().getConfig(BrooklynServerConfig.ENTITIES_STOP_WAIT_PROVISIONING_TIMEOUT))
+                    .run();
+            machine = Machines.findUniqueMachineLocation(entity().getLocations());
+        }
+
         Task<List<?>> stoppingProcess = null;
         if (canStop(stopProcessMode, entity())) {
             stoppingProcess = Tasks.parallel("stopping",
--- a/software/base/src/test/java/org/apache/brooklyn/entity/ExpungingJcloudsLocationLiveTest.java
+++ b/software/base/src/test/java/org/apache/brooklyn/entity/ExpungingJcloudsLocationLiveTest.java
@@ -1,0 +1,186 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.brooklyn.entity;
+
+import com.google.common.base.Predicate;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Iterables;
+import org.apache.brooklyn.api.entity.EntitySpec;
+import org.apache.brooklyn.core.entity.Attributes;
+import org.apache.brooklyn.core.entity.Entities;
+import org.apache.brooklyn.core.entity.EntityAsserts;
+import org.apache.brooklyn.core.entity.lifecycle.Lifecycle;
+import org.apache.brooklyn.core.internal.BrooklynProperties;
+import org.apache.brooklyn.core.location.LocationConfigKeys;
+import org.apache.brooklyn.core.location.cloud.CloudLocationConfig;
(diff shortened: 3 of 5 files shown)


software/base/src/test/java/org/apache/brooklyn/entity/EC2ExpungeLiveTest.java
```diff
@@ -0,0 +1,128 @@
...
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            Exceptions.propagate(e);
+        }
+
+        long beginTime = System.currentTimeMillis();
+        Entities.destroyCatching(app);
+        LOG.info("Time for expunging: {}", System.currentTimeMillis() - beginTime);
+
+        NodeMetadata nodeMetadata = Iterables.getFirst(((AWSEC2ComputeService) jcloudsLocation.getComputeService()).listNodesDetailsMatching(new Predicate<ComputeMetadata>() {
```
andreaturli: I think you can simplify with something like
```
import static com.google.common.base.Predicates.*;
import static org.jclouds.compute.predicates.NodePredicates.*;
import static com.google.common.collect.Iterables.*;

...

Iterable<? extends NodeMetadata> billedNodes = filter(jcloudsLocation.getComputeService().listNodesDetailsMatching(all()), and(withGroup(group), not(TERMINATED))));
```

software/base/src/test/java/org/apache/brooklyn/entity/EC2ExpungeLiveTest.java
```diff
@@ -0,0 +1,128 @@
...
+     * Verifies the behavior described in
+     * <a href="https://issues.apache.org/jira/browse/BROOKLYN-264">BROOKLYN-264 Stop app while VM still being provisioned: vm is left running when app is expunged</a>
+     * <ul>
+     *     <li>ApplicationResource.launch</li>
+     *     <li>wait a few seconds and EntityResponse.expunge</li>
+     *     <li>assert the image is on the cloud</li>
+     * </ul>
+     */
+    @Test(groups = {"Live"})
+    public void testExpunge() {
```
bostko: BTW the whole test passes for 3:45 seconds.
If we exclude the `listNodes` call, users will wait approximately 3 and a half minutes for expunge if they immediately click deploy and expunge.

software/base/src/main/java/org/apache/brooklyn/entity/software/base/lifecycle/MachineLifecycleEffectorTasks.java
```diff
@@ -708,7 +710,21 @@ protected void doStop(ConfigBag parameters, Callable<StopMachineDetails<Integer>
 
         DynamicTasks.queue("pre-stop", new PreStopCustomTask());
 
+        // BROOKLYN-263:
+        // With this change the stop effector will always wait for machinelocation to provision, so it can stop it.
+        // However in some situations if the stop effector is called in less than 8 seconds after start was called,
+        // it may be safe to skip this wait and proceed with stop.
+        // Hesitant about parameterising this.
         Maybe<MachineLocation> machine = Machines.findUniqueMachineLocation(entity().getLocations());
+        if (machine.isAbsent()) {
```
aledsage: This feels dangerous, for if start were to fail and not get a machine - we'd wait a long time, presumably.

software/base/src/test/java/org/apache/brooklyn/entity/EC2ExpungeLiveTest.java
```diff
@@ -0,0 +1,128 @@
...
+                .configure(EmptySoftwareProcess.PROVISIONING_PROPERTIES.subKey(CloudLocationConfig.INBOUND_PORTS.getName()), ImmutableList.of(22)));
+
+        app.addLocations(ImmutableList.of(jcloudsLocation));
+
+        EntityManagementUtils.start(app);
+
+        try {
+            // Number of seconds we wait for an AWS EC2 instance to be created.
+            // The test succeeds if stop effector is called 200 seconds or more later
+            Thread.sleep(16 * 1000);
```
aledsage: This kind of timing feels brittle to include in a test. But I guess we do need to wait some length of time, so it's not terminating it while it was still constructing the template (rather than in the middle of actually provisioning).

software/base/src/test/java/org/apache/brooklyn/entity/EC2ExpungeLiveTest.java
```diff
@@ -0,0 +1,128 @@
...
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
```
aledsage: Don't bother catching and rethrowing. Just declare all your test methods as `throws Exception {`.

software/base/src/test/java/org/apache/brooklyn/entity/EC2ExpungeLiveTest.java
```diff
@@ -0,0 +1,128 @@
...
+            Exceptions.propagate(e);
+        }
+
+        long beginTime = System.currentTimeMillis();
+        Entities.destroyCatching(app);
+        LOG.info("Time for expunging: {}", System.currentTimeMillis() - beginTime);
```
aledsage: Use guava's `Stopwatch`, rather than this. Can also use `Duration.of(...)` to get a much nicer string representation of the duration.

aledsage:
My comments so far are bike-shedding.

Looking at the big picture, I need to think more about:

* How should we trigger the delete, so that it safely handles success and failure of provisioning? Options include:
  * As you've done here, have `stop()` wait for the machine to be set; but abort if start() fails.
  * Mark the entity's internal state as "requiring machine-release", and have the start() code immediately release the machine. But we'd still need `stop()` to wait for that to be done, otherwise we might go on to unmanage the entity and thus cancel the tasks.

* For speed:
  * Can we use the jclouds event bus to be told of the VM's id long before jclouds has returned from the provisioning call? @andreaturli what do you think?
  * Can we tell the `JcloudsLocation` instance that the provisioning has been aborted (this would allow us to avoid waiting-for-sshable, etc). Not quite sure how best to do that.

We can ignore the speed issue for now, unless there is a quick win that greatly speeds things up. The main thing is to get it functionally correct, and with clean code.

bostko:
Unfortunately this change dramatically slow down unit tests :(. 
Stopping EmptySoftwareProcess is a common task for most of the tests. Adding a five minute timeout on stop makes `mvn clean install` run for hours.

bostko:
I updated the PR however I do not know why tests in `rest-resources` are failing.

software/base/src/main/java/org/apache/brooklyn/entity/software/base/lifecycle/MachineLifecycleEffectorTasks.java
```diff
@@ -709,7 +713,24 @@ protected void doStop(ConfigBag parameters, Callable<StopMachineDetails<Integer>
...
+        // BROOKLYN-263:
+        // With this change the stop effector will always wait for machinelocation to provision, so it can stop it.
+        // However in some situations if the stop effector is called in less than 8 seconds after start was called,
+        // it may be safe to skip this wait and proceed with stop.
+        // Hesitant about parameterising this.
         Maybe<MachineLocation> machine = Machines.findUniqueMachineLocation(entity().getLocations());
+        /*Do not retry if non-jclouds location has started.*/
+        if (Boolean.TRUE.equals(entity().sensors().get(Sensors.newBooleanSensor("jclouds.provisioningStarted")))
+                && machine.isAbsent()) {
+            org.jclouds.util.Predicates2.retry(new Predicate<MachineLifecycleEffectorTasks>() {
```
bostko: What do you think about using `Predicates2.retry` across Apache Brooklyn. Can you suggest a guava alternative or what do you think about adding such function in `brooklyn-core`?

aledsage: It feels like you should use our existing `org.apache.brooklyn.util.repeat.Repeater`. My personal opinion is that the builder pattern for the Repeater is easier to read, compared to the `Predicates2.retry` approach. The latter is probably only a good idea when you want to pass it to something that expects a `Predicate`.

software/base/src/test/java/org/apache/brooklyn/entity/EC2ExpungeLiveTest.java
```diff
@@ -0,0 +1,144 @@
...
+        mgmt = new LocalManagementContextForTests(brooklynProperties);
+        super.setUp();
+    }
+
+    /**
+     Verifies that when {@link org.apache.brooklyn.entity.software.base.lifecycle.MachineLifecycleEffectorTasks#doStop(ConfigBag, Callable<?>)} is called
+     it will not retry and wait for a MachineLocation to appear.
+     */
+    @Test
+    public void testExpungeNonJcloudsLocation() throws Exception {
```
bostko: This test is for simple entities used in unit tests (have to improve naming here).
I am trying to add a similar Integration test for a BYON location.

bostko:
I think I have much better assertions now.

software/base/src/test/java/org/apache/brooklyn/entity/ExpungingJcloudsLocationLiveTest.java
```diff
@@ -0,0 +1,187 @@
...
+     * Verifies the behavior described in
+     * <a href="https://issues.apache.org/jira/browse/BROOKLYN-264">BROOKLYN-264 Stop app while VM still being provisioned: vm is left running when app is expunged</a>
+     * <ul>
+     *     <li>ApplicationResource.launch</li>
+     *     <li>wait a few seconds and EntityResponse.expunge</li>
+     *     <li>assert the image is on the cloud</li>
+     * </ul>
+     */
+    @Test(groups = {"Live"})
+    public void verifyJclousMachineIsExpungedWhenStoppedImmediatelyAfterStart() {
```
bostko: I have to add a test for one more case. Will see how straight it is to handle it.
When an exception occurs during deployment, in other words the execution goes here https://github.com/apache/brooklyn-server/blob/master/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java#L1062-L1105
and `destroyOnFailure` is false then the machine will not be expunged.

bostko: Supporting this case will need won't be one line change.
What  do you think about leaving this for a later PR?

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -735,7 +737,11 @@ protected MachineLocation obtainOnce(ConfigBag setup) throws NoMachinesAvailable
                     if (LOG.isDebugEnabled())
                         LOG.debug("NOTE: unused flags passed to obtain VM in "+setup.getDescription()+": "
                                 + Sanitizer.sanitize(setup.getUnusedConfig()));
-                
+                Object callerContext = setup.get(JcloudsLocationConfig.CALLER_CONTEXT);
+                if (callerContext != null && callerContext instanceof Entity) {
+                    Entity callerContextEntity = (Entity)setup.get(JcloudsLocationConfig.CALLER_CONTEXT);
```
aledsage: Minor: don't call `setup.get()` again - just use your `callerContext` variable.

software/base/src/main/java/org/apache/brooklyn/entity/software/base/lifecycle/MachineLifecycleEffectorTasks.java
```diff
@@ -709,7 +712,26 @@ protected void doStop(ConfigBag parameters, Callable<StopMachineDetails<Integer>
...
+                && machine.isAbsent()) {
+            entity().sensors().set(Attributes.JCLOUDS_PROVISIONING_STARTED, null);
+            Repeater.create("Wait for a machine to appear")
+                    .until(new Callable<Boolean>() {
+                        @Override
+                        public Boolean call() throws Exception {
+                            return Machines.findUniqueMachineLocation(entity().getLocations()).isPresent();
+                        }
+                    })
+                    .every(Duration.THIRTY_SECONDS)
```
aledsage: Feels like strange times to use. Why only check every thirty seconds? That means we might take 29 seconds longer than we should have. The call to `getLocations()` should be cheap, so doing it every second seems fine.

aledsage: And to be able to test it fast (in unit tests, rather than live tests), we'd want to set this very low. But let's not worry about that quite yet.

software/base/src/main/java/org/apache/brooklyn/entity/software/base/lifecycle/MachineLifecycleEffectorTasks.java
```diff
@@ -709,7 +712,26 @@ protected void doStop(ConfigBag parameters, Callable<StopMachineDetails<Integer>
...
+                    .limitTimeTo(Duration.FIVE_MINUTES)
```
aledsage: Why wait for a hard-coded 5 minutes? Should be configurable. 5 minutes is not long enough on some clouds.

software/base/src/main/java/org/apache/brooklyn/entity/software/base/lifecycle/MachineLifecycleEffectorTasks.java
```diff
@@ -709,7 +712,26 @@ protected void doStop(ConfigBag parameters, Callable<StopMachineDetails<Integer>
 
         DynamicTasks.queue("pre-stop", new PreStopCustomTask());
 
+        // BROOKLYN-263:
+        // With this change the stop effector will wait for Location to provision so it can stop it.
+        // It will not retry if non-jclouds location is used in the entity.
         Maybe<MachineLocation> machine = Machines.findUniqueMachineLocation(entity().getLocations());
+        if (Boolean.TRUE.equals(entity().sensors().get(Attributes.JCLOUDS_PROVISIONING_STARTED))
+                && machine.isAbsent()) {
+            entity().sensors().set(Attributes.JCLOUDS_PROVISIONING_STARTED, null);
```
aledsage: This name feels misleading. It's sort-of more about whether the provisioning is running, rather than started. I'd be inclined to use an enum to indicate the exact state. The value could be one of PROVISIONING, ABORTING, or null.

See later comment (not added yet!) about telling the jclouds provisioning to not do anything else (e.g. currently it will wait for ssh'able etc, which we can abort earlier).

bostko: @aledsage do you have an idea how can I tell the stop task that the provisioning started other than the sensor flag. I feel like the code will need serious refactoring in order stop task to know earlier about the jclouds node.
The entity knows only about the machine location. It doesn't know about the LocationProvisioner.

bostko:
@aledsage do you have an idea how can I tell the stop task that the provisioning started other than the sensor flag. I feel like the code will need serious refactoring in order stop task to know earlier about the jclouds node.
The entity knows only about the machine location. It doesn't know about the LocationProvisioner.

Do you thing it is reasonable assigning the location provisioner to the sensor instead true or null?

bostko:
- I marked the sensor as "transient"
- I added a code which removes the sensor immidiately after a machine location is added
- I added an enum class for the `jclouds.provisioning.started` flag.<br>
  However I think with this code it could stay a boolean

aledsage:
I've been looking at how we can do this more generally (entirely in `MachineLifecycleEffectorTasks`, so it's not jclouds-specific). That makes the code easier to follow and easier to test. The former because the places we set the sensors is all in the same place; the latter because we can do it all with unit tests, rather than needing to provision VMs in live tests.

I have a commit that can go on top of this commit. I'll merge yours now, and then submit mine as a new PR.

PR #212, opened 2016-06-21
Repeat ssh tests

sjcorbett:
`TestSshCommand` test repeatedly runs its command and assertions until either all the assertions pass or the configured timeout expires. Previously the class ran the command once and then ran all the assertions repeatedly until either they passed or the timeout expired.


test-framework/src/test/java/org/apache/brooklyn/test/framework/TestSshCommandIntegrationTest.java
```diff
@@ -58,12 +62,39 @@
...
+        t.start();
+
+        // Times out in one second.
+        EntityAsserts.assertAttributeContinuallyNotEqualTo(testWithCmd, Attributes.SERVICE_STATE_ACTUAL, Lifecycle.RUNNING);
+
+        // Replace command with a valid one and the test should start to pass.
+        testWithCmd.config().set(COMMAND, "true");
+        EntityAsserts.assertAttributeEqualsEventually(testWithCmd, Attributes.SERVICE_STATE_ACTUAL, Lifecycle.RUNNING);
+
+        t.join();
```
aledsage: I'd prefer a timeout in the join - so if we introduce something that causes the thread to not return, then the test would fail rather than hanging. But no strong feelings (given we call app.start all over the place in the main thread for tests).

aledsage:
Yikes, well spotted! Merging now.

PR #213, opened 2016-06-22
TestSshCommand checks all assertions rather than throwing first

sjcorbett:
Fixes build by restoring behaviour changed in #212, which broke `TestSshCommandTest.shouldNotBeUpIfAssertionsFail` because only one failure was in the exception message.


sjcorbett:
Jenkins failure: `java.lang.OutOfMemoryError: unable to create new native thread`

sjcorbett:
Going to merge this since it fixes the build.

PR #214, opened 2016-06-22
BROOKLYN-305: handle invalid xml chars in attribute vals

aledsage:
I'm not convinced by this approach (calling a special method `XmlUtil.xpathHandlingIllegalChars()`, which retries the xpath having escaped the illegal characters). It works well enough for us because we are just making very basic use of xpath - it is primarily to just extact the entity id from the xml file, before we try to fully deserialize it.

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/mgmt/persist/BrooklynMementoPersisterToObjectStore.java
+++ b/core/src/main/java/org/apache/brooklyn/core/mgmt/persist/BrooklynMementoPersisterToObjectStore.java
@@ -294,7 +294,7 @@ public class BrooklynMementoPersisterToObjectStore implements BrooklynMementoPer
                     exceptionHandler.onLoadMementoFailed(type, "memento "+id+" read error", e);
                 }
                 
-                String xmlId = (String) XmlUtil.xpath(contents, "/"+type.toCamelCase()+"/id");
+                String xmlId = (String) XmlUtil.xpathHandlingIllegalChars(contents, "/"+type.toCamelCase()+"/id");
                 String safeXmlId = Strings.makeValidFilename(xmlId);
                 if (!Objects.equal(id, safeXmlId))
                     LOG.warn("ID mismatch on "+type.toCamelCase()+", "+id+" from path, "+safeXmlId+" from xml");
@@ -334,7 +334,7 @@ public class BrooklynMementoPersisterToObjectStore implements BrooklynMementoPer
 
                 class XPathHelper {
                     private String get(String innerPath) {
-                        return (String) XmlUtil.xpath(contents, prefix+innerPath);
+                        return (String) XmlUtil.xpathHandlingIllegalChars(contents, prefix+innerPath);
                     }
                 }
                 XPathHelper x = new XPathHelper();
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/apache/brooklyn/core/mgmt/persist/BrooklynMementoPersisterToObjectStore.java
```diff
@@ -294,7 +294,7 @@ public void visit(BrooklynObjectType type, String id, String contentsSubpath) th
                     exceptionHandler.onLoadMementoFailed(type, "memento "+id+" read error", e);
                 }
                 
-                String xmlId = (String) XmlUtil.xpath(contents, "/"+type.toCamelCase()+"/id");
+                String xmlId = (String) XmlUtil.xpathHandlingIllegalChars(contents, "/"+type.toCamelCase()+"/id");
```
bostko: Would `xpathHandlingIllegalChars` escape mechanism will be appropariate for filenames?
Ideally if a function like `File.escapeSpecialChars` exists then this should look like `File.escapeSpecialChars(XmlUtil.xpathHandlingIllegalChars(contents, "/"+type.toCamelCase()+"/id"))`

core/src/test/java/org/apache/brooklyn/util/core/xstream/XmlUtilTest.java
```diff
@@ -20,15 +20,65 @@
...
+import com.google.common.base.Optional;
+
 
 public class XmlUtilTest {
 
     @Test
     public void testXpath() throws Exception {
         String xml = "<a><b>myb</b></a>";
-        assertEquals(XmlUtil.xpath(xml, "/a/b[text()]"), "myb");
+        for (int i = 0; i < 2; i++) {
```
duncangrant: Why twice?

aledsage: One of the commits changes `XmlUtil` to use thread-local storage (rather than re-creating the `DocumentBuilder` on every invocation). By calling it twice, we confirm that thread-local storage isn't messing things up (e.g. if you removed the `documentBuilder.reset()` then it would probably break).

core/src/main/java/org/apache/brooklyn/core/mgmt/persist/BrooklynMementoPersisterToObjectStore.java
aledsage: We're making the assumption that the `id` field etc won't have crazy characters in it (because the Brooklyn framework generates those ids). The place we can get crazy characters is in attributes and config values that are set by the user.

duncangrant:
LGTM

aledsage:

We can later look at including in our persisted state:

    <?xml version=\"1.1\" encoding=\"UTF-8\"?>

But a very quick test gave the error below, which I still need to look into:

    com.thoughtworks.xstream.io.StreamException:  : only 1.0 is supported as <?xml version not '1.1' (position: START_DOCUMENT seen <?xml version="1.1"... @1:19)

PR #216, opened 2016-06-23
Serialize environment values - implement for all entities

neykov:
Builds on https://github.com/apache/brooklyn-server/pull/208


grkvlt:
@neykov this is useful, although I don't much like the fact we have to do it explicitly everywhere. can't see a way of centralising it, unless we implement some kind of post-processor with a `Function<T,String> toString` as an option for every `ConfigKey` and a sensible default?

grkvlt:
@neykov happy to merge as-is, anyway

neykov:
That's very specific to bash environment serialization so doing a `toString` for config keys in general would need information on how the value would be used. I think it becomes too complex for little gain. In future I agree we could add object specific serialization when need arises.

grkvlt:
@neykov OK, merging

PR #218, opened 2016-06-23
Yaml entity refs: support $brooklyn:self()

aledsage:
- Adds proper test coverage
- Fixes sibling/descendant/ancestor so will never return self

### DIFF
--- a/camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/methods/DslComponent.java
+++ b/camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/methods/DslComponent.java
@@ -120,12 +120,10 @@ public class DslComponent extends BrooklynDslDeferredSupplier<Entity> {
                 case SCOPE_ROOT:
                     return Entities.catalogItemScopeRoot(entity);
                 case DESCENDANT:
-                    entitiesToSearch = Entities.descendants(entity);
-                    entitiesToSearch = Iterables.filter(entitiesToSearch, notSelfPredicate);
+                    entitiesToSearch = Entities.descendantsWithoutSelf(entity);
                     break;
                 case ANCESTOR:
-                    entitiesToSearch = Entities.ancestors(entity);
-                    entitiesToSearch = Iterables.filter(entitiesToSearch, notSelfPredicate);
+                    entitiesToSearch = Entities.ancestorsWithoutSelf(entity);
                     break;
                 case SIBLING:
                     entitiesToSearch = entity.getParent().getChildren();
--- a/camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/methods/DslComponent.java
+++ b/camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/methods/DslComponent.java
@@ -28,6 +28,7 @@ import org.apache.brooklyn.api.sensor.AttributeSensor;
 import org.apache.brooklyn.api.sensor.Sensor;
 import org.apache.brooklyn.camp.brooklyn.BrooklynCampConstants;
 import org.apache.brooklyn.camp.brooklyn.spi.dsl.BrooklynDslDeferredSupplier;
+import org.apache.brooklyn.camp.brooklyn.spi.dsl.methods.DslComponent.Scope;
 import org.apache.brooklyn.core.config.ConfigKeys;
 import org.apache.brooklyn.core.entity.Entities;
 import org.apache.brooklyn.core.entity.EntityInternal;
@@ -44,6 +45,8 @@ import org.apache.brooklyn.util.text.StringEscapes.JavaStringEscapes;
 import com.google.common.base.Objects;
 import com.google.common.base.Optional;
 import com.google.common.base.Preconditions;
+import com.google.common.base.Predicate;
+import com.google.common.base.Predicates;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Iterables;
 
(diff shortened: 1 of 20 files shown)


robertgmoss:
@aledsage test failure in `EntitiesYamlTest.testMultipleReferences()`

aledsage:
Note the change to `test-referencing-entities.yaml` is to fix `EntitiesYamlTest.testMultipleReferences()`. That test relied on `$brooklyn:descendants("app1")` including itself in the search. I changed it so that it did not include itself (given one can easily do `$brooklyn:self()` instead). Please shout out if you think that's a bad idea @robertgmoss @neykov (but note that it does differ now compared to the behaviour of `Entities.descendants(entity)`).

robertgmoss:
Makes sense not to have self included in `$brooklyn:descendants`, but I find it contrary to expectations that `Entities.descendants(entity)` does not mirror this.  I presume this is to preserve backwards compatibility?  To me it would make sense to call it `descendantsOrSelf()`.

grkvlt:
@aledsage but isn't `$brooklyn:self().attributeWhenReady("foo")` exactly the same as `$brooklyn:attributeWhenReady("foo")` or am I missing something clever?

grkvlt:
@aledsage assuming you're talking about `producer: $brooklyn:component("this", "ignored")` being replaced with `producer: $brooklyn:self()` isn't the default for an enricher to have the producer be the current entity context, anyway?

aledsage:
@grkvlt we briefly changed the behaviour of `Propagator` to require that the producer be passed in (because previously the defaults were crazy - if no configuration, then propagate all your own sensors to yourself, thus causing an infinite loop!). We reverted that back to have the default producer being oneself, with the proper checks that there are no infinite loops.

But the point of this PR still stands for me: that we had now way to refer to self when supplying a config key value (for whatever use-cases of other enrichers, entities, etc). Therefore I'd argue this is still useful to include.

---
I'll also do as @robertgmoss suggests and rename `Entities.descendants()` to `Entities.descendantsAndSelf()`, and same for `Entities.ancestors()` (deprecating the old version).

camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/methods/DslComponent.java
```diff
@@ -100,30 +103,36 @@ protected EntityInternal getEntity() {
...
                 case ROOT:
-                    return getEntity().getApplication();
+                    return entity.getApplication();
                 case SCOPE_ROOT:
-                    return Entities.catalogItemScopeRoot(getEntity());
+                    return Entities.catalogItemScopeRoot(entity);
                 case DESCENDANT:
-                    entitiesToSearch = Entities.descendants(getEntity());
+                    entitiesToSearch = Entities.descendantsWithoutSelf(entity);
+                    entitiesToSearch = Iterables.filter(entitiesToSearch, notSelfPredicate);
```
robertgmoss: is this required? previous line suggests self will not be included

camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/methods/DslComponent.java
```diff
@@ -100,30 +103,36 @@ protected EntityInternal getEntity() {
...
                     break;
                 case ANCESTOR:
-                    entitiesToSearch = Entities.ancestors(getEntity());
+                    entitiesToSearch = Entities.ancestorsWithoutSelf(entity);
+                    entitiesToSearch = Iterables.filter(entitiesToSearch, notSelfPredicate);
```
robertgmoss: same

core/src/main/java/org/apache/brooklyn/core/entity/Entities.java
```diff
@@ -655,34 +677,112 @@ public static boolean isDescendant(Entity ancestor, Entity potentialDescendant)
...
+        }
+    }
+
+    
+    
+    
+    
+    
+    
+
```
robertgmoss: excessive whitespace

launcher/src/test/java/org/apache/brooklyn/launcher/blueprints/Windows7zipBlueprintLiveTest.java
```diff
@@ -81,7 +81,7 @@ public void test7zip() throws Exception {
         
         Predicate<Application> asserter = new Predicate<Application>() {
             @Override public boolean apply(Application app) {
-                VanillaWindowsProcess entity = Iterables.getOnlyElement(Entities.descendants(app, VanillaWindowsProcess.class));
+                VanillaWindowsProcess entity = Iterables.getOnlyElement(Entities.descendantsAndSelf(app, VanillaWindowsProcess.class));
```
robertgmoss: The author of this test clearly only wanted self here - is it worth adding a self method to `Entities`? (or if this comes up a lot a `rootElement()` method on Application? - not necessary for this PR)

core/src/test/java/org/apache/brooklyn/core/mgmt/rebind/RebindEntityTest.java
```diff
@@ -310,7 +310,7 @@ public void testEntityTags() throws Exception {
         origE.tags().addTag(origApp);
 
         newApp = rebind();
-        MyEntity newE = Iterables.getOnlyElement( Entities.descendants(newApp, MyEntity.class) );
+        MyEntity newE = Iterables.getOnlyElement( Entities.descendantsAndSelf(newApp, MyEntity.class) );
```
robertgmoss: same

launcher/src/test/java/org/apache/brooklyn/launcher/blueprints/Windows7zipBlueprintLiveTest.java
robertgmoss: Ignore me, it might not be the root being searched for.

aledsage:
Thanks @robertgmoss - incorporated those comments; merging now.

PR #219, opened 2016-06-23
Added identity method to DSL

grkvlt:
Adds a new function to the DSL that returns an entities ID. Useful for unique identifiers and references in blueprints. The new function can be used as follows:

```YAML
shell.env:
  ENTITY_ID: $brooklyn:entityId()
  OTHER_ID: $brooklyn:entity("other-entity").entityId()
```


camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/IdentityYamlTest.java
```diff
@@ -0,0 +1,82 @@
...
+
+@Test
+public class IdentityYamlTest extends AbstractYamlTest {
+    private static final Logger log = LoggerFactory.getLogger(IdentityYamlTest.class);
+
+    private static final ConfigKey<String> TEST_ENTITY_ONE_ID = ConfigKeys.newStringConfigKey("testentityone.id");
+    private static final ConfigKey<String> TEST_ENTITY_TWO_ID = ConfigKeys.newStringConfigKey("testentitytwo.id");
+
+    protected Iterable<? extends Entity> setupAndCheckTestEntityInBasicYamlWith() throws Exception {
+        Entity app = createAndStartApplication(loadYaml("test-entity-identity.yaml"));
```
aledsage: I have a personal dislike for this style of test, where the tiny amount of yaml is kept separately from the code that uses it. My rule of thumb is that it's only worth having it in a separate file if the yaml reaches some level of length/complexity (e.g. a dozen lines).

I'd write this as the java below. Note that it strips out all the extra assertions that aren't needed, or that are already covered in dozens of other tests.

    public void testBrooklynIdentityFunction() throws Exception {
    Entity app = createAndStartApplication(
            "services:",
            "- type: "+TestEntity.class.getName(),
            "  id: testentityone",
            "  name: Test Entity One",
            "  brooklyn.config:",
            "    testentityone.id: $brooklyn:identity()",
            "    testentitytwo.id: $brooklyn:entity(\"testentitytwo\").identity()",
            "- type: "+TestEntity.class.getName(),
            "  id: testentitytwo",
            "  name: test Entity Two");

    Entity entity1 = Iterables.find(testEntities, EntityPredicates.displayNameEqualTo("test Entity One"));
    Entity entity2 = Iterables.find(testEntities, EntityPredicates.displayNameEqualTo("test Entity Two"));

    assertEquals(entity1.config().get(TEST_ENTITY_ONE_ID), entity1.getId());
    assertEquals(entity1.config().get(TEST_ENTITY_TWO_ID), entity2.getId());
    }

aledsage:
@grkvlt the code looks good.

However, I'm unsure of the name "identity". We already have "id" in yaml, which means something entirely different (i.e. it is the user-supplied identifier for referring to other entities within yaml). I'd worry that "identity" will make people think that it will return the id.

Also, the pedant in me thinks that the internal id is an *identifier* rather than the identity.

Howe about `internalId()` instead?

neykov:
Agree there could be confusion between the different kind of IDs. Feels wrong for users to use `internalId`. What about `entityId()` or `brooklynId()`?

grkvlt:
@aledsage @neykov I will change the name to `entityId()` instead, and refactor the test

grkvlt:
@aledsage @neykov I'd like to merge this, unless there are any other issues?

PR #220, opened 2016-06-24
Added a percentage enricher

Graeme-Miller:


core/src/main/java/org/apache/brooklyn/enricher/stock/PercentageEnricher.java
```diff
@@ -0,0 +1,127 @@
...
+import org.apache.brooklyn.core.enricher.AbstractEnricher;
+import org.apache.brooklyn.util.collections.MutableMap;
+import org.apache.brooklyn.util.javalang.JavaClassNames;
+
+public class PercentageEnricher extends AbstractEnricher implements SensorEventListener<Number> {
+
+    private static final Logger LOG = LoggerFactory.getLogger(PercentageEnricher.class);
+
+    public static final ConfigKey<AttributeSensor<? extends Number>> SOURCE_CURRENT_SENSOR = ConfigKeys.newConfigKey(new TypeToken<AttributeSensor<? extends Number>>() {
+    }, "enricher.sourceCurrentSensor");
```
grkvlt: Don't split the line for the `TypeToken` block, just do`new TypeToken() { }, "enricher.blah"` and also add a description to the key definition.

core/src/main/java/org/apache/brooklyn/enricher/stock/PercentageEnricher.java
```diff
@@ -0,0 +1,127 @@
...
+    @Override
+    public void setEntity(EntityLocal entity) {
+        super.setEntity(entity);
+
+        this.sourceCurrentSensor = getConfig(SOURCE_CURRENT_SENSOR);
+        this.sourceTotalSensor = getConfig(SOURCE_TOTAL_SENSOR);
+        this.targetSensor = getConfig(TARGET_SENSOR);
+        this.producer = getConfig(PRODUCER) == null ? entity : getConfig(PRODUCER);
+
+        if (sourceCurrentSensor == null) {
```
grkvlt: Better doing these at assignment as follows:
```Java
sourceCurrentSensor = Preconditions.checkNotNull(getConfig(SOURCE_TOTAL_SENSOR), "Enricher %s has no %s", JavaClassNames.simpleClassName(this), SOURCE_CURRENT_SENSOR.getName());
```

core/src/main/java/org/apache/brooklyn/enricher/stock/PercentageEnricher.java
```diff
@@ -0,0 +1,127 @@
...
+        Double currentDouble = current.doubleValue();
+        Double totalDouble = total.doubleValue();
+
+        if (currentDouble > totalDouble) {
+            LOG.debug("Current is greater than total, returning");
+            return;
+        }
+
+        if (currentDouble < 0 || totalDouble < 0) {
+            LOG.debug("Current {"+currentDouble+"}  or total {"+totalDouble+"} is negative, returning");
```
grkvlt: Prefer using formatting in log method, as in `LOG.debug("Current value ({}) or total ({}) is negative, returning", currentDouble, totalDouble)`

core/src/main/java/org/apache/brooklyn/enricher/stock/PercentageEnricher.java
```diff
@@ -0,0 +1,127 @@
...
+        if (total == null || total.equals(0)) {
+            LOG.debug("total {" + total + "} is null or zero, returning");
+            return;
+        }
+
+        Double currentDouble = current.doubleValue();
+        Double totalDouble = total.doubleValue();
+
+        if (currentDouble > totalDouble) {
+            LOG.debug("Current is greater than total, returning");
```
grkvlt: Not necessarily wrong. The ratio will just be greater than 1.0

core/src/main/java/org/apache/brooklyn/enricher/stock/PercentageEnricher.java
```diff
@@ -0,0 +1,127 @@
...
+        }
+
+        if (currentDouble < 0 || totalDouble < 0) {
+            LOG.debug("Current {"+currentDouble+"}  or total {"+totalDouble+"} is negative, returning");
+            return;
+        }
+
+        if (current.equals(0)) {
+            LOG.debug("current is zero, setting percent to zero");
+            result = 0d;
```
grkvlt: This gives the same result as `0d / totalDouble` i.e. `0d` so don't need the check, just always do the division.

core/src/main/java/org/apache/brooklyn/enricher/stock/PercentageEnricher.java
```diff
@@ -0,0 +1,127 @@
...
+import org.apache.brooklyn.api.sensor.AttributeSensor;
+import org.apache.brooklyn.api.sensor.SensorEvent;
+import org.apache.brooklyn.api.sensor.SensorEventListener;
+import org.apache.brooklyn.config.ConfigKey;
+import org.apache.brooklyn.core.config.ConfigKeys;
+import org.apache.brooklyn.core.enricher.AbstractEnricher;
+import org.apache.brooklyn.util.collections.MutableMap;
+import org.apache.brooklyn.util.javalang.JavaClassNames;
+
+public class PercentageEnricher extends AbstractEnricher implements SensorEventListener<Number> {
```
grkvlt: Add some JavaDoc explaining what this does

core/src/test/java/org/apache/brooklyn/enricher/stock/PercentageEnricherTest.java
```diff
@@ -0,0 +1,275 @@
...
+        EntityAsserts.assertAttributeEqualsEventually(app, targetSensor, null);
+    }
+
+    @Test
+    public void totalLessThanCurrentTest() {
+        addEnricher();
+
+        app.sensors().set(currentSensor, 50d);
+        app.sensors().set(totalSensor, 25d);
+        EntityAsserts.assertAttributeEqualsEventually(app, targetSensor, null);
```
grkvlt: So, the result here should be 50/25 = 2.0d

core/src/test/java/org/apache/brooklyn/enricher/stock/PercentageEnricherTest.java
```diff
@@ -0,0 +1,275 @@
...
+    }
+
+    @Test
+    public void oneHundredPercentTest() {
+        addEnricher();
+
+        app.sensors().set(currentSensor, 50d);
+        app.sensors().set(totalSensor, 50d);
+        EntityAsserts.assertAttributeEqualsEventually(app, targetSensor, 100d);
```
grkvlt: This should be 50/50 = 1.0d

core/src/test/java/org/apache/brooklyn/enricher/stock/PercentageEnricherTest.java
```diff
@@ -0,0 +1,275 @@
...
+        EntityAsserts.assertAttributeEqualsEventually(app, targetSensor, null);
+    }
+
+    @Test
+    public void totalDoubleMaxValue() {
+        addEnricher();
+
+        app.sensors().set(currentSensor, Double.MAX_VALUE);
+        app.sensors().set(totalSensor, Double.MAX_VALUE);
+        EntityAsserts.assertAttributeEqualsEventually(app, targetSensor, 100d);
```
grkvlt: Again, result should be 1.0d

core/src/test/java/org/apache/brooklyn/enricher/stock/PercentageEnricherTest.java
```diff
@@ -0,0 +1,275 @@
...
+        app.enrichers().add(EnricherSpec.create(PercentageEnricher.class)
+                .configure(PercentageEnricher.SOURCE_CURRENT_SENSOR, currentSensor)
+                .configure(PercentageEnricher.SOURCE_TOTAL_SENSOR, totalSensor)
+                .configure(PercentageEnricher.TARGET_SENSOR, targetSensor)
+                .configure(PercentageEnricher.PRODUCER, producer)
+        );
+
+        producer.sensors().set(currentSensor, 25d);
+        producer.sensors().set(totalSensor, 50d);
+        EntityAsserts.assertAttributeEqualsEventually(app, targetSensor, 50d);
```
grkvlt: Should be 0.5d

core/src/main/java/org/apache/brooklyn/enricher/stock/PercentageEnricher.java
```diff
@@ -0,0 +1,127 @@
...
+
+    private static final Logger LOG = LoggerFactory.getLogger(PercentageEnricher.class);
+
+    public static final ConfigKey<AttributeSensor<? extends Number>> SOURCE_CURRENT_SENSOR = ConfigKeys.newConfigKey(new TypeToken<AttributeSensor<? extends Number>>() {
+    }, "enricher.sourceCurrentSensor");
+
+    public static final ConfigKey<AttributeSensor<? extends Number>> SOURCE_TOTAL_SENSOR = ConfigKeys.newConfigKey(new TypeToken<AttributeSensor<? extends Number>>() {
+    }, "enricher.sourceTotalSensor");
+
+    public static final ConfigKey<AttributeSensor<Double>> TARGET_SENSOR = ConfigKeys.newConfigKey(new TypeToken<AttributeSensor<Double>>() {
```
grkvlt: Add a `RendererHint` that will format this as a percentage in the `setEntity()` method, like this:
```Java
RendererHints.register(config().get(TARGET_SENSOR), RendererHints.displayValue(MathFunctions.percent(2)));
```

core/src/main/java/org/apache/brooklyn/enricher/stock/PercentageEnricher.java
geomacy: At the same time can you split it into two lines for readability, on Github if nowhere else!

core/src/main/java/org/apache/brooklyn/enricher/stock/PercentageEnricher.java
```diff
@@ -0,0 +1,127 @@
...
+            throw new IllegalArgumentException("Enricher " + JavaClassNames.simpleClassName(this) + " has no " + SOURCE_CURRENT_SENSOR.getName());
+        }
+        if (sourceTotalSensor == null) {
+            throw new IllegalArgumentException("Enricher " + JavaClassNames.simpleClassName(this) + " has no " + SOURCE_TOTAL_SENSOR.getName());
+        }
+        if (targetSensor == null) {
+            throw new IllegalArgumentException("Enricher " + JavaClassNames.simpleClassName(this) + " has no " + TARGET_SENSOR.getName());
+        }
+
+        if (targetSensor.equals(sourceCurrentSensor)) {
```
aledsage: This is presumably only a cycle if `entity.equals(producer)`.

core/src/main/java/org/apache/brooklyn/enricher/stock/PercentageEnricher.java
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    @Override
+    public void onEvent(SensorEvent<Number> event) {
+        Number current = producer.sensors().get(sourceCurrentSensor);
+        Number total = producer.sensors().get(sourceTotalSensor);
+        Double result = null;
+
+        if (current == null) {
+            LOG.debug("Current is null, returning");
```
aledsage: Give more context in the log message. Imagine that this is buried inside a log file, with a dozen entities that have this kind of enricher and you want to figure out from the customer's log what this refers to. I'd do something like:

```
LOG.debug("{} not updating target sensor of {}, because current is null", this, entity);
```

Same for your other log messages below.

core/src/main/java/org/apache/brooklyn/enricher/stock/PercentageEnricher.java
aledsage: Agreed - percentages above 100% are reasonable in some situations (e.g. over-committing virtual cores, etc).

core/src/test/java/org/apache/brooklyn/enricher/stock/PercentageEnricherTest.java
```diff
@@ -0,0 +1,275 @@
...
+    }
+
+    @Test
+    public void currentNullTest() {
+        addEnricher();
+
+        app.sensors().set(currentSensor, null);
+        app.sensors().set(totalSensor, 100d);
+        EntityAsserts.assertAttributeEqualsEventually(app, targetSensor, null);
```
aledsage: Do you want to test that it set the value to null, or that it didn't change the pre-existing value? I think it will do the latter. If you want to test that, then you want to initialise it to something first. Same for other tests.

core/src/test/java/org/apache/brooklyn/enricher/stock/PercentageEnricherTest.java
```diff
@@ -0,0 +1,275 @@
...
+                .configure(PercentageEnricher.SOURCE_CURRENT_SENSOR, currentSensor)
+                .configure(PercentageEnricher.SOURCE_TOTAL_SENSOR, totalSensor));
+    }
+
+    @Test
+    public void testDifferentProducer() {
+        EntitySpec<TestApplication> appSpec = EntitySpec.create(TestApplication.class)
+                .configure(BrooklynConfigKeys.SKIP_ON_BOX_BASE_DIR_RESOLUTION, shouldSkipOnBoxBaseDirResolution());
+
+        TestApplication producer = mgmt.getEntityManager().createEntity(appSpec);
```
aledsage: Personal preference to just create a child entity as the producer, rather than a new app - it's less code:

```
Entity producer = app.addChild(EntitySpec.create(TestEntity.class));
```

core/src/test/java/org/apache/brooklyn/enricher/stock/PercentageEnricherTest.java
```diff
@@ -0,0 +1,275 @@
...
+    public void totalDoubleMaxValue() {
+        addEnricher();
+
+        app.sensors().set(currentSensor, Double.MAX_VALUE);
+        app.sensors().set(totalSensor, Double.MAX_VALUE);
+        EntityAsserts.assertAttributeEqualsEventually(app, targetSensor, 100d);
+    }
+
+    //SETUP TESTS
+    @Test(expectedExceptions = IllegalArgumentException.class)
```
aledsage: Also want to test that it throws an exception if entity == producer && targetSensor == sourceSensor.

software/base/src/main/java/org/apache/brooklyn/entity/machine/AddMachineMetrics.java
```diff
@@ -75,6 +77,12 @@ public static void addMachineMetricsEnrichers(EntityLocal entity) {
         entity.enrichers().add(EnricherSpec.create(YamlRollingTimeWindowMeanEnricher.class)
                 .configure(YamlRollingTimeWindowMeanEnricher.SOURCE_SENSOR, MachineAttributes.USED_MEMORY_DELTA_PER_SECOND_LAST)
                 .configure(YamlRollingTimeWindowMeanEnricher.TARGET_SENSOR, MachineAttributes.USED_MEMORY_DELTA_PER_SECOND_IN_WINDOW));
+
+        entity.enrichers().add(EnricherSpec.create(PercentageEnricher.class)
+                .configure(PercentageEnricher.SOURCE_CURRENT_SENSOR, MachineAttributes.USED_MEMORY)
+                .configure(PercentageEnricher.SOURCE_TOTAL_SENSOR, MachineAttributes.TOTAL_MEMORY)
+                .configure(PercentageEnricher.TARGET_SENSOR, MachineAttributes.USED_MEMORY_PERCENT));
```
aledsage: I'd also do `.configure(PercentageEnricher.SUPPRESS_DUPLICATES, true)`.

aledsage:
Looks good - clean code and good test coverage. Only minor comments.

aledsage:
The jenkins failure is for the reason Andrew pointed out:
```
org.apache.brooklyn.util.exceptions.PropagatedRuntimeException: failed succeeds-eventually, 69 attempts, 30001ms elapsed: AssertionError: entity=Application[m0gi2x3r]; attribute=Sensor: target (java.lang.Double) expected [100.0] but found [1.0]
```

core/src/main/java/org/apache/brooklyn/enricher/stock/PercentageEnricher.java
```diff
@@ -0,0 +1,127 @@
...
+import org.apache.brooklyn.core.config.ConfigKeys;
+import org.apache.brooklyn.core.enricher.AbstractEnricher;
+import org.apache.brooklyn.util.collections.MutableMap;
+import org.apache.brooklyn.util.javalang.JavaClassNames;
+
+public class PercentageEnricher extends AbstractEnricher implements SensorEventListener<Number> {
+
+    private static final Logger LOG = LoggerFactory.getLogger(PercentageEnricher.class);
+
+    public static final ConfigKey<AttributeSensor<? extends Number>> SOURCE_CURRENT_SENSOR = ConfigKeys.newConfigKey(new TypeToken<AttributeSensor<? extends Number>>() {
```
neykov: Can you add `@SuppressWarnings("serial")` to all `new TypeToken` places.

core/src/test/java/org/apache/brooklyn/enricher/stock/PercentageEnricherTest.java
Graeme-Miller: I want to test the situation where the currentSensor has not been set at all. 
No exceptions should be thrown, the target should just be set to null.

There is a test later, validThenInvalid, which checks that a sensor having it's value updated to something invalid will not overwrite the previous result.

grkvlt:
@Graeme-Miller looks much better now, happy to merge

aledsage:
LGTM; merging.

PR #221, opened 2016-06-26
Misc cleanup of javalang, type coercion, jsonya, and Maybe

ahgittin:
big tidy is https://github.com/apache/brooklyn-server/commit/ec4da197d5caf3ac9bb63ab94924718aae83c615 moving most of our static `TypeCoercions` to a non-static `utils` impl (but not changing how we use `TypeCoercions` ... apart from the exception it throws)

others are as described in the individual commits, nothing too exciting, and big improvements to tests and javadocs

encountered while looking at how we can improve yaml type inference


core/src/main/java/org/apache/brooklyn/util/core/flags/TypeCoercions.java
```diff
@@ -53,443 +36,177 @@
...
-     * as well as <code>String</code> coercion to:
-     * <ul>
-     * <li> {@link Set}, {@link List}, {@link Map} and similar -- parses as YAML
-     * <li> {@link Date} -- parses using {@link Time#parseDate(String)}
-     * <li> {@link Duration} -- parses using {@link Duration#parse(String)}
-     * </ul>
-     */
-    public static <T> T coerce(Object value, Class<T> targetType) {
-        return coerce(value, TypeToken.of(targetType));
+    static TypeCoercerExtensible coercer;
```
grkvlt: I'm a bit wary of having our singleton instance as a non-final package scoped field here, is `private static final` better?

grkvlt:
@ahgittin Nice, looks like a good refactor. Minor comment about singletons...?

ahgittin:
good idea @grkvlt , done

PR #223, opened 2016-06-28
VanillaSoftwareProcess tests

aledsage:
For "	Don’t call postInstallCommand if skipInstall=true", why did we previous still call postInstall @grkvlt ?

The bulk of this PR is the unit testing of VanillaSoftwareProcess, using the `RecordSshTool` to stub out all the ssh commands.

Note the "Adds EffectorUtilsTest" is unrelated - I meant to include it in a PR a while ago, but accidentally missed it! We previously didn't have any tests that covered these methods directly.

### DIFF
--- a/core/src/test/java/org/apache/brooklyn/core/mgmt/internal/EffectorUtilsTest.java
+++ b/core/src/test/java/org/apache/brooklyn/core/mgmt/internal/EffectorUtilsTest.java
@@ -1,0 +1,119 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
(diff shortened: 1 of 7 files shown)


grkvlt:
@aledsage not sure about `postInstall()` but I wonder if there was an entity that relied on that behaviour in Clocker? I will check before we merge this...

neykov:
It was something to do with entities setting object properties on `postInstall`, later required in the launch step. We got NPEs if `postInstall` isn't called.

neykov:
LGTM

aledsage-tmp:
@grkvlt any objections to me merging this? I'll do so later today, unless you say otherwise.

PR #225, opened 2016-06-30
Improve test cases

aledsage:


test-framework/src/test/java/org/apache/brooklyn/test/framework/TestSshCommandTest.java
```diff
@@ -211,11 +213,9 @@ public void shouldFailOnUnmatchedStderr() {
 
     @Test
     public void shouldNotBeUpIfAssertionsFail() {
-        Map<String, Object> equalsOne = MutableMap.of();
-        equalsOne.put(EQUALS, 1);
+        Map<String, ?> equalsOne = ImmutableMap.of(EQUALS, 1);
```
bostko: Where is this used?

test-framework/src/test/java/org/apache/brooklyn/test/framework/TestSshCommandTest.java
```diff
@@ -278,6 +274,28 @@ public void shouldFailIfTestEntityHasNoMachine() throws Exception {
     }
     
     @Test
+    public void shouldFailFastIfNoCommand() throws Exception {
+        Map<String, ?> equalsZero = ImmutableMap.of(EQUALS, 0);
+        
+        TestSshCommand test = app.createAndManageChild(EntitySpec.create(TestSshCommand.class)
+                .configure(TIMEOUT, Duration.ONE_MINUTE)
```
bostko: The test passes for 0.5 s on my side so I think it is reasonable to shorten the limit to ten seconds?

test-framework/src/test/java/org/apache/brooklyn/test/framework/TestSshCommandTest.java
```diff
@@ -278,6 +274,28 @@ public void shouldFailIfTestEntityHasNoMachine() throws Exception {
...
+            app.start(ImmutableList.<Location>of());
+            Asserts.shouldHaveFailedPreviously();
+        } catch (Exception e) {
+            Asserts.expectedFailureContains(e, "Must specify exactly one of download.url and command");
+            Duration elapsed = Duration.of(stopwatch);
+            Asserts.assertTrue(elapsed.isShorterThan(Duration.THIRTY_SECONDS), "elapsed="+elapsed);
+        }
+
+        assertServiceFailed(test);
+    }
```
bostko: What do you think about adding the opposite test:
```java
    @Test
    public void testTargetEntityByIdWithDelayedEntityCreationAndNoTargetResolutionTimeout() {
        final Duration entityCreationDelay = Duration.millis(250);
        ExecutorService executor = Executors.newCachedThreadPool();

        try {
            executor.submit(new Runnable() {
                @Override
                public void run() {
                    Time.sleep(entityCreationDelay);
                    TestEntity target = app.addChild(EntitySpec.create(TestEntity.class)
                            .configure(BrooklynCampConstants.PLAN_ID, "myTargetId"));
                    target.sensors().set(STRING_SENSOR, "myval");
                }
            });

            app.addChild(EntitySpec.create(TestSensor.class)
                    .configure(TestSensor.TARGET_ID, "myTargetId")
                    .configure(TestSensor.SENSOR_NAME, STRING_SENSOR.getName())
                    .configure(TestSensor.ASSERTIONS, ImmutableList.of(ImmutableMap.of("equals", "myval"))));


            app.start(ImmutableList.<Location>of());

        } catch (Exception e) {
            Throwable noSuchElementException = Exceptions.getFirstThrowableOfType(e, NoSuchElementException.class);
            Asserts.assertNotNull(noSuchElementException);
            Asserts.assertTrue(noSuchElementException.getMessage().equals("No entity matching id myTargetId"));
        } finally {
            executor.shutdownNow();
        }
    }
```

bostko:
LGTM

test-framework/src/test/java/org/apache/brooklyn/test/framework/TestSshCommandTest.java
aledsage: We've seen on apache jenkins that tests some times run extremely slowly (possibly due to thread starvation). I'm not sure we've ever reached 10 seconds, but certainly several seconds. Our general default timeout is 30 seconds in `Asserts.DEFAULT_LONG_TIMEOUT`, given these are not performance tests.

The reason I put it as one minute here is that if the fail-fast doesn't happen, then the existing default is 30 seconds. I don't want us to pass the test with a false-positive when it was really just a slow timeout, and the defaults being applied. But I also feel like leaving the "normal" default time in the assertion a few lines lower down. I could/should have referenced that constant rather than re-declaring 30 seconds again.

test-framework/src/test/java/org/apache/brooklyn/test/framework/TestSshCommandTest.java
aledsage: It's used a few lines lower down, in `makeAssertions(equalsOne, equals255)`.

test-framework/src/test/java/org/apache/brooklyn/test/framework/TestSshCommandTest.java
aledsage: Not convinced by that opposite test (it's really a race, whether we manage to spot that the target is not there before it is created).

However, I am adding a couple of tests motivated by this: that we fail if the target does not exist (in both the case where there is and is not a TARGET_RESOLUTION_TIMEOUT).

aledsage:
Thanks @bostko - merging now.

PR #226, opened 2016-06-30
Expose assertServiceFailed(Entity) and assertServiceHealthy(Entity)

bostko:
I am using the same code in
https://github.com/apache/brooklyn-server/pull/211/files#diff-a1f3deedd5b53826884793055b989b87R96

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/entity/EntityAsserts.java
+++ b/core/src/main/java/org/apache/brooklyn/core/entity/EntityAsserts.java
@@ -32,17 +32,19 @@ import org.apache.brooklyn.api.sensor.AttributeSensor;
 import org.apache.brooklyn.api.sensor.SensorEvent;
 import org.apache.brooklyn.api.sensor.SensorEventListener;
 import org.apache.brooklyn.config.ConfigKey;
+import org.apache.brooklyn.core.entity.lifecycle.Lifecycle;
+import org.apache.brooklyn.core.entity.lifecycle.ServiceStateLogic;
 import org.apache.brooklyn.test.Asserts;
 import org.apache.brooklyn.util.time.Duration;
 
 import java.util.Collection;
 import java.util.Map;
 import java.util.Set;
-import java.util.concurrent.Callable;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
 
-import static org.apache.brooklyn.test.Asserts.assertEquals;
+import static org.apache.brooklyn.core.entity.trait.Startable.SERVICE_UP;
+import static org.apache.brooklyn.test.Asserts.*;
 
 /**
  * Convenience class containing assertions that may be made about entities.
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/brooklyn/core/entity/EntityAsserts.java
```diff
@@ -223,4 +225,14 @@ public void run() {
         });
     }
 
+    public static void assertServiceFailed(Entity test) {
+        assertFalse(test.sensors().get(SERVICE_UP), "Service should be down");
+        assertEquals(ServiceStateLogic.getExpectedState(test), Lifecycle.ON_FIRE, "Service should be marked on fire");
```
aledsage: Why is this using `getExpectedState`, rather than `assertAttributeEquals(test, Attributes.SERVICE_STATE_ACTUAL, Lifecycle.ON_FIRE)`?

core/src/main/java/org/apache/brooklyn/core/entity/EntityAsserts.java
```diff
@@ -223,4 +225,14 @@ public void run() {
         });
     }
 
+    public static void assertServiceFailed(Entity test) {
+        assertFalse(test.sensors().get(SERVICE_UP), "Service should be down");
```
aledsage: I prefer the error message you'd get if you called `assertAttributeEquals(test, SERVICE_UP, false)`. That would also cope better with the sensor being null, giving us a nicer message.

aledsage:
@bostko I'm unsure about this. The problem is whether `assertServiceFailed` and `assertServiceHealthy` are general enough for all entities. The name "service" also looks out of place, compared to the other methods (but I do see that it agrees with the name in "SERVICE_UP").

At the very least, we'd need clear javadoc to say what is being asserted. And probably should mark it `@Beta`.

I suspect there are a lot of places in our code that we do this pair of assertions, so overall I do lean towards including it.

aledsage:
@bostko why is it checking the value of `SERVICE_STATE_EXPECTED`, rather than the value of `SERVICE_STATE_ACTUAL`?

bostko:
I made it this way because `ServiceStateLogic.getExpectedState` is used in sever tests .
See `TestSshCommandIntegrationTest`, `SimpleShellCommandDeprecatedIntegrationTest`.
When I think about it I see that both are used and may be it is reasonable to add assertion for `SERVICE_STATE_EXPECTED`, at least for `assertServiceHealthy`
@aledsage what do you think?

aledsage:
@bostko we could include `SERVICE_STATE_EXPECTED` in the `assertServiceHealthy` - no strong feelings. It probably will never say actual is "running" if expected is not also something like running.

Can you also add javadoc, then good to merge. For example:

    /**
      * Asserts sensors {@code service.isUp} is true, and that {@code service.state} is "running".
      * Setting these sensors is common behaviour for entities, but depends on the particular entity 
      * implementation.
      */

And something similar for `assertServiceFailed`.

test-framework/src/test/java/org/apache/brooklyn/test/framework/TestSshCommandTest.java
```diff
@@ -30,6 +30,8 @@
 import static org.apache.brooklyn.test.framework.TestSshCommand.DOWNLOAD_URL;
 import static org.apache.brooklyn.test.framework.TestSshCommand.SHELL_ENVIRONMENT;
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.apache.brooklyn.core.entity.EntityAsserts.assertServiceFailed;
```
aledsage: This name doesn't match the new method names you've used - `assertEntityHealthy` etc.

Presumably you'd also need to update the calling code.

aledsage:
Thanks @bostko - LGTM; merging.

PR #227, opened 2016-07-01
Fixes BrooklynFeatureEnablement setting defaults

aledsage:
Previously if a downstream project set a default during static init, 
that was indistinguishable from subsequent explicit calls to 
setEnablement (or from brooklyn properties). Therefore the 
brooklyn.properties didn’t overwrite the default.

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/BrooklynFeatureEnablement.java
+++ b/core/src/main/java/org/apache/brooklyn/core/BrooklynFeatureEnablement.java
@@ -28,15 +28,22 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.google.common.annotations.Beta;
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Objects;
 import com.google.common.collect.Maps;
 
 /**
  * For enabling/disabling experimental features.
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/apache/brooklyn/core/BrooklynFeatureEnablement.java
```diff
@@ -182,27 +211,38 @@ public static boolean setEnablement(String property, boolean val) {
...
+                LOG.debug("Default enablement of "+property+" set to "+val);
             }
         }
     }
-    
+
+    @VisibleForTesting
     static void clearCache() {
         synchronized (MUTEX) {
             FEATURE_ENABLEMENTS.clear();
```
nakomis: Unrelated to this PR, but if `clearCache` is called, there is no mechanism for downstream projects to re-add their defaults

nakomis:
One comment unrelated to this PR. Other than that, code looks good. Is there an documentation that explains the explicit precedence?

aledsage:
@nakomis regarding docs, good question. There's not, except for the javadoc. Most end users won't touch this stuff, but agree it's still worth including in the docs.

Note that `clearCache` is only used by tests. Because of the limitation you pointed out, I marked it as `@VisibleForTesting` to discourage its use elsewhere.

Merging now.

PR #228, opened 2016-07-01
Misc OSGi fixes

neykov:
* adds missing bundles for persistence
* enable persistence by default
* add missing entityType for policies (required)
* remove deprecated config key, causing duplicate name warning in logs
* remove TargetableTestComponent from catalog.bom - it's not a user-facing entity, meant to be subclassed


grkvlt:
@neykov happy to merge, looks correct to me

PR #229, opened 2016-07-03
Merge brooklyn.parameters in YAML files

grkvlt:
Implements behaviour required to merge `brooklyn.parameters` sections of specifications in YAML files, allowing addition of new parameters to entities without having to specify the complete list again.

### DIFF
--- a/api/src/main/java/org/apache/brooklyn/api/internal/AbstractBrooklynObjectSpec.java
+++ b/api/src/main/java/org/apache/brooklyn/api/internal/AbstractBrooklynObjectSpec.java
@@ -27,6 +27,16 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.common.annotations.Beta;
+import com.google.common.base.Objects;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Iterables;
+import com.google.common.collect.Maps;
+
 import org.apache.brooklyn.api.mgmt.EntityManager;
 import org.apache.brooklyn.api.mgmt.Task;
 import org.apache.brooklyn.api.objs.BrooklynObject;
@@ -35,18 +45,9 @@ import org.apache.brooklyn.config.ConfigKey;
 import org.apache.brooklyn.config.ConfigKey.HasConfigKey;
 import org.apache.brooklyn.util.collections.MutableSet;
 import org.apache.brooklyn.util.exceptions.Exceptions;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
-import com.google.common.annotations.Beta;
-import com.google.common.base.Objects;
-import com.google.common.collect.ImmutableList;
-import com.google.common.collect.ImmutableList.Builder;
-import com.google.common.collect.ImmutableSet;
-import com.google.common.collect.Iterables;
-import com.google.common.collect.Maps;
-
-/** Defines a spec for creating a {@link BrooklynObject}.
+/**
+ * Defines a spec for creating a {@link BrooklynObject}.
  * <p>
  * In addition to the contract defined by the code,
  * subclasses should provide a public static <code>create(Class)</code>
@@ -54,7 +55,8 @@ import com.google.common.collect.Maps;
  * <p>
  * The spec is then passed to type-specific methods,
  * e.g. {@link EntityManager#createEntity(org.apache.brooklyn.api.entity.EntitySpec)}
- * to create a managed instance of the target type. */
+ * to create a managed instance of the target type.
+ */
 public abstract class AbstractBrooklynObjectSpec<T,SpecT extends AbstractBrooklynObjectSpec<T,SpecT>> implements Serializable {
 
     private static final long serialVersionUID = 3010955277740333030L;
@@ -134,7 +136,7 @@ public abstract class AbstractBrooklynObjectSpec<T,SpecT extends AbstractBrookly
         Iterables.addAll(this.tags, tagsToReplace);
         return self();
     }
-    
+
     // TODO which semantics are correct? replace has been the behaviour;
     // add breaks tests and adds unwanted parameters,
     // but replacing will cause some desired parameters to be lost.
@@ -153,17 +155,16 @@ public abstract class AbstractBrooklynObjectSpec<T,SpecT extends AbstractBrookly
     @Beta
     public SpecT parametersAdd(List<? extends SpecParameter<?>> parameters) {
         // parameters follows immutable pattern, unlike the other fields
-        Builder<SpecParameter<?>> result = ImmutableList.<SpecParameter<?>>builder();
-        if (this.parameters!=null)
-            result.addAll(this.parameters);
-        result.addAll( checkNotNull(parameters, "parameters") );
-        this.parameters = result.build();
+        Set<SpecParameter<?>> params = MutableSet.<SpecParameter<?>>copyOf(this.parameters);
+        params.removeAll(parameters);
+        params.addAll(parameters);
+        this.parameters = ImmutableList.copyOf(params);
         return self();
     }
     /** replaces parameters with the given */
     @Beta
     public SpecT parametersReplace(List<? extends SpecParameter<?>> parameters) {
-        this.parameters = ImmutableList.copyOf( checkNotNull(parameters, "parameters") );
+        this.parameters = ImmutableList.copyOf(checkNotNull(parameters, "parameters"));
         return self();
     }
 
@@ -302,12 +303,12 @@ public abstract class AbstractBrooklynObjectSpec<T,SpecT extends AbstractBrookly
     }
 
     public <V> SpecT removeConfig(ConfigKey<V> key) {
-        config.remove( checkNotNull(key, "key") );
+        config.remove(checkNotNull(key, "key"));
         return self();
     }
 
     public <V> SpecT removeFlag(String key) {
-        flags.remove( checkNotNull(key, "key") );
+        flags.remove(checkNotNull(key, "key"));
         return self();
     }
 
@@ -323,7 +324,7 @@ public abstract class AbstractBrooklynObjectSpec<T,SpecT extends AbstractBrookly
     public Map<String, ?> getFlags() {
         return Collections.unmodifiableMap(flags);
     }
-    
+
     /**
      * @return Read-only configuration values
      */
--- a/api/src/main/java/org/apache/brooklyn/api/internal/AbstractBrooklynObjectSpec.java
+++ b/api/src/main/java/org/apache/brooklyn/api/internal/AbstractBrooklynObjectSpec.java
@@ -155,10 +155,14 @@ public abstract class AbstractBrooklynObjectSpec<T,SpecT extends AbstractBrookly
     @Beta
     public SpecT parametersAdd(List<? extends SpecParameter<?>> parameters) {
         // parameters follows immutable pattern, unlike the other fields
-        Set<SpecParameter<?>> params = MutableSet.<SpecParameter<?>>copyOf(this.parameters);
-        params.removeAll(parameters);
-        params.addAll(parameters);
-        this.parameters = ImmutableList.copyOf(params);
+        Set<SpecParameter<?>> params = MutableSet.<SpecParameter<?>>copyOf(parameters);
+        Set<SpecParameter<?>> current = MutableSet.<SpecParameter<?>>copyOf(this.parameters);
+        current.removeAll(params);
+
+        this.parameters = ImmutableList.<SpecParameter<?>>builder()
+                .addAll(params)
+                .addAll(current)
+                .build();
         return self();
     }
     /** replaces parameters with the given */
--- a/camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/creation/BrooklynEntityDecorationResolver.java
+++ b/camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/creation/BrooklynEntityDecorationResolver.java
@@ -21,6 +21,9 @@ package org.apache.brooklyn.camp.brooklyn.spi.creation;
 import java.util.List;
 import java.util.Map;
 
+import com.google.common.annotations.Beta;
+import com.google.common.collect.ImmutableList;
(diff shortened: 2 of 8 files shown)


camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/catalog/SpecParameterUnwrappingTest.java
```diff
@@ -173,17 +173,128 @@ public void testDepentantCatalogsOverrideParameters(Class<? extends BrooklynObje
...
+        CatalogItem<?, ?> item = catalog.getCatalogItem(SYMBOLIC_NAME, TEST_VERSION);
+        AbstractBrooklynObjectSpec<?,?> spec = createSpec(item);
+        List<SpecParameter<?>> params = spec.getParameters();
+        assertEquals(params.size(), 1);
+        assertTrue(Iterables.tryFind(params, nameEqualTo("simple")).isPresent());
+        assertTrue(Iterables.tryFind(params, labelEqualTo("override")).isPresent());
+    }
+
+    @Test
+    public void testDependantCatalogConfigOverridesParameters() {
```
grkvlt: Note only for entities, as `brooklyn.config` is split into flags and config for policies and locations, so the parameter logic does not allow detection of existing config here.

core/src/main/java/org/apache/brooklyn/core/objs/BasicSpecParameter.java
```diff
@@ -384,5 +387,73 @@ public static WeightedParameter getFieldConfig(ConfigKey<?> config, Field config
...
+            }
+        };
+    }
+
+    /**
+     * Returns true if the {@link SpecParameter#getLabel() label} is the same as
+     * the specified string.
+     */
+    public static Predicate<SpecParameter<?>> labelEqualTo(final String label) {
+        return new Predicate<SpecParameter<?>>() {
```
aledsage: Don't use anonymous inner classes, in case they get persisted.

aledsage: I'd be tempted to move the public methods to a separate `SpecParameterPredicates` class.

core/src/main/java/org/apache/brooklyn/core/objs/BasicSpecParameter.java
```diff
@@ -384,5 +387,73 @@ public static WeightedParameter getFieldConfig(ConfigKey<?> config, Field config
...
+    public static void addParameters(AbstractBrooklynObjectSpec<?, ?> spec, List<? extends SpecParameter<?>> explicitParams, BrooklynClassLoadingContext loader) {
+        if (explicitParams.size() > 0) {
+            List<SpecParameter<?>> currentParams = MutableList.copyOf(spec.getParameters());
+            Map<ConfigKey<?>, ?> currentConfig = spec.getConfig();
+            for (final SpecParameter<?> param : explicitParams) {
+                Optional<ConfigKey<?>> configExists = Iterables.tryFind(currentConfig.keySet(), ConfigPredicates.nameEqualTo(param.getConfigKey().getName()));
+                if (!configExists.isPresent()) {
+                    Optional<SpecParameter<?>> paramExists = Iterables.tryFind(currentParams, sameName(param));
+                    if (paramExists.isPresent()) {
+                        currentParams.remove(paramExists.get());
```
aledsage: Not sure I follow. I'd have though we still want to have the parameter (e.g. so we get the metadata such as its description). I can see how removing it would mean that it would pick up the config value rather than the param's default value, but is there not a better way to do this?

grkvlt: @aledsage not sure, can you suggest anything?

aledsage: @grkvlt I wondered about the same pattern as is used for config (see `BrooklynComponentTemplateResolver.decorateSpec`). Not sure if the `configureEntityConfig` does anything significantly different from what you have though, just in a different place.

My gut high-level feel was that you'd always overwrite the inherited SpecParameter if there was one with the same name (irrespective of whether an additional config value is defined). Then when we look up the config value, we'd figure out the right ConfigKey (see `EntityConfigMap.getConfig`, and how it looks up "ownKey" - that should return the config key produced from our SpecParameter). Then we will use that as the default.

The area that this PR does not seem to address (in the test cases at least, as far as I can see) is the second kind of inheritance: an entity inheriting config from its parent. What should this do? Should we find the `SpecParameter` of our parent entity, and use that default? Or are you leaving that for a separate PR?

But I really don't like the code in there (or our camp parsing in general - hard to know exactly what to do to improve it though).

grkvlt: OK, I think that might work, if I remove the code that deals with `ConfigKey` then things will be OK, because the actual config will still override a parameter with a default, no matter where in the hierarchy it is set. I'll add a test case for the inherited config case. I assume you mean something like this:

```YAML
- type: entity
  id: parent
  brooklyn.config:
    key: value
  brooklyn.children:
    - type: entity
      id: child
      brooklyn.parameters:
        - name: key
          default: other
```

So, the value of `$brooklyn:entity("child").config("key")` will be `value` because it inherits from the parent entity, yes?

camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/creation/BrooklynEntityDecorationResolver.java
```diff
@@ -181,19 +181,17 @@ protected void addDecorationFromJsonMap(Map<?, ?> decorationJson, List<EntityIni
...
-            // inherit except where it is set as config and then add the new explicit ones
-            if (!explicitParams.isEmpty()) {
-                entitySpec.parameters(explicitParams);
-            }
-            if (entitySpec.getParameters().isEmpty()) {
-                entitySpec.parameters(BasicSpecParameter.fromSpec(instantiator.loader.getManagementContext(), entitySpec));
+            if (explicitParams.size() > 0) {
+                BasicSpecParameter.addParameters(entitySpec, explicitParams, instantiator.loader);
+            } else {
+                BasicSpecParameter.addParameters(entitySpec, ImmutableList.<SpecParameter<?>>of(), instantiator.loader);
```
neykov: No need for an `if-else` here. Above case handles all code paths.

core/src/main/java/org/apache/brooklyn/core/objs/BasicSpecParameter.java
```diff
@@ -384,5 +387,34 @@ public static WeightedParameter getFieldConfig(ConfigKey<?> config, Field config
...
+     * updated list.
+     *
+     * @see EntitySpec#parameters(List)
+     */
+    public static void addParameters(AbstractBrooklynObjectSpec<?, ?> spec, List<? extends SpecParameter<?>> explicitParams, BrooklynClassLoadingContext loader) {
+        if (explicitParams.size() > 0) {
+            List<SpecParameter<?>> currentParams = MutableList.copyOf(spec.getParameters());
+            Map<ConfigKey<?>, ?> currentConfig = spec.getConfig();
+            for (final SpecParameter<?> param : explicitParams) {
+                Optional<ConfigKey<?>> configExists = Iterables.tryFind(currentConfig.keySet(), ConfigPredicates.nameEqualTo(param.getConfigKey().getName()));
```
neykov: Should add parameters even if a config key with the same name exists. Config keys coming from yaml are too generic if the config key doesn't exist on the class. An explicit parameter with type information is always preferred. And even if a config key with the same name exists on the class, a parameter from yaml should override it.

core/src/main/java/org/apache/brooklyn/core/objs/BasicSpecParameter.java
```diff
@@ -384,5 +387,34 @@ public static WeightedParameter getFieldConfig(ConfigKey<?> config, Field config
...
+                if (!configExists.isPresent()) {
+                    Optional<SpecParameter<?>> paramExists = Iterables.tryFind(currentParams, SpecParameterPredicates.sameName(param));
+                    if (paramExists.isPresent()) {
+                        currentParams.remove(paramExists.get());
+                    }
+                    currentParams.add(param);
+                }
```
neykov: Could use a `LinkedHashSet` to simplify the code.

core/src/main/java/org/apache/brooklyn/core/objs/BasicSpecParameter.java
```diff
@@ -384,5 +387,34 @@ public static WeightedParameter getFieldConfig(ConfigKey<?> config, Field config
...
+     * {@link AbstractBrooklynObjectSpec spec}. Inherits parent parameters except
+     * where there is an existing {@link ConfigKey config key}.
+     * <p>
+     * Implemented by explicitly replacing the existing parameters with an
+     * updated list.
+     *
+     * @see EntitySpec#parameters(List)
+     */
+    public static void addParameters(AbstractBrooklynObjectSpec<?, ?> spec, List<? extends SpecParameter<?>> explicitParams, BrooklynClassLoadingContext loader) {
+        if (explicitParams.size() > 0) {
```
neykov: Shouldn't be one or the other, but both instead. First add from spec (if no spec parameters), then merge with explicit parameters.

core/src/main/java/org/apache/brooklyn/core/objs/BasicSpecParameter.java
```diff
@@ -384,5 +387,34 @@ public static WeightedParameter getFieldConfig(ConfigKey<?> config, Field config
...
+     * Adds the given list of {@link SpecParameter parameters} to the provided
+     * {@link AbstractBrooklynObjectSpec spec}. Inherits parent parameters except
+     * where there is an existing {@link ConfigKey config key}.
+     * <p>
+     * Implemented by explicitly replacing the existing parameters with an
+     * updated list.
+     *
+     * @see EntitySpec#parameters(List)
+     */
+    public static void addParameters(AbstractBrooklynObjectSpec<?, ?> spec, List<? extends SpecParameter<?>> explicitParams, BrooklynClassLoadingContext loader) {
```
neykov: Isn't this helper better moved in the `parametersAdd` method of `AbstractBrooklynobjectSpec`?

camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/catalog/SpecParameterUnwrappingTest.java
```diff
@@ -158,18 +157,18 @@ public void testDepentantCatalogsInheritParameters(Class<? extends BrooklynObjec
                 "      - simple",
                 "  - id: " + SYMBOLIC_NAME,
                 "    item:",
-                "      type: paramItem:" + TEST_VERSION);
+                "      type: paramItem");
```
neykov: Why remove the version?

neykov:
+1 for the change. I've personally felt the pain of not being able to merge parameters.
I think that we need to discuss with a broader audience how the merge happens - suggest mailing to the dev list.
What I'm concerned about is mostly how the parameters look in the UI when merged so the input of blueprint authors will be very valuable. Should override parameters be appended or prepended? Should they be merged with those coming from the class?

camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/creation/BrooklynEntityDecorationResolver.java
grkvlt: True, not sure what I was thinking there...!

core/src/main/java/org/apache/brooklyn/core/objs/BasicSpecParameter.java
grkvlt: I was trying to deal with the situation where a config key exists on the class, and a value is set in the YAML for it, and did not want the parameter to override that value. I now think that this won't be a problem - the default value will change (perhaps) but the actual value set will be the value from the YAML brooklyn.config setting. I'll write some more tests for this.

camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/catalog/SpecParameterUnwrappingTest.java
grkvlt: Don't remember; I think it just didn't work properly with it set, and other tests didn't have it?

grkvlt:
@neykov I have updated this with the suggested changes

neykov:
@grkvlt Changes look good and can be merged as is. Still think it's a good idea to gather some input on the mailing list, or at least increase exposure of the upcoming change before merging. For example - are we prepending or appending parameters (matters for UI).

grkvlt:
@neykov I implemented your suggestion (first add from spec if no spec parameters then merge with explicit parameters) and it also now places the current entities params at the start of the list. If there is no objection from the mailing list, we should be able to merge soon.

grkvlt:
See [message](http://mail-archives.apache.org/mod_mbox/brooklyn-dev/201607.mbox/%3CCAEK0Ph2S4WJ7pV6XNr21wJGPfikX5tQYgm3at1dNDoBwoacy6Q%40mail.gmail.com%3E) on <mailto:dev@brooklyn.apache.org> for discussion.

api/src/main/java/org/apache/brooklyn/api/internal/AbstractBrooklynObjectSpec.java
```diff
@@ -153,17 +155,20 @@ public SpecT parameters(List<? extends SpecParameter<?>> parameters) {
...
-            result.addAll(this.parameters);
-        result.addAll( checkNotNull(parameters, "parameters") );
-        this.parameters = result.build();
+        Set<SpecParameter<?>> params = MutableSet.<SpecParameter<?>>copyOf(parameters);
+        Set<SpecParameter<?>> current = MutableSet.<SpecParameter<?>>copyOf(this.parameters);
+        current.removeAll(params);
+
+        this.parameters = ImmutableList.<SpecParameter<?>>builder()
+                .addAll(params)
+                .addAll(current)
```
neykov: I really like how this turned out. We can use it to reorder/pull up parameters we are interested in.

neykov:
Thanks @grkvlt. Merging.

PR #230, opened 2016-07-05
Fix setting the hostname on non-centos machines

neykov:
Failure in the command will abort machine obtain, so make sure the command reports successfull result on non-centos machines.

### DIFF
--- a/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
+++ b/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
@@ -924,10 +924,10 @@ public class JcloudsLocation extends AbstractCloudMachineProvisioningLocation im
                         executeCommandThrowingOnError(
                                 (SshMachineLocation)machineLocation,
                                 "Generate hostname " + node.getName(),
-                                Arrays.asList("sudo hostname " + node.getName(),
-                                        "sudo sed -i \"s/HOSTNAME=.*/HOSTNAME=" + node.getName() + "/g\" /etc/sysconfig/network",
-                                        "sudo bash -c \"echo 127.0.0.1   `hostname` >> /etc/hosts\"")
-                        );
+                                ImmutableList.of(BashCommands.chainGroup(
+                                        String.format("echo '127.0.0.1 %s' | ( %s )", node.getName(), BashCommands.sudo("tee -a /etc/hosts")),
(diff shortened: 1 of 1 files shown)


locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -924,9 +924,10 @@ protected MachineLocation obtainOnce(ConfigBag setup) throws NoMachinesAvailable
                         executeCommandThrowingOnError(
                                 (SshMachineLocation)machineLocation,
                                 "Generate hostname " + node.getName(),
-                                Arrays.asList("sudo hostname " + node.getName(),
-                                        "sudo sed -i \"s/HOSTNAME=.*/HOSTNAME=" + node.getName() + "/g\" /etc/sysconfig/network",
-                                        "sudo bash -c \"echo 127.0.0.1   `hostname` >> /etc/hosts\"")
+                                ImmutableList.of(BashCommands.chainGroup(
+                                        "sudo bash -c \"echo 127.0.0.1   " + node.getName() + " >> /etc/hosts\"",
```
grkvlt: prefer `"echo 127.0.0.1 " + node.getName() + " | sudo tee -a /etc/hosts"` to this

grkvlt:
Why not fix [SetHostnameCustomizer](https://github.com/apache/brooklyn-server/blob/master/software/base/src/main/java/org/apache/brooklyn/entity/machine/SetHostnameCustomizer.java) for non CentOS/RHEL instead?

grkvlt:
Or, copy [JcloudsHostnameCustomizer](https://github.com/brooklyncentral/clocker/blob/master/docker/src/main/java/clocker/docker/entity/util/JcloudsHostnameCustomizer.java) from Clocker, which works on most Linux OS types?

grkvlt:
@neykov @aledsage I _think_ `SetHostnameCustomizer` is meant to be used after this command executes in `JcloudsLocation` to fix additional CentOS/RHL issues, right?

neykov:
Yeah, looks like the customizer is an additional step. Can we just use it from provisioning.properties instead of making it  an initiailzer?

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
neykov: Good suggestion, will implement it.

neykov:
@grkvlt addressed comments.

ahgittin:
feels like we should be able to combine some of the commands between the two files

added TODO comments in https://github.com/apache/brooklyn-server/commit/257926534101749280853455dd7e0684eaa56c27

(and merged)

PR #231, opened 2016-07-05
Fix jmxmp connector in karaf

aledsage:
e.g. tomcat failed to reach service-up because it couldn't create the JMXMP connector due to classloading problems.

### DIFF
--- a/karaf/apache-brooklyn/src/main/resources/etc/org.apache.brooklyn.osgilauncher.cfg
+++ b/karaf/apache-brooklyn/src/main/resources/etc/org.apache.brooklyn.osgilauncher.cfg
@@ -38,22 +38,22 @@
 #ignorePersistenceErrors=true
 
 # The high availability mode. Possible values are:
-# - disabled: management node works in isolation - will not cooperate with any other standby/master nodes in management plane;
-# - auto: will look for other management nodes, and will allocate itself as standby or master based on other nodes' states;
-# - master: will startup as master - if there is already a master then fails immediately;
-# - standby: will start up as lukewarm standby with no state - if there is not already a master then fails immediately,
+# - DISABLED: management node works in isolation - will not cooperate with any other standby/master nodes in management plane;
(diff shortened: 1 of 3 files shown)


justinThompson:
This also resolves the issues of policies not being registered into the catalog

aledsage:
@justinThompson I extracted the catalog.bom policy issue into its own PR, and have merged it. Rebased this PR again against master.

The failing test here is possibly unrelated. I'll investigate further, but jenkins is also rebuilding:

```
2016-07-05 23:05:51,943 INFO  TESTNG FAILED: "Surefire test" - org.apache.brooklyn.camp.brooklyn.ConfigYamlTest.testDeferredSupplierToAttributeWhenReady() finished in 180 ms
java.lang.IllegalStateException
	at org.apache.brooklyn.util.guava.IllegalStateExceptionSupplier.get(IllegalStateExceptionSupplier.java:52)
	at org.apache.brooklyn.util.guava.IllegalStateExceptionSupplier.get(IllegalStateExceptionSupplier.java:26)
	at org.apache.brooklyn.util.guava.Maybe$Absent.getException(Maybe.java:276)
	at org.apache.brooklyn.util.guava.Maybe$Absent.get(Maybe.java:273)
	at org.apache.brooklyn.camp.brooklyn.ConfigYamlTest.testDeferredSupplierToAttributeWhenReady(ConfigYamlTest.java:207)
```

aledsage:
Confirmed that test failure is unrelated. The problem is that when `entity.config().getNonBlocking()` is called with a deferred supplier value, it will kick off a task and then wait just a few millis for that taks to execute `deferredSupplier.get()`. If it times out, then it returns `Maybe.absent`. However, on apache jenkins the machine is often slow so the task doesn't complete in the given number of millis (even though `deferredSupplier.get()` doesn't need to block for anything).

neykov:
LGTM, merging

PR #234, opened 2016-07-06
Fix Entities.UNCHANGED marker, for rebind

aledsage:
Previously, after rebind then enrichers using this would have their
own copy of their object (due to xstream deserialisation). Instead,
use an enum.

This does not fix the problem for existing persisted state of enrichers
that have their own object!

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/entity/Entities.java
+++ b/core/src/main/java/org/apache/brooklyn/core/entity/Entities.java
@@ -144,19 +144,26 @@ public class Entities {
             "access.cert",
             "access.key");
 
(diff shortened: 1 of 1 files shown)


ahgittin:
very nice fix.  i should think in most cases we don't persist these values, they are intermediate returns to determine whether to change.  but there might be a few cases where we do, eg initial values, however in those cases they should get the real value as soon as the enricher runs, so worth keeping an eye out.

aledsage:
@ahgittin we persist them as part of the enricher returned by `ServiceNotUpLogic.newEnricherForServiceUpIfNotUpIndicatorsEmpty()`. The default value in the computing function is `Entities.UNCHANGED`. So it might be in customer persisted state :-(

PR #238, opened 2016-07-07
Don't serialize NodeMetadata in location config

neykov:


aledsage:
jenkins failure is caused by the failing test `AssemblyTest.checkBrooklynCoreFeature`. That looks unrelated to these changes. I'm going to hope that it's fixed by https://github.com/apache/brooklyn-server/pull/239, but it's probably not as the jenkins test output doesn't suggest there were any persistence/rebind errors.

aledsage:
LGTM; merging.

PR #239, opened 2016-07-07
Enable persistence for tests and point it at the target folder

neykov:


karaf/itest/src/test/java/org/apache/brooklyn/KarafTestUtils.java
```diff
@@ -44,7 +46,9 @@
         configureConsole().ignoreLocalConsole(),
         logLevel(LogLevel.INFO),
         features(karafStandardFeaturesRepository(), "eventadmin"),
-        junitBundles()
+        junitBundles(),
+        editConfigurationFilePut("etc/org.apache.brooklyn.osgilauncher.cfg", "persistMode", "AUTO"),
+        editConfigurationFilePut("etc/org.apache.brooklyn.osgilauncher.cfg", "persistenceDir", new File("target/paxexam/persistence/" + Identifiers.makeRandomId(6) + "/").getAbsolutePath()),
```
aledsage: Looks good. I presume target will get deleted at some point (e.g. on the next clean build), especially when run on local machines? So no need for tearDown to delete the folder?

With that assumption, very happy to merge. Doing so now.

neykov: That was my thinking, yes. Isn't worth making it more complex just to do the cleanup.

PR #240, opened 2016-07-08
Use entity's classpath when loading resources

neykov:
In addition to any libraries provided by the catalog item.

We are actually assigning the wrong catalog item ID to the entity (should be the inner-most instead of the outer-most). That's more of a workaround, than fixing the root cause.


aledsage:
@neykov I'm not sure what the best thing to do is here long term. You could imagine having a catalog item X that has library L, which includes an inner entity from another catalog item Y. And that in X, it sets a config value on the inner entity (where the type of the value is a class in L, e.g. using `$brooklyn:object`). On rebind, for us to fully deserialize the file for the inner entity, it will need access to library L. If we changed the catalog reference so that the inner entity knew only about Y and not X, then it wouldn't have access to L.

We could let it have the chain of catalog items. But that might lead to conflicting dependencies.

Much longer term, we probably want to encourage state to be simple - primitives/strings/maps etc wherever possible.

I'm happy to merge this now. Except for the failing jenkins test ({{org.apache.brooklyn.AssemblyTest.checkBrooklynCoreFeature}}). Any ideas how we fix that? It has failed in a couple of different PR builds.

I'll leave you to merge this when you're happy with it.

PR #241, opened 2016-07-08
catalog-classes.bom: adds brooklyn-test-framework entities

aledsage:


karaf/init/src/main/resources/catalog-classes.bom
```diff
@@ -505,3 +505,36 @@ brooklyn.catalog:
...
+
+  - brooklyn.libraries:
+    - name: org.apache.brooklyn.brooklyn-test-framework
+      version: "0.10.0.SNAPSHOT" # BROOKLYN_VERSION
+    itemType: entity
+    items:
+    - id: org.apache.brooklyn.test.framework.TestSshCommand
+      item:
+        type: org.apache.brooklyn.test.framework.TestSshCommand
+    - id: org.apache.brooklyn.test.framework.SimpleShellCommandTest
```
geomacy: I suppose this is needed even though it's been renamed as TestSshCommand, because the old and now deprecated class exists?  Or would it not be ok to leave it out here, and let TestSshCommand be the only one available in the catalog?

geomacy:
One question, but LGTM.

karaf/init/src/main/resources/catalog-classes.bom
aledsage: Good question. That depends on what backwards compatibility we want. I think we want to minimise the amount of breakage when someone is upgrading so probably include it. But we want to hide it so folk aren't offered it.

We have a `CatalogItem.setDeprecated`. We should probably change our yaml parser to accept `deprecated: true`, and to set that on the catalog item. But not in this PR. See https://issues.apache.org/jira/browse/BROOKLYN-315

PR #242, opened 2016-07-08
OSGi fallback resource loading

neykov:


core/src/main/java/org/apache/brooklyn/core/mgmt/classloading/JavaBrooklynClassLoadingContext.java
```diff
@@ -133,17 +129,11 @@ public boolean equals(Object obj) {
...
     @Override
     public Iterable<URL> getResources(String name) {
-        Enumeration<URL> resources;
-        try {
-            resources = getClassLoader().getResources(name);
-        } catch (IOException e) {
-            throw Exceptions.propagate(e);
-        }
-        return Collections.list(resources);
+        return new ClassLoaderUtils(loader).getResources(name);
```
aledsage: ClassLoader.getResources() should return an empty Iterable. At first blush, it looks like this might throw an exception, or perhaps return null?

core/src/main/java/org/apache/brooklyn/util/core/ClassLoaderUtils.java
```diff
@@ -136,6 +143,33 @@ public ClassLoaderUtils(Class<?> callingClass, @Nullable ManagementContext mgmt)
...
+            throw new ClassNotFoundException("Class " + className + " could not be loaded from bundle " + toBundleString(symbolicName, version));
+        }
+    }
+
+    public URL getResource(String name) {
+        return load(name, ResourceLoaderDispatcher.INSTANCE);
+    }
+
+    public Iterable<URL> getResources(String name) {
+        return load(name, MultipleResourceLoaderDispatcher.INSTANCE);
```
aledsage: The contract of `getResources(name)` is that it "finds all the resources with the given name". Does that mean we should not terminate the search in he first bundle/loader that we find something, but instead search the other places as well?!

Also I think that load(String, LoaderDispatcher) just does a null check. So if MultipleResourceLoaderDispatcher returns an empty Iterable, then we'd terminate the search early.

aledsage:
I like the way you've shared the code - very nice!

Can you add some basic tests, similar to those in ClassLoaderUtils, for getResource and getResources?

neykov:
Addressed comments in a separate commit, to be squashed before merge.

neykov:
Also dispatcher is not the right term for the loaders, any better suggestions?

aledsage:
@neykov LGTM; waiting for jenkins to confirm it works and then will merge.

neykov:
Will squash the changes in that case.

PR #243, opened 2016-07-09
Fix AssemblyTest (increase timeout for FeaturesService)

aledsage:
@neykov setting these timeouts seems to fix `AssemblyTest` for me locally. I've added the timeout to everywhere else that we do `@Inject` in itests (I saw a failure in `OsgiLauncherTest` for `gave up waiting for service org.apache.brooklyn.api.mgmt.ManagementContext`).


aledsage:
I'm going to merge this now, as the existing `AssemblyTest` failures are causing problems in other PRs. The only code changes are to tests, to increase a timeout, so it is low risk.

PR #244, opened 2016-07-09
Update Karaf logging configuration.

geomacy:
Updates org.ops4j.pax.logging.cfg to be configured similarly to the logback.xml configuration with regular launcher.

### DIFF
--- a/karaf/apache-brooklyn/src/main/resources/etc/org.ops4j.pax.logging.cfg
+++ b/karaf/apache-brooklyn/src/main/resources/etc/org.ops4j.pax.logging.cfg
@@ -18,22 +18,32 @@
 ################################################################################
 
 # Root logger
-log4j.rootLogger=INFO, out, osgi:VmLogAppender
+log4j.rootLogger=INFO, out, debugFile, osgi:VmLogAppender
 log4j.throwableRenderer=org.apache.log4j.OsgiThrowableRenderer
 
 # CONSOLE appender not used by default
 log4j.appender.stdout=org.apache.log4j.ConsoleAppender
 log4j.appender.stdout.layout=org.apache.log4j.PatternLayout
-log4j.appender.stdout.layout.ConversionPattern=%d{ABSOLUTE} | %-5.5p | %-16.16t | %-32.32C %4L | %X{bundle.id} - %X{bundle.name} - %X{bundle.version} | %m%n
+log4j.appender.stdout.layout.ConversionPattern=%d{ABSOLUTE} %-5.5p %3.3X{bundle.id} %-30.30c{2} [%-16.16t] %m%n
 
-# File appender
+# Info file appender
 log4j.appender.out=org.apache.log4j.RollingFileAppender
 log4j.appender.out.layout=org.apache.log4j.PatternLayout
-log4j.appender.out.layout.ConversionPattern=%d{ABSOLUTE} | %-5.5p | %-16.16t | %-32.32C %4L | %X{bundle.id} - %X{bundle.name} - %X{bundle.version} | %m%n
-log4j.appender.out.file=${karaf.home}/log/brooklyn.log
+log4j.appender.out.layout.ConversionPattern=%d{ABSOLUTE} %-5.5p %3.3X{bundle.id} %-30.30c{2} [%-16.16t] %m%n
+log4j.appender.out.file=${karaf.home}/log/brooklyn.info.log
+log4j.appender.out.Threshold=INFO
 log4j.appender.out.append=true
-log4j.appender.out.maxFileSize=1MB
+log4j.appender.out.maxFileSize=100MB
 log4j.appender.out.maxBackupIndex=10
+
+# Debug file appender
+log4j.appender.debugFile=org.apache.log4j.RollingFileAppender
+log4j.appender.debugFile.layout=org.apache.log4j.PatternLayout
+log4j.appender.debugFile.layout.ConversionPattern=%d{ABSOLUTE} %-5.5p %3.3X{bundle.id} %-30.30c{2} [%-16.16t] %m%n
+log4j.appender.debugFile.file=${karaf.home}/log/brooklyn.debug.log
+log4j.appender.debugFile.append=true
+log4j.appender.debugFile.maxFileSize=100MB
+log4j.appender.debugFile.maxBackupIndex=10
 
 # Sift appender
 log4j.appender.sift=org.apache.log4j.sift.MDCSiftingAppender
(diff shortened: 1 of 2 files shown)


karaf/apache-brooklyn/src/main/resources/etc/org.ops4j.pax.logging.cfg
```diff
@@ -18,29 +18,74 @@
...
 log4j.appender.out.maxFileSize=1MB
 log4j.appender.out.maxBackupIndex=10
 
+# Debug file appender
+log4j.appender.debugFile=org.apache.log4j.RollingFileAppender
+log4j.appender.debugFile.layout=org.apache.log4j.PatternLayout
+log4j.appender.debugFile.layout.ConversionPattern=%d{ABSOLUTE} %-5.5p %3.3X{bundle.id} %-30.30c{2} [%-16.16t] %m%n
+log4j.appender.debugFile.file=${karaf.home}/log/brooklyn.debug.log
+log4j.appender.debugFile.append=true
+log4j.appender.debugFile.maxFileSize=1MB
```
aledsage: 1MB seems very small. It was 100MB in logback: https://github.com/apache/brooklyn-server/blob/rel/apache-brooklyn-0.9.0/logging/logback-includes/src/main/resources/brooklyn/logback-appender-file.xml#L60-L62

karaf/apache-brooklyn/src/main/resources/etc/org.ops4j.pax.logging.cfg
```diff
@@ -18,29 +18,74 @@
...
+
+
+# some loggers are very noisy however, exclude them, or make them file only
+log4j.logger.org.apache.cxf=ERROR
+log4j.logger.net.schmizz=WARN
+log4j.logger.org.eclipse.jetty=WARN
+log4j.logger.org.reflections.Reflections=WARN
+log4j.logger.com.sun.jersey.server.impl.application=WARN
+log4j.logger.org.apache.whirr.service.ComputeCache=WARN
+log4j.logger.jclouds.ssh=WARN
```
aledsage: I think we do want this. In https://github.com/apache/brooklyn-server/blob/rel/apache-brooklyn-0.9.0/logging/logback-includes/src/main/resources/brooklyn/logback-logger-excludes.xml#L40-L43, it sets it to only be written to the file. Is it possible to do anything like that here? Could we perhaps have a different log level specifically for the `log4j.appender.out`?

Same for org.apache.http.impl.client: in the logback config, it is written only to the file (rather than setting it to warn).

karaf/apache-brooklyn/src/main/resources/etc/org.ops4j.pax.logging.cfg
```diff
@@ -18,29 +18,74 @@
...
+log4j.logger.org.apache.http.impl.client=WARN
+log4j.logger.javax.management.remote=WARN
+log4j.logger.com.wordnik.swagger=WARN
+log4j.logger.org.mongodb.driver=WARN
+log4j.logger.org.apache.brooklyn.location.jclouds.JcloudsPropertiesFromBrooklynProperties=WARN
```
aledsage: Why set `JcloudsPropertiesFromBrooklynProperties ` to warn?

karaf/apache-brooklyn/src/main/resources/etc/org.ops4j.pax.logging.cfg
```diff
@@ -18,29 +18,74 @@
...
+log4j.appender.sift.appender.append=true
+
+
+# Logger configuration
+
+# The following properties turn on quite verbose DEBUG logging for Brooklyn-relevant loggers
+log4j.logger.brooklyn=DEBUG
+log4j.logger.org.apache.brooklyn=DEBUG
+log4j.logger.org.jclouds=DEBUG
+log4j.logger.jclouds=DEBUG
```
aledsage: Also set `log4j.logger.io.cloudsoft.winrm4j.winrm.WinRmTool=DEBUG` (as is done in `logback-logger-excludes.xml` - that is a very strange place to set the logging to debug!)

aledsage:
Thanks @geomacy - only a few minor comments.

karaf/apache-brooklyn/src/main/resources/etc/org.ops4j.pax.logging.cfg
geomacy: Both the loggers here are file loggers, there's no console output - I thought this might be a good default, but I shall copy the comment about "a bit noisy at info" and leave them at info level, then users looking at this file will be able to conveniently turn them off if needed.

karaf/apache-brooklyn/src/main/resources/etc/org.ops4j.pax.logging.cfg
geomacy: When I started Brooklyn up and pointed a web server at it, but didn't actually do anything, I got bunches of log lines from that class every few seconds, not really telling me anything but filling up the logs, e.g.

JcloudsPropertiesFromBrooklynProperties.getProviderName NOT changing provider from openstack-nova to candidate bluebox

This setting will avoid the logs filling up with these messages.

geomacy: although INFO might be better

karaf/apache-brooklyn/src/main/resources/etc/org.ops4j.pax.logging.cfg
geomacy: Should that not be org.apache.brooklyn.util.core.internal.winrm.WinRmTool in Brooklyn?

karaf/apache-brooklyn/src/main/resources/etc/org.ops4j.pax.logging.cfg
```diff
@@ -18,29 +18,74 @@
 ################################################################################
```
aledsage: Can we remove the "Licensed to Apache" header from the AMP files please.

geomacy: ah! good point - cut-n-paste...

karaf/apache-brooklyn/src/main/resources/etc/org.ops4j.pax.logging.cfg
aledsage: Feels like we should fix the code doing the logging, rather than treating that brooklyn class as a special case of not being logged at debug.

geomacy: OK - I'll add that - I guess could change the code to live TRACE rather than remove the line altogether.

karaf/apache-brooklyn/src/main/resources/etc/org.ops4j.pax.logging.cfg
aledsage: We'll already have the o.a.b WinRmTool at debug. The reason I say about the io.cloudsoft class is becaues that is the existing logback configuration: https://github.com/apache/brooklyn-server/blob/master/logging/logback-includes/src/main/resources/brooklyn/logback-logger-excludes.xml#L71

geomacy: Yes I'm just wondering if that's right, or something mistakenly added to Brooklyn?

karaf/apache-brooklyn/src/main/resources/etc/org.ops4j.pax.logging.cfg
aledsage: @geomacy Not sure why this comment is in the brooklyn repo - I thought I made the comment in a private AMP repo on a different PR?! Deleting these comments as they have nothing to do with this brooklyn code/PR.

karaf/apache-brooklyn/src/main/resources/etc/org.ops4j.pax.logging.cfg
aledsage: I think it's right - we are using the apache 2 licensed java winrm library at `io.cloudsoft.winrm4j`. I presume the debug logging that it writes out was deemed useful (in the same way as we find the jclouds debug logging is useful).

geomacy: Grand will add that here too then.

geomacy:
Changes made and squashed.

aledsage:
LGTM; merging now.

PR #245, opened 2016-07-09
BROOKLYN-313: log.debug exceptions thrown by effectors

aledsage:
Also see https://github.com/apache/brooklyn-library/pull/49 (but please review and merge this first).

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/mgmt/internal/EffectorUtils.java
+++ b/core/src/main/java/org/apache/brooklyn/core/mgmt/internal/EffectorUtils.java
@@ -278,9 +278,7 @@ public class EffectorUtils {
                 mgmtSupport.getEntityChangeListener().onEffectorCompleted(eff);
             }
         } catch (Exception e) {
-            handleEffectorException(entity, eff, e);
-            // (won't return below)
-            return null;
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/apache/brooklyn/core/mgmt/internal/EffectorUtils.java
```diff
@@ -319,11 +317,12 @@ private static RuntimeException propagate(Entity entity, Effector<?> effector, T
                 
             EffectorCallPropagatedRuntimeException result = new EffectorCallPropagatedRuntimeException(entity, effector, throwable);
             log.warn(Exceptions.collapseText(result));
+            log.debug(makeMessage(entity, effector), throwable);
```
bostko: If I understand correctly, this line makes the difference.

bostko:
LGTM

core/src/main/java/org/apache/brooklyn/core/mgmt/internal/EffectorUtils.java
aledsage: Correct - everything else is just minor code tidy or tests.

PR #246, opened 2016-07-11
Fix adding multiple RendererHints for percentage enricher

grkvlt:
Stops long stream of `WARN` log messages when using `PercentageEnricher` on a sensor that already has a `RendererHint` applied


aledsage:
Test failure is unrelated - `RebindJmxFeedTest.testJmxFeedIsPersistedWithPreCreatedJmxHelper` fails with "java.rmi.server.ExportException: Port already in use: 40124". We should look at using port 0, so it chooses any free port, rather than us finding a free port and then telling it to use that port.

aledsage:
LGTM; merging.

PR #247, opened 2016-07-11
Add logo urls to karaf catalog-classes.bom

aledsage:
Add `iconUrl: ...` for things in karaf's catalog-classes.bom

These use references to external URLs, rather than things on the classpath, because we don't want to rely on having started the catalog's bundles (or classloading from the "right" bundle) when returning icons.

### DIFF
--- a/karaf/init/src/main/resources/catalog-classes.bom
+++ b/karaf/init/src/main/resources/catalog-classes.bom
@@ -118,6 +118,7 @@ brooklyn.catalog:
       item:
         type: org.apache.brooklyn.entity.software.base.SameServerEntity
     - id: org.apache.brooklyn.entity.chef.ChefEntity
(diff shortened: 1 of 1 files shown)


Graeme-Miller:
lgtm, but can you re-run the test, it failed with OutOfMemory exception

neykov:
OOME fixed in https://github.com/apache/brooklyn-server/pull/250

aledsage:
I rebased to include #250; jenkins is building again.

aledsage:
Relies on https://github.com/apache/brooklyn-library/pull/50 for some of these icons.

Using `classpath://...` for the urls to get them from the local amp, rather than having to reach out to the public internet. We can do this, now that resource loading in karaf is working.

aledsage:
apache/brooklyn-library#50 is now merged; merging this.

PR #248, opened 2016-07-12
brooklyn-server karaf: don’t depend on library

aledsage:
Don’t depend on bundles that are in brooklyn-library when building
brooklyn-server: they won’t be available yet.

Long-term solution is to move all of this into brooklyn-dist, and 
then we can re-enable these.

### DIFF
--- a/karaf/features/src/main/feature/feature.xml
+++ b/karaf/features/src/main/feature/feature.xml
@@ -336,11 +336,16 @@
         <bundle>mvn:org.apache.brooklyn/brooklyn-test-framework/${project.version}</bundle>
     </feature>
 
-    <!-- Don't load all bundles out of the box in the long run. Let users cherry pick what they need. -->
-    <feature name="brooklyn-software-all" version="${project.version}" description="Brooklyn All Software Entities">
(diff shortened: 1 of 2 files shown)


karaf/features/src/main/feature/feature.xml
```diff
@@ -279,76 +279,16 @@
...
-    </feature>
-
-    <feature name="brooklyn-software-monitoring" version="${project.version}" description="Brooklyn Monitoring Software Entities">
-        <bundle>mvn:org.apache.brooklyn/brooklyn-software-monitoring/${project.version}</bundle>
-        <feature>brooklyn-core</feature>
-        <feature>brooklyn-api</feature>
-        <feature>brooklyn-software-base</feature>
-        <bundle dependency="true">mvn:com.google.guava/guava/${guava.version}</bundle>
-    </feature>
-
```
neykov: Let's leave these features in, just don't include them in the boot bundles (only change `brooklyn-software-all`). This should be enough not to depend on `brooklyn-library` at build time.

neykov:
LGTM, merging

PR #249, opened 2016-07-12
Return when PercentageEnricher cannot compute result

grkvlt:
Does not pollute `INFO` log with exception stack traces. Also replaces check for zero with a `DoubleMath.fuzzyCompare()` call using an epsilon of 1x10^-12 to allow for floating point inaccuracies.

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/enricher/stock/PercentageEnricher.java
+++ b/core/src/main/java/org/apache/brooklyn/enricher/stock/PercentageEnricher.java
@@ -21,7 +21,9 @@ package org.apache.brooklyn.enricher.stock;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.google.common.base.Objects;
 import com.google.common.base.Preconditions;
+import com.google.common.math.DoubleMath;
 import com.google.common.reflect.TypeToken;
 
 import org.apache.brooklyn.api.entity.Entity;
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/brooklyn/enricher/stock/PercentageEnricher.java
```diff
@@ -98,18 +101,27 @@ public void setEntity(EntityLocal entity) {
...
-        if (totalDouble.compareTo(0d) == 0) {
-            LOG.debug("Can't calculate Enricher ({}) value for entity ({}) as total from producer ({}) is zero", new Object[]{JavaClassNames.simpleClassName(this), entity, producer});
+        if (DoubleMath.fuzzyEquals(totalDouble, 0d, EPSILON)) {
+            LOG.trace("Can't calculate percentage value for entity {} as total from producer {} is zero", entity, producer);
             return;
         }
-
-        if (currentDouble < 0 || totalDouble < 0) {
-            LOG.debug("Can't calculate Enricher ({}) value for entity ({}) as Current ({})  or total ({}) value from producer ({}) is negative, returning", new Object[]{JavaClassNames.simpleClassName(this), entity, currentDouble, totalDouble, producer});
+        if (currentDouble < 0d || totalDouble < 0d) {
```
aledsage: Not sure why we disallow negative percentages, but that was already the case. So your PR looks good.

PR #250, opened 2016-07-12
Fix tests out of memory crashes

neykov:
Note that this will make the project fail on 32 bit JDKs, at least on Windows.


aledsage:
LGTM; merging.

PR #252, opened 2016-07-12
catalog-classes.bom: fix test-framework bundle name

aledsage:


neykov:
Should fix the generated `symbolicName` eventually for `test-framework` to strip the `brooklyn` prefix.

PR #254, opened 2016-07-12
Fix memory leak where unused class references are kept around

neykov:
The BrooklynTypes cache was keeping references to classes indefinitely, preventing the OSGi class loaders and frameworks from being GCed.


aledsage:
LGTM - well found @neykov ! Once jenkins has finished its built, very happy to merge.

PR #255, opened 2016-07-13
BROOKLYN-316: fix use of karaf jetty LoginService

aledsage:


karaf/jetty-config/src/main/resources/jetty.xml
```diff
@@ -56,6 +56,9 @@
                         <Item>org.apache.brooklyn.rest.security.jaas.BrooklynLoginModule$RolePrincipal</Item>
                     </Array>
                 </Set>
+                <Set name="identityService">
+                    <New class="org.eclipse.jetty.security.DefaultIdentityService"/>
+                </Set>
```
neykov: That's the default if none is specified (see [code](https://github.com/eclipse/jetty.project/blob/jetty-9.3.x/jetty-jaas/src/main/java/org/eclipse/jetty/jaas/JAASLoginService.java#L178)).

m4rkmckenna: @neykov But if two threads are invoking that code an `java.lang.IllegalStateException` is thrown

neykov:
Reading the the jira issue, the change makes much more sense now.
Shouldn't you add the package `org.eclipse.jetty.security` to the imports instead?

m4rkmckenna:
LGTM / tested

aledsage:
@neykov you said: "add the package org.eclipse.jetty.security to the imports instead." Do you mean in karaf/jetty-config/pom.xml ? I've added that now.

Merging now.

PR #256, opened 2016-07-13
Convert versions to OSGi format before passing them to the OSGi runtime.

neykov:


utils/common/src/main/java/org/apache/brooklyn/util/osgi/OsgiUtils.java
```diff
@@ -98,4 +101,97 @@ public static String getVersionedId(Bundle b) {
         return Maybe.of(new VersionedName(parts[0], v));
     }
 
+    public static String toOsgiVersion(String version) {
```
aledsage: Worth adding a test that demonstrates the behaviour (even though the code is copy-pasted). I'd also like to see that a version such as `0.10.0-20160713.1653` is converted to `0.10.0.20160713_1653`. Note the last dot is converted to an underscore.

PR #257, opened 2016-07-13
Comment out catalog items coming from brooklyn-libraries in Karaf dist

neykov:
brooklyn-libraries features are not installed at boot so can't reference them in the default catalog. The reason is that brooklyn-libraries is not yet built when the Karaf dist is assembled.


geomacy:
I've raised PRs to move the karaf distribution into brooklyn-dist, see 

https://github.com/apache/brooklyn-server/pull/271
https://github.com/apache/brooklyn-library/pull/53
https://github.com/apache/brooklyn-dist/pull/41

As part of that I split out the catalog and feature files for the brooklyn library stuff, so if the above are accepted then this PR may no longer be needed.

neykov:
Closing in favour of https://github.com/apache/brooklyn-server/pull/271

PR #258, opened 2016-07-13
Add shell.env config for SSH sensor and effector

grkvlt:
Means behaviour of sensors and effectors can be controlled with more granularity

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/effector/ssh/SshCommandEffector.java
+++ b/core/src/main/java/org/apache/brooklyn/core/effector/ssh/SshCommandEffector.java
@@ -19,34 +19,39 @@
 package org.apache.brooklyn.core.effector.ssh;
 
 import java.util.Map;
+import java.util.concurrent.ExecutionException;
+
+import com.google.common.base.Preconditions;
+import com.google.common.base.Predicates;
+import com.google.common.collect.Maps;
 
 import org.apache.brooklyn.api.effector.Effector;
 import org.apache.brooklyn.api.effector.ParameterType;
 import org.apache.brooklyn.config.ConfigKey;
 import org.apache.brooklyn.core.config.ConfigKeys;
+import org.apache.brooklyn.core.config.MapConfigKey;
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/brooklyn/core/effector/ssh/SshCommandEffector.java
```diff
@@ -65,38 +69,43 @@ public Body(Effector<?> eff, ConfigBag params) {
...
+            } catch (InterruptedException | ExecutionException e) {
+                Exceptions.propagateIfFatal(e);
+            }
+
             // finally set the parameters we've been passed; this will repeat declared parameters but to no harm,
             // it may pick up additional values (could be a flag defining whether this is permitted or not)
-            env.putAll(Strings.toStringMap(params.getAllConfig()));
-            
-            SshEffectorTasks.SshEffectorTaskFactory<String> t = SshEffectorTasks.ssh(command)
+            env.putAll(Strings.toStringMap(params.getAllConfig(), ""));
```
neykov: Remove `EFFECTOR_SHELL_ENVIRONMENT` if present first. Why implement it at all if all effector parameters are passed to the environment?

neykov:
Not convinced we need the `shell.env` for the effector, what's the use case? Otherwise LGTM.

core/src/main/java/org/apache/brooklyn/core/effector/ssh/SshCommandEffector.java
grkvlt: We need it because it makes the syntax the same between the SSH sensor and effector initialisers, and it has the normal and well understood semantics of a config key?

neykov:
Thanks for the update @grkvlt. PR looks good, can be merged after resolving merge conflicts.

neykov:
Conflicts probably because of the merge of https://github.com/apache/brooklyn-server/pull/216, needs same changed carried across.

grkvlt:
@neykov can you take a quick look at this, now that it is using the `ShellEnvironmentSerializer`?

PR #259, opened 2016-07-13
Create OptionalEntity to allow configurable addition of child entities

grkvlt:
Allows entity creation to be controlled using a configuration key value.

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/entity/stock/ConditionalEntity.java
+++ b/core/src/main/java/org/apache/brooklyn/entity/stock/ConditionalEntity.java
@@ -1,0 +1,61 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.brooklyn.entity.stock;
+
+import com.google.common.annotations.Beta;
+import com.google.common.reflect.TypeToken;
+
+import org.apache.brooklyn.api.entity.Entity;
+import org.apache.brooklyn.api.entity.EntitySpec;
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/apache/brooklyn/entity/stock/OptionalEntity.java
```diff
@@ -0,0 +1,56 @@
...
+ *       $brooklyn:entitySpec:
+ *         type: load-balancer
+ *         brooklyn.config:
+ *           proxy.port: 8080
+ *           loadbalancer.serverpool: $brooklyn:entity("servers")
+ * </pre>
+ */
+@Beta
+@ImplementedBy(OptionalEntityImpl.class)
+public interface OptionalEntity extends BasicStartable {
```
neykov: Would a `ConditionalEntity` be a better name?

core/src/main/java/org/apache/brooklyn/entity/stock/OptionalEntityImpl.java
```diff
@@ -0,0 +1,41 @@
...
+
+    @Override
+    public void start(Collection<? extends Location> locations) {
+        Boolean create = config().get(CREATE_OPTIONAL_ENTITY);
+        if (Boolean.TRUE.equals(create)) {
+            EntitySpec<?> spec = Preconditions.checkNotNull(config().get(OPTIONAL_ENTITY_SPEC));
+            addChild(EntitySpec.create(spec));
+        }
+        super.start(locations);
+    }
```
neykov: Check if the entity is already created - to handle start -> stop -> start/restart.

grkvlt: Good point.

core/src/test/java/org/apache/brooklyn/entity/stock/OptionalEntityTest.java
```diff
@@ -0,0 +1,80 @@
...
+import org.apache.brooklyn.api.entity.EntitySpec;
+import org.apache.brooklyn.api.location.LocationSpec;
+import org.apache.brooklyn.api.mgmt.ManagementContext;
+import org.apache.brooklyn.core.entity.Entities;
+import org.apache.brooklyn.core.location.SimulatedLocation;
+import org.apache.brooklyn.core.test.entity.LocalManagementContextForTests;
+import org.apache.brooklyn.core.test.entity.TestApplication;
+import org.apache.brooklyn.core.test.entity.TestEntity;
+
+public class OptionalEntityTest {
```
aledsage-tmp: You can extend BrooklynAppUnitTestSupport to save you writing a few lines here.

aledsage-tmp:
@grkvlt agree with @neykov about the name ConditionalEntity. And similar rename in config keys. Also worth adding test for restart.

grkvlt:
@aledsage-tmp @neykov I changed the name and made the suggested changes

neykov:
Thanks @grkvlt, merging.

PR #260, opened 2016-07-14
Fix tests expecting specific Brooklyn version

neykov:


neykov:
Changing tests only - merging.

PR #261, opened 2016-07-14
Test target sensor set from targetId

grkvlt:


test-framework/src/main/java/org/apache/brooklyn/test/framework/RelativeEntityTestCase.java
```diff
@@ -52,12 +51,14 @@
 @ImplementedBy(RelativeEntityTestCaseImpl.class)
 public interface RelativeEntityTestCase extends TargetableTestComponent {
 
+    @SetFromFlag("anchor")
     AttributeSensorAndConfigKey<Entity, Entity> ANCHOR = ConfigKeys.newSensorAndConfigKey(Entity.class,
-            "anchor",
+            "test.anchor.entity",
```
aledsage-tmp: I don't think we should rename all these. The semantics for SetFromFlag is different (for inheritance) so this is not backwards compatible. I think we should leave as-is. The old name was fine. We don't need to make the key names "slightly unique".

aledsage-tmp:
It was very hard to see what has changed due to the many formatting and rename changes.

The one line change for the pr subject looks sensible.
The configure key renames I disagree with.
The formatting changes: no strong feelings.

test-framework/src/main/java/org/apache/brooklyn/test/framework/RelativeEntityTestCase.java
grkvlt: OK, I'll revert this

neykov:
Thanks for the updates @grkvlt. LGTM, merging.

neykov:
Fixed a NPE in https://github.com/apache/brooklyn-server/commit/7dd9d1ba25b3c681efabfe911f9c1c645753ece4

grkvlt:
@neykov thanks for the merge && the fixup!

PR #263, opened 2016-07-15
Adds a new ChildSoftwareProcess entity

grkvlt:


software/base/src/main/java/org/apache/brooklyn/entity/software/base/ChildSoftwareProcess.java
```diff
@@ -0,0 +1,58 @@
...
+package org.apache.brooklyn.entity.software.base;
+
+import com.google.common.annotations.Beta;
+
+import org.apache.brooklyn.api.entity.ImplementedBy;
+import org.apache.brooklyn.location.ssh.SshMachineLocation;
+
+/**
+ * An entity that has the same functionality as {@link VanillaSoftwareProcess}
+ * but starts in the same {@link SshMachineLocation machine} as its parent.
```
aledsage-tmp: I thought this would happen automatically with vanillasoftwareprocess? The parent is given a ProvisioningLocation and creates the machine; the child is passed the machine in the start() call. Maybe things changed with the way it now looks in the parent's locations (to find those passed in via an EntitySpec)?

software/base/src/main/java/org/apache/brooklyn/entity/software/base/ChildSoftwareProcess.java
```diff
@@ -0,0 +1,58 @@
...
+ * services:
+ *   - type: org.apache.brooklyn.entity.software.base.VanillaSoftwareProcess
+ *     brooklyn.config:
+ *       install.command: |
+ *         wget http://example.org/service.tgz
+ *         tar zxf service.tgz
+ *       launch.command: |
+ *         service.sh
+ *     brooklyn.children:
+ *       - type: org.apache.brooklyn.entity.software.base.ChildSoftwareProcess
```
aledsage-tmp: I'm not sure what the best thing is here. I see advantages to what you did for being explicit, but it's less reusable. If any vanillasoftwaprocess could be used, either independently or as a child (I.e. Part of setting up a machine) then that would seem most flexible/reusable/composable.

software/base/src/main/java/org/apache/brooklyn/entity/software/base/ChildSoftwareProcessImpl.java
```diff
@@ -0,0 +1,51 @@
...
+import org.apache.brooklyn.core.location.Machines;
+import org.apache.brooklyn.location.ssh.SshMachineLocation;
+import org.apache.brooklyn.util.guava.Maybe;
+
+public class ChildSoftwareProcessImpl extends VanillaSoftwareProcessImpl implements ChildSoftwareProcess {
+
+    @Override
+    public void init() {
+        // Set the child start mode so that we are started after the parent finishes
+        getParent().config().set(CHILDREN_STARTABLE_MODE, ChildStartableMode.FOREGROUND_LATE);
```
aledsage-tmp: I'd only set this if it doesn't already have an explicit value.

software/base/src/main/java/org/apache/brooklyn/entity/software/base/ChildSoftwareProcess.java
grkvlt: No, you're right. Not sure what I was thinking here. I'm closing this PR...

PR #264, opened 2016-07-18
Orphaned locations cleaner

iyovcheva:
Created a cleaner for orphaned locations, enrichers, policies and feeds. It does not affect the current persisted state but does a copy of it removing the orphaned data. The criteria for keeping a location is to have any reference to it in any of the existing entities, enrichers, policies, feeds or another location which is already referenced in the mentioned.


aledsage-tmp:
There are 114 changed/added files. That seems excessive (though struggling to review properly on iPad because of that).

Instead of checking in an entire persisted state folder (eg including brooklyn-server/launcher/src/test/resources/orphaned-locations-test-data/data-without-orphaned-locations/catalog/org.apache.brooklyn.entity.cm.ansible.AnsibleEntity_0.10.0-SNAPSHOT), I'd recommend we have far more focused test cases with just a minimal file. Or even construct the persisted state on the fly.

core/src/main/java/org/apache/brooklyn/core/mgmt/rebind/transformer/impl/DeleteOrphanedLocationsTransformer.java
```diff
@@ -94,6 +126,86 @@ public boolean containsAny(Collection<?> container, Iterable<?> contenders) {
...
+            String entityString = (String) entity.getValue();
+            result.addAll(Strings.getAllOccurrencesBetween(entityString, locationProxyPrefix, locationProxySuffix));
+            result.addAll(Strings.getAllOccurrencesBetween(entityString, locationPrefix, locationSuffix));
+            result.addAll(Strings.getAllOccurrencesBetween(entityString, locationPrefixWinEOF, locationSuffixWinEOF));
+        }
+
+        return result;
+    }
+
+    private Set<String> searchLocationsToKeepInLocations(Map<String, String> locations, Set<String> alreadyReferencedLocations) {
```
aledsage-tmp: Also need to keep parent to and children.

core/src/main/java/org/apache/brooklyn/core/mgmt/rebind/transformer/impl/DeleteOrphanedLocationsTransformer.java
```diff
@@ -94,6 +126,86 @@ public boolean containsAny(Collection<?> container, Iterable<?> contenders) {
...
+    }
+
+    private Set<String> searchLocationsToKeepInEntities(Map<String, String> entities) {
+        Set<String> result = MutableSet.of();
+
+        for(Map.Entry entity: entities.entrySet()) {
+
+            String locationProxyPrefix = "<locationProxy>";
+            String locationProxySuffix = "</locationProxy>";
+            String locationPrefix = "<locations>\n" +
```
aledsage-tmp: Perhaps use xpath for this. See XmlUtils.

launcher-common/pom.xml
```diff
@@ -79,6 +79,11 @@
             <artifactId>brooklyn-locations-jclouds</artifactId>
             <version>${project.version}</version>
         </dependency>
+        <dependency>
+            <groupId>org.apache.brooklyn</groupId>
+            <artifactId>brooklyn-core</artifactId>
+            <version>0.10.0-SNAPSHOT</version>
```
aledsage-tmp: Why do we now need this dependency? Also, use project.version.

launcher/pom.xml
```diff
@@ -218,6 +218,17 @@
             <classifier>tests</classifier>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.apache.brooklyn</groupId>
+            <artifactId>brooklyn-core</artifactId>
+            <version>0.10.0-SNAPSHOT</version>
```
aledsage-tmp: Use project.version

launcher/src/test/java/org/apache/brooklyn/launcher/CleanOrphanedLocationsLiveTest.java
```diff
@@ -0,0 +1,247 @@
...
+import org.testng.annotations.AfterMethod;
+import org.testng.annotations.BeforeMethod;
+import org.testng.annotations.Test;
+
+import org.apache.brooklyn.core.internal.BrooklynProperties;
+
+import java.util.Map;
+import java.util.Set;
+
+public class CleanOrphanedLocationsLiveTest {
```
aledsage-tmp: It's very hard to tell from the test code what cases are actually being tested. I'll look at writing some test cases later, to see how we can do this in a more explicit way.

aledsage:
@iyovcheva I've created https://github.com/iyovcheva/brooklyn-server/pull/1 to show how we can write unit tests for this.

The absolute must is that we "do no harm". We must never delete a location that is referenced by anything else. These tests show that we are deleting referenced locations.

core/src/main/java/org/apache/brooklyn/core/mgmt/rebind/transformer/impl/DeleteOrphanedLocationsTransformer.java
iyovcheva: Added them as well, but the comment won't show as outdated.

launcher/src/test/java/org/apache/brooklyn/launcher/CleanOrphanedLocationsLiveTest.java
```diff
@@ -0,0 +1,257 @@
...
+        for (String location: allReferencedLocationsFoundByTransformer) {
+            notReferencedLocationsDueToTransormer.remove(location);
+        }
+
+        Asserts.assertEquals(notReferencedLocationsDueToTransormer, orphanedLocations);
+        Asserts.assertEquals(locationsNotToKeepDueToTransormer.keySet(), orphanedLocations);
+
+
+        BrooklynMementoRawData transformedMemento = transformer.transform(mementoRawData);
+        Asserts.assertFalse(EqualsBuilder.reflectionEquals(mementoRawData, transformedMemento));
```
bostko: @iyovcheva you can use `CleanOrphanedLocationsTest#assertTransformDeletes` and `CleanOrphanedLocationsTest#assertTransformIsNoop`
I'll try to PR against your PR with so we can minify the size of methods in `CleanOrphanedLocationsLiveTest`

server-cli/src/main/java/org/apache/brooklyn/cli/Main.java
```diff
@@ -827,6 +827,82 @@ public ToStringHelper string() {
...
+                description = "The directory to copy persistence data to without orphaned locations")
+        public String destinationDir;
+
+        @Option(name = { "--destinationLocation" }, title = "persistence location",
+                description = "The location spec for an object store to copy data to")
+        public String destinationLocation;
+
+        @Option(name = { "--transformations" }, title = "transformations",
+                description = "local transformations file, to be applied to the copy of the data before uploading it")
+        public String transformations;
```
aledsage: The `transformations` field isn't used. We should delete it.

server-cli/src/main/java/org/apache/brooklyn/cli/Main.java
```diff
@@ -827,6 +827,82 @@ public ToStringHelper string() {
...
+            checkNotNull(destinationDir, "orphanedReferencesTmpDir");
+
+            BrooklynLauncher launcher;
+            failIfArguments();
+
+            try {
+                log.info("Retrieving and copying persisted state to " + destinationDir + " without orphaned locations");
+
+
+                PersistMode persistMode = PersistMode.AUTO;
```
aledsage: I wonder if we should use `PersistMode.REBIND` - i.e. fail if there is no persisted state in the given directory.

server-cli/src/main/java/org/apache/brooklyn/cli/Main.java
```diff
@@ -931,6 +1007,7 @@ public ToStringHelper string() {
                         HelpCommand.class,
                         cliInfoCommand(),
                         GeneratePasswordCommand.class,
+                        CleanOrphanedLocationsCommand.class,
```
aledsage: This should now be called `CleanOrphanedStateCommand`.

server-cli/src/main/java/org/apache/brooklyn/cli/Main.java
```diff
@@ -827,6 +827,82 @@ public ToStringHelper string() {
         }
     }
 
+    @Command(name = "clean-orphaned-locations", description = "Removes existing orphaned locations")
```
aledsage: How about name "clean-orphaned-state" and description "Removes existing orphaned persisted state (e.g. locations, policies, enrichers and feeds)".

server-cli/src/main/java/org/apache/brooklyn/cli/Main.java
```diff
@@ -827,6 +827,82 @@ public ToStringHelper string() {
...
+    public static class CleanOrphanedLocationsCommand extends BrooklynCommandCollectingArgs {
```
aledsage: Long term, it would be good to have a dry run mode. But not in this PR.

server-cli/src/main/java/org/apache/brooklyn/cli/Main.java
```diff
@@ -827,6 +827,82 @@ public ToStringHelper string() {
...
+                description = "The location spec for an object store to copy data to")
+        public String destinationLocation;
+
+        @Option(name = { "--transformations" }, title = "transformations",
+                description = "local transformations file, to be applied to the copy of the data before uploading it")
+        public String transformations;
+
+        @Override
+        public Void call() throws Exception {
+            checkNotNull(destinationDir, "orphanedReferencesTmpDir");
```
aledsage: Change "orphanedReferencesTmpDir" to "destinationDir" - that will be the NullPointerException message.

aledsage:
Merging now. I'll address my most recent comments in a new PR later this evening.

PR #265, opened 2016-07-18
Extend XmlUtil#xpath signature

bostko:


bostko:
PR Included in https://github.com/apache/brooklyn-server/pull/264 so closing this.

aledsage:
@bostko worth adding unit test(s) when making changes like this They should be easy to write.

PR #266, opened 2016-07-18
BROOKLYN-264: refactor wait-for-provision

aledsage:
We now mark the internal state as provisioning entirely in 
MachineLifecycleEffectorTasks, and that is also where we wait for it
on stop(). This also allows us to test it as a unit test, rather than
needing a jclouds live test.

@bostko can you please review?

This builds on https://github.com/apache/brooklyn-server/pull/211, which is now merged.


software/base/src/main/java/org/apache/brooklyn/entity/software/base/lifecycle/MachineLifecycleEffectorTasks.java
```diff
@@ -379,20 +418,30 @@ public MachineLocation call() throws Exception {
             if (!(location instanceof LocalhostMachineProvisioningLocation))
                 log.info("Starting {}, obtaining a new location instance in {} with ports {}", new Object[]{entity(), location, flags.get("inboundPorts")});
             entity().sensors().set(SoftwareProcess.PROVISIONING_LOCATION, location);
+            Transition expectedState = entity().sensors().get(Attributes.SERVICE_STATE_EXPECTED);
+            
+            // BROOKLYN-263: see corresponding code in doStop()
+            if (expectedState != null && (expectedState.getState() == Lifecycle.STOPPING || expectedState.getState() == Lifecycle.STOPPED)) {
```
bostko: Good check, but should it be in this PR?

software/base/src/main/java/org/apache/brooklyn/entity/software/base/lifecycle/MachineLifecycleEffectorTasks.java
```diff
@@ -444,9 +493,6 @@ public void run() {
                 }
             }
             entity().addLocations(ImmutableList.of((Location) machine));
-            if (entity().getAttribute(Attributes.JCLOUDS_PROVISIONING_RUNNING) != null) {
```
bostko: Why you do not call
```
entity().sensors().remove(PROVISIONING_TASK_STATE)
entity().sensors().remove(PROVISIONED_MACHINE)
```
Otherwise these sensors will stay after machine provisioning completes.

software/base/src/test/java/org/apache/brooklyn/entity/ExpungingJcloudsLocationLiveTest.java
```diff
@@ -1,186 +0,0 @@
...
-        // Also removes scriptHeader (e.g. if doing `. ~/.bashrc` and `. ~/.profile`, then that can cause "stdin: is not a tty")
-        brooklynProperties.remove("brooklyn.ssh.config.scriptHeader");
-
-        mgmt = new LocalManagementContextForTests(brooklynProperties);
-        super.setUp();
-    }
-
-    @Test
-    public void verifyExpungingMockedEntityIsQuick() throws Exception {
-        final EmptySoftwareProcess emptySoftwareProcess = app.createAndManageChild(EntitySpec.create(EmptySoftwareProcess.class));
```
bostko: This is still a valid test which verifies that there is no bug in the stop functionality.
If it the waitForSshMachine location takes 5 minutes or more, then the stop code can make executing unit tests to last for hours.

software/base/src/main/java/org/apache/brooklyn/entity/software/base/lifecycle/MachineLifecycleEffectorTasks.java
```diff
@@ -379,20 +418,30 @@ public MachineLocation call() throws Exception {
...
+            entity().sensors().set(PROVISIONING_TASK_STATE, ProvisioningTaskState.RUNNING);
+            
             MachineLocation machine;
             try {
                 machine = Tasks.withBlockingDetails("Provisioning machine in " + location, new ObtainLocationTask(location, flags));
-                if (machine == null)
-                    throw new NoMachinesAvailableException("Failed to obtain machine in " + location.toString());
-            } catch (Exception e) {
-                throw Exceptions.propagate(e);
+                entity().sensors().set(PROVISIONED_MACHINE, machine);
```
bostko: I was thinking about this option but doesn't feel like the best option.
The downside I see is that someone could start using this sensor instead of `getLocations()`.

software/base/src/main/java/org/apache/brooklyn/entity/software/base/lifecycle/MachineLifecycleEffectorTasks.java
aledsage: I think it's fine to be in this PR. It relates to concurrent calls to start and stop.

software/base/src/main/java/org/apache/brooklyn/entity/software/base/lifecycle/MachineLifecycleEffectorTasks.java
aledsage: I'm willing to take that risk - the description says "internal" etc. The problem is that otherwise there is a big gap between when provisioning returns and when we add the machine to the location. I worry that start() will not always go through all those steps if there has been a concurrent call to stop. It might skip executing those subsequent tasks. With the change I made, we set DONE and the machine in a finally block so it removes that risk.

software/base/src/test/java/org/apache/brooklyn/entity/ExpungingJcloudsLocationLiveTest.java
aledsage: Good point - my SoftwareProcessStopsDuringStartTest.testSequentialStartThenStop doesn't time how long the call took.

software/base/src/main/java/org/apache/brooklyn/entity/software/base/lifecycle/MachineLifecycleEffectorTasks.java
aledsage: I'll rename the sensors to include "internal" in their names, as well as in the description.

The reason I didn't like this approach was that if `obtain()` failed, we would wait for the max time before stop continued - there was no way for it to terminate early. I'll add a test for that.

aledsage: One big reason that someone might call stop() while it's still starting is they realise their provisioning parameters were wrong, so want to abort. Therefore continuing to support that use-case seems important.

software/base/src/test/java/org/apache/brooklyn/entity/software/base/SoftwareProcessStopsDuringStartJcloudsLiveTest.java
```diff
@@ -0,0 +1,231 @@
...
+        EntityAsserts.assertEntityHealthy(entity);
+        assertEquals(entity.getAttribute(MachineLifecycleEffectorTasks.INTERNAL_PROVISIONING_TASK_STATE), MachineLifecycleEffectorTasks.ProvisioningTaskState.DONE);
+        assertEquals(entity.getAttribute(MachineLifecycleEffectorTasks.INTERNAL_PROVISIONED_MACHINE), Machines.findUniqueMachineLocation(entity.getLocations(), SshMachineLocation.class).get());
+
+        executeInLimitedTime(new Callable<Void>() {
+            public Void call() {
+                Entities.destroy(app);
+                return null;
+            }
+        }, Asserts.DEFAULT_LONG_TIMEOUT.toMilliseconds(), TimeUnit.MILLISECONDS);
```
bostko: Isn't this too big value?

PR #267, opened 2016-07-19
Fix for BROOKLYN-314

aledsage:
Try-catch in getOptionalNode/getOptionalImage (e.g. if called by user
in rebind code, then don’t fail outright if cloud throws exception


nakomis:
Looks good, and pretty straightforward. Could you add a test for the rebind issue described in BROOKLYN-314

aledsage:
@nakomis not particularly easily is the answer, for writing a test to recreate BROOKLYN-314! Would need to stub-out the a JcloudsLocation to simulate the failure, and to write a custom entity and an associated driver that calls `machine.getOsDetails()` in its constructor.

I'll have more of a look at doing that, but am going to merge this now.

PR #269, opened 2016-07-21
Improve test entity asserts (notEqual, actual state in failure)

neykov:
* Adds a `notEqual` assert type
* Includes the actual state in assert failures


grkvlt:
LGTM, merging

PR #270, opened 2016-07-22
State transformations: support deletions

aledsage:


nakomis:
Tested catalog deletions will YAML similar to following:

aledsage:
Thanks @nakomis - merging.

PR #271, opened 2016-07-22
Move karaf distribution to brooklyn-dist.

geomacy:
This PR is the first of three that together move the karaf distribution to brooklyn-dist and separate out the software from brooklyn-library into separate catalog bom and feature files.


karaf/features/src/main/feature/feature.xml
```diff
@@ -276,60 +276,6 @@
         <feature>jclouds-api-filesystem</feature>
 
         <bundle start-level="85">mvn:org.apache.brooklyn/brooklyn-locations-jclouds/${project.version}</bundle>
-        <feature>brooklyn-software-winrm</feature>
```
neykov: `winrm` is part of core. Don't see where do you include the feature back (in -library and -dist PRs)?

neykov:
LGTM (apart from the missing winrm dependency)

karaf/features/src/main/feature/feature.xml
geomacy: I had meant to move it to brooklyn-library and then realised it didn't belong there, must have missed putting it back here.  Will fix.

geomacy:
Review comments applied, squashed and pushed.

duncangrant:
I've tested this - all seems to work

PR #272, opened 2016-07-22
orphaned-state deletion: keep PortForwardManagers

aledsage:


core/src/main/java/org/apache/brooklyn/core/mgmt/rebind/transformer/impl/DeleteOrphanedStateTransformer.java
```diff
@@ -91,93 +107,219 @@ public BrooklynMementoRawData transform(BrooklynMementoRawData input) {
...
+                    feeds.add(id);
+                }
+            }
+            
+            // Initial pass of locations (from the roots of reachability)
+            for (Map.Entry<String, String> entry : input.getLocations().entrySet()) {
+                String id = entry.getKey();
+                String locationState = entry.getValue();
+                if (locationState.contains("PortForwardManager")) {
+                    locations.add(id);
```
iyovcheva: Will we keep all locations of that type? Can't we search by <pfm> reference?

core/src/main/java/org/apache/brooklyn/core/mgmt/rebind/transformer/impl/DeleteOrphanedStateTransformer.java
```diff
@@ -64,12 +68,24 @@ protected DeleteOrphanedStateTransformer(DeleteOrphanedStateTransformer.Builder
...
+     * will we delete it.
+     */
     @Override
     public BrooklynMementoRawData transform(BrooklynMementoRawData input) {
-        Map<String, String> locationsToKeep = findLocationsToKeep(input);
-        Map<String, String> enrichersToKeep = copyRetainingKeys(input.getEnrichers(), findAllReferencedEnrichers(input));
-        Map<String, String> policiesToKeep = copyRetainingKeys(input.getPolicies(), findAllReferencedPolicies(input));
-        Map<String, String> feedsToKeep = copyRetainingKeys(input.getFeeds(), findAllReferencedFeeds(input));
+        ReferencedState stateReferencedFromXpath = new ReachabilityXpathInspector().inspect(input);
+        ReferencedState stateToKeepFromGrep = new ReachabilityGrepInspector().inspect(input);
```
bostko: Isn't it better if `ReachabilityGrepInspector` is after `locationsToKeep`, `enrichersToKeep`, `policiesToKeep`, `feedsToKeep` are produced.
This way it could also grep inside those for any references.

PR #275, opened 2016-07-27
Resolve all environment variables

grkvlt:
Use the config resolver on the complete environment map before serializing to strings in SSH command sensor and effector.


neykov:
LGTM, merging.

PR #276, opened 2016-07-28
Effector for opening inbound ports in security group

bostko:
Caveat: the effector will return error when called before machine location is provisioned

### DIFF
--- a/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/NetworkingEffectors.java
+++ b/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/NetworkingEffectors.java
@@ -1,0 +1,100 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.brooklyn.location.jclouds.networking;
+
+import com.google.common.base.Optional;
+import com.google.common.base.Preconditions;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Range;
+import com.google.common.reflect.TypeToken;
+import org.apache.brooklyn.api.effector.Effector;
+import org.apache.brooklyn.api.location.Location;
+import org.apache.brooklyn.config.ConfigKey;
+import org.apache.brooklyn.core.config.ConfigKeys;
+import org.apache.brooklyn.core.effector.EffectorBody;
+import org.apache.brooklyn.core.effector.Effectors;
+import org.apache.brooklyn.location.jclouds.JcloudsMachineLocation;
+import org.apache.brooklyn.util.collections.MutableList;
+import org.apache.brooklyn.util.core.config.ConfigBag;
+import org.apache.brooklyn.util.net.Cidr;
+import org.apache.brooklyn.util.net.Networking;
+import org.jclouds.net.domain.IpPermission;
+import org.jclouds.net.domain.IpProtocol;
+
+import java.util.List;
+
+import static com.google.common.base.Predicates.instanceOf;
+import static com.google.common.collect.Iterables.tryFind;
+import static org.apache.brooklyn.core.location.Locations.getLocationsCheckingAncestors;
+
+public class NetworkingEffectors {
+    // Intentionally not use CloudLocationConfig.INBOUND_PORTS to make richer syntax and rename it to differ it from the first in a ConfigBag
+    public static final ConfigKey<List<String>> INBOUND_PORTS_LIST = ConfigKeys.newConfigKey(new TypeToken<List<String>>() {}, "inbound.ports.list",
+            "Ports to open from the effector", ImmutableList.<String>of());
+    public static final ConfigKey<IpProtocol> INBOUND_PORTS_LIST_PROTOCOL = ConfigKeys.newConfigKey(new TypeToken<IpProtocol>() {}, "inbound.ports.list.protocol",
+            "Protocol for ports to open. Possible values: TCP, UDP, ICMP, ALL.", IpProtocol.TCP);
+
+    public static final ConfigKey<JcloudsMachineLocation> JCLOUDS_MACHINE_LOCATIN = ConfigKeys.newConfigKey(JcloudsMachineLocation.class, "jcloudsMachineLocation");
+
+    @SuppressWarnings("unchecked")
+    public static final Effector<Iterable<IpPermission>> OPEN_INBOUND_PORTS_IN_SECURITY_GROUP_EFFECTOR = (Effector<Iterable<IpPermission>>)(Effector<?>)Effectors.effector(Iterable.class, "openPortsInSecurityGroup")
+                .parameter(INBOUND_PORTS_LIST)
+                .parameter(INBOUND_PORTS_LIST_PROTOCOL)
+                .description("Open ports in Cloud Security Group. If called before machine location is provisioned, it will fail.")
+                .impl(new OpenPortsInSecurityGroupBody())
+                .build();
+
+    @SuppressWarnings("rawtypes")
+    private static class OpenPortsInSecurityGroupBody extends EffectorBody<Iterable> {
+        @Override
+        public Iterable<IpPermission> call(ConfigBag parameters) {
+            List<String> rawPortRules = parameters.get(INBOUND_PORTS_LIST);
+            IpProtocol ipProtocol = parameters.get(INBOUND_PORTS_LIST_PROTOCOL);
+            JcloudsMachineLocation jcloudsMachineLocation = parameters.get(JCLOUDS_MACHINE_LOCATIN);
+            Preconditions.checkNotNull(ipProtocol, INBOUND_PORTS_LIST_PROTOCOL.getName() + " cannot be null");
+            Preconditions.checkNotNull(rawPortRules, INBOUND_PORTS_LIST.getName() + " cannot be null");
+            MutableList.Builder<IpPermission> ipPermissionsBuilder = MutableList.builder();
+            for (Range<Integer> portRule : Networking.portRulesToRanges(rawPortRules).asRanges()) {
+                ipPermissionsBuilder.add(
+                        IpPermission.builder()
+                                .ipProtocol(ipProtocol)
+                                .fromPort(portRule.lowerEndpoint())
+                                .toPort(portRule.upperEndpoint())
+                                .cidrBlock(Cidr.UNIVERSAL.toString())
+                                .build());
+            }
+            JcloudsLocationSecurityGroupCustomizer customizer = JcloudsLocationSecurityGroupCustomizer.getInstance(entity());
+
+            if (jcloudsMachineLocation == null) {
+                Optional<Location> jcloudsMachineLocationOptional = tryFind(
+                        (Iterable<Location>) getLocationsCheckingAncestors(null, entity()),
+                        instanceOf(JcloudsMachineLocation.class));
+                if (!jcloudsMachineLocationOptional.isPresent()) {
+                    throw new IllegalArgumentException("Tried to execute open ports effector on an entity with no JcloudsMachineLocation");
+                } else {
+                    jcloudsMachineLocation = (JcloudsMachineLocation)jcloudsMachineLocationOptional.get();
+                }
+            }
+            Iterable<IpPermission> ipPermissionsToAdd = ipPermissionsBuilder.build();
+            customizer.addPermissionsToLocation(jcloudsMachineLocation, ipPermissionsToAdd);
+            return ipPermissionsToAdd;
+        }
+    }
+
+}
--- a/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/NetworkingEffectors.java
+++ b/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/NetworkingEffectors.java
@@ -18,6 +18,7 @@
  */
 package org.apache.brooklyn.location.jclouds.networking;
 
+import com.google.common.annotations.Beta;
 import com.google.common.base.Optional;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableList;
@@ -52,6 +53,7 @@ public class NetworkingEffectors {
 
     public static final ConfigKey<JcloudsMachineLocation> JCLOUDS_MACHINE_LOCATIN = ConfigKeys.newConfigKey(JcloudsMachineLocation.class, "jcloudsMachineLocation");
 
+    @Beta
     @SuppressWarnings("unchecked")
     public static final Effector<Iterable<IpPermission>> OPEN_INBOUND_PORTS_IN_SECURITY_GROUP_EFFECTOR = (Effector<Iterable<IpPermission>>)(Effector<?>)Effectors.effector(Iterable.class, "openPortsInSecurityGroup")
                 .parameter(INBOUND_PORTS_LIST)
--- a/software/base/pom.xml
+++ b/software/base/pom.xml
@@ -70,6 +70,11 @@
             <artifactId>brooklyn-policy</artifactId>
             <version>${project.version}</version>
         </dependency>
+        <dependency>
+            <groupId>org.apache.brooklyn</groupId>
+            <artifactId>brooklyn-locations-jclouds</artifactId>
+            <version>${project.version}</version>
+        </dependency>
 
         <dependency>
             <groupId>org.freemarker</groupId>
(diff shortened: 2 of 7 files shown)


core/src/main/java/org/apache/brooklyn/util/core/BrooklynNetworkUtils.java
```diff
@@ -39,4 +43,22 @@ public static InetAddress getLocalhostInetAddress() {
                 Networking.getLocalHost()), InetAddress.class);
     }
 
+    // TODO it does not add adjacent intervals: {[22, 22], [23, 23]} is not merged to {[22, 23]}
+    public static RangeSet<Integer> portRulesToRanges(Collection<String> portRules) {
+        RangeSet<Integer> result = TreeRangeSet.create();
+        for (String portRule : portRules) {
+            if (portRule.contains("-")) {
+                String[] fromTo = portRule.split("-");
+                assert fromTo.length == 2;
```
nakomis: Assertions are disabled by default, use `Preconditions.checkState`

aledsage: Don't use `assert` for validating user input. Instead we'd want to throw a nicer `IllegalArgumentException` to tell the caller that they passed in the wrong input.

Only use `assert` when it's about the internal invariants of a class (e.g. validating args in private methods, etc).

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/NetworkingEffectors.java
```diff
@@ -0,0 +1,80 @@
...
+    private static class OpenPortsInSecurityGroupBody extends EffectorBody<Void> {
+        @Override
+        public Void call(ConfigBag parameters) {
+            List<String> rawPortRules = parameters.get(INBOUND_PORTS_EFFECTOR);
+            Preconditions.checkNotNull(rawPortRules, "ports cannot be null");
+            MutableList.Builder<IpPermission> ipPermissionsBuilder = MutableList.builder();
+            for (Range<Integer> portRule : BrooklynNetworkUtils.portRulesToRanges(rawPortRules).asRanges()) {
+                ipPermissionsBuilder.add(
+                        IpPermission.builder()
+                                .ipProtocol(IpProtocol.TCP)
```
aledsage: We want this to support both UPD and TCP, so the effector to indicate which is required.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/NetworkingEffectors.java
```diff
@@ -0,0 +1,80 @@
...
+
+    public static Effector<Void> openPortsInSecurityGroupEffector() {
+        return Effectors.effector(Void.class, "openPortsInSecurityGroup")
+                .parameter(INBOUND_PORTS_EFFECTOR)
+                .description("Open ports in Cloud Security Group")
+                .impl(new OpenPortsInSecurityGroupBody())
+                .build();
+    }
+
+    private static class OpenPortsInSecurityGroupBody extends EffectorBody<Void> {
```
aledsage: We'd also like this to work in other clouds, such as downsteram projects that use vCloudDirector via the brooklyn-networking `SubnetTier`. I need to think about that more, for how we structure that.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/NetworkingEffectors.java
```diff
@@ -0,0 +1,80 @@
...
+import org.apache.brooklyn.util.core.BrooklynNetworkUtils;
+import org.apache.brooklyn.util.core.config.ConfigBag;
+import org.apache.brooklyn.util.net.Cidr;
+import org.jclouds.net.domain.IpPermission;
+import org.jclouds.net.domain.IpProtocol;
+
+import java.util.List;
+
+public class NetworkingEffectors {
+    public static final ConfigKey<List<String>> INBOUND_PORTS_EFFECTOR = ConfigKeys.newConfigKey(new TypeToken<List<String>>() {}, "inbound.ports.effector",
```
aledsage: "inbound.ports.effector" seems like a strange name. This isn't an effector, I believe: it's a parameter for the effector, yes?

software/base/src/main/java/org/apache/brooklyn/entity/software/base/SoftwareProcess.java
```diff
@@ -147,6 +147,8 @@
     @SetFromFlag("runDir")
     AttributeSensorAndConfigKey<String,String> RUN_DIR = BrooklynConfigKeys.RUN_DIR;
 
+    ConfigKey<Boolean> ADD_OPEN_PORTS_EFFECTOR = ConfigKeys.newBooleanConfigKey("effector.add.openPorts", "Flag which adds effector for opening ports through Cloud security groups");
```
aledsage: I prefer giving an explicit default, rather than null.

(But I do see you handle null in SoftwareProcessImpl.init, so the code does work).

aledsage:
It would be good to have test(s). Have you thought about how to write those?

aledsage:
I've merged #298 which has your first commit. Can you rebase against master please?

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/NetworkingEffectors.java
```diff
@@ -0,0 +1,88 @@
...
+
+public class NetworkingEffectors {
+    // Intentionally not use CloudLocationConfig.INBOUND_PORTS to make richer syntax and rename it to differ it from the first in a ConfigBag
+    public static final ConfigKey<List<String>> INBOUND_PORTS_LIST = ConfigKeys.newConfigKey(new TypeToken<List<String>>() {}, "inbound.ports.list",
+            "Ports to open from the effector", ImmutableList.<String>of());
+    public static final ConfigKey<IpProtocol> INBOUND_PORTS_LIST_PROTOCOL = ConfigKeys.newConfigKey(new TypeToken<IpProtocol>() {}, "inbound.ports.list.protocol",
+            "Protocol for ports to open. Possible values: TCP, UDP, ICMP, ALL.", IpProtocol.TCP);
+
+    public static Effector<Iterable<IpPermission>> openPortsInSecurityGroupEffector() {
+        return (Effector<Iterable<IpPermission>>)Effectors.effector(Iterable.class, "openPortsInSecurityGroup")
```
nakomis: This fails to build in IntelliJ, replacing with `return (Effector<Iterable<IpPermission>>)Effectors.effector((new TypeToken<Iterable<IpPermission>>(){}).getRawType(), "openPortsInSecurityGroup")` would keep IDEs happy, but it builds in Maven, so no strong feelings

nakomis: Or even better, overloading `Effectors.effector` to take a TypeToken :-)

rdowner:
retest this please

rdowner:
(above was a command to jenkins, not to the submitter!)
(but it looks like it didn't work anyway)

nakomis:
Manually tested successfully with the following YAML:

```
name: Simple Netcat Server & Effector
location: aws-ec2:eu-west-1
services:
  name: Simple Netcat Server
  launch.command: |
    echo hello | nc -l 4321 &
    echo $! > $PID_FILE
  brooklyn.config:
    effector.add.openPorts: true
```

I'm getting a 404 from Jenkins when I try to look a the build failure - I think Jenkins only keeps the logs for so long. It also builds locally - @bostko can you do a force push on your branch to re-trigger the jenkins build

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/NetworkingEffectors.java
```diff
@@ -0,0 +1,90 @@
...
+
+    private interface EffectorResult extends Iterable<IpPermission> {}
+
+    private static class OpenPortsInSecurityGroupBody extends EffectorBody<Iterable<IpPermission>> {
+        @Override
+        public Iterable<IpPermission> call(ConfigBag parameters) {
+            List<String> rawPortRules = parameters.get(INBOUND_PORTS_LIST);
+            IpProtocol ipProtocol = parameters.get(INBOUND_PORTS_LIST_PROTOCOL);
+            Preconditions.checkNotNull(ipProtocol);
+            Preconditions.checkNotNull(rawPortRules, "ports cannot be null");
```
neykov: use the parameter name `inbound.ports.list`

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/NetworkingEffectors.java
```diff
@@ -0,0 +1,90 @@
...
+                ipPermissionsBuilder.add(
+                        IpPermission.builder()
+                                .ipProtocol(ipProtocol)
+                                .fromPort(portRule.lowerEndpoint())
+                                .toPort(portRule.upperEndpoint())
+                                .cidrBlock(Cidr.UNIVERSAL.toString())
+                                .build());
+            }
+            JcloudsLocationSecurityGroupCustomizer customizer = JcloudsLocationSecurityGroupCustomizer.getInstance(entity());
+            Optional<Location> jcloudsMachineLocation = Iterables.tryFind(entity().getLocations(), Predicates.instanceOf(JcloudsMachineLocation.class));
```
neykov: Use `Locations.getLocationsCheckingAncestors(locations, entity);` instead - the location could be declared on parent entities (SameServerEntity).

software/base/pom.xml
```diff
@@ -116,7 +116,6 @@
             <groupId>org.apache.brooklyn</groupId>
             <artifactId>brooklyn-locations-jclouds</artifactId>
             <version>${project.version}</version>
-            <scope>test</scope>
```
neykov: Why remove it from here, but leave all following dependencies the same. If you need a (non-test) dependency on `brooklyn-locations-jclouds` add a new one instead.

software/base/src/main/java/org/apache/brooklyn/entity/software/base/SoftwareProcess.java
```diff
@@ -147,6 +147,9 @@
     @SetFromFlag("runDir")
     AttributeSensorAndConfigKey<String,String> RUN_DIR = BrooklynConfigKeys.RUN_DIR;
 
+    ConfigKey<Boolean> ADD_OPEN_PORTS_EFFECTOR = ConfigKeys.newBooleanConfigKey("effector.add.openPorts",
```
neykov: You are using "inbound" and "open" interchangeably. Better stick to one of them for consistency.

neykov: I think it's better if we use `AddEffector` interceptor instead.

software/base/src/test/java/org/apache/brooklyn/entity/software/base/location/NetworkingEffectorsLiveTests.java
```diff
@@ -0,0 +1,107 @@
...
+        Optional<Location> jcloudsMachineLocation = Iterables.tryFind(emptySoftwareProcess.getLocations(), Predicates.instanceOf(JcloudsMachineLocation.class));
+        if (!jcloudsMachineLocation.isPresent()) {
+            throw new IllegalArgumentException("Tried to execute open ports effector on an entity with no JcloudsMachineLocation");
+        }
+        ComputeService computeService = ((JcloudsMachineLocation)jcloudsMachineLocation.get()).getParent().getComputeService();
+        String nodeId = ((JcloudsMachineLocation)jcloudsMachineLocation.get()).getNode().getId();
+        final SecurityGroupExtension securityApi = computeService.getSecurityGroupExtension().get();
+
+        Effector<Iterable<IpPermission>> openPortsInSecurityGroup = (Effector<Iterable<IpPermission>>)EffectorUtils.findEffectorDeclared(emptySoftwareProcess, "openPortsInSecurityGroup").get();
+        Task<Iterable<IpPermission>> task = EffectorUtils.invokeEffectorAsync(emptySoftwareProcess, openPortsInSecurityGroup, ImmutableMap.of("inbound.ports.list", "234,324,550-1050"));
```
neykov: Could use the Java config key instead of name.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/NetworkingEffectors.java
```diff
@@ -0,0 +1,90 @@
...
+import java.util.List;
+
+public class NetworkingEffectors {
+    // Intentionally not use CloudLocationConfig.INBOUND_PORTS to make richer syntax and rename it to differ it from the first in a ConfigBag
+    public static final ConfigKey<List<String>> INBOUND_PORTS_LIST = ConfigKeys.newConfigKey(new TypeToken<List<String>>() {}, "inbound.ports.list",
+            "Ports to open from the effector", ImmutableList.<String>of());
+    public static final ConfigKey<IpProtocol> INBOUND_PORTS_LIST_PROTOCOL = ConfigKeys.newConfigKey(new TypeToken<IpProtocol>() {}, "inbound.ports.list.protocol",
+            "Protocol for ports to open. Possible values: TCP, UDP, ICMP, ALL.", IpProtocol.TCP);
+
+    public static Effector<Iterable<IpPermission>> openPortsInSecurityGroupEffector() {
```
neykov: Turn this into a constant to make it easier to call the effector in Java.

neykov:
LGTM, minor comments only.

software/base/pom.xml
bostko: I need a non-test dependency on `brooklyn-locations-jclouds`. Tests also need code from `brooklyn-locations-jclouds` so adding a simple dependency with default compile scope is enough I think?
@neykov WDYT?

bostko:
I added a new commit with `InboundPortsJcloudsLocationCustomizer`:
This is the yaml I tested with:
```yaml
location:
  AWS Oregon:
    hardwareId: t1.micro
    customizers:
    - $brooklyn:object:
        type: org.apache.brooklyn.location.jclouds.networking.InboundPortsJcloudsLocationCustomizer
        object.fields: {inboundPortsList: "23,55", inboundPortsListProtocol: UDP}
    - $brooklyn:object:
        type: org.apache.brooklyn.location.jclouds.networking.InboundPortsJcloudsLocationCustomizer
        object.fields: {inboundPortsList: "200-300,500", inboundPortsListProtocol: TCP}
services:
```

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/NetworkingEffectors.java
bostko: @neykov this should be fixed now.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/NetworkingEffectors.java
bostko: wdyt about using `tryFind((Iterable<Location>) getLocationsCheckingAncestors(null, entity()),                       instanceOf(JcloudsMachineLocation.class))`

software/base/src/main/java/org/apache/brooklyn/entity/software/base/SoftwareProcess.java
bostko: Why to use AddEffector? Isn't it interchangeable with  `Effectors.effector`?

software/base/src/test/java/org/apache/brooklyn/entity/software/base/location/NetworkingEffectorsLiveTests.java
bostko: Fixed

software/base/src/main/java/org/apache/brooklyn/entity/software/base/SoftwareProcess.java
neykov: My idea was that the class `AddEffector` is a generic way to attach effectors to entities in yaml. Then no need for an ad-hoc flag to attach the effector.
Looking at the implementation though it's not straightforward to use it in YAML currently so ignore.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/NetworkingEffectors.java
```diff
@@ -0,0 +1,103 @@
...
+                    jcloudsMachineLocation = (JcloudsMachineLocation)jcloudsMachineLocationOptional.get();
+                }
+            }
+            Iterable<IpPermission> ipPermissionsToAdd = ipPermissionsBuilder.build();
+            customizer.addPermissionsToLocation(jcloudsMachineLocation, ipPermissionsToAdd);
+            return ipPermissionsToAdd;
+        }
+    }
+
+    public static void main(String[] args) {
```
aledsage: Suggest we delete this `main` method.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/NetworkingEffectors.java
```diff
@@ -0,0 +1,100 @@
...
+import org.jclouds.net.domain.IpPermission;
+import org.jclouds.net.domain.IpProtocol;
+
+import java.util.List;
+
+import static com.google.common.base.Predicates.instanceOf;
+import static com.google.common.collect.Iterables.tryFind;
+import static org.apache.brooklyn.core.location.Locations.getLocationsCheckingAncestors;
+
+public class NetworkingEffectors {
```
aledsage: Worth including `@beta`, I think.

I was imagining that we could tell it to open certain TCP ports and certain UDP ports in a single effector call. But I see that your approach works (one can make multiple effector calls), and is probably simpler than having a map from protocol to a list of port ranges (or whatever scheme one comes up with). So I'm happy with this.

aledsage:
@bostko looks good to me. I'll merge now. Thanks for the amazingly fast update to add `@Beta`, even late at night!

What are your thoughts on `openIptables`? It could well be that the port(s) will still be locked down in the iptables rules. Do you think that would be a separate effector (or script or whatever) to open those?

bostko:
In terms of user prospective I think it should be optional and may be within the same effector.

PR #277, opened 2016-07-28
Fix TestCases, and add TestSensor.abortConditions

aledsage:


test-framework/src/main/java/org/apache/brooklyn/test/framework/TargetableTestComponentImpl.java
```diff
@@ -116,4 +125,14 @@ private static Entity getTargetById(final ExecutionContext executionContext, fin
...
     }
+
+    protected void setUpAndRunState(boolean up, Lifecycle status) {
+        if (up) {
+            sensors().set(SERVICE_UP, up);
+            setExpectedState(this, status);
+        } else {
+            setExpectedState(this, status);
+            sensors().set(SERVICE_UP, up);
+        }
```
neykov: As you are updating this anyway would be nice to make the `service.isUp` + `service.state` managed through the notUp + problems maps, instead of directly. 
Current approach is valid as well so happy to go this way if you think it's out of scope.

neykov:
After a very quick look I think that's very useful and something I felt the need for. When previously thinking about the problem the approach I came up with is having `validConditions` instead of `abortConditions`. This way you only define the expected states the blueprint will go through instead of any combination of invalid state. For `service.state` that would usually be `STARTING`, `RUNNING`, but where it becomes even more useful is asserting on any random sensor. For example resizing a cluster and asserting the new size - the expected sizes are the initial one and the end one. Using the `abortConditions` here would not be too practical.
Or may be both are useful - would be easier to use one or the other for different use cases.

test-framework/src/main/java/org/apache/brooklyn/test/framework/TargetableTestComponentImpl.java
aledsage: Agreed we should change that. Will it mean that app.start() can return when the service.isUp still equals false? I believe so - e.g. the `TestCase` just calls start sequentially on its children, without waiting for service.isUp.

If so, we'll also need to change our test assertions to "eventually".

I'll hold off doing that until another PR.

aledsage:
@neykov interesting discussion. The "validConditions" makes a lot of sense. I was imagining we'd list abortConditions as {on-fire, stopping, stopped, destroyed} (which is quite verbose compared to your `validConditions`).

I also like the idea of being able to assert about a different sensor. We can do that in the Java code's `DependentConfiguration` (unrelated, but I used that as inspiration). It's a bit more fiddly to wire in here.

I'm going to leave in the `abortConditions`, and think about adding `validConditions` as well (in a different PR). My one concern about `validConditions` is that it sounds like it could be a valid termination condition, rather than us keeping going until the assert succeeds.

test-framework/src/main/java/org/apache/brooklyn/test/framework/TargetableTestComponentImpl.java
neykov: We are waiting for `service.isUp` in `SoftwareProcess` so makes sense to do the same here for consistency.

aledsage:
To clarify what I meant by "validConditions sounds like it could be a valid termination condition"...

I mean that if someone is writing a test-case or is reading someone else's test-case, then "validConditions" might seem like it could be used as the condition that must be satisfied for the test to complete. So they might try to use that rather than "assertions". But I'm sure we can avoid that with suitable config descriptions. And I can't think of a better name than "validConditions" (without getting ridiculously technical like "validTransientConditions" or some such).

---
Build was previously failing because I'd missed the `AbortError.java` file from my commits! Fixed that now.

aledsage:
`testPersistenceVolumeFastNoTrigger` failed in jenkins PR build, but that is unrelated to this PR. I've added a comment to `testPersistenceVolumeFastNoTrigger` to explain what I think the problem is. I've increased the "persistPeriod" to make it more likely to pass.

aledsage:
Merging this now. I agree that `validConditions` is also worth having. We can add that later.

PR #278, opened 2016-07-28
rest api: entity/spec returns unquoted text/x-yaml

aledsage:
Previously it returned json:
* it wrapped it in quotes.
* it did strange things with \n: the result rendered as “\n” rather than actual new-line characters (though it was more complicated than that - an angular UI did recognise the “\n” as a single character).


m4rkmckenna:
LGTM / Tested

aledsage:
Thanks @m4rkmckenna - merging now.

PR #279, opened 2016-07-29
Allow provisioning.properties to reference resolvable config

grkvlt:
Enables constructions like the following:
```YAML
- type: vanilla-software-process
  brooklyn.config:
    provisioning.properties:
      minRam: $brooklyn:scopeRoot().config("entity.min.ram")
```

### DIFF
--- a/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
+++ b/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
@@ -28,7 +28,6 @@ import static org.jclouds.util.Throwables2.getFirstThrowableOfType;
 import java.io.ByteArrayOutputStream;
 import java.io.File;
 import java.io.IOException;
-import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.security.KeyPair;
@@ -42,6 +41,7 @@ import java.util.Map;
 import java.util.NoSuchElementException;
 import java.util.Set;
 import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Semaphore;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
@@ -64,6 +64,7 @@ import org.apache.brooklyn.config.ConfigKey.HasConfigKey;
 import org.apache.brooklyn.core.BrooklynVersion;
 import org.apache.brooklyn.core.config.ConfigUtils;
 import org.apache.brooklyn.core.config.Sanitizer;
+import org.apache.brooklyn.core.entity.EntityInternal;
 import org.apache.brooklyn.core.location.AbstractLocation;
 import org.apache.brooklyn.core.location.BasicMachineMetadata;
 import org.apache.brooklyn.core.location.LocationConfigKeys;
@@ -528,7 +529,7 @@ public class JcloudsLocation extends AbstractCloudMachineProvisioningLocation im
 
     public ComputeService getComputeService(ConfigBag config) {
         ComputeServiceRegistry registry = getConfig(COMPUTE_SERVICE_REGISTRY);
-        return registry.findComputeService(ResolvingConfigBag.newInstanceExtending(getManagementContext(), config), true);
+        return registry.findComputeService(config, true);
     }
 
     /** @deprecated since 0.7.0 use {@link #listMachines()} */ @Deprecated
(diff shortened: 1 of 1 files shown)


aledsage:
@grkvlt test failures look plausibly related: failures in `JcloudsLocationTest`.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -527,7 +531,7 @@ public ComputeService getComputeService(Map<?,?> flags) {
 
     public ComputeService getComputeService(ConfigBag config) {
         ComputeServiceRegistry registry = getConfig(COMPUTE_SERVICE_REGISTRY);
-        return registry.findComputeService(ResolvingConfigBag.newInstanceExtending(getManagementContext(), config), true);
```
aledsage: Why are you removing the `ResolvingConfigBag`?

aledsage: This was added to fix "BROOKLYN-249: JcloudsLocation resolves external config" (see commit 99dd3ba4fc507988bdecd4e57c2b36cbb465b2a5)

aledsage:
The jenkins test failures look plausibly related: `JcloudsLocationTest` tests are failing.

@grkvlt please add test(s) for this. See the existing tests like those in `JcloudsLocationExternalConfigYamlTest`. Can you also confirm that those live tests work (they don't create any VMs, but they go through talking to the cloud for the list of images etc to create the `Template`).

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
grkvlt: By the time `getComputeService()` is called, I will already have set up the `ConfigBag` to be a resolving config, see `private ConfigBag getConfigBag(Map<?, ?> flags)` method below. The addition of a new `ResolvingConfigBag` was redundant. In essence I'm solving the same problem as **BROOKLYN-249** and also adding the ability to resolve config through the calling entity context, if it exists.

grkvlt:
@aledsage I've fixed the failing tests (there may not be a caller context, so do not rely on its existence) and will add some more to illustrate the use of resolvable entity config in a jclouds location context.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -49,6 +50,68 @@
 import javax.annotation.Nullable;
 import javax.xml.ws.WebServiceException;
 
+import org.apache.commons.lang3.ArrayUtils;
```
aledsage: Can we avoid re-ordering all the imports?

aledsage:
@grkvlt you said that you'd "add some more [tests] to illustrate the use of resolvable entity config in a jclouds location context". Is that something you plan to do for this PR? Do you know when you'll have time to do that?

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
grkvlt: But this way is more logically defensible!

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -603,21 +607,37 @@ public MachineLocation obtain(Map<?,?> flags, TemplateBuilder tb) throws NoMachi
         return obtain(MutableMap.builder().putAll(flags).put(TEMPLATE_BUILDER, tb).build());
     }
 
-    /** core method for obtaining a VM using jclouds;
+    private ConfigBag getConfigBag(Map<?, ?> flags) {
```
ahgittin: call this `getConfigBagExtendedWithFlagsResolved` ?

ahgittin:
I think this resolves and stores config too early.  We need to ensure the raw unresolved values are given to the `JcloudsSshMachineLocation` (and a few other places).  See https://github.com/apache/brooklyn-server/pull/194.

Can you check whether `ExternalConfigYamlTest` passes?  They are integration tests (probably shouldn't be!) but I think would illustrate why we can't just resolve everything.

The use case you introduce @grkvlt is useful.  I think we should handle this either by ensuring we resolve each access eg in `buildTemplate` or by maintaining two maps in the code path, `newItemConfigResolved` and `newItemConfigRaw` (instead of `setup`).

Also agree we want a test for this and please undo the huge mad import order change (under what ordering is the new one more sensible??).

Note that the semantics flip once a location (or any adjunct or entity) is managed, `BrooklynObject.config().get()` returns resolved values.  A more elegant but much bigger fix might be to create the `JcloudsSshMachineLocation` early, pass it unresolved config, and then for it to call its own config at which point resolving on access would be a sensible default.  This makes it more like entities (which tend to create the thing they represent, as opposed to the parent creating the thing first which is what `JcloudsLocation` is doing).

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
ahgittin: @grkvlt just came across another import rearrange you did in 33b0b1c879a9ed07bcdc3138893cd8017e211cfa

can you configure your import order as per
http://brooklyn.apache.org/v/latest/dev/env/ide/index.html#intellij-idea ?

grkvlt: I have done this for all my Brooklyn projects now, thanks for the pointer, @ahgittin

grkvlt: BTW, my ordering was to keep the `org.apache.brooklyn` and `io.brooklyn` and `brooklyn` imports at the bottom, with more generic imports above. But, whatever, am happy to use the standard, have changed my IDE config.

aledsage:
See https://issues.apache.org/jira/browse/BROOKLYN-349 for a description of the underlying problem.

Also see https://github.com/apache/brooklyn-server/pull/341 for an alternative fix and a test for that fix. In PR #341 the resolution is deferred until the value is actually requested. This was a bit more fiddly to implement, but more because it highlighted a problem where the `DslComponent`'s task lost the context of the entity that called it.

We'll only want one of this PR or #341 to be merged, I think. @grkvlt @ahgittin @neykov I'm keen to hear your thoughts on it.

aledsage:
@grkvlt https://github.com/apache/brooklyn-server/pull/341 is now merged, so this PR is no longer needed. Can you close it please (assuming you agree)?

PR #280, opened 2016-07-29
Add special flag decoration to parameters to enable EntitySpec values

grkvlt:


aledsage:
@grkvlt Can you add a test to demonstrate what you're fixing please?

camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/creation/BrooklynEntityDecorationResolver.java
```diff
@@ -180,20 +182,23 @@ protected void addDecorationFromJsonMap(Map<?, ?> decorationJson, List<EntityIni
...
+
         protected SpecParameterResolver(BrooklynYamlTypeInstantiator.Factory instantiator) { super(instantiator); }
 
         @Override
         protected String getDecorationKind() { return "Spec Parameter initializer"; }
 
         @Override
-        public void decorate(EntitySpec<?> entitySpec, ConfigBag attrs) {
+        public void decorate(EntitySpec<?> entitySpec, ConfigBag attrs, Set<String> encounteredRegisteredTypeIds) {
+            transformer = new BrooklynComponentTemplateResolver.SpecialFlagsTransformer(instantiator.loader, encounteredRegisteredTypeIds);
```
neykov: Would it work if you transform `attrs` before passing it to `BasicSpecParameter.fromConfigList` so it doesn't need to know about the `transformer`?

neykov:
From reading the sources my understanding is that this allows parameters to have specs as default values. Something that is a common pattern in the java blueprints. Nice idea.
LGTM, but agree with @aledsage that it needs test coverage.

aledsage:
@grkvlt when will you have a chance to look at this PR again?

I tried writing a test for you (adding it to `ConfigParametersYamlTest`), but it fails:

```
    @Test
    public void testConfigParameterWithEntitySpecAsDefault() throws Exception {
        addCatalogItems(
                "brooklyn.catalog:",
                "  itemType: entity",
                "  items:",
                "  - id: entity-with-keys",
                "    item:",
                "      type: "+TestEntity.class.getName(),
                "      brooklyn.parameters:",
                "      - name: my.param.key",
                "        type: "+EntitySpec.class.getName(),
                "        default: ",
                "          $brooklyn:entitySpec:",
                "          - type: "+BasicApplication.class.getName());

        String yaml = Joiner.on("\n").join(
                "services:",
                "- type: entity-with-keys");
        
        Entity app = createStartWaitAndLogApplication(yaml);
        TestEntity entity = (TestEntity) Iterables.getOnlyElement(app.getChildren());
        
        Object defaultVal = entity.config().get(entity.getEntityType().getConfigKey("my.param.key"));
        assertEquals(defaultVal instanceof EntitySpec, "defaultVal="+defaultVal);
        assertEquals(((EntitySpec<?>)defaultVal).getType(), BasicApplication.class, "defaultVal="+defaultVal);
        
        Entity child = entity.addChild((EntitySpec<?>)defaultVal);
        assertTrue(child instanceof BasicApplication, "child="+child);
    }
```

The error I get is:

```
org.apache.brooklyn.util.javalang.coerce.ClassCoercionException: Cannot coerce type org.apache.brooklyn.camp.brooklyn.spi.creation.EntitySpecConfiguration to org.apache.brooklyn.api.entity.EntitySpec (org.apache.brooklyn.camp.brooklyn.spi.creation.EntitySpecConfiguration@5a35672e): no adapter known
```

aledsage:
The failing jenkins test looks unrelated - `LoadBalancingPolicySoakTest.testLoadBalancingQuickTest`. We really need to deal with that one!

grkvlt:
@aledsage The "no adapter known" error is exactly what I was getting, that prompted this change. Thanks for the test code, I have added it as you suggested, and it passes as expected. This should be ready to merge now.

camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/creation/BrooklynEntityDecorationResolver.java
grkvlt: Unfortunately not, it would involve essentially duplicating the code in `parseParameter` which I don't think we should do.

grkvlt:
@aledsage @neykov I'm going to merge this now, as the tests have been added as requested and are passing

ahgittin:
as seen in https://github.com/cloudsoft/brooklyn-tosca/pull/113 this breaks backwards compatibility.  can we have methods matching the old signature passing the `null`?

PR #281, opened 2016-07-29
Prevent VanillaSoftwareProcess commands from being inherited by children

grkvlt:


sjcorbett:
+1 to this, it has caught me out in the past. It potentially breaks existing blueprints though so should be raised on the mailing list.

The test failures are valid: ConfigInheritanceYamlTest.testInheritsParentConfig and SpecParameterUnwrappingTest.testParameterDefaultsUsedInConfig.

aledsage:
Suggest you add the following tests to `org.apache.brooklyn.entity.software.base.VanillaSoftwareProcessTest`:

    @Test
    public void testCmdNotInheritedByChildrenOfSoftwareProcess() throws Exception {
        app.createAndManageChild(EntitySpec.create(VanillaSoftwareProcess.class)
                .configure(VanillaSoftwareProcess.PRE_INSTALL_COMMAND, "preInstallCommand")
                .configure(VanillaSoftwareProcess.LAUNCH_COMMAND, "parentLaunchCommand")
                .configure(VanillaSoftwareProcess.CHILDREN_STARTABLE_MODE, ChildStartableMode.FOREGROUND)
                .child(EntitySpec.create(VanillaSoftwareProcess.class)
                        .configure(VanillaSoftwareProcess.LAUNCH_COMMAND, "childLaunchCommand")));
        app.start(ImmutableList.of(loc));

        assertExecHasOnlyOnce(RecordingSshTool.getExecCmds(), "childLaunchCommand");
        assertExecHasOnlyOnce(RecordingSshTool.getExecCmds(), "preInstallCommand");
    }

    @Test
    public void testCmdNotInheritedFromParentOfSoftwareProcess() throws Exception {
        app.config().set(VanillaSoftwareProcess.PRE_INSTALL_COMMAND, "preInstallCommand");
        app.createAndManageChild(EntitySpec.create(VanillaSoftwareProcess.class)
                .configure(VanillaSoftwareProcess.LAUNCH_COMMAND, "launchCommand"));
        app.start(ImmutableList.of(loc));

        assertExecHasNever(RecordingSshTool.getExecCmds(), "preInstallCommand");
    }

    protected void assertExecHasNever(List<ExecCmd> actuals, String expectedCmd) {
        assertExecHasExactly(actuals, expectedCmd, 0);
    }
    
    protected void assertExecHasOnlyOnce(List<ExecCmd> actuals, String expectedCmd) {
        assertExecHasExactly(actuals, expectedCmd, 1);
    }
    
    protected void assertExecHasExactly(List<ExecCmd> actuals, String expectedCmd, int expectedCount) {
        String errMsg = "actuals="+actuals+"; expected="+expectedCmd;
        int count = 0;
        for (ExecCmd actual : actuals) {
            for (String subActual : actual.commands) {
                if (subActual.matches(expectedCmd)) {
                    count++;
                    break;
                }
            }
        }
        assertEquals(count, expectedCount, errMsg);
    }

software/base/src/main/java/org/apache/brooklyn/entity/software/base/SoftwareProcess.java
```diff
@@ -241,6 +246,7 @@
     @SetFromFlag("runtimeTemplates")
     MapConfigKey<String> RUNTIME_TEMPLATES = new MapConfigKey.Builder<String>(String.class, "templates.runtime")
             .description("Mapping of templates, to be filled in and copied before customisation, to destination name relative to runDir") 
+            .parentInheritance(ConfigInheritance.NONE)
```
neykov: Not sure about the file uploads. I suggest to leave them as is. The worst that can happen is to end up with some unused files on the machine. I'd even go as far as changing to merge behaviour :).
+1 for the rest.

grkvlt:
@sjcorbett @neykov I'll fix the tests. I'm not convinced the file uploads should be inherited, so I'll ask the `brooklyn-dev` list to convince me...

aledsage:
@grkvlt test failure since your last commit looks like a real failure: `ConfigInheritanceYamlTest.testInheritsParentConfig`.

aledsage:
@grkvlt will you get a chance to look at this again soon?

Hopefully the tests I wrote and pasted into the comment will work for you. They certainly failed in the right way without your changes.

And the jenkins test failure still sounds relevant:

```
2016-08-11 15:05:21,757 INFO  TESTNG FAILED: "Surefire test" - org.apache.brooklyn.camp.brooklyn.ConfigInheritanceYamlTest.testInheritsParentConfig() finished in 229 ms
java.lang.AssertionError: entity=EmptySoftwareProcessImpl{id=vp8rqasm0m}; configKey=templates.preinstall[ConfigKey:java.util.Map] expected [{file:///tmp/ConfigInheritanceYamlTest5144609806604208325.txt=myfile}] but found [null]
	at org.apache.brooklyn.test.Asserts.fail(Asserts.java:724)
	at org.apache.brooklyn.test.Asserts.failNotEquals(Asserts.java:114)
	at org.apache.brooklyn.test.Asserts.assertEquals(Asserts.java:436)
	at org.apache.brooklyn.core.entity.EntityAsserts.assertConfigEquals(EntityAsserts.java:60)
	at org.apache.brooklyn.camp.brooklyn.ConfigInheritanceYamlTest.assertEmptySoftwareProcessConfig(ConfigInheritanceYamlTest.java:805)
	at org.apache.brooklyn.camp.brooklyn.ConfigInheritanceYamlTest.testInheritsParentConfig(ConfigInheritanceYamlTest.java:227)
```

grkvlt:
@aledsage I added the tests you suggested, and updated the failing test pointed out by @sjcorbett so I think this should be able to be merged now

ahgittin:
Just what that option is intended for!  Merging...

ahgittin:
I've discovered a problem with this /cc @grkvlt @sjcorbett @aledsage @neykov.

Some blueprints set some of these keys e.g. `pre.install.command` as anonymous config on a non-software-process parent entity (ie where there is no corresponding config key) for the express intention that it get picked up by multiple different software process descendants.  (One common example is machine preparation, e.g. setting up monitoring agents and mounting an extra disk.)

But with this change the config is excluded and those commands no longer run.

(I also notice that `Entity.getAllConfig()` and the UI config view both include ancestor config which is not actually inherited; seems we aren't correctly attending to inheritance there.)

A proper way might be (A) to emit a "machine-sshable" sensor as part of `SoftwareProcess` and define an `InvokeEffector...` policy at the parent to (A.A) invoke an `SshCommandEffector` (with added capability of specifying the emitter as the effector target) and then emit another sensor on the emitter which unblocks an install latch there), or to (A.B) add and start a child at the emitter, and the child runs the ssh commands and emits the sensor.

However it would be very nice to support (B) existing practice eg pre-install config being specified in the parent, especially if we can do so in a way which addresses the problem this PR fixes where we don't want pre-install-command passed from parent software processes to children software processes.  As this change is only very recently introduced I think we should limit the backwards-incompatibility it introduces before the next release, especially to support (B).  (We can do what is necessary for (A.A) and (A.B) and suggest people migrate from (B) to (A.B) or (A.A) as best practice but as (B) is so easy I don't want to lose it.)

As a concrete way of doing this we could extend how we deal with `ConfigInheritance` so that we attend to different behaviours from parents/ancestors/bequeathers and children/descendants/inheritors.

Also it's a bit of a smell that we have an `enum` and a `class` that are basically the same, and both acting as hints with the logic for dealing with inheritance at the callers. Could this be cleaned up, eg with logic in this class?  (Especially if we are adding new modes!)

More to follow.

ahgittin:
Thinking aloud...

Where a key is defined at a parent/ancestor/bequeather it could specify

* is this inheritable?  (if false, descendants will never see it and all other options are ignored.)
* what conflict-resolution strategy should inheritors apply?  eg overwrite or deep merge

Where a key is defined at a child/descendant/inheritor it could specify:

* what conflict-resolution-strategy should be applied?  (as above, or following bequeather's recommendations, or perhaps "ignore inherited")
* should local default values be attended to?  (this is low priority but having this and saying true would allow us to drop an "ignore inherited" strategy because it would be the same as overwriting; and we get the option of merging default values for free)

Based on this I propose we change inheritance to three attributes:

* reinheritable?  true/false; if false, children/descendants/inheritors will never see it; default true
* conflict-resolution-strategy?  overwrite or deep_merge or null; default null meaning caller supplies
* use-local-default-value? true/false; if true, overwrite above means "always ignore"; default false

(And still more to follow.)

ahgittin:
Implementation-wise I suggest we deprecate `InheritanceMode` in favour of a new `StandardInheritanceModes` whose constructors set the values above, with the following upgrade patterns:

```
InheritanceMode.NONE = // one of these two depending on context
    // this should probably be the more common
    // (fixes the issue observed today with software process commands)
    StandardInheritanceModes.NOT_REINHERITED(false,overwrite,false)
    // but for backwards compatibility we'll set it equal to this mode (with a warning)
    StandardInheritanceModes.NEVER_INHERITED(false,overwrite,true)

InheritanceMode.IF_NO_EXPLICIT_VALUE =
    StandardInheritanceModes.OVERWRITE(true,overwrite,false)

InheritanceMode.DEEP_MERGE =
    StandardInheritanceModes.DEEP_MERGE(true,deep_merge,false)
```

We also change `ConfigInheritance` to be an interface with these new methods:

    /** used by callers to determine whether to export values for inheritance */
    boolean ConfigInheritance.isReinheritable(ConfigKey)

    /** returns the value respecting inheritance and using the defaultMode if keys do not define inheritance*/
    Object ConfigInheritance.resolve(@Nullable ConfigKey parentKey, @Nullable Object parentValue, @Nullable ConfigKey childKey, @Nullable Object childValue, @NonNull ConfigInheritance defaultMode);

All old methods are deprecated.  `StandardInheritanceModes` implements`ConfigInheritance` and code is rewritten to use the new methods.  I think this will cut down on repeated code dealing with inheritance elsewhere and if ever needed allow different strategies to be applied.

Big questions:
* Does this sound sensible?
* Are there any situations where we'd need more information than is currently passed to the proposed`resolve(...)` method?

@sjcorbett @grkvlt @aledsage @neykov your feedback especially desired

aledsage:
@ahgittin @grkvlt @sjcorbett @neykov This all sounds scarily complicated to describe to users, with the actual behaviour of inheritance for a given parent/child entity pair being unclear to the user until they dig into that entity's definition.

---
I worry about the pattern of using the `pre.install.command` on the parent entity. Whether that works depends on the child entity. If the child entity has its own value for `pre.install.command` then the parent's value would have been ignored (including in Brooklyn 0.9.0).

I can see that your suggestion of "conflict-resolution-strategy" being supplied by the caller would "fix" that, but I think it would be extremely hard to use.

---
Maybe I'm missing something in your proposal, but I think there is a fundamental problem that makes it hard to use: the inheritance mode / conflict-resolution mode is being defined in the config key declaration; but the user of the blueprint is only thinking about supplying the value in the `brooklyn.config` seciton.

---
To make this feasible to use/understand in YAML blueprints, I wonder if we really need the feature discussed in the email thread "[PROPOSAL] separate shell.env per command".

That proposal could be extended to support something like that below (but with more sensible key/values for "inherited_by_children", "conflict_resolution", etc)?

```
  brooklyn.config:
    pre.install:
      command: |
        sudo yum install -y my-favourite-monitoring
      inherited_by_children: true
      conflict_resolution: prefix
```

ahgittin:
@aledsage good observations

firstly yes this is complicated and we should avoid people needing to know about it in most cases -- eg use the policy approach, and/or come up with new bash entities which allow setting environment variables per command (although that wouldn't help in this instance) and/or putting arbitrary numbers of commands in an effector (i've been thinking of doing this in a map, using initd sequencing convention or precondition/postcondition markers).

very elegant idea to allow config keys to define their own strategy, also it allows us to do resolution/merges for lists if we ever needed that.  i think my refactor allows this (whereas previously it was not allowed).  but i think it just moves the complexity, and in any case it wouldn't solve the problem that this PR has broken users' blueprints.

but i think we should continue to have default/standard/simple inheritance modes -- no major change there -- it's just that the current semantics of `NONE` are wrong for most cases this PR set it (`NOT_REINHERITED` vs `NEVER_INHERITED`).  i think the new `StandardInheritanceModes` are pretty clear to a first approximation, and it's only if you are at a weird edge case (which we'd try to avoid) that you'd need to dig deeper, and if you do dig deeper the separation of inheritance attributes i think would be clearer now.  (though it's still so fresh it could probably be improved.)

aledsage:
@ahgittin (cc @grkvlt @sjcorbett @neykov) am I understanding your terminology correctly (of `NOT_REINHERITED` vs `NEVER_INHERITED`) that the former is about "child inheritance" and the latter is about "parent inheritance"? That is, `NOT_REINHERITED` means whether the value would be passed to children. Whereas `NEVER_INHERITED` is about whether the value defined on a parent would be inherited?

It feels like like we want separate configuration values for the three circumstance (!):
1. type-inheritance (hopefully it's ok as-is)
2. parent-inheritance
3. child-inheritance

For the case of `pre.install.command`, we'd want to change it back to have default parent-inheritance set to true (overwritten if child has any value); and child-inheritance set to never.

Is that roughly what you mean? Does that sound sensible?

ahgittin:
@aledage I think child-inheritance is separate to the other two.  And "parent-inheritance" is badly named (since parent-child exists also in type definition hierarchy); it should be something like "runtime-management-inheritance".

You're correct about the distinction `NOT_REINHERITED` vs `NEVER_INHERITED`.  But I think this distinction applies either for type definition and runtime management resolution (call these "inheritance contexts" ?).

The `SoftwareProcess` commands should have the default type-definition inheritance (`OVERWRITE`) so extended types can selectively change their parent's values.  But runtime management should be special `NOT_REINHERITED` to avoid managers polluting their children *if* the managers use those keys.

Ah yes it is complicated ... but if we get the thinking right here the good thing is that most users will just get the right behaviour (TM).

grkvlt:
@aledsage @ahgittin I'm not sure if your proposal will still allow the behaviour now relied on by Kubernetes, for example, where children of a `VanillaSoftwareProcess` do *not* want to see the parent's config. Is this still going to work?

ahgittin:
Comments from discussion:
* Being able to define resolution at the config key would be useful, eg to say to prepend to a list
* May need to know *who* defined the value eg so we can use its loader to resolve URLs and DSL commands (more notes below)
* Explicitly document the semantics, both of config inheritance and how loaders are used
* Might want a way to define config on specific children/descendants (Aled suggested, but Alex doesn't like, better to concentrate on validation)

Related issues with resolution through inheritance
  * Type inheritance -- need to access ancestor bundles; probably aggregating bundles (as it would be a lot of work to record the bundle of the supertype where the key is defined)
    * Distinction between "outer" case where we just need to aggregate super-type bundles, and 
    * "composite" case where we need to include the bundle where a type is being *used* (eg as a child); NB composite case could be tricky to implement as we might need to record `catalog-item-id-where-being-used` (possibly a list?) in addition to the `catalog-item-id` of an entity
  * Management inheritance -- should detect which runtime ancestor sets the value and resolve according to it (but not access ancestor bundles directly); if in places that's really hard (because a URL string is being passed down the line), could use DSL command eg `$brooklyn:local-url("classpath://start.sh")` or allow bundle-symbolicname to be supplied in a url `osgi:bundlename:/start.sh`)

The above is true for URLs and also for executing DSL commands.  (@neykov can you paste your illustrations of the three cases)

aledsage:
@grkvlt yes, that should still work: where children of a VanillaSoftwareProcess do not want to see the parent's config. The config's value would "not be re-inherited". That is, if the VanillaSoftwareProcess declares the config key (i.e. uses its value) then it will not be passed down to the children.

ahgittin:
@grkvlt all we'll do for `SoftwareProcess` instances is change `NONE` from `NEVER_INHERIT` to `NOT_REINHERITED` so yes, `SP` children of `SP` still won't see parent's commands, but what it will mean is that if you stick a `pre.install.command` on the root node (where it isn't recognised) it will now be available to all of the oldest `SP` descendants -- in other words it will run once on each machine (unless they define their own `pre.install.command`)

PR #282, opened 2016-07-30
Added attribute to DSL for immediate return of sensor data

grkvlt:
Allows constructs like the following, where `cluster.first` usually has the boolean false value, and therefore `attributeWhenReady` will block waiting for it to become true. This lets us use boolean valued sensors more effectively.
```YAML
memberSpec:
  $brooklyn:entitySpec:
    type: vanilla-software-process
      brooklyn.config:
        first: $brooklyn:parent().attribute("cluster.first")
```

### DIFF
--- a/camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/methods/BrooklynDslCommon.java
+++ b/camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/methods/BrooklynDslCommon.java
@@ -120,6 +120,10 @@ public class BrooklynDslCommon {
         return new DslComponent(Scope.THIS, "").config(keyName);
     }
 
+    public static BrooklynDslDeferredSupplier<?> attribute(String sensorName) {
+        return new DslComponent(Scope.THIS, "").attribute(sensorName);
+    }
+
     public static BrooklynDslDeferredSupplier<?> attributeWhenReady(String sensorName) {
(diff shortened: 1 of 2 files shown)


neykov:
Should we fix `attributeWhenReady` instead? It's confusing to have it wait indefinitely on `false`.
Having another `attribute`, very similar to `attributeWhenReady` complicates things even further.

Suggest renaming `attributeWhenRedy` to `attributeWhenTruth` and changing `attributeWhenReady` to wait on `notNull && notEmpty`.

aledsage:
+1 to suggestion from @neykov - having `attributeWhenReady` and `attribute` is extremely confusing. I initially expected your `attribute` DSL function to give me the immediate value, rather than blocking at all.

@grkvlt This is one for the mailing list as it will break backwards compatibility. Some people currently rely on `attribueWhenReady("service.isUp")` to wait for true.

grkvlt:
@aledsage what about changing the ready predicate for `attribute(...)` to `Predicates.alwaysTrue()` which would then return the current sensor value, including `null`?

aledsage:
FYI this was discussed on the mail dev@brooklyn.apache.org mailing list with subject "[PROPOSAL] $brooklyn:attributeWhenReady semantics".

ahgittin:
@grkvlt can you close this issue as it definitely isn't what we want; any breakage to backwards compatibility here will be very bad for users

continue discussion on ML

grkvlt:
@ahgittin @aledsage closing this, based on feedback from mailing list

PR #283, opened 2016-07-30
Make dontRequireTtyForSudo idempotent

grkvlt:
Prevent race condition with multiple entities (for example children of a `SoftwareProcess` or `SameServerEntity`) that share the same configuration, and could try to make this change at the same time. Still not fully resilient in the face of bad timing, but better than before. Perhaps some kind of lock file could be created using `flock` at the start of the script?


aledsage:
LGTM; merging.

PR #284, opened 2016-08-02
Add TestCase.ON_ERROR_SPEC

aledsage:
Instantiates and executes this entity if there is an error while executing the test-cases.


drigodwin:
I also get the `testInvokeEffectorErrorCollapsedNicely` test failure

drigodwin:
Looks good to me! Thanks @aledsage

aledsage:
Thanks @drigodwin - as per our offline conversation, I've also added: 1) more logging for if we do or don't create the on-error entity; 2) the on_error_spec is now not inherited by child entities, so we'll only instantiate it once if a child test-case fails.

Rebased to one commit. Will merge once jenkins completes its tests.

PR #285, opened 2016-08-03
Adding HttpExecutor to be used for interchangeable implementations

ygy:


utils/common/src/main/java/org/apache/brooklyn/util/http/HttpExecutor.java
```diff
@@ -0,0 +1,37 @@
...
+import org.apache.http.client.HttpClient;
+import org.apache.http.impl.client.LaxRedirectStrategy;
+import org.apache.http.params.HttpParams;
+
+@SuppressWarnings("deprecation")
+public interface HttpExecutor extends HttpClient {
+
+    void setParams(HttpParams httpParams);
+
+    void setRedirectStrategy(LaxRedirectStrategy laxRedirectStrategy);
```
aledsage: I'm not convinced we want to tie this interface to the `org.apache.http` classes. I was assuming we'd make it more general. And that we would not extend `HttpClient`, which would presumably make it a lot harder to supply implementations which do not use `org.apache.http`.

utils/common/src/main/java/org/apache/brooklyn/util/http/HttpExecutorFactory.java
```diff
@@ -0,0 +1,25 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.brooklyn.util.http;
+
+public class HttpExecutorFactory {
```
aledsage: I was assuming this would be an interface, with an implementation that used the `org.apache.http` client.

utils/common/src/main/java/org/apache/brooklyn/util/http/HttpTool.java
```diff
@@ -319,7 +318,7 @@ public HttpClientBuilder trustSelfSigned() {
             return this;
         }
         public HttpClient build() {
-            final DefaultHttpClient httpClient = new DefaultHttpClient(clientConnectionManager);
+            final HttpExecutor httpClient = new HttpExecutorImpl(clientConnectionManager);
```
aledsage: I assumed we'd use the factory, rather than calling the constructor directly.

ygy:
@nakomis @neykov Can you review this PR?

utils/common/src/main/java/org/apache/brooklyn/util/http/HttpExecutorFactory.java
```diff
@@ -0,0 +1,65 @@
...
+import org.apache.brooklyn.util.exceptions.Exceptions;
+import org.apache.brooklyn.util.http.executor.HttpExecutor;
+import org.apache.brooklyn.util.javalang.Reflections;
+import org.apache.brooklyn.util.text.StringPredicates;
+import org.apache.brooklyn.util.text.Strings;
+
+import com.google.common.base.Optional;
+import com.google.common.collect.Iterables;
+
+public class HttpExecutorFactory {
```
neykov: Should be an interface

utils/common/src/main/java/org/apache/brooklyn/util/http/HttpExecutorFactory.java
```diff
@@ -0,0 +1,65 @@
...
+
+    public <T, S> HttpExecutor getHttpExecutor(Map<?, ?> props) {
+        HttpExecutor httpExecutor;
+        
+        Optional<Map<?,?>> executorProps =  (Optional<Map<?, ?>>) Iterables.tryFind(props.keySet(), StringPredicates.isStringStartingWith(HTTP_EXECUTOR_CLASS_PROPERTIES_PREFIX));
+
+        if (executorProps.isPresent()) {
+            Map<?, ?> instanceProps = executorProps.get();
+
+            String httpExecutorClass = (String) instanceProps.get(HTTP_EXECUTOR_CLASS);
```
neykov: You are filtering for `HTTP_EXECUTOR_CLASS_PROPERTIES_PREFIX`, but expect `HTTP_EXECUTOR_CLASS` to be included - I think not?
Also if only `HTTP_EXECUTOR_CLASS` is included in the config `isPresent()` above will be `false`.

utils/common/src/main/java/org/apache/brooklyn/util/http/HttpExecutorFactory.java
```diff
@@ -0,0 +1,65 @@
...
+            Map<String, String> httpExecutorProps = MutableMap.of();
+            for (Entry<?, ?> entry: instanceProps.entrySet()) {
+                String keyName = Strings.removeFromStart((String)entry.getKey(), HTTP_EXECUTOR_CLASS_PROPERTIES_PREFIX);
+                httpExecutorProps.put(keyName, (String)entry.getValue());
+            }
+
+            Reflections reflections = new Reflections(HttpExecutorFactory.class.getClassLoader());
```
neykov: Better use `ClassLoaderUtils` instead of `Reflections` (here and elsewhere in the PR) to be able to use some classloading goodies in OSGi land.

core/src/main/java/org/apache/brooklyn/core/location/LocationConfigKeys.java
```diff
@@ -35,7 +35,9 @@
     public static final ConfigKey<String> DISPLAY_NAME = ConfigKeys.newStringConfigKey("displayName");
     public static final ConfigKey<Boolean> ENABLED = ConfigKeys.newBooleanConfigKey("enabled", "Whether the location is enabled for listing and use "
         + "(only supported for selected locations)", true);
-    
+
+    public static final ConfigKey<String> EXTENSION = ConfigKeys.newStringConfigKey("extension");
```
neykov: Better make it Map<String, Object> to
  * allow multiple extensions to be registered
  * use $brooklyn:object to construct the extensions (though brooklyn DSL doesn't work too well in location config). If it's a string then could treat it as class name and instantiate it. 

wdyt?

utils/common/src/main/java/org/apache/brooklyn/util/http/executor/HttpRequest.java
```diff
@@ -0,0 +1,128 @@
...
+
+@Beta
+public interface HttpRequest {
+    
+    // TODO Should we use InputStream for body (rather than byte[]). That would mean we don't have
+    // to hold the entire body in-memory, which might be important for really big payloads!
+    
+    @Beta
+    public static class Builder {
+        protected boolean isHttps;
```
neykov: The protocol is included in the `URI`, better skip this one.

utils/common/src/main/java/org/apache/brooklyn/util/http/executor/HttpRequest.java
```diff
@@ -0,0 +1,128 @@
...
+
+import com.google.common.annotations.Beta;
+import com.google.common.collect.ArrayListMultimap;
+import com.google.common.collect.Multimap;
+
+@Beta
+public interface HttpRequest {
+    
+    // TODO Should we use InputStream for body (rather than byte[]). That would mean we don't have
+    // to hold the entire body in-memory, which might be important for really big payloads!
```
neykov: +1

neykov: One reason not to do it is that consuming the stream might be async (?) so it's not clear when to close the stream.
Whatever we end up doing it's best to be consistent - why accept byte[] in the request but return stream in the response?

core/src/main/java/org/apache/brooklyn/feed/http/HttpFeed.java
```diff
@@ -275,7 +293,23 @@ public HttpFeed() {
...
         Map<String,String> baseHeaders = ImmutableMap.copyOf(checkNotNull(builder.headers, "headers"));
-        
+
+        HttpExecutor httpExecutor = null;
+        if (builder.httpExecutor != null) {
+            httpExecutor = builder.httpExecutor;
+        } else {
+            HttpExecutorFactory httpExecutorFactory = null;
+            Collection<Location> locations = builder.entity.getLocations();
+            Maybe<MachineLocation> location =  Machines.findUniqueMachineLocation(locations, MachineLocation.class);
```
neykov: Use `Locations.getLocationsCheckingAncestors(locations, entity)` instead, the location could be coming from a parent entity.

core/src/main/java/org/apache/brooklyn/feed/http/HttpFeed.java
```diff
@@ -275,7 +293,23 @@ public HttpFeed() {
...
+            if (location.isPresent() && location.get().hasExtension(HttpExecutorFactory.class)) {
+                httpExecutorFactory = location.get().getExtension(HttpExecutorFactory.class);
+                Map<String, Object> httpExecutorProps = location.get().getAllConfig(true);
+                httpExecutor = httpExecutorFactory.getHttpExecutor(httpExecutorProps);
```
neykov: Minor, could pass the location and the implementation will then decide whether to do `.getAllConfig(true);` or somthing simpler.

core/src/main/java/org/apache/brooklyn/feed/http/HttpFeed.java
```diff
@@ -275,7 +293,23 @@ public HttpFeed() {
...
+            } else {
+                httpExecutor = HttpExecutorImpl.newInstance();
```
neykov: `new HttpExecutorImpl()` instead

neykov: Ignore.

utils/common/src/main/java/org/apache/brooklyn/util/http/HttpExecutorFactory.java
```diff
@@ -0,0 +1,65 @@
...
+
+            String httpExecutorClass = (String) instanceProps.get(HTTP_EXECUTOR_CLASS);
+            Map<String, String> httpExecutorProps = MutableMap.of();
+            for (Entry<?, ?> entry: instanceProps.entrySet()) {
+                String keyName = Strings.removeFromStart((String)entry.getKey(), HTTP_EXECUTOR_CLASS_PROPERTIES_PREFIX);
+                httpExecutorProps.put(keyName, (String)entry.getValue());
+            }
+
+            Reflections reflections = new Reflections(HttpExecutorFactory.class.getClassLoader()); 
+            httpExecutor = (HttpExecutor) reflections.loadInstance(httpExecutorClass, httpExecutorProps);
```
neykov: Don't see this constructor in `HttpExecutorImpl`

utils/common/src/main/java/org/apache/brooklyn/util/http/executor/apacheclient/HttpExecutorImpl.java
```diff
@@ -0,0 +1,106 @@
...
+        HttpConfig config = (request.config() != null) ? request.config() : DEFAULT_CONFIG;
+        Credentials creds = (request.credentials() != null) ? new UsernamePasswordCredentials(request.credentials().username(), request.credentials().password()) : null;
+        HttpClient httpClient = HttpTool.HttpClientBuilder.fromBuilder(baseBuilder)
+                .uri(request.uri())
+                .https(request.isHttps())
+                .credential(Optional.fromNullable(creds))
+                .laxRedirect(config.laxRedirect())
+                .trustSelfSigned(config.trustSelfSigned())
+                .trustAll(config.trustAll())
+                .build();
```
neykov: The `baseBuilder` is always overridden by the properties here. Don't see a point of having it at all.
Executors are initialized with map of properties which are not directly related to the requests (i.e. HttpClientBuilder).

neykov:
Looks like a great start.

core/src/main/java/org/apache/brooklyn/feed/http/HttpFeed.java
```diff
@@ -331,52 +366,45 @@ protected void preStart() {
...
+                    HttpResponse response =  pollInfo.httpExecutor.execute(new HttpRequest.Builder()
+                            .headers(pollInfo.headers)
+                            .uri(pollInfo.uriProvider.get())
+                            .credentials(creds)
+                            .method(pollInfo.method)
+                            .body(pollInfo.body)
+                            .build());
+                    return createHttpToolRespose(response);
+                }};
+                getPoller().scheduleAtFixedRate(pollJob, new DelegatingPollHandler(handlers), minPeriod);
```
bostko: Wrong indentation here.

core/src/main/java/org/apache/brooklyn/feed/http/HttpFeed.java
```diff
@@ -222,15 +238,17 @@ protected void finalize() {
     }
     
     private static class HttpPollIdentifier {
+        final HttpExecutor httpExecutor;
```
bostko: @ygy you do not need it here.

bostko:
A few fixes in here https://github.com/ygy/brooklyn-server/pull/3

core/src/main/java/org/apache/brooklyn/feed/http/HttpFeed.java
```diff
@@ -331,52 +365,45 @@ protected void preStart() {
...
-        return builder.build();
     }
 
     @SuppressWarnings("unchecked")
     protected Poller<HttpToolResponse> getPoller() {
-        return (Poller<HttpToolResponse>) super.getPoller();
+        return  (Poller<HttpToolResponse>) super.getPoller();
+    }
+
+    private HttpToolResponse createHttpToolRespose(HttpResponse response) throws IOException {
```
iyovcheva: very minor -  error in syntax - "response"

core/src/main/java/org/apache/brooklyn/core/location/AbstractLocation.java
```diff
@@ -718,6 +719,35 @@ public boolean removeChild(Location child) {
...
+                    Class<?> extensionClassType =  new ClassLoaderUtils(this, getManagementContext()).loadClass(extension.getKey());
+
+                    if (!hasExtension(extensionClassType)) {
+                        Object extensionClass = new ClassLoaderUtils(this, getManagementContext()).loadClass(extension.getValue()).newInstance();
+                        addExtension((Class)extensionClassType, extensionClass);
+                    }
+                }
+            }
+        } catch (Exception e) {
+            LOG.error("Location extension can not be loaded {}", e);
```
aledsage: Minor: this will call `LOG.error(String msg, Throwable t)` so the `{}` has no effect. That might be what we want, to ensure the stacktrace definitely gets logged (depending what happens to the propagated exception).

If we do want just the exception's toString, would need to change it to something like:

    LOG.error("Location extension can not be loaded (rethrowing): {}", new Object[] {e});

aledsage:
LGTM - a few comments from @neykov that are not yet addressed. Have those been discussed?

core/src/main/java/org/apache/brooklyn/util/http/executor/HttpExecutorFactoryImpl.java
```diff
@@ -0,0 +1,75 @@
...
+            Map<?, ?> executorProps = Maps.filterKeys(props, StringPredicates.isStringStartingWith(HTTP_EXECUTOR_CLASS_CONFIG_PREFIX));
+            if (executorProps.size() > 0) {
+                Map<String, String> httpExecutorProps = MutableMap.of();
+                for (Entry<?, ?> entry: executorProps.entrySet()) {
+                    String keyName = Strings.removeFromStart((String)entry.getKey(), HTTP_EXECUTOR_CLASS_CONFIG_PREFIX);
+                    httpExecutorProps.put(keyName, TypeCoercions.coerce(entry.getValue(), String.class));
+                }
+
+                try {
+                    httpExecutor = (HttpExecutor) new ClassLoaderUtils(HttpExecutorFactoryImpl.class).loadClass(httpExecutorClass).getConstructor(Map.class).newInstance(httpExecutorProps);
```
bostko: Shouldn't this come from location extension?

bostko: You can benefit from using a factory pattern by implementing cache. May be better in a separate factory.

core/src/test/java/org/apache/brooklyn/feed/http/HttpFeedTest.java
```diff
@@ -71,12 +71,12 @@
 
     private static final long TIMEOUT_MS = 10*1000;
     
-    private BetterMockWebServer server;
-    private URL baseUrl;
+    protected BetterMockWebServer server;
```
bostko: It worth's adding in this PR tests for HttpExecutor but without HttpFeed.

utils/common/src/main/java/org/apache/brooklyn/util/http/HttpTool.java
```diff
@@ -250,6 +254,21 @@ public static HttpClientBuilder httpClientBuilder() {
         private boolean trustAll;
         private boolean trustSelfSigned;
 
+        public static HttpClientBuilder fromBuilder(HttpClientBuilder other) {
+            HttpClientBuilder result = httpClientBuilder();
+            result.clientConnectionManager = other.clientConnectionManager;
+            result.httpParams = other.httpParams;
```
bostko: HttpClient should be configured with `clientConnectionManager` possibly the `HttpExecutorFactory` and other similar client properties.
A client can be used against multiple endpoints so including `uri`, `port` and `credentials` is not appropriate for the HttpClientBulder in my opinion.

utils/common/src/main/java/org/apache/brooklyn/util/http/executor/HttpRequest.java
bostko: I think the end consumer who is using HttpResponse and HttpRequest should take care for closing the Stream.
Please check `org.apache.http.HttpEntity`.
In essence HttpExecutor is a pattern for providing external httpClients.
For deciding what to return in HttpRequest and HttpResponse Stream or byte[], I think we should better take a look at exposing it in a way which is implemented in most of the httpClients like Apache Http Client.

ygy: That's a very good idea. I'll create a new ticket for the subsequent extending the API.

utils/common/src/main/java/org/apache/brooklyn/util/http/executor/apacheclient/HttpExecutorImpl.java
ygy: The baseBuilder provides configuration for several more parameters which are implementation dependent (e.g. apache client)

core/src/main/java/org/apache/brooklyn/util/http/executor/HttpExecutorFactoryImpl.java
bostko: Better use `getClass()` to highlight that you are using the class from `this`.

utils/common/src/main/java/org/apache/brooklyn/util/http/HttpTool.java
bostko: `httpParams` is leftover code which is deprecated so it can be removed now but it is ok to remove it later as well.

core/src/main/java/org/apache/brooklyn/feed/http/HttpFeed.java
```diff
@@ -275,7 +294,23 @@ public HttpFeed() {
...
         Map<String,String> baseHeaders = ImmutableMap.copyOf(checkNotNull(builder.headers, "headers"));
-        
+
+        HttpExecutor httpExecutor;
+        if (builder.httpExecutor != null) {
+            httpExecutor = builder.httpExecutor;
+        } else {
+            HttpExecutorFactory httpExecutorFactory = null;
+            Collection<? extends Location> locations = Locations.getLocationsCheckingAncestors(builder.entity.getLocations(), builder.entity);
+            Maybe<Location> location =  Machines.findUniqueElement(locations, Location.class);
```
neykov: All of the elements here are of type `Location`.  It's possible to have two locations per entity - one coming from `location: xx` property in YAML  (which is the description of how to provision a new machine) and another `SshMachineLocation` (which points to a specific VM with IPs created from the previous one). I believe only the former will have the extensions when coming from yaml (they will not be inherited). Looking for the first location to have the required extension should be good enough.

core/src/main/java/org/apache/brooklyn/feed/http/HttpFeed.java
```diff
@@ -331,52 +365,54 @@ protected void preStart() {
...
-        return (Poller<HttpToolResponse>) super.getPoller();
+        return  (Poller<HttpToolResponse>) super.getPoller();
+    }
+
+    @SuppressWarnings("unchecked")
+    private HttpToolResponse createHttpToolRespose(HttpResponse response) throws IOException {
+        int responseCode = response.code();
+        if (responseCode == 400) { // Unprocessable Entity - https://stackoverflow.com/questions/6123425/rest-response-code-for-invalid-data
+            throw new IOException(" Unprocessable Entity: " + response.reasonPhrase());
+        }
```
neykov: Why are you interpreting the response here? Let the caller do it - 400 is not any special than other responses. Don't understand the error message as well - which entity are you referring to?

core/src/main/java/org/apache/brooklyn/feed/http/HttpFeed.java
neykov: Better get the `SshMachineLocation` here (there might or might not be another location) and get the config from it. Inheritance should take care of getting the config from the right location.

core/src/main/java/org/apache/brooklyn/location/byon/ByonLocationResolver.java
```diff
@@ -108,6 +109,7 @@ protected ConfigBag extractConfig(Map<?,?> locationFlags, String spec, LocationR
         MutableMap<String, Object> defaultProps = MutableMap.of();
         defaultProps.addIfNotNull("user", user);
         defaultProps.addIfNotNull("port", port);
+        defaultProps.addIfNotNull(LocationConfigKeys.EXTENSIONS.getName(), config.get(LocationConfigKeys.EXTENSIONS));
```
neykov: Don't see how this is used, remove.

core/src/test/java/org/apache/brooklyn/feed/http/HttpFeedTest.java
ygy: Addressed in 4d905a5120a263c41e7cb3d3e69c3f2cd1a7ccaa

core/src/main/java/org/apache/brooklyn/feed/http/HttpFeed.java
ygy: Addressed. `MachineLocation` is used to cover windows entities too.

core/src/main/java/org/apache/brooklyn/feed/http/HttpFeed.java
ygy: Response code has to be interpreted so it is backward compatible with `HttpToolResponse` which depends on `org.apache.http.HttpResponse`. 
`HttpToolResponse` checks for response code `0` in `getResponseCode()`.

You are correct about code `400` in general. HTTP code `422` will be used instead. I've put a comment from the RFC why it is used.

core/src/main/java/org/apache/brooklyn/location/byon/ByonLocationResolver.java
ygy: It is needed so the extensions are supported for BYON locations too.

utils/common/src/main/java/org/apache/brooklyn/util/http/executor/HttpRequest.java
```diff
@@ -0,0 +1,130 @@
...
+
+    @Beta
+    public static class Builder {
```
bostko: Minor: Doesn't make sence to have a builder for an implementation in the interface.

core/src/main/java/org/apache/brooklyn/core/location/AbstractLocation.java
```diff
@@ -718,6 +719,35 @@ public boolean removeChild(Location child) {
...
+                    Class<?> extensionClassType =  new ClassLoaderUtils(this, getManagementContext()).loadClass(extension.getKey());
+
+                    if (!hasExtension(extensionClassType)) {
+                        Object extensionClass = new ClassLoaderUtils(this, getManagementContext()).loadClass(extension.getValue()).newInstance();
+                        addExtension((Class)extensionClassType, extensionClass);
+                    }
+                }
+            }
+        } catch (Exception e) {
+            LOG.error("Location extension can not be loaded (rethrowing): {}", new Object[] {e});
```
neykov: Could catch + rethrow inside the loop to get access to the failing extension pair and log it.

core/src/main/java/org/apache/brooklyn/feed/http/HttpFeed.java
```diff
@@ -331,52 +366,62 @@ protected void preStart() {
...
+        /* From https://tools.ietf.org/html/rfc4918#section-11.2
+        The 422 (Unprocessable Entity) status code means the server
+        understands the content type of the request entity (hence a
+        415(Unsupported Media Type) status code is inappropriate), and the
+        syntax of the request entity is correct (thus a 400 (Bad Request)
+        status code is inappropriate) but was unable to process the contained
+        instructions. */
+        if (responseCode == 422) {
+            throw new IOException(" Unprocessable Entity: " + response.reasonPhrase());
+        }
```
neykov: Still don't see why we need special handling of 422? (to discuss).

core/src/main/java/org/apache/brooklyn/location/byon/ByonLocationResolver.java
neykov: After looking in more details: This adds the `EXTENSIONS` key to the machine spec. If this is skipped, i.e. the machine locations don't have the key, then wouldn't they get it from the parent location through inheritance? Why is BYON any more special than jclouds in handling this case? Is there a test which covers this case (and fails if the line is dropped)?

core/src/main/java/org/apache/brooklyn/util/http/executor/HttpExecutorFactory.java
```diff
@@ -0,0 +1,29 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.brooklyn.util.http.executor;
+
+import java.util.Map;
+
+public interface HttpExecutorFactory {
+    public static final String HTTP_EXECUTOR_CLASS = "httpExecutorClass";
```
neykov: Nit: `HTTP_EXECUTOR_CLASS_CONFIG` would be more appropriate (or `HTTP_EXECUTOR_CLASS_PROPERTY`), the former following the pattern of he next property.

core/src/main/java/org/apache/brooklyn/util/http/executor/HttpExecutorFactoryImpl.java
```diff
@@ -0,0 +1,75 @@
...
+        HttpExecutor httpExecutor;
+
+        String httpExecutorClass = (String) props.get(HTTP_EXECUTOR_CLASS);
+        if (httpExecutorClass != null) {
+            Map<?, ?> executorProps = Maps.filterKeys(props, StringPredicates.isStringStartingWith(HTTP_EXECUTOR_CLASS_CONFIG_PREFIX));
+            if (executorProps.size() > 0) {
+                Map<String, String> httpExecutorProps = MutableMap.of();
+                for (Entry<?, ?> entry: executorProps.entrySet()) {
+                    String keyName = Strings.removeFromStart((String)entry.getKey(), HTTP_EXECUTOR_CLASS_CONFIG_PREFIX);
+                    httpExecutorProps.put(keyName, TypeCoercions.coerce(entry.getValue(), String.class));
```
neykov: Why force all executor params to be strings?

core/src/main/java/org/apache/brooklyn/util/http/executor/HttpExecutorFactoryImpl.java
```diff
@@ -0,0 +1,75 @@
...
+                }
+
+                try {
+                    httpExecutor = (HttpExecutor) new ClassLoaderUtils(getClass()).loadClass(httpExecutorClass).getConstructor(Map.class).newInstance(httpExecutorProps);
+                } catch (Exception e) {
+                    throw Exceptions.propagate(e);
+                }
+            } else {
+                LOG.error("Missing parameters for: " + HTTP_EXECUTOR_CLASS);
+                throw Exceptions.propagate(new IllegalArgumentException("Missing parameters for: " + HTTP_EXECUTOR_CLASS));
```
neykov: Some executor implementations don't need additional arguments (i.e. `HttpExecutorImpl`). Don't treat this as an error - create the requested executed with an empty map instead.

utils/common/src/main/java/org/apache/brooklyn/util/http/executor/HttpRequest.java
```diff
@@ -0,0 +1,130 @@
...
+    /**
+     * The payload of the request, or null if no payload (e.g. for GET requests). 
+     */
+    @Nullable
+    byte[] body();
+    
+    Multimap<String, String> headers();
+
+    @Nullable
+    Credentials.BasicAuth credentials();
```
neykov: Don't like how `BasicAuth` is hardcoded in the API. Username + password could be used for digest or a number of other schemes. The credentials are (mostly) orthogonal to the scheme.
* Rename `BasicAuth` to `UsernamePassword(Credentials)`. 
* Change Credentials to interface (with getUser, getPassword method) and implement it in the above.
* Use Credentials in the API.

utils/common/src/main/java/org/apache/brooklyn/util/http/executor/HttpResponseImpl.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    @Override
+    public Multimap<String, String> headers() {
+        return headers;
+    }
+
+    @Override
+    public long getContentLength() {
+        return contentLength;
```
neykov: In the streaming case could have `Content-Length` set in the headers.

utils/common/src/main/java/org/apache/brooklyn/util/http/executor/apacheclient/HttpExecutorImpl.java
```diff
@@ -0,0 +1,113 @@
...
+@Beta
+public class HttpExecutorImpl implements HttpExecutor {
+
+    private static final byte[] EMPTY_BYTE_ARRAY = new byte[0];
+    
+    private static final HttpConfig DEFAULT_CONFIG = HttpConfig.builder()
+            .laxRedirect(false)
+            .trustAll(false)
+            .trustSelfSigned(false)
+            .build();
```
neykov: Are these settings universally supported between clients?

utils/common/src/main/java/org/apache/brooklyn/util/http/executor/apacheclient/HttpExecutorImpl.java
neykov: Who is supposed to pass the builder to the executor? The contract is to pass a map of options to the constructor. The `baseBuilder` is always empty.

utils/common/src/main/java/org/apache/brooklyn/util/http/executor/apacheclient/HttpExecutorImpl.java
```diff
@@ -0,0 +1,113 @@
...
+        case "head":
+            response = HttpTool.httpHead(httpClient, request.uri(), request.headers());
+            break;
+        case "post":
+            response = HttpTool.httpPost(httpClient, request.uri(), request.headers(), orEmpty(request.body()));
+            break;
+        case "put":
+            response = HttpTool.httpPut(httpClient, request.uri(), request.headers(), orEmpty(request.body()));
+            break;
+        case "delete":
```
neykov: Add as upper-case constants to the interface.

utils/common/src/main/java/org/apache/brooklyn/util/http/executor/apacheclient/HttpResponseWrapper.java
```diff
@@ -0,0 +1,109 @@
...
+    @Override
+    public long getContentLength() {
+        byte[] content = getContentImpl();
+        return (content == null) ? -1 : content.length;
+    }
+
+    @Override
+    public InputStream getContent() {
+        byte[] content = getContentImpl();
+        return (content == null) ? null : new ByteArrayInputStream(content);
```
neykov: Isn't there API to get the stream directly?

core/src/main/java/org/apache/brooklyn/feed/http/HttpFeed.java
```diff
@@ -331,52 +366,55 @@ protected void preStart() {
...
+    }
+
+    @SuppressWarnings("unchecked")
+    private HttpToolResponse createHttpToolRespose(HttpResponse response) throws IOException {
+        int responseCode = response.code();
+
+        Map<String,? extends List<String>> headers = (Map<String, List<String>>) (Map<?, ?>) response.headers().asMap();
+
+        byte[] content = null;
+        if (response.getContentLength() > 0) {
```
neykov: Better copy the current logic from [HttpToolResponse](https://github.com/apache/brooklyn-server/blob/master/utils/common/src/main/java/org/apache/brooklyn/util/http/HttpToolResponse.java#L69-L81) as it's been battle-tested.

core/src/main/java/org/apache/brooklyn/util/http/executor/HttpExecutorFactoryImpl.java
```diff
@@ -0,0 +1,73 @@
...
+
+        String httpExecutorClass = (String) props.get(HTTP_EXECUTOR_CLASS_CONFIG);
+        if (httpExecutorClass != null) {
+            Map<String, String> httpExecutorProps = MutableMap.of();
+            Map<?, ?> executorProps = Maps.filterKeys(props, StringPredicates.isStringStartingWith(HTTP_EXECUTOR_CLASS_CONFIG_PREFIX));
+            if (executorProps.size() > 0) {
+
+                for (Entry<?, ?> entry: executorProps.entrySet()) {
+                    String keyName = Strings.removeFromStart((String)entry.getKey(), HTTP_EXECUTOR_CLASS_CONFIG_PREFIX);
+                    httpExecutorProps.put(keyName, TypeCoercions.coerce(entry.getValue(), String.class));
```
neykov: Why assume all keys are `String`s? This is implementation dependent.

core/src/test/java/org/apache/brooklyn/util/http/executor/HttpExecutorImplTest.java
```diff
@@ -0,0 +1,210 @@
...
+
+    @BeforeMethod(alwaysRun=true)
+    public void setUp() throws Exception {
+        factory = new HttpExecutorFactoryImpl();
+        server = BetterMockWebServer.newInstanceLocalhost();
+        server.play();
+        baseUrl = server.getUrl("/");
+    }
+
+    @AfterMethod
```
neykov: Personal preference to include `alwaysRuns=true`

core/src/main/java/org/apache/brooklyn/core/location/AbstractLocation.java
```diff
@@ -718,6 +719,35 @@ public boolean removeChild(Location child) {
...
+            for (Map.Entry<String, String> extension: extensions.entrySet()) {
+                try {
+                    Class<?> extensionClassType =  new ClassLoaderUtils(this, getManagementContext()).loadClass(extension.getKey());
+
+                    if (!hasExtension(extensionClassType)) {
+                        Object extensionClass = new ClassLoaderUtils(this, getManagementContext()).loadClass(extension.getValue()).newInstance();
+                        addExtension((Class)extensionClassType, extensionClass);
+                    }
+                } catch (Exception e) {
+                    LOG.error("Location extension can not be loaded (rethrowing): {}", new Object[] {extension.getKey(), extension.getValue(), e});
```
neykov: Update the message to match the argument count

neykov:
Almost there, only concerned about the content length being available on responses.
I'd like to see tests covering the extensions config key in jclouds and byon locations - can be done in another PR.

core/src/main/java/org/apache/brooklyn/feed/http/HttpFeed.java
```diff
@@ -407,14 +410,26 @@ private HttpToolResponse createHttpToolRespose(HttpResponse response) throws IOE
...
+        final long durationMillisOfFullContent;
+        final long startTime = System.currentTimeMillis();
+
+        ByteArrayOutputStream out = new ByteArrayOutputStream();
+        durationMillisOfFirstResponse = Duration.sinceUtc(startTime).toMilliseconds();
+        try {
+            ByteStreams.copy(response.getContent(), out);
+            content = out.toByteArray();
+        } catch (IOException e) {
+            throw Throwables.propagate(e);
```
neykov: In Brooklyn the recommended is `Exceptions.propagate`, but this works as well in this case.

ygy:
Jenkins, retest this please

neykov:
Thanks @ygy, merging.

PR #287, opened 2016-08-03
Add new member tracking policy with configurable SSH commands

grkvlt:
Allows easy creation of new load balancer entities by defining tracking policies in YAML that execute the required commands over SSH when an entity is added or removed.

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/entity/AbstractEntity.java
+++ b/core/src/main/java/org/apache/brooklyn/core/entity/AbstractEntity.java
@@ -1698,6 +1698,7 @@ public abstract class AbstractEntity extends AbstractBrooklynObject implements E
             CatalogUtils.setCatalogItemIdOnAddition(AbstractEntity.this, policy);
             policiesInternal.add((AbstractPolicy)policy);
             ((AbstractPolicy)policy).setEntity(AbstractEntity.this);
+            ConfigConstraints.assertValid(policy);
 
             getManagementSupport().getEntityChangeListener().onPolicyAdded(policy);
             sensors().emit(AbstractEntity.POLICY_ADDED, new PolicyDescriptor(policy));
@@ -1776,6 +1777,7 @@ public abstract class AbstractEntity extends AbstractBrooklynObject implements E
             CatalogUtils.setCatalogItemIdOnAddition(AbstractEntity.this, enricher);
             enrichersInternal.add((AbstractEnricher) enricher);
             ((AbstractEnricher)enricher).setEntity(AbstractEntity.this);
+            ConfigConstraints.assertValid(enricher);
             
             getManagementSupport().getEntityChangeListener().onEnricherAdded(enricher);
             // TODO Could add equivalent of AbstractEntity.POLICY_ADDED for enrichers; no use-case for that yet
--- a/core/src/main/java/org/apache/brooklyn/core/entity/internal/EntityConfigMap.java
+++ b/core/src/main/java/org/apache/brooklyn/core/entity/internal/EntityConfigMap.java
@@ -26,6 +26,13 @@ import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Set;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.common.base.Predicate;
+import com.google.common.collect.Maps;
+import com.google.common.collect.Sets;
+
 import org.apache.brooklyn.api.mgmt.ExecutionContext;
 import org.apache.brooklyn.api.mgmt.Task;
 import org.apache.brooklyn.config.ConfigInheritance;
@@ -35,7 +42,6 @@ import org.apache.brooklyn.core.config.Sanitizer;
 import org.apache.brooklyn.core.config.StructuredConfigKey;
 import org.apache.brooklyn.core.config.internal.AbstractConfigMapImpl;
 import org.apache.brooklyn.core.entity.AbstractEntity;
-import org.apache.brooklyn.core.location.AbstractLocation;
 import org.apache.brooklyn.util.collections.CollectionMerger;
 import org.apache.brooklyn.util.collections.MutableMap;
 import org.apache.brooklyn.util.core.config.ConfigBag;
@@ -44,12 +50,6 @@ import org.apache.brooklyn.util.core.flags.SetFromFlag;
 import org.apache.brooklyn.util.core.flags.TypeCoercions;
 import org.apache.brooklyn.util.core.internal.ConfigKeySelfExtracting;
 import org.apache.brooklyn.util.guava.Maybe;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
-import com.google.common.base.Predicate;
-import com.google.common.collect.Maps;
-import com.google.common.collect.Sets;
 
 public class EntityConfigMap extends AbstractConfigMapImpl {
 
@@ -96,7 +96,7 @@ public class EntityConfigMap extends AbstractConfigMapImpl {
         
         // TODO Confirm desired behaviour: I switched this around to get ownKey.getParentInheritance first.
         ConfigInheritance inheritance = ownKey.getParentInheritance();
-        if (inheritance==null) inheritance = key.getInheritance(); 
+        if (inheritance==null) inheritance = key.getParentInheritance();
         if (inheritance==null) {
             // TODO we could warn by introducing a temporary "ALWAYS_BUT_WARNING" instance
             inheritance = getDefaultParentInheritance(); 
@@ -132,7 +132,11 @@ public class EntityConfigMap extends AbstractConfigMapImpl {
 
         // Get own value
         if (((ConfigKeySelfExtracting<T>)key).isSet(ownConfig)) {
-            ownValue = Maybe.of(((ConfigKeySelfExtracting<T>)key).extractValue(ownConfig, exec));
+            Map<ConfigKey<?>, ?> ownCopy = null;
+            synchronized (ownConfig) {
+                ownCopy = MutableMap.copyOf(ownConfig);
+            }
+            ownValue = Maybe.of(((ConfigKeySelfExtracting<T>) key).extractValue(ownCopy, exec));
         } else if (localConfigBag.containsKey(key)) {
             // TODO configBag.get doesn't handle tasks/attributeWhenReady - it only uses TypeCoercions
             // Precedence ordering has changed; previously we'd prefer an explicit isSet(inheritedConfig)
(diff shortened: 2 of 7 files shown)


core/src/main/java/org/apache/brooklyn/core/entity/AbstractEntity.java
```diff
@@ -1698,6 +1698,7 @@ public void add(Policy policy) {
             CatalogUtils.setCatalogItemIdOnAddition(AbstractEntity.this, policy);
             policiesInternal.add((AbstractPolicy)policy);
             ((AbstractPolicy)policy).setEntity(AbstractEntity.this);
+            ConfigConstraints.assertValid(policy);
```
aledsage: Seems strange to add + activate the policy, and only then assert that it is valid. But I guess it might not be fully initialised before `setEntity` is called, for us to validate it?

Hopefully the policy was instantiated using the normal mechanism (so went through `InternalPolicyFactory.createPolicy`, which will have called `assertValid`.

Are there situations where this code catches something we were previously missing? If so, should we write a **test** for it?

core/src/main/java/org/apache/brooklyn/core/entity/internal/EntityConfigMap.java
```diff
@@ -132,7 +132,11 @@ public EntityConfigMap(AbstractEntity entity, Map<ConfigKey<?>, Object> storage)
 
         // Get own value
         if (((ConfigKeySelfExtracting<T>)key).isSet(ownConfig)) {
-            ownValue = Maybe.of(((ConfigKeySelfExtracting<T>)key).extractValue(ownConfig, exec));
+            Map<ConfigKey<?>, ?> ownCopy = null;
```
aledsage: Very minor: personal preference to not set its value to null when we're about to set its value immediately below that. Let the compiler tell us if someone changing it makes a mistake so the value no longer gets initialised.

core/src/main/java/org/apache/brooklyn/core/objs/AbstractEntityAdjunct.java
```diff
@@ -421,6 +421,7 @@ public void setDisplayName(String name) {
     public void setEntity(EntityLocal entity) {
         if (destroyed.get()) throw new IllegalStateException("Cannot set entity on a destroyed entity adjunct");
         this.entity = entity;
+        this.execution = ((EntityInternal) entity).getExecutionContext();
```
aledsage: I'm guessing you hit a problem that made you add this line - can we capture in a test what that problem was, rather than just fixing it?

I'm fine with merging this time, but for future reference I am really keen that we improve our test coverage for when we hit problems like this.

core/src/main/java/org/apache/brooklyn/core/objs/proxy/InternalPolicyFactory.java
```diff
@@ -125,7 +124,6 @@ public InternalPolicyFactory(ManagementContextInternal managementContext) {
             for (Map.Entry<ConfigKey<?>, Object> entry : spec.getConfig().entrySet()) {
                 pol.config().set((ConfigKey)entry.getKey(), entry.getValue());
             }
-            ConfigConstraints.assertValid(pol);
```
aledsage: Ah, I see you've removed the `assertValid` here. Why? Is that because the policy/enricher won't have finished initialising until `setEntity` etc is called? Is that an indication of a poorly implemented policy? Or the right thing to do?

core/src/main/java/org/apache/brooklyn/entity/group/SshCommandMembershipTrackingPolicy.java
```diff
@@ -0,0 +1,105 @@
...
+import org.apache.brooklyn.util.core.task.Tasks;
+import org.apache.brooklyn.util.exceptions.Exceptions;
+import org.apache.brooklyn.util.guava.Maybe;
+import org.apache.brooklyn.util.text.Strings;
+
+/**
+ * Policy which tracks membership of a group, and executes SSH commands
+ * on MEMBER{ADDED,REMOVED} events, as well as SERVICE_UP {true,false} for those members.
+ */
+public class SshCommandMembershipTrackingPolicy extends AbstractMembershipTrackingPolicy {
```
aledsage: test? (preferably using `RecordingSshTool`, so it's a unit test).

core/src/main/java/org/apache/brooklyn/entity/group/SshCommandMembershipTrackingPolicy.java
```diff
@@ -0,0 +1,105 @@
...
+        // Set things from the entities defined shell environment, overriding with our config
+        Map<String, Object> env = MutableMap.of();
+        env.putAll(MutableMap.copyOf(entity.config().get(BrooklynConfigKeys.SHELL_ENVIRONMENT)));
+        env.putAll(MutableMap.copyOf(config().get(BrooklynConfigKeys.SHELL_ENVIRONMENT)));
+
+        // Try to resolve the configuration in the env Map
+        try {
+            env = (Map<String, Object>) Tasks.resolveDeepValue(env, Object.class, ((EntityInternal) entity).getExecutionContext());
+        } catch (InterruptedException | ExecutionException e) {
+            Exceptions.propagateIfFatal(e);
```
aledsage: Never swallow exceptions. We should log the exception if we're not propagating it, yes?

core/src/main/java/org/apache/brooklyn/core/entity/AbstractEntity.java
grkvlt: No, this asserts that the `ConfigConstraints` on the policy are valid. It had to be moved to cope with the situation where some config keys had values that were resolvable only using the *entities* context. So it must happen *after* `setEntity()` is called on the policy. I fixed enrichers too, just for luck.

core/src/main/java/org/apache/brooklyn/core/entity/internal/EntityConfigMap.java
grkvlt: But the compiler won't compile if you don't? Or is that just my misleading Eclipse settings...?

core/src/main/java/org/apache/brooklyn/core/objs/AbstractEntityAdjunct.java
grkvlt: The problem was that `execution` is *NEVER* set otherwise, and it's used in `getContext()` that's what.

core/src/main/java/org/apache/brooklyn/core/objs/proxy/InternalPolicyFactory.java
grkvlt: Because of unresolvable config that needs to know the entity context, which is a valid thing.

core/src/main/java/org/apache/brooklyn/entity/group/SshCommandMembershipTrackingPolicy.java
grkvlt: OK!

core/src/main/java/org/apache/brooklyn/entity/group/SshCommandMembershipTrackingPolicy.java
grkvlt: probably should be `throw Exceptions.propagate()` then?

core/src/main/java/org/apache/brooklyn/entity/group/SshCommandMembershipTrackingPolicy.java
```diff
@@ -0,0 +1,105 @@
...
+    protected void onEntityEvent(EventType type, Entity entity) {
+        LOG.trace("Event {} received for {} in {}", new Object[] { type, entity, getGroup() });
+        String command = config().get(UPDATE_COMMAND);
+        if (Strings.isNonBlank(command)) {
+            execute(command);
+        }
+    }
+
+    public void execute(String command) {
+        Maybe<SshMachineLocation> machine = Machines.findUniqueMachineLocation(entity.getLocations(), SshMachineLocation.class);
```
neykov: Suggest to use `Locations.getLocationsCheckingAncestors(...)` here. For example when an entity is nested in a `SameServerEntity`.

core/src/main/java/org/apache/brooklyn/entity/group/SshCommandMembershipTrackingPolicy.java
```diff
@@ -0,0 +1,105 @@
...
+    public static final ConfigKey<String> UPDATE_COMMAND = ConfigKeys.newStringConfigKey("update.command", "Command to run on membership change events");
+
+    /**
+     * Called when a member is updated or group membership changes.
+     */
+    @Override
+    protected void onEntityEvent(EventType type, Entity entity) {
+        LOG.trace("Event {} received for {} in {}", new Object[] { type, entity, getGroup() });
+        String command = config().get(UPDATE_COMMAND);
+        if (Strings.isNonBlank(command)) {
```
neykov: Can you add `type` as an environment variable. Would be useful for scripts to know whether this is add/remove/change event.

grkvlt:
@aledsage @neykov I added a test and implemented the suggested changes; should be ready for merge now

core/src/test/java/org/apache/brooklyn/entity/group/SshCommandMembershipTrackingPolicyTest.java
```diff
@@ -0,0 +1,113 @@
...
+    @Override
+    public void tearDown() throws Exception {
+        super.tearDown();
+        RecordingSshTool.clear();
+    }
+
+    @Test
+    public void testCommandExecutedWithEnvironment() throws Exception {
+        TestEntity member = entityManager.createEntity(EntitySpec.create(TestEntity.class).parent(group));
+        Duration.seconds(1).countdownTimer().waitForExpiry();
```
aledsage: Why wait for a second? That might well not be long enough on Apache Jenkins for it to work every time! And it's too long on a normal laptop.

I'd have written:

```
protected void assertExecSizeEventually(final int expectedSize) {
    Asserts.succeedsEventually(new Runnable() {
        public void run() {
            assertEquals(RecordingSshTool.getExecCmds().size(), expectedSize);
        }});
}
```

Same below when you wait a second and assert that the size is 2 or 3. (hence why worth extracting it into a method).

aledsage:
LGTM - one very minor comment for how to dramatically reduce the test time from 3 seconds to a few milliseconds. It adds up to a surprisingly high number when too many of our tests did sleeps!

core/src/test/java/org/apache/brooklyn/entity/group/SshCommandMembershipTrackingPolicyTest.java
grkvlt: Good idea, have done exactly this

neykov:
LGTM from me as well, merging.

grkvlt:
Thanks @neykov

PR #288, opened 2016-08-03
BROOKLYN-323: Simplify Logout api

bostko:
- removed TEMPORARY_REDIRECT step
- fixed redirecting for firefox
- change logout call to a GET request

### DIFF
--- a/rest/rest-api/src/main/java/org/apache/brooklyn/rest/api/LogoutApi.java
+++ b/rest/rest-api/src/main/java/org/apache/brooklyn/rest/api/LogoutApi.java
@@ -18,6 +18,7 @@
  */
 package org.apache.brooklyn.rest.api;
 
+import javax.ws.rs.GET;
 import javax.ws.rs.POST;
 import javax.ws.rs.Path;
 import javax.ws.rs.PathParam;
@@ -27,22 +28,24 @@ import io.swagger.annotations.Api;
 import io.swagger.annotations.ApiOperation;
 import io.swagger.annotations.ApiParam;
 import io.swagger.annotations.ApiResponse;
-import io.swagger.annotations.ApiResponses;
 
 @Path("/logout")
 @Api("Logout")
 public interface LogoutApi {
-
     @POST
     @ApiOperation(value = "Request a logout and clean session")
-    @ApiResponses(value = {
-            @ApiResponse(code = 307, message = "Redirect to /logout/user, keeping the request method")
-    })
+    @ApiResponse(code = 401, message = "Unauthorized")
     Response logout();
 
     @POST
+    @Path("/redirect")
+    @ApiOperation(value = "Intermediate redirect for browsers in order to make sure browser will reload its DOM")
+    Response redirectToLogout();
+
+    @POST
     @Path("/{user}")
-    @ApiOperation(value = "Logout and clean session if matching user logged")
+    @ApiOperation(value = "Deprecated: Logout and clean session if matching user logged")
+    @Deprecated
     Response logoutUser(
         @ApiParam(value = "User to log out", required = true)
         @PathParam("user") final String user);
--- a/rest/rest-resources/src/main/java/org/apache/brooklyn/rest/resources/LogoutResource.java
+++ b/rest/rest-resources/src/main/java/org/apache/brooklyn/rest/resources/LogoutResource.java
@@ -37,32 +37,45 @@ public class LogoutResource extends AbstractBrooklynRestResource implements Logo
     @Context UriInfo uri;
 
     @Override
-    public Response logout() {
-        WebEntitlementContext ctx = (WebEntitlementContext) Entitlements.getEntitlementContext();
-        URI dest = uri.getBaseUriBuilder().path(LogoutApi.class).path(LogoutApi.class, "logoutUser").build(ctx.user());
+    public Response redirectToLogout() {
+        URI dest = uri.getBaseUriBuilder().path(LogoutApi.class).build();
 
-        // When execution gets here we don't know whether this is the first fetch of logout() or a subsequent one
-        // with a re-authenticated user. The only way to tell is compare if user names changed. So redirect to an URL
-        // which contains the user name.
-        return Response.status(Status.TEMPORARY_REDIRECT)
-                .header("Location", dest.toASCIIString())
+        // Return response with Javascript which will make an asynchronous POST request to the logout method.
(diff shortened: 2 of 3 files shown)


bostko:
Details about this change in https://issues.apache.org/jira/browse/BROOKLYN-323

rest/rest-api/src/main/java/org/apache/brooklyn/rest/api/LogoutApi.java
```diff
@@ -27,22 +28,20 @@
 import io.swagger.annotations.ApiOperation;
 import io.swagger.annotations.ApiParam;
 import io.swagger.annotations.ApiResponse;
-import io.swagger.annotations.ApiResponses;
 
 @Path("/logout")
 @Api("Logout")
 public interface LogoutApi {
 
-    @POST
```
bostko: Do you think I should keep a `@POST` method for backwards compatibility?

nakomis:
Tested alongside https://github.com/apache/brooklyn-ui/pull/30 and this resolves the issue on Windows 10 on both IE and Edge. Behaviour on Chrome on OS X remains unchanged

LGTM

sjcorbett:
@bostko Can you describe why we should change `/logout` from `POST` to `GET`? For one it means that if I can find a way to inject HTML into the page I can log every user that loads it out. It also risks browsers pre-fetching the page and users being unexpectedly logged out (as described in http://stackoverflow.com/questions/3521290/logout-get-or-post).

sjcorbett:
Test failure unrelated: SpecParameterUnwrappingTest.testParameterDefaultsUsedInConfig. Will check locally.

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/resources/LogoutResource.java
```diff
@@ -37,32 +37,44 @@
...
-            doLogout();
-
-            return Response.status(Status.UNAUTHORIZED)
-                    .header("WWW-Authenticate", "Basic realm=\"webconsole\"")
-                    .build();
-        } else {
-            return Response.temporaryRedirect(uri.getAbsolutePathBuilder().replacePath("/").build()).build();
-        }
+        return Response.status(Status.FOUND)
+                .header("Location", uri.getBaseUriBuilder().path(LogoutApi.class).path(LogoutApi.class, "logout").build())
```
sjcorbett: Are both calls of `path` required?

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/resources/LogoutResource.java
```diff
@@ -37,32 +37,44 @@
...
 
     @Override
+    public Response redirectToLogout() {
+        URI dest = uri.getBaseUriBuilder().path(LogoutApi.class).build();
+
+        return Response.status(Status.OK)
+                .entity(String.format("<!DOCTYPE html>\n<body>\n" +
+                        "<script>\n" +
+                        "(function(c){var a=new window.XMLHttpRequest;" +
+                        // https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/open
```
sjcorbett: Could you also add a comment describing what the purpose of the two lines below? Inconvenient to have to piece it together from the Mozilla link. And (nitpick) is there any reason it needs to be so densely formatted?

sjcorbett:
Couple of questions / suggestions to answers then should be good to merge.

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/resources/LogoutResource.java
```diff
@@ -37,32 +37,45 @@
...
+
+        // Return response with Javascript which will make an asynchronous POST request to the logout method.
+        // (When calling logout it is important to use wrong username and password in order to make browser forget the old ones)
+        return Response.status(Status.OK)
+                .entity(String.format("<!DOCTYPE html>\n<body>\n" +
+                        "<script>\n" +
+                        "var a=new window.XMLHttpRequest;" +
+                        "a.open('POST','%1$s',0,'user','wrong'+(new Date).getTime().toString());a.send(\"\");\n" +
+                        "window.location.href='/';</script></body>", dest.toASCIIString()))
+                .build();
```
neykov: No need to go through an intermediate page here. Could execute the same Javascript in https://github.com/apache/brooklyn-ui/pull/30/files#diff-d317a26aaa3190463c4c316d72ee9b6cR192 and remove the REST call.

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/resources/LogoutResource.java
```diff
@@ -37,32 +37,45 @@
...
+        }
+
+        URI dest = uri.getBaseUriBuilder().build();
+
+        return Response.status(Status.UNAUTHORIZED)
+                .header("WWW-Authenticate", "Basic realm=\"webconsole\"")
+                // For Status 403, HTTP Location header may be omitted.
+                // Location is best to be used for http status 302 https://tools.ietf.org/html/rfc2616#section-10.3.3
                 .header("Location", dest.toASCIIString())
+                .entity("<script>window.location.replace(\"/\");</script>")
```
neykov: I think we should keep previous behaviour here and for `logoutUser`. Changes in https://github.com/apache/brooklyn-ui/pull/30 don't actually use the `lougout` call anyway - they rely on javascript to invalidate the authentication header.
Why it's needed (for clients that use it):
  * a client posts to /logout
  * server returns 401
  * browser displays an authentication dialog
  * browser retries request with new credentials (against /logout)
  * server return 401 (because that's all it does)
  * credentials appear to be invalid

How current code fixes this:
  * a client posts to /logout
  * server redirects to /logout/username
  * server returns 401
  * browser display an authentication dialog
  * browser retries request with new credentials (against /logout/username)
  * server returns 200 if new username doesn't match username in URL or 401 if it matches

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/resources/LogoutResource.java
bostko: This is needed to reload the whole DOM.

bostko:
Thank you @neykov . 
Closing PR since all changes are now made in brooklyn-ui https://github.com/apache/brooklyn-ui/pull/30

PR #289, opened 2016-08-04
Fix non-deterministic test of SpecParameter by choosing correct entity

grkvlt:
Fixes `SpecParameterUnwrappingTest#testParameterDefaultsUsedInConfig()`


aledsage:
LGTM - the previous version of the test is most confusing as to what the expected behaviour was! The new assertions look much more correct. Merging now.

PR #291, opened 2016-08-05
Conveniences to make groups easier to configure in YAML

alasdairhodge:
* Extends `$brooklyn:object()` syntax to support parameterised constructors and static factory methods
* Adds variants of `EntityPredicates.config*()` and `attribute*()` that accept strings
* Adds variants of `EntityFunctions.config()` and `attribute()` that accept strings
* Adds variants of `EntityFunctions.config()` and `attribute()` that accept format specifier

### DIFF
--- a/camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/methods/BrooklynDslCommon.java
+++ b/camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/methods/BrooklynDslCommon.java
@@ -19,6 +19,7 @@
 package org.apache.brooklyn.camp.brooklyn.spi.dsl.methods;
 
 import static com.google.common.base.Preconditions.checkNotNull;
+import static org.apache.brooklyn.camp.brooklyn.spi.dsl.DslUtils.resolved;
 
 import java.util.Arrays;
 import java.util.Iterator;
@@ -38,10 +39,8 @@ import org.apache.brooklyn.camp.brooklyn.BrooklynCampReservedKeys;
 import org.apache.brooklyn.camp.brooklyn.spi.creation.BrooklynYamlTypeInstantiator;
 import org.apache.brooklyn.camp.brooklyn.spi.creation.EntitySpecConfiguration;
 import org.apache.brooklyn.camp.brooklyn.spi.dsl.BrooklynDslDeferredSupplier;
-import org.apache.brooklyn.camp.brooklyn.spi.dsl.DslUtils;
 import org.apache.brooklyn.camp.brooklyn.spi.dsl.methods.DslComponent.Scope;
 import org.apache.brooklyn.core.config.external.ExternalConfigSupplier;
-import org.apache.brooklyn.core.entity.AbstractEntity;
 import org.apache.brooklyn.core.entity.EntityDynamicType;
 import org.apache.brooklyn.core.entity.EntityInternal;
 import org.apache.brooklyn.core.mgmt.internal.ExternalConfigSupplierRegistry;
@@ -66,6 +65,7 @@ import org.slf4j.LoggerFactory;
 
 import com.google.common.base.Function;
 import com.google.common.base.Objects;
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Iterables;
 import com.google.common.collect.Lists;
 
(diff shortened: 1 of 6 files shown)


camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/methods/BrooklynDslCommon.java
```diff
@@ -327,25 +337,74 @@ public FormatString(String pattern, Object[] args) {
...
+        private final String factoryMethodName;
+        private final Class<?> type;
+        private final List<Object> constructorArgs, factoryMethodArgs;
+        private final Map<String, Object> fields, config;
+
+        public DslObject(
+                String typeName,
+                List<Object> constructorArgs,
+                Map<String, Object> fields,
+                Map<String, Object> config) {
```
aledsage: I'd use `Map<String, ?>` if we can get away with it - makes it easier for the caller.
But this is pretty internal stuff, so I don't really care.

camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/methods/BrooklynDslCommon.java
```diff
@@ -384,24 +445,48 @@ public Object apply(List<Object> input) {
...
+            } catch (Exception e) {
+                throw Exceptions.propagate(e);
+            }
+        }
+
+        public static Object create(Class<?> type, String factoryMethodName, List<Object> factoryMethodArgs, Map<String,?> fields, Map<String,?> config) {
+            try {
+                Object bean;
+                try {
+                    bean = TypeCoercions.coerce(fields, type);
```
aledsage: I don't follow this line. Why are we trying to coerce the fields (which is a map) to the expected object type?

Do we expect to support someone using `$brooklyn:object: { type: Map, object.fields: { a: b } }` to instantiate a map with contents `{a: b}`?!

camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/ObjectsYamlTest.java
```diff
@@ -275,6 +288,57 @@ public void testBrooklynObjectWithFunction() throws Exception {
         Assert.assertTrue(testObjectObject instanceof SimpleTestPojo, "Expected a SimpleTestPojo: "+testObjectObject);
     }
 
+    @Test
+    public void testBrooklynObjectWithParameterisedConstructor() throws Exception {
```
aledsage: Looks good. Would also be nice to have a test that used a deferred supplier. But given there are no existing tests in `ObjectsYamlTest` that do that, I see why you didn't either!

We can do that in a separate PR.

aledsage:
@alasdairhodge looks good. Just one comment that I'm interested in before this is merged: what is the purpose of calling `TypeCoercions.coerce(fields, type)`?

camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/methods/BrooklynDslCommon.java
alasdairhodge: Yeah, I don't really know why that was in the previous version either: I can't think of a sensible use-case, and the unit tests don't cover it. I opted to preserve it rather than risk breaking some obscure corner case, but I'd be *very* happy to see it removed.

camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/ObjectsYamlTest.java
alasdairhodge: +1

camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/methods/BrooklynDslCommon.java
alasdairhodge: I originally attempted this throughout, but it caused issues with the `Map.put` calls on lines 435, 437, 443, 445, as well as the call to `Reflections.invokeMethodFromArgs` on line 483.

alasdairhodge:
Thanks, Aled. Suggest improving test coverage and tidying up the bizarre coersion stuff in a subsequent PR.

alasdairhodge:
Just one other thing: do we have a naming convention for config keys? Don't want to start a religious war over the camelCase `factoryMethod` ;o)

aledsage:
@alasdairhodge we don't have a consistent convention for config key names. We have three approaches being used:
* dot-separated - normally used as a hierarchy, but sometimes dots used in the same way as camelCase (but it shouldn't be)
* underscores
* camelCase

Agree would be nice to have consistency! I'm fine with your camelCase.

Merging now.

PR #292, opened 2016-08-08
Shared location customizer

duncangrant:
A location customizer that creates a single shared security group for each entity it is applied to as well as an individual entity.  The shared entity allows full tcp and udp traffic within members of the group.

This PR includes 85c65ed078060cd01f74732b7cc4bddef8a58006 from @bostko 's #276.

The implementation use the JcloudsLocationSecurityGroupCustomizer but has a no arg constructor so can be used in yaml.

### DIFF
--- a/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/SharedLocationSecurityGroupCustomizer.java
+++ b/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/SharedLocationSecurityGroupCustomizer.java
@@ -1,0 +1,185 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.brooklyn.location.jclouds.networking;
+
+import java.util.List;
+
+import javax.annotation.Nullable;
+
+import org.apache.brooklyn.location.jclouds.BasicJcloudsLocationCustomizer;
+import org.apache.brooklyn.location.jclouds.JcloudsLocation;
+import org.apache.brooklyn.location.jclouds.JcloudsMachineLocation;
+import org.apache.brooklyn.util.net.Cidr;
+import org.apache.brooklyn.util.net.Networking;
+import org.jclouds.compute.ComputeService;
+import org.jclouds.compute.domain.Template;
+import org.jclouds.net.domain.IpPermission;
+import org.jclouds.net.domain.IpProtocol;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Function;
+import com.google.common.base.Strings;
+import com.google.common.base.Suppliers;
+import com.google.common.collect.FluentIterable;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Range;
+import com.google.common.collect.RangeSet;
+
+/**
+ * Configures a shared security group on Jclouds locations
+ * <p>
+ * Is based on {@link JcloudsLocationSecurityGroupCustomizer} but can be instantiated
+ * in yaml e.g.
+ * <p>
+ * <pre>
+ * {@code
+ * services:
+ * - type: org.apache.brooklyn.entity.software.base.EmptySoftwareProcess
+ *   brooklyn.config:
+ *     provisioning.properties:
+ *       customizers:
+ *       - $brooklyn:object:
+ *         type: org.apache.brooklyn.location.jclouds.networking.SharedLocationSecurityGroupCustomizer
+ *         object.fields: {locationName: "test", tcpPortRanges: ["900-910", "915", "22"], cidr: "82.40.153.101/24"}
+ * }
+ * </pre>
+*/
+public class SharedLocationSecurityGroupCustomizer extends BasicJcloudsLocationCustomizer {
+
+    private String locationName = null;
+
+    private String cidr = null;
+
+    /**
+     * We track inbound ports from the template to open them once we've created the new
+     * security groups
+     */
+    private int[] inboundPorts;
+
+    private RangeSet<Integer> tcpPortRanges;
+    private RangeSet<Integer> udpPortRanges;
+
+    /**
+     * The location name is appended to the name of the shared SG - use if you need distinct shared SGs within the same location
+     *
+     * @param locationName appended to the name of the SG
+     */
+    public void setLocationName(String locationName) {
+        this.locationName = locationName;
+    }
+
+    /**
+     * Extra ports to be opened on the entities in the SG
+     *
+     * @param tcpPortRanges
+     */
+    public void setTcpPortRanges(List<String> tcpPortRanges) {
+        this.tcpPortRanges = Networking.portRulesToRanges(tcpPortRanges);
+    }
+
+    public void setUdpPortRanges(ImmutableList<String> udpPortRanges) {
+        this.udpPortRanges = Networking.portRulesToRanges(udpPortRanges);
+    }
+
+    /**
+     * Set to restict the range that the ports that are to be opened can be accessed from
+     *
+     * @param cidr
+     */
+    public void setCidr(String cidr) {
+        this.cidr = cidr;
+    }
+
+    @Override
+    public void customize(JcloudsLocation location, ComputeService computeService, Template template) {
+        super.customize(location, computeService, template);
+
+        inboundPorts = template.getOptions().getInboundPorts();
+
+        final JcloudsLocationSecurityGroupCustomizer instance = getInstance(getSharedGroupId(location));
+        if (cidr != null) instance.setSshCidrSupplier(Suppliers.ofInstance(new Cidr(cidr)));
+        instance.customize(location, computeService, template);
+    }
+
+    @Override
+    public void customize(JcloudsLocation location, ComputeService computeService, JcloudsMachineLocation machine) {
+        super.customize(location, computeService, machine);
+
+        final JcloudsLocationSecurityGroupCustomizer instance = getInstance(getSharedGroupId(location));
+
+        ImmutableList.Builder<IpPermission> builder = ImmutableList.<IpPermission>builder();
+        builder.addAll(getIpPermissions(instance, tcpPortRanges, IpProtocol.TCP));
+        builder.addAll(getIpPermissions(instance, udpPortRanges, IpProtocol.UDP));
+
+        if (inboundPorts != null) {
+            for (int inboundPort : inboundPorts) {
+                IpPermission ipPermission = IpPermission.builder()
+                        .fromPort(inboundPort)
+                        .toPort(inboundPort)
+                        .ipProtocol(IpProtocol.TCP)
+                        .cidrBlock(instance.getBrooklynCidrBlock())
+                        .build();
+                builder.add(ipPermission);
+            }
+        }
+        instance.addPermissionsToLocation(machine, builder.build());
+    }
+
+    private List<IpPermission> getIpPermissions(JcloudsLocationSecurityGroupCustomizer instance, RangeSet<Integer> portRanges, IpProtocol protocol) {
+        List<IpPermission> ipPermissions = ImmutableList.<IpPermission>of();
+        if (portRanges != null) {
+             ipPermissions =
+                    FluentIterable
+                            .from(portRanges.asRanges())
+                            .transform(portRangeToPermission(instance, protocol))
+                            .toList();
+        }
+        return ipPermissions;
+    }
+
+    private Function<Range<Integer>, IpPermission> portRangeToPermission(final JcloudsLocationSecurityGroupCustomizer instance, final IpProtocol protocol) {
+        return new Function<Range<Integer>, IpPermission>() {
+            @Nullable
+            @Override
+            public IpPermission apply(@Nullable Range<Integer> integerRange) {
+                IpPermission extraPermission = IpPermission.builder()
+                        .fromPort(integerRange.lowerEndpoint())
+                        .toPort(integerRange.upperEndpoint())
+                        .ipProtocol(protocol)
+                        .cidrBlock(instance.getBrooklynCidrBlock())
+                        .build();
+                return extraPermission;
+            }
+        };
+    }
+
+    private String getSharedGroupId(JcloudsLocation location) {
+        return (Strings.isNullOrEmpty(locationName))
+                ? location.getId()
+                : locationName + "-" + location.getId();
+    }
+
+    @VisibleForTesting
+    JcloudsLocationSecurityGroupCustomizer getInstance(String sharedGroupId) {
+        return JcloudsLocationSecurityGroupCustomizer.getInstance(sharedGroupId);
+    }
+}
+
+
+
--- a/locations/jclouds/src/test/java/org/apache/brooklyn/location/jclouds/networking/SharedLocationSecurityGroupCustomizerTest.java
+++ b/locations/jclouds/src/test/java/org/apache/brooklyn/location/jclouds/networking/SharedLocationSecurityGroupCustomizerTest.java
@@ -1,0 +1,150 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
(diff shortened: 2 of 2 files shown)


core/src/main/java/org/apache/brooklyn/util/core/BrooklynNetworkUtils.java
```diff
@@ -19,24 +19,52 @@
...
+        return TypeCoercions.coerce(JavaGroovyEquivalents.elvis(BrooklynServiceAttributes.LOCALHOST_IP_ADDRESS.getValue(),
                 Networking.getLocalHost()), InetAddress.class);
     }
 
+    public static RangeSet<Integer> portRulesToRanges(Collection<String> portRules) {
+        RangeSet<Integer> result = TreeRangeSet.create();
+        for (String portRule : portRules) {
+            if (portRule.contains("-")) {
+                String[] fromTo = portRule.split("-");
+                Preconditions.checkState(fromTo.length == 2);
```
aledsage: I know this came from @bostko 's commit, but if might merge it here...

I'd much prefer us to give a message. This will currently just throw an `IllegalStateException` with no messsage. Someone would need to look at the code to understand what went wrong. I'd change this to something like:

    Preconditions.checkState(fromTo.length == 2, "Invalid port range '%s'", portRule);

core/src/main/java/org/apache/brooklyn/util/core/BrooklynNetworkUtils.java
```diff
@@ -19,24 +19,52 @@
...
+    /**
+     * returns a IP address for localhost paying attention to a system property to prevent lookup in some cases
+     */
     public static InetAddress getLocalhostInetAddress() {
-        return TypeCoercions.coerce(JavaGroovyEquivalents.elvis(BrooklynServiceAttributes.LOCALHOST_IP_ADDRESS.getValue(), 
+        return TypeCoercions.coerce(JavaGroovyEquivalents.elvis(BrooklynServiceAttributes.LOCALHOST_IP_ADDRESS.getValue(),
                 Networking.getLocalHost()), InetAddress.class);
     }
 
+    public static RangeSet<Integer> portRulesToRanges(Collection<String> portRules) {
```
aledsage: This should probably go in `Networking` instead of `BrooklynNetworkUtils`.

The difference between the two classes seems to be:

* `Networking` is for generic network utility methods.
* `BrooklynNetworkUtils` is for utility methods that rely on some other part(s) of Brooklyn, or seem too custom in how they are used/configured to be considered a "common utility". For example `getLocalhostExternalIp()` will load the resource `external-ip-address-resolvers.txt` that tells us about external service(s) to use, such as http://jsonip.com/, and will contact that in a time-limited way to get + cache the ip).

Worth adding javadoc to the two classes to that affect, and to include in the latter a reference to `{@link org.apache.brooklyn.util.net.Networking}`.

core/src/main/java/org/apache/brooklyn/util/core/BrooklynNetworkUtils.java
```diff
@@ -19,24 +19,52 @@
...
+                result.add(closedRange(fromTo[0], fromTo[1]));
+            } else {
+                result.add(closedRange(portRule, portRule));
+            }
+        }
+        return result;
+    }
+
+    private static Range<Integer> closedRange(String from, String to) {
+        return Range.closed(Integer.parseInt(from), Integer.parseInt(to));
```
aledsage: Should we fail-fast with numbers that are out of range (i.e. negative or greater than `Networking.MAX_PORT_NUMBER`)? Any use-case for a kind of port that can exceed 65535 (I don't think there is)?

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/SharedLocationSecurityGroupCustomizer.java
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    @Override
+    public void customize(JcloudsLocation location, ComputeService computeService, TemplateOptions templateOptions) {
+        super.customize(location, computeService, templateOptions);
+        inboundPorts = templateOptions.getInboundPorts();
+    }
+
+    @Override
+    public void customize(JcloudsLocation location, ComputeService computeService, Template template) {
```
neykov: Nit: One would usually override one or the other method (above one and this one) - they are called in succession with the above one being just a convenience, see https://github.com/apache/brooklyn-server/blob/master/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java#L1540-L1541.

bostko:
@duncangrant I addressed @aledsage 's comments in my PR.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/SharedLocationSecurityGroupCustomizer.java
```diff
@@ -0,0 +1,172 @@
...
+ * Is based on {@link JcloudsLocationSecurityGroupCustomizer} but can be instantiated
+ * in yaml e.g.
+ * <p>
+ * services:
+ * - type: org.apache.brooklyn.entity.software.base.EmptySoftwareProcess
+ * brooklyn.config:
+ * provisioning.properties:
+ * customizers:
+ * - $brooklyn:object:
+ * type: org.apache.brooklyn.location.jclouds.networking.SharedLocationSecurityGroupCustomizer
```
aledsage: indent seems wrong.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/SharedLocationSecurityGroupCustomizer.java
```diff
@@ -0,0 +1,172 @@
...
+ * object.fields: {locationName: "test", portRanges: ["900-910","915", "22"], cidr: "82.40.153.101/24"}
```
aledsage: nit-pick: consistency of whether there's a space after the comma.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/SharedLocationSecurityGroupCustomizer.java
aledsage: +1; you can hopefully do `template.getTemplateOptions().getInboundPorts()` here.

The other method is intended for when you want to modify the templateOptions, before it is used.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/SharedLocationSecurityGroupCustomizer.java
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    @Override
+    public void customize(JcloudsLocation location, ComputeService computeService, JcloudsMachineLocation machine) {
+        super.customize(location, computeService, machine);
+
+        final JcloudsLocationSecurityGroupCustomizer instance = getInstance(getSharedGroupId(location));
+
+        if (portRanges != null) {
+            RangeSet<Integer> portRanges = BrooklynNetworkUtils.portRulesToRanges(this.portRanges);
```
aledsage: I'd parse these when they are first set, rather than when they are about to be used - fail fast (rather than after someone has waited for their VM to be provisioned).

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/SharedLocationSecurityGroupCustomizer.java
```diff
@@ -0,0 +1,172 @@
...
+import com.google.common.collect.Range;
+import com.google.common.collect.RangeSet;
+
+/**
+ * Configures a shared security group on Jclouds locations
+ * <p>
+ * Is based on {@link JcloudsLocationSecurityGroupCustomizer} but can be instantiated
+ * in yaml e.g.
+ * <p>
+ * services:
```
aledsage: Surround with the `<pre>` etc - for an example, see javadoc of https://github.com/apache/brooklyn-server/blob/master/core/src/main/java/org/apache/brooklyn/feed/ssh/SshFeed.java

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/SharedLocationSecurityGroupCustomizer.java
```diff
@@ -0,0 +1,172 @@
...
+        }
+    }
+
+    private String getSharedGroupId(JcloudsLocation location) {
+        return (Strings.isNullOrEmpty(locationName))
+                ? location.getId()
+                : locationName + "-" + location.getId();
+    }
+
+    //Override-able in Tests
```
aledsage: I like the Guava annotation `@VisibleForTesting`

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/SharedLocationSecurityGroupCustomizer.java
```diff
@@ -0,0 +1,172 @@
...
+ * services:
+ * - type: org.apache.brooklyn.entity.software.base.EmptySoftwareProcess
+ * brooklyn.config:
+ * provisioning.properties:
+ * customizers:
+ * - $brooklyn:object:
+ * type: org.apache.brooklyn.location.jclouds.networking.SharedLocationSecurityGroupCustomizer
+ * object.fields: {locationName: "test", portRanges: ["900-910","915", "22"], cidr: "82.40.153.101/24"}
+ */
+public class SharedLocationSecurityGroupCustomizer extends BasicJcloudsLocationCustomizer {
```
aledsage: This is good for this PR, but should we push any of this functionality into `JcloudsLocationSecurityGroupCustomizer` (which this delegates to)?

Is the JcloudsLocationSecurityGroupCustomizer idempotent for if we try adding the same port range multiple times?

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/SharedLocationSecurityGroupCustomizer.java
```diff
@@ -0,0 +1,172 @@
...
+                    FluentIterable
+                            .from(portRanges.asRanges())
+                            .transform(new Function<Range<Integer>, IpPermission>() {
+                                @Nullable
+                                @Override
+                                public IpPermission apply(@Nullable Range<Integer> integerRange) {
+                                    IpPermission extraPermission = IpPermission.builder()
+                                            .fromPort(integerRange.lowerEndpoint())
+                                            .toPort(integerRange.upperEndpoint())
+                                            .ipProtocol(IpProtocol.TCP)
```
aledsage: Can we support both tcp and upd ports? I presume it would be fairly easy (e.g. change the `portRanges` field and setter to be `tcpPortRanges`, and add a similar field+setter for `udpPortRanges`)?

locations/jclouds/src/test/java/org/apache/brooklyn/location/jclouds/networking/SharedLocationSecurityGroupCustomizerTest.java
```diff
@@ -0,0 +1,121 @@
...
+public class SharedLocationSecurityGroupCustomizerTest {
+
+    TestSharedLocationSecurityGroupCustomizer customizer;
+    JcloudsLocationSecurityGroupCustomizer sgCustomizer;
+    ComputeService computeService;
+    Location location;
+    SecurityGroupExtension securityApi;
+    private JcloudsLocation jcloudsLocation = new JcloudsLocation(MutableMap.of("deferConstruction", true));
+
+    @BeforeMethod
```
aledsage: personal preference to always use `@BeforeMethod(alwaysRun=true)`. If someone later adds (or marks) a test as `group="Integration"` then the setUp wouldn't run (but it would work in the Eclipse IDE as that seems to always run the `@BeforeMethod` no matter what the test group).

locations/jclouds/src/test/java/org/apache/brooklyn/location/jclouds/networking/SharedLocationSecurityGroupCustomizerTest.java
```diff
@@ -0,0 +1,121 @@
...
+        verify(sgCustomizer, never()).setSshCidrSupplier(any(Supplier.class));
+    }
+
+
+    @Test
+    public void testPermissionsSetFromIpRanges() {
+        customizer.setPortRanges(ImmutableList.of("99-100"));
+        when(sgCustomizer.getBrooklynCidrBlock()).thenReturn("10.10.10.10/24");
+        customizer.customize(jcloudsLocation, computeService, mock(JcloudsMachineLocation.class));
+        ArgumentCaptor<List> listArgumentCaptor = ArgumentCaptor.forClass(List.class);
```
aledsage: Nice unit tests! I hadn't seen `ArgumentCaptor` before. Looks really useful!

aledsage:
Thanks @duncangrant - finished reviewing.

For the first commit, @bostko has incorporated those comments. I've created https://github.com/apache/brooklyn-server/pull/298 to get that merged asap (and I have added a few minor tweaks to that PR as well).

For your `SharedLocationSecurityGroupCustomizer`, it looks good. a few very minor comments. The only significant one is whether we add udp support as well. The reason I think it's worth it now is that it will affect our naming: `portRanges` will be ambiguous if we add udp in the future.

If you want, you could just name it "tcpPortRanges" now, and defer adding udp. Up to you.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/SharedLocationSecurityGroupCustomizer.java
duncangrant: I don't think there is anything worth pushing into JcloudsLocationSecurityGroupCustomizer.  I would have like to just modify it to have a no args constructor that would let us work with it in yaml but couldn't quite see a way to do it without changing its behaviour.

It tries to be idempotent but then it also seems to catch (and ignore) the error if it does try to add a range a second time so possibly that can happen but I can't see where.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/SharedLocationSecurityGroupCustomizer.java
duncangrant: @grkvlt suggested that too so I'll add it.

aledsage:
@duncangrant I've merged #298 which has the first commit. Can you rebase against master please.

duncangrant:
@aledsage @neykov I think I've addressed all your comments

aledsage:
@duncangrant there are merge conflicts - can you rebase against master please? (Jenkins won't run the tests until that conflict is resolved).

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/SharedLocationSecurityGroupCustomizer.java
```diff
@@ -0,0 +1,190 @@
...
+        }
+        instance.addPermissionsToLocation(machine, builder.build());
+    }
+
+    private List<IpPermission> getIpPermissions(JcloudsLocationSecurityGroupCustomizer instance, RangeSet<Integer> tcpPortRanges, IpProtocol protocol) {
+        List<IpPermission> ipPermissions = ImmutableList.<IpPermission>of();
+        if (tcpPortRanges != null) {
+             ipPermissions =
+                    FluentIterable
+                            .from(tcpPortRanges.asRanges())
```
aledsage: Can you rename this variable to `portRanges`, because it is called for both tcp and udp so is a bit confusing.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/SharedLocationSecurityGroupCustomizer.java
```diff
@@ -0,0 +1,190 @@
...
+                            .transform(portRangeToPermission(instance, protocol))
+                            .toList();
+        }
+        return ipPermissions;
+    }
+
+    private Function<Range<Integer>, IpPermission> portRangeToPermission(final JcloudsLocationSecurityGroupCustomizer instance, final IpProtocol udp) {
```
aledsage: Can you rename the field `udp` to `protocol` as it could have the value TCP or UDP, so is a bit confusing.

aledsage:
@duncangrant code looks good. A couple of minor parameter renames would make it clearer. Once merge conflicts are dealt with and jenkins confirms tests pass, then good to merge (ideally with the renames also done).

duncangrant:
@aledsage I've done the renaming and fixed the merge conflicts

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/SharedLocationSecurityGroupCustomizer.java
```diff
@@ -0,0 +1,190 @@
...
+
+        if (cidr != null) instance.setSshCidrSupplier(Suppliers.ofInstance(new Cidr(cidr)));
+
+        instance.customize(location, computeService, template);
+    }
+
+    @Override
+    public void customize(JcloudsLocation location, ComputeService computeService, JcloudsMachineLocation machine) {
+        super.customize(location, computeService, machine);
+
```
grkvlt: why

are

there

lots

of blank lines?

duncangrant:
I've removed some whitespace and squashed commits.  @grkvlt Do you think we can merge this now?

aledsage:

PR #293, opened 2016-08-09
Move tests temporary folders to the target folder

neykov:
On Windows some leftover files remain in the `osgi/cache` folder, triggering the rat plugin failure. It's a temporary folder so move it to the build folder.


aledsage:
LGTM; merging.

PR #295, opened 2016-08-10
Change conditional entity to reflect service state of optional child

grkvlt:
Adds an enricher that propagates `service.isUp` from the child entity, if one is created. This allows other entities to use the `service.isUp` sensor of the `ConditionalEntity` as a latch, without needing to know the identity (or even presence) of the child. A configuration setting enables propagation of other sensors from the child.

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/entity/stock/ConditionalEntityImpl.java
+++ b/core/src/main/java/org/apache/brooklyn/entity/stock/ConditionalEntityImpl.java
@@ -33,10 +33,12 @@ import org.apache.brooklyn.core.entity.BrooklynConfigKeys;
 import org.apache.brooklyn.core.entity.Entities;
 import org.apache.brooklyn.core.entity.lifecycle.Lifecycle;
 import org.apache.brooklyn.core.entity.lifecycle.ServiceStateLogic;
+import org.apache.brooklyn.core.entity.lifecycle.ServiceStateLogic.ComputeServiceIndicatorsFromChildrenAndMembers;
 import org.apache.brooklyn.core.entity.trait.Startable;
 import org.apache.brooklyn.core.location.Locations;
 import org.apache.brooklyn.enricher.stock.Enrichers;
 import org.apache.brooklyn.util.collections.MutableSet;
+import org.apache.brooklyn.util.collections.QuorumCheck;
 import org.apache.brooklyn.util.exceptions.Exceptions;
 import org.apache.brooklyn.util.time.Duration;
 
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/apache/brooklyn/entity/stock/ConditionalEntityImpl.java
```diff
@@ -19,25 +19,72 @@
...
+                        enrichers().add(Enrichers.builder().propagatingAllButUsualAnd().from(child).build());
+                    } else {
+                        enrichers().add(Enrichers.builder().propagating(sensors).from(child).build());
+                    }
+                }
+            }
+
+            // Start child if create flag is set; otherwise just set service.isUp
+            if (Boolean.TRUE.equals(create)) {
+                Entities.invokeEffectorWithArgs(this, child, Startable.START, locations).getUnchecked();
```
neykov: Why did you have to pass locations explicitly, won't it pick them from the parent by default? What if there's a location embedded in the spec?

core/src/main/java/org/apache/brooklyn/entity/stock/ConditionalEntityImpl.java
```diff
@@ -19,25 +19,72 @@
...
+            addLocations(locations);
+            locations = Locations.getLocationsCheckingAncestors(locations, this);
+            LOG.info("Starting entity "+this+" at "+locations);
+
+            // Child not yet created; Entity spec is present; Create flag is true
+            if (child == null && spec != null && Boolean.TRUE.equals(create)) {
+                child = addChild(EntitySpec.create(spec));
+                sensors().set(CONDITIONAL_ENTITY, child);
+
+                // Add enrichers for sensor propagateion
```
neykov: Typo propagateion

core/src/main/java/org/apache/brooklyn/entity/stock/ConditionalEntityImpl.java
```diff
@@ -19,25 +19,72 @@
...
+                enrichers().add(Enrichers.builder().propagating(Startable.SERVICE_UP).from(child).build());
```
neykov: This approach to propagating the `SERVICE_UP` sensor could in theory conflict with the logic which updates it based on the `service.notUp.indicators` key. Suggest using instead:
```
        enrichers().add(ServiceStateLogic.newEnricherFromChildrenUp()
                .suppressDuplicates(true)
                .configure(ComputeServiceIndicatorsFromChildrenAndMembers.UP_QUORUM_CHECK, QuorumCheck.QuorumChecks.all()));
```

It will add an entry to the `notUp` map if there are any problems with the children and update `state.isUp` correspondingly.

If you would like additionally to propagate the running state (i.e. wait for children startup) could replace the above with
```
        enrichers().add(ServiceStateLogic.newEnricherFromChildren()
                .suppressDuplicates(true)
                .configure(ComputeServiceIndicatorsFromChildrenAndMembers.UP_QUORUM_CHECK, QuorumCheck.QuorumChecks.all()));
```

core/src/main/java/org/apache/brooklyn/entity/stock/ConditionalEntityImpl.java
```diff
@@ -19,25 +19,72 @@
...
+                        enrichers().add(Enrichers.builder().propagating(sensors).from(child).build());
+                    }
+                }
+            }
+
+            // Start child if create flag is set; otherwise just set service.isUp
+            if (Boolean.TRUE.equals(create)) {
+                Entities.invokeEffectorWithArgs(this, child, Startable.START, locations).getUnchecked();
+            } else {
+                sensors().set(Attributes.SERVICE_UP, true);
```
neykov: Similar to the previous comment on `SERVICE_UP` propagation you could control the sensor value by adding/removing values in the `notUp` map. `AbstractApplication` is a good example of the approach.

neykov:
LGTM, could improve `SERVICE_UP` handling by following the `AbstractApplication` example.

grkvlt:
@neykov I added the enricher you suggested, it looks like a better way of doing things, thanks

PR #299, opened 2016-08-12
fix extra ssh key data

ahgittin:


locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -1013,10 +1013,22 @@ protected MachineLocation obtainOnce(ConfigBag setup) throws NoMachinesAvailable
...
-                                "Authorizing ssh keys",
+                                "Authorizing ssh keys from URLs",
                                 ImmutableList.of(new AuthorizeRSAPublicKeys(extraKeyDataToAuth).render(org.jclouds.scriptbuilder.domain.OsFamily.UNIX)));
                     }
                 }
+                
+                String extraKeyDataToAuth = setup.get(EXTRA_PUBLIC_KEY_DATA_TO_AUTH);
+                if (extraKeyDataToAuth!=null && !extraKeyDataToAuth.isEmpty()) {
+                    if (windows) {
+                        LOG.warn("Ignoring flag EXTRA_PUBLIC_KEY_DATA_TO_AUTH on Windows location", machineLocation);
```
sjcorbett: It would be great if there were some way to communicate these messages to the user/tool doing the deployment. Or even to raise it before deployment.

aledsage:
The jenkins build failure is unrelated (e.g. `Could not start Jetty server on port 9,998: Address already in use`).

LGTM; merging.

PR #300, opened 2016-08-13
Location DSL Updates

grkvlt:
- Adds a `$brooklyn:location()` command to the DSL to enable retrieval of location config keys
- Adds a `$brooklyn:locationSpec:` command to the DSL to enable defining `LocationSpec<?>` objects in YAML
- Updates `DynamicFabric` with configuration key that accepts a list of `LocationSpec<?>` objects that will be used as extra locations to create clusters
- Removed deprecated `EntityFactory` usage from fabric

### DIFF
--- a/camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/creation/LocationSpecConfiguration.java
+++ b/camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/creation/LocationSpecConfiguration.java
@@ -1,0 +1,58 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.brooklyn.camp.brooklyn.spi.creation;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import java.util.Map;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.common.collect.Maps;
+
+import org.apache.brooklyn.api.location.LocationSpec;
+
+/**
+ * Captures the {@link LocationSpec} configuration defined in YAML. 
+ * 
+ * This class does not parse that output; it just stores it.
+ */
+public class LocationSpecConfiguration {
+
+    @SuppressWarnings("unused")
+    private static final Logger LOG = LoggerFactory.getLogger(LocationSpecConfiguration.class);
+
+    private Map<String, Object> specConfiguration;
+
+    public LocationSpecConfiguration(Map<String, ?> specConfiguration) {
+        this.specConfiguration = Maps.newHashMap(checkNotNull(specConfiguration, "specConfiguration"));
+    }
+
+    public Map<String, Object> getSpecConfiguration() {
+        return specConfiguration;
+    }
+    
+    /**
+     * Allows BrooklynComponentTemplateResolver to traverse the configuration and resolve any entity specs
+     */
+    public void setSpecConfiguration(Map<String, Object> specConfiguration) {
+       this.specConfiguration =  specConfiguration;
+    }
+}
--- a/camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/methods/DslComponent.java
+++ b/camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/methods/DslComponent.java
@@ -32,7 +32,9 @@ import com.google.common.collect.Iterables;
 import com.google.common.util.concurrent.Callables;
 
 import org.apache.brooklyn.api.entity.Entity;
+import org.apache.brooklyn.api.location.Location;
 import org.apache.brooklyn.api.mgmt.Task;
+import org.apache.brooklyn.api.objs.BrooklynObject;
 import org.apache.brooklyn.api.sensor.AttributeSensor;
 import org.apache.brooklyn.api.sensor.Sensor;
 import org.apache.brooklyn.camp.brooklyn.BrooklynCampConstants;
@@ -41,19 +43,22 @@ import org.apache.brooklyn.core.config.ConfigKeys;
 import org.apache.brooklyn.core.entity.Entities;
 import org.apache.brooklyn.core.entity.EntityInternal;
 import org.apache.brooklyn.core.entity.EntityPredicates;
+import org.apache.brooklyn.core.location.Machines;
+import org.apache.brooklyn.core.location.internal.LocationInternal;
 import org.apache.brooklyn.core.mgmt.BrooklynTaskTags;
 import org.apache.brooklyn.core.mgmt.internal.EntityManagerInternal;
 import org.apache.brooklyn.core.sensor.DependentConfiguration;
 import org.apache.brooklyn.core.sensor.Sensors;
+import org.apache.brooklyn.location.ssh.SshMachineLocation;
 import org.apache.brooklyn.util.core.task.TaskBuilder;
 import org.apache.brooklyn.util.core.task.Tasks;
 import org.apache.brooklyn.util.guava.Maybe;
 import org.apache.brooklyn.util.text.StringEscapes.JavaStringEscapes;
 
-public class DslComponent extends BrooklynDslDeferredSupplier<Entity> {
+public class DslComponent<O extends BrooklynObject> extends BrooklynDslDeferredSupplier<O> {
 
     private static final long serialVersionUID = -7715984495268724954L;
-    
+
     private final String componentId;
     private final DslComponent scopeComponent;
     private final Scope scope;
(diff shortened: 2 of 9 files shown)


aledsage:
@grkvlt the nine test failures in jenkins look related - e.g. `org.apache.brooklyn.entity.group.DynamicFabricTest.testDynamicFabricCreatesAndStartsEntityWhenGivenManyLocations`.

camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/creation/LocationSpecConfiguration.java
```diff
@@ -0,0 +1,58 @@
...
+ */
+public class LocationSpecConfiguration {
+
+    @SuppressWarnings("unused")
+    private static final Logger LOG = LoggerFactory.getLogger(LocationSpecConfiguration.class);
+
+    private Map<String, Object> specConfiguration;
+
+    public LocationSpecConfiguration(Map<String, ?> specConfiguration) {
+        this.specConfiguration = Maps.newHashMap(checkNotNull(specConfiguration, "specConfiguration"));
```
aledsage: Personal preference for `newLinkedHashMap` (so that logging/display order is consistent) with what the user supplies.

Seems strange to take a copy here, and to assert non-null; but then in `setSpecConfiguration` we don't copy it. And if subsequently using the setter, then will it really pass non-null to the constructor?

camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/methods/DslComponent.java
```diff
@@ -84,51 +89,72 @@ public DslComponent(DslComponent scopeComponent, Scope scope, String componentId
...
+        public ObjectInScopeFinder(DslComponent scopeComponent, Scope scope, String componentId) {
             this.scopeComponent = scopeComponent;
             this.scope = scope;
             this.componentId = componentId;
         }
 
         protected EntityInternal getEntity() {
             if (scopeComponent!=null) {
-                return (EntityInternal)scopeComponent.get();
+                return (EntityInternal) scopeComponent.get();
```
aledsage: Seems strange to cast this to `EntityInternal`, now that you've generalised it to `ObjectInScopeFinder`.

aledsage: Ah, ignore me - I missed the method name!

camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/methods/DslComponent.java
```diff
@@ -84,51 +89,72 @@ public DslComponent(DslComponent scopeComponent, Scope scope, String componentId
...
+            } else {
+                throw new IllegalStateException("Scope component must be set");
+            }
+        }
+
         @Override
-        public Entity call() throws Exception {
-            Iterable<Entity> entitiesToSearch = null;
+        public O call() throws Exception {
             EntityInternal entity = getEntity();
```
aledsage: I don't follow what the expectation of this class is now. Do we expect the `scopeComponent` to always be an entity? Or sometimes a location? If it calls `call()` then it will fail for anything but an entity, I presume.

camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/methods/DslComponent.java
```diff
@@ -145,81 +171,88 @@ public Entity call() throws Exception {
...
+        return new DslComponent(this, Scope.LOCATION, null);
+    }
+
     // DSL words which return things
 
     public BrooklynDslDeferredSupplier<?> entityId() {
-        return new EntityId(this);
+        return new Identity(this);
     }
-    protected static class EntityId extends BrooklynDslDeferredSupplier<Object> {
```
aledsage: Is this rename going to be backwards compatible? Could `EntityId` be in anyone's persisted state?

aledsage:
No tests have been added/updated - e.g. for the DSL to lookup locations, and for testing what you're adding to `DynamicFabric`.

camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/methods/DslComponent.java
```diff
@@ -84,51 +89,72 @@ public DslComponent(DslComponent scopeComponent, Scope scope, String componentId
...
+        }
+
         @Override
-        public Entity call() throws Exception {
-            Iterable<Entity> entitiesToSearch = null;
+        public O call() throws Exception {
             EntityInternal entity = getEntity();
+
+            if (scope == Scope.LOCATION) {
+                Maybe<SshMachineLocation> machine = Machines.findUniqueMachineLocation(entity.getLocations(), SshMachineLocation.class);
```
aledsage: Does this mean it will fail if used for a non-machine location? e.g. if called on the top-level app that has a `JcloudsLocation` instance?

aledsage:
@grkvlt finished reviewing - a few minor comments; the biggest is about the failing tests, and lack of new tests to cover the new functionality.

camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/creation/LocationSpecConfiguration.java
neykov: +1 for ordered map - snameyaml's maps are already ordered.

camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/methods/DslComponent.java
neykov: It makes sense to expose only machine locations - our hierarchy of parent "spec" locations is confusing to users (well me). Better hide them altogether - it's the machine locations that are interesting. This has the nice effect for spinning up machines that the `$brooklyn:location().locationId()` will return only when the VM is provisioned. Instead of returning immediately with the ID of the spec location.

Can be improved by:
  * working with `MachineLocation` to accomodate the winrm case
  * looking at parent entities for the location if there's non at this level - for the case of nested softwareprocesses/sameserverentity.

core/src/main/java/org/apache/brooklyn/entity/group/DynamicFabricImpl.java
```diff
@@ -134,29 +125,39 @@ public void start(Collection<? extends Location> locsO) {
...
 
-            // finally (and usually) we create new entities for locations passed in
+            // now look at the configured location specs
+            List<LocationSpec<?>> locationSpecs = MutableList.copyOf(config().get(LOCATION_SPECS));
+            if (locationSpecs.size() > 0) {
+                // use these in addition to locations passed to start method
+                for (LocationSpec<?> spec : locationSpecs) {
+                    Location location = getManagementContext().getLocationManager().createLocation(spec);
+                    locationsToStart.add(location);
+                }
```
neykov: Where are these locations unmanaged?

neykov:
Changes look good (apart from Aled's comments). Very useful!
I prefer if we don't re-use DslComponent for implementing both entity and location behaviour. The parser supports object chaining so it could be implemented similar to `attributeWhenReady` but returning a `DslLocation` instead.

core/src/main/java/org/apache/brooklyn/entity/group/DynamicFabric.java
```diff
@@ -18,39 +18,41 @@
...
     public static final ConfigKey<EntitySpec<?>> MEMBER_SPEC = ConfigKeys.newConfigKey(
-            new TypeToken<EntitySpec<?>>() {}, "dynamiccfabric.memberspec", "entity spec for creating new cluster members", null);
+            new TypeToken<EntitySpec<?>>() {}, "dynamiccfabric.memberspec", "Entity specification for creating new cluster members", null);
 
-    @SetFromFlag("factory")
-    public static final ConfigKey<EntityFactory<?>> FACTORY = ConfigKeys.newConfigKey(
-        new TypeToken<EntityFactory<?>>() {}, "dynamicfabric.factory", "factory for creating new cluster members", null);
+    @SetFromFlag("locationSpecs")
+    public static final ConfigKey<List<LocationSpec<?>>> LOCATION_SPECS = ConfigKeys.newConfigKey(
+            new TypeToken<List<LocationSpec<?>>>() {}, "dynamiccfabric.locationspecs", "List of location specifications for starting new cluster members", null);
```
neykov: Persistence/backwards compatibility concerns.

core/src/main/java/org/apache/brooklyn/entity/group/DynamicFabric.java
```diff
@@ -67,9 +69,7 @@
     public static final AttributeSensor<Lifecycle> SERVICE_STATE_ACTUAL = Attributes.SERVICE_STATE_ACTUAL;
 
     public void setMemberSpec(EntitySpec<?> memberSpec);
-    
-    public void setFactory(EntityFactory<?> factory);
-
```
neykov: Should we go through a deprecation cycle here?

ahgittin:
is there an ambiguity eg at a `SoftwareProcess` if i do `$brooklyn:location()` do i get the ssh machine location or the cloud location ... and similar eg at a `Fabric` if there are multiple locations.

note the config keys in `LocationInternal` eg `spec.final` which would be quite useful to retrieve the location spec

neykov:
@ahgittin I think `$brooklyn:location()` should be referencing the `MachineLocation`. Blocking until available. Perhaps we could add `.getParent()` to get the "location spec".

ahgittin:
@neykov yes referencing the `MachineLocation` is probably what people want right now but conceptually it is a hack to hard code that.  i think we should **not** hack things into the `$` DSL:  they will be hard to undo.

i'd prefer to look at building out the DSL around *relations* and use that.  eg a jquery-ish `$brooklyn:relation("RUNNING_ON_MACHINE_LOCATION").wait("30s").requireUnique().config("host.address")`.

in the interim what about the simple approach of a policy that publishes the `MachineLocation` as a sensor and the DSL allowing `attributeWhenReady("my.machine").config("spec.final")` ?

neykov:
Makes sense @ahgittin.

grkvlt:
I am going to close this, as the `DslComponent` code has moved too much to merge, and I think @neykov's idea of a `DslLocation` might bea better way of handling this concept. Will revisit later.

PR #304, opened 2016-08-19
Adds a target configuration to the SSH command executing member tracker

grkvlt:
Adds `ENTITY`, `MEMBER` and `ALL_MEMBERS` as targets for `SshCommandMembershipTrackingPolicy`, updates `Joiner` to handle map entries as key-value pairs, adds `MapAggregator` to generate maps from entity sensor pairs and adds new `ENTITY_ID`, `APPLICATION_ID` and `CATALOG_ID` default sensors to entities.

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/entity/AbstractEntity.java
+++ b/core/src/main/java/org/apache/brooklyn/core/entity/AbstractEntity.java
@@ -196,7 +196,11 @@ public abstract class AbstractEntity extends AbstractBrooklynObject implements E
             "entity.group.added", "Group dynamically added to entity");
     public static final BasicNotificationSensor<Group> GROUP_REMOVED = new BasicNotificationSensor<Group>(Group.class,
             "entity.group.removed", "Group dynamically removed from entity");
-    
+
+    public static final AttributeSensor<String> ENTITY_ID = Attributes.ENTITY_ID;
+    public static final AttributeSensor<String> APPLICATION_ID = Attributes.APPLICATION_ID;
+    public static final AttributeSensor<String> CATALOG_ID = Attributes.CATALOG_ID;
+
     static {
         RendererHints.register(Entity.class, RendererHints.displayValue(EntityFunctions.displayName()));
     }
@@ -1615,7 +1619,7 @@ public abstract class AbstractEntity extends AbstractBrooklynObject implements E
     // -------- INITIALIZATION --------------
 
     /**
-     * Default entity initialization, just calls {@link #initEnrichers()}.
+     * Default entity initialization sets ID sensors and calls {@link #initEnrichers()}.
      */
     public void init() {
         super.init();
@@ -1623,6 +1627,9 @@ public abstract class AbstractEntity extends AbstractBrooklynObject implements E
         if (Strings.isNonBlank(getConfig(DEFAULT_DISPLAY_NAME))) {
             setDefaultDisplayName(getConfig(DEFAULT_DISPLAY_NAME));
         }
+        sensors().set(ENTITY_ID, getId());
+        sensors().set(APPLICATION_ID, getApplicationId());
+        sensors().set(CATALOG_ID, getCatalogItemId());
     }
     
     /**
--- a/core/src/main/java/org/apache/brooklyn/enricher/stock/Enrichers.java
+++ b/core/src/main/java/org/apache/brooklyn/enricher/stock/Enrichers.java
@@ -161,7 +161,11 @@ public class Enrichers {
         }
         /** as {@link #combining(Collection)} but the collection of values comes from the given sensor on multiple entities */
         public <S> AggregatorBuilder<S, Object> aggregating(AttributeSensor<S> val) {
-            return new AggregatorBuilder<S,Object>(val);
+            return new AggregatorBuilder<S, Object>(val);
+        }
+        /** as {@link #combining(Collection)} but the collection of values comes from the given sensor on multiple entities */
+        public <S> AggregatorBuilder<S, Object> aggregating(AttributeSensor<S> keySensor, AttributeSensor<S> valueSensor) {
+            return new AggregatorBuilder<S, Object>(keySensor, valueSensor);
         }
         /** creates an {@link UpdatingMap} enricher: 
          * {@link UpdatingMapBuilder#from(AttributeSensor)} and {@link UpdatingMapBuilder#computing(Function)} are required
@@ -182,13 +186,15 @@ public class Enrichers {
 
 
     protected abstract static class AbstractAggregatorBuilder<S, T, B extends AbstractAggregatorBuilder<S, T, B>> extends AbstractEnricherBuilder<B> {
-        protected final AttributeSensor<S> aggregating;
+        protected AttributeSensor<S> aggregating;
+        protected AttributeSensor<S> keySensor;
+        protected AttributeSensor<S> valueSensor;
         protected AttributeSensor<T> publishing;
         protected Entity fromEntity;
         /** @deprecated since 0.7.0, kept for backwards compatibility for rebind, but not used otherwise */
         @Deprecated protected Function<? super Collection<S>, ? extends T> computing;
         // use supplier so latest values of other fields can be used
-        protected Supplier<Function<? super Collection<S>, ? extends T>> computingSupplier;
+        protected Supplier<Function<? super Collection<S>, ? extends T>> computingSupplier = Suppliers.ofInstance(null);
         protected Boolean fromMembers;
         protected Boolean fromChildren;
         protected Boolean excludingBlank;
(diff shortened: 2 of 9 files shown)


core/src/main/java/org/apache/brooklyn/entity/group/SshCommandMembershipTrackingPolicy.java
```diff
@@ -25,7 +25,9 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.google.api.client.repackaged.com.google.common.base.Preconditions;
```
sjcorbett: We have a checkstyle rule to forbid this package.

grkvlt: Argh, stupid Eclipse import command...

core/src/main/java/org/apache/brooklyn/entity/group/SshCommandMembershipTrackingPolicy.java
```diff
@@ -62,6 +69,38 @@
...
+        ALL_MEMBERS;
+
+        private static Converter<String, String> converter = CaseFormat.LOWER_CAMEL.converterTo(CaseFormat.UPPER_UNDERSCORE);
+
+        public static ExecutionTarget fromString(String name) {
+            Maybe<ExecutionTarget> parsed = tryFromString(name);
+            return parsed.get();
+        }
+
+        public static Maybe<ExecutionTarget> tryFromString(String name) {
```
ahgittin: use `TypeCoercions.coerce` instead to reduce code here and handle case insensitive etc

core/src/main/java/org/apache/brooklyn/core/entity/AbstractEntity.java
```diff
@@ -196,7 +196,11 @@
             "entity.group.added", "Group dynamically added to entity");
     public static final BasicNotificationSensor<Group> GROUP_REMOVED = new BasicNotificationSensor<Group>(Group.class,
             "entity.group.removed", "Group dynamically removed from entity");
-    
+
+    public static final AttributeSensor<String> ENTITY_ID = Attributes.ENTITY_ID;
+    public static final AttributeSensor<String> APPLICATION_ID = Attributes.APPLICATION_ID;
+    public static final AttributeSensor<String> CATALOG_ID = Attributes.CATALOG_ID;
```
ahgittin: why make these sensors?  is there a better way to achieve whatever the goal is?

this info is already present as fields and it seems to me a bit of a smell to put the same data in multiple places.

core/src/main/java/org/apache/brooklyn/enricher/stock/Enrichers.java
```diff
@@ -293,42 +307,33 @@ protected String getDefaultUniqueTag() {
             return "aggregator:"+publishing.getName();
         }
         public EnricherSpec<?> build() {
-            Predicate<Object> valueFilter;
```
ahgittin: won't removing this break every aggregator which relies on `valueFilter` to deliver `excludingBlank` behaviour?

is it better to keep the if block and fail if both value-filter and excluding blank are set (or construct a new `and` filter) ?

core/src/main/java/org/apache/brooklyn/enricher/stock/Joiner.java
```diff
@@ -92,36 +96,48 @@ protected Object compute(SensorEvent<T> event) {
...
                     if (maximum!=null && maximum>=0) {
                         if (c1.size()>=maximum) break;
                     }
-                    c1.appendIfNotNull(Strings.toString(ci));
+                    if (ci instanceof Map.Entry) {
+                        String key = Strings.toString(((Map.Entry) ci).getKey());
+                        Object value = ((Map.Entry) ci).getValue();
+                        String keyValueSeparator = config().get(KEY_VALUE_SEPARATOR);
+                        if (value != null) {
+                            c1.append(String.format("%s%s%s", key, keyValueSeparator, Strings.toString(value)));
```
ahgittin: add a `// TODO we might want to handle QUOTE=true specially for this case, to quote keys and values if needed` ?

core/src/main/java/org/apache/brooklyn/core/entity/AbstractEntity.java
ahgittin: simpler to handle the requirement that entity ID is the key in `MapAggregator` could be for entity ID to be the default key.  that feels like a sensible common default, and means we don't (yet) need to add these artificial sensors.

core/src/main/java/org/apache/brooklyn/entity/group/SshCommandMembershipTrackingPolicy.java
grkvlt: But the methods here are _for_ `TypeCoercions` to work... If I use it, won't there be an infinite recursion?

core/src/main/java/org/apache/brooklyn/core/entity/AbstractEntity.java
grkvlt: It isn't possible to access these fields in YAML though - the _only_ things we can access that way are attributes and config, so exporting as sensor data seems the right way to expose them. I did add an `entityId()` command to the DSL, but this is only available where we have the entity available as context to execute the command, which is not possible inside enrichers etc, where (again) all we can access is sensor data on some (as yet unspecified) entity. I think its perfectly reasonable to add these, since sensor/attribute data is the main and often only way of obtaining information in blueprints.

core/src/main/java/org/apache/brooklyn/enricher/stock/Enrichers.java
grkvlt: No, it's OK, this is actually fixing broken (although mostly working) code. The code now properly configures the enricher with the passed in `excludingBlank` and `valueFilter` values, and the aggregators that use this functionality manage the configuration of the filter themselves (see `getDefaultVlaueFilter()` method in `Aggregator.java` and the line in `AbstracxtAggregator.java` in `setEntityLoadingConfig()` which uses this to set the `valueFilter` field.) It might be useful to have a check that one of the two config keys, `EXCLUDING_BLANK` or `VALUE_FILTER` is set, though.

core/src/main/java/org/apache/brooklyn/enricher/stock/Joiner.java
grkvlt: Good idea, the thought passed my mind as well...

core/src/main/java/org/apache/brooklyn/core/entity/AbstractEntity.java
grkvlt: Probably a nice idea to have a default key, though - will add that. But can just set it as `ENTITY_ID` I guess?

core/src/main/java/org/apache/brooklyn/entity/group/SshCommandMembershipTrackingPolicy.java
ahgittin: Yes it might given the naming convention.  But do you need this at all?  Are the strategies in `EnumTypeCoercions` insufficient?  It already does `CaseFormat.LOWER_CAMEL.converterTo(CaseFormat.UPPER_UNDERSCORE)` among others.

core/src/main/java/org/apache/brooklyn/core/entity/AbstractEntity.java
ahgittin: @aledsage @neykov wdyt about adding these three pieces of information as sensors?

I can see the pragmatic reason for doing so...

+1 to default

core/src/main/java/org/apache/brooklyn/enricher/stock/Enrichers.java
ahgittin: Aha the removed lines are redundant because of `Aggregator.getDefaultValueFilter()`.

Possibly worth promoting that method to `AbstractAggregator` but I agree with your analysis.

ahgittin:
would it make sense to split up `SshCommandMembershipTrackingPolicy` so that we have`InvokeEffectorMembershipTrackingPolicy` and `SshMultiEntityCommandEffector` -- where the latter has the configurable target you've introduced here (and where `TARGET` defaults to the current entity if it has an `SshMachineLocation`, to sshable members if it's a group, and to sshable children otherwise;  and could be set to be an `Entity` or the enum you provide) ?

(then to handle DNS i think we don't even need a membership tracking policy if we have a suitably configured group and an aggregator on hostnames -- all we need is an `InvokeEffectorOnSensorChangePolicy`)

ahgittin:
@grkvlt happy to merge this with the addition of a few TODO comments on the topics mentioned, unless you think you'll have bandwidth to merge this in the next day or so -- LMK

ahgittin:
merged, and remarks above addressed in c57fb6004aa096488675507bef2e05df8776a590

core/src/main/java/org/apache/brooklyn/core/entity/AbstractEntity.java
neykov: +1 to add, it's a better approach than extending the DSL from YAML point of view.

PR #305, opened 2016-08-23
BROOKLYN-338: BrooklynApi fixes for karaf

aledsage:
These aren’t enough to fix BROOKLYN-338, but it’s a start:
* Ensures rest-client imports the resteasy marshallers package,
  which it calls with reflection (via the resteasy ProxyBuilder).
* Passes the ProxyBuilder a classloader that has access to the 
  rest-api classes and to rest-client’s classloader (which thus 
  can access the marshallers).

### DIFF
--- a/rest/rest-client/pom.xml
+++ b/rest/rest-client/pom.xml
@@ -144,5 +144,20 @@
             <scope>test</scope>
         </dependency>
     </dependencies>
-    
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.felix</groupId>
+                <artifactId>maven-bundle-plugin</artifactId>
+                <extensions>true</extensions>
+                <configuration>
+                    <instructions>
+                        <Import-Package>*,org.jboss.resteasy.client.core.marshallers</Import-Package>
+                    </instructions>
+                </configuration>
+            </plugin>
+        </plugins>
+    </build>
+
 </project>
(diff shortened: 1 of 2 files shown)


rest/rest-client/pom.xml
```diff
@@ -144,5 +144,21 @@
...
         </dependency>
     </dependencies>
-    
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.felix</groupId>
+                <artifactId>maven-bundle-plugin</artifactId>
+                <version>2.5.3</version>
```
neykov: Remove version, already defined (as `2.5.4`) in `brooklyn-parent`.

neykov:
LGTM, build failure is unrelated.

aledsage:
Jenkins failure was for `OsTest.testDeleteRecursivelyEmptyDir`, failing for the 1000 attempts like:

    2016-08-23 22:38:29,719 WARN  Attempt to create temp dir failed /tmp/OsTest-IicA. Either an IO error (disk full, no rights) or someone else created the folder after the !exists() check.

Presumably "just" the build machine had run out of disk space! So assuming it wasn't our tests that filled up the disk, then this error is not related to our code.

---
I've incorporated Svet's comment and squashed the commits.

neykov:
Tests are passing locally, merging.

PR #307, opened 2016-08-25
ensures locations are made available to location registry when catalog is reset

ahgittin:
(longer term the location registry should be removed in favour of pure catalog,
but this is a tactical fix)

also minor fix w dep config seeing null values


neykov:
LGTM, merging.

PR #308, opened 2016-08-25
Invoke on sensor and ssh to multiple

ahgittin:
These fixes make https://github.com/brooklyncentral/brooklyn-dns work neatly.

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/effector/ssh/SshCommandEffector.java
+++ b/core/src/main/java/org/apache/brooklyn/core/effector/ssh/SshCommandEffector.java
@@ -18,15 +18,14 @@
  */
 package org.apache.brooklyn.core.effector.ssh;
 
+import java.util.Collection;
 import java.util.Map;
 import java.util.concurrent.ExecutionException;
 
-import com.google.common.base.Preconditions;
-import com.google.common.base.Predicates;
-import com.google.common.collect.Maps;
-
 import org.apache.brooklyn.api.effector.Effector;
 import org.apache.brooklyn.api.effector.ParameterType;
+import org.apache.brooklyn.api.entity.Entity;
+import org.apache.brooklyn.api.entity.Group;
 import org.apache.brooklyn.config.ConfigKey;
 import org.apache.brooklyn.core.config.ConfigKeys;
 import org.apache.brooklyn.core.config.MapConfigKey;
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/apache/brooklyn/policy/InvokeEffectorOnSensorChange.java
```diff
@@ -0,0 +1,71 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.common.base.Preconditions;
+
+/**
+ * Invokes the given effector when the policy changes.
+ * 
+ * TODO
+ * * support parameters
```
grkvlt: I suggest when removing the now deprecated `SshCommandMembershipTrackingPolicy` we use a similar technique, and add the entity that changed and the sensor data that caused the change as arguments.

core/src/main/java/org/apache/brooklyn/policy/InvokeEffectorOnSensorChange.java
```diff
@@ -0,0 +1,71 @@
...
+ * * support conditions
+ * * allow to be triggered by sensors on members
+ */
+public class InvokeEffectorOnSensorChange extends AbstractPolicy implements SensorEventListener<Object> {
+    
+    private static final Logger LOG = LoggerFactory.getLogger(InvokeEffectorOnSensorChange.class);
+
+    public static final ConfigKey<Object> SENSOR = ConfigKeys.newConfigKey(Object.class,
```
grkvlt: Would have preferred `sensorsToTrack` that takes a list of sensors

core/src/main/java/org/apache/brooklyn/policy/InvokeEffectorOnSensorChange.java
```diff
@@ -0,0 +1,71 @@
...
+
+/**
+ * Invokes the given effector when the policy changes.
+ * 
+ * TODO
+ * * support parameters
+ * * support conditions
+ * * allow to be triggered by sensors on members
+ */
+public class InvokeEffectorOnSensorChange extends AbstractPolicy implements SensorEventListener<Object> {
```
grkvlt: Needs tests

core/src/main/java/org/apache/brooklyn/core/effector/ssh/SshCommandEffector.java
```diff
@@ -18,36 +18,57 @@
...
     public static final ConfigKey<String> EFFECTOR_COMMAND = ConfigKeys.newStringConfigKey("command");
     public static final ConfigKey<String> EFFECTOR_EXECUTION_DIR = SshCommandSensor.SENSOR_EXECUTION_DIR;
     public static final MapConfigKey<Object> EFFECTOR_SHELL_ENVIRONMENT = BrooklynConfigKeys.SHELL_ENVIRONMENT;
 
+    public static enum ExecutionTarget {
+        ENTITY,
+        MEMBERS,
+        CHILDREN
+    }
+    public static final ConfigKey<ExecutionTarget> EXECUTION_TARGET = ConfigKeys.newConfigKey(ExecutionTarget.class, "executionTarget",
```
grkvlt: Needs tests for these arguments, but admittedly the `SshCMTP` didn't have explicit tests for these either

core/src/main/java/org/apache/brooklyn/entity/group/SshCommandMembershipTrackingPolicy.java
```diff
@@ -61,11 +64,16 @@
...
+ * (which in turn is aggregated over members) instead of children sensors directly).
+ * If there is a need for {@link #EVENT_TYPE} or {@link #MEMBER_ID} then we'll want
+ * InvokeEffectorOnSensorChange to be able to subscribe to members/children etc as well,
+ * and pass parameters based on source.entityId and source.sensor.
  */
-// TODO might make sense to split up behaviour into two classes,
-// an InvokeEffectorMembershipTrackingPolicy and an SshMultiEntityCommandEffector -- 
-// where the latter has the configurable target introduced here
-@Beta
+@Deprecated
```
grkvlt: Actually, this could be safely removed, since the only usage of it has been updated to use the new policy defined here.

grkvlt:
LGTM, but see comments

core/src/main/java/org/apache/brooklyn/policy/InvokeEffectorOnSensorChange.java
ahgittin: happy for that to be added, along with a debounce, but then we'll also want a way to pass the sensor and originator as parameters to the effector ... so i've started simple

feel free to add when needed

i like having both `sensor` and `sensorsToTrack` and they are additive

core/src/main/java/org/apache/brooklyn/policy/InvokeEffectorOnSensorChange.java
ahgittin: yeah...

PR #310, opened 2016-08-28
Child entity customizer

grkvlt:
Extends the jclouds location customizer to be a `BasicConfigurableObject` allowing better configuration etc. when created using `$brooklyn:object` in the DSL, and add a new `ChildEntityCustomizer` that creates a child entity before the main entity has started up, allowing more flexible configuration.

### DIFF
--- a/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/SudoTtyFixingCustomizer.java
+++ b/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/SudoTtyFixingCustomizer.java
@@ -30,19 +30,15 @@ import com.google.common.base.Preconditions;
 /**
  * Wraps Brooklyn's sudo-tty mitigations in a {@link JcloudsLocationCustomizer} for easy(-ish) consumption
  * in YAML blueprints:
+ * <pre>{@code
+ * name: My App
+ * brooklyn.config:
+ *   provisioning.properties:
+ *     customizerType: SudoTtyFixingCustomizer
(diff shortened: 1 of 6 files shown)


locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/SudoTtyFixingCustomizer.java
```diff
@@ -30,19 +30,15 @@
...
- *     provisioning.properties:
- *       customizerType: SudoTtyFixingCustomizer
- *   services: ...
- * </pre>
- *
- * <p>This class should be seen as a temporary workaround and might disappear completely if/when Brooklyn takes care of this automatically.
- *
- * <p>See
- * <a href='http://unix.stackexchange.com/questions/122616/why-do-i-need-a-tty-to-run-sudo-if-i-can-sudo-without-a-password'>http://unix.stackexchange.com/questions/122616/why-do-i-need-a-tty-to-run-sudo-if-i-can-sudo-without-a-password</a>
+ * @see <a href='http://unix.stackexchange.com/questions/122616/why-do-i-need-a-tty-to-run-sudo-if-i-can-sudo-without-a-password'>http://unix.stackexchange.com/questions/122616/why-do-i-need-a-tty-to-run-sudo-if-i-can-sudo-without-a-password</a>
```
neykov: Btw this is already part of `JcloudsLocation`. Should we deprecate this now? Any use-cases where the location config can't be used?

PR #313, opened 2016-08-31
Adds deprecated decorate method to preserve backwards compatibility

grkvlt:
Adds `#decorate(EntitySpec<?>, ConfigBag)` as an overloaded deprecated method in `BrooklynEntityDecorationResolver` class, for backwards compatibility.

See https://github.com/cloudsoft/brooklyn-tosca/pull/113

### DIFF
--- a/camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/creation/BrooklynEntityDecorationResolver.java
+++ b/camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/creation/BrooklynEntityDecorationResolver.java
@@ -22,9 +22,9 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
-import com.google.common.annotations.Beta;
 import com.google.common.base.Function;
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
 
 import org.apache.brooklyn.api.entity.EntityInitializer;
 import org.apache.brooklyn.api.entity.EntitySpec;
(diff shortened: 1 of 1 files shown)


camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/creation/BrooklynEntityDecorationResolver.java
```diff
@@ -46,22 +46,26 @@
...
-    
+
     protected BrooklynEntityDecorationResolver(BrooklynYamlTypeInstantiator.Factory instantiator) {
         this.instantiator = instantiator;
     }
-    
+
+    /** @deprected since 0.10.0 - use {@link #decorate(EntitySpec<?>, ConfigBag, Set<String>)} */
+    @Deprecated
+    public abstract void decorate(EntitySpec<?> entitySpec, ConfigBag attrs);
```
ahgittin: make this `public final` not abstract with the implementation passing the empty `ImmutableSet` ?

camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/creation/BrooklynEntityDecorationResolver.java
```diff
@@ -46,22 +46,26 @@
 
 /**
  * Pattern for resolving "decorations" on service specs / entity specs, such as policies, enrichers, etc.
+ *
  * @since 0.7.0
  */
-@Beta
```
ahgittin: agree

ahgittin:
one tweak then good to merge

PR #314, opened 2016-08-31
Fix winrm4j in Karaf

neykov:
winrm4j version update has fixes for OSGi. Also adding an extension bundle to httpcomponents-asyncclient so that it's able to see `org.apache.http.ssl package` (that's fixed in later revision releases of the jar).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -155,7 +155,7 @@
         <maxmind.version>0.8.1</maxmind.version>
         <maxmind-db.version>0.3.4</maxmind-db.version>
         <jna.version>4.0.0</jna.version>
-        <winrm4j.version>0.3.5</winrm4j.version>
+        <winrm4j.version>0.4.0</winrm4j.version>
         <karaf.version>4.0.4</karaf.version>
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -155,7 +155,7 @@
         <maxmind.version>0.8.1</maxmind.version>
         <maxmind-db.version>0.3.4</maxmind-db.version>
         <jna.version>4.0.0</jna.version>
-        <winrm4j.version>0.3.5</winrm4j.version>
+        <winrm4j.version>0.4.0-SNAPSHOT</winrm4j.version>
```
geomacy: Why snapshot, is that a temporary thing until we can get a 4.0.something built?

neykov: that's right - to facilitate testing until we get a winrm4j release out

ahgittin: jenkins failure

`[WARNING] The POM for io.cloudsoft.windows:winrm4j:jar:0.4.0-SNAPSHOT is missing, no dependency information available`

let's get `0.4.0` out (not snapshot) and simply reference it ?

ahgittin:
excellent workaround

PR #315, opened 2016-08-31
Try alternative Java versions

drigodwin:
This tries incrementing the Java version we're installing if the current one fails. This primarily addresses the problem of Java 7 not being available by default on Ubuntu 16.04.

Tested with Tomcat7 on GCE, AWS & Bluebox on Ubuntu 14 / 16, Centos 6 / 7


ahgittin:
great simple fix.  longer term we should support ranges.

PR #316, opened 2016-08-31
Set default minRam to 1gb

aledsage:
Also changes JcloudsLocation’s template building so that if
a config key has a default value then that will be used, if there is
no explicit value (or empty value) set.


neykov:
LGTM

PR #317, opened 2016-09-01
Changed jackson version to 2.7.5

Graeme-Miller:


sjcorbett:
Can you test that `brooklyn-tosca` works with this change please?

neykov:
Since we are upgrading suggest doing this change as well https://github.com/neykov/brooklyn-server/commit/62613c05c2ba08c3ad9055d1fa3006f100a58b38.

I also tried bumping to latest releases available (with same major number) in https://github.com/neykov/brooklyn-server/commit/da1073484ec24b555fed0c0f40b89e561dc929a8 - seems to work fine deploying kubernetes.

grkvlt:
@Graeme-Miller happy to merge once you cherry-pick https://github.com/neykov/brooklyn-server/commit/62613c05c2ba08c3ad9055d1fa3006f100a58b38 into this PR as well

sjcorbett:
@grkvlt @Graeme-Miller I am not happy for this to be merged before you have confirmed or we have at least discussed what happens to the `brooklyn-tosca` dist.

geomacy:
@sjcorbett, @Graeme-Miller, I will pull this and test it out with brooklyn-tosca when I've done my next commit.

geomacy:
👍 LGTM; I have pulled this PR, built brooklyn with it and then brooklyn-tosca on top of that (actually using https://github.com/cloudsoft/brooklyn-tosca/pull/117) and I can deploy and undeploy a sample application successfully.

sjcorbett:
@geomacy did you use the dist tgz that brooklyn-tosca builds? I will also test shortly.

geomacy:
@sjcorbett good point, no, I was using the Karaf distro.

ahgittin:
out of curiosity is there a reason for this bump?  if there is can you say what it is for posterity?  otoh if it's just good hygiene that's fine although probably we should do a sweep as i'm sure there are other stale versions.

sjcorbett:
@ahgittin another downstream project has a dependency on https://github.com/fabric8io/kubernetes-client which requires the later version.

neykov:
The version bump was held back on purpose due to incompatibility with brooklyn-tosca transitive dependencies. Now that we are running in Karaf we can finally update (to a semi-recent version).

Agree there are lots of stale library versions in there which need updating.

sjcorbett:
I've tested Brooklyn 0.10.0-SNAPSHOT master + this PR against brooklyn-tosca master. The standard Tomcat/MySQL blueprint deployed fine from Alien4Cloud. The brooklyn-tosca dist includes these jars in its lib directory:
```
jackson-annotations-2.7.5.jar
jackson-core-2.7.5.jar
jackson-databind-2.7.5.jar
jackson-dataformat-xml-2.4.5.jar
jackson-dataformat-yaml-2.4.5.jar
jackson-datatype-joda-2.4.5.jar
jackson-jaxrs-base-2.7.5.jar
jackson-jaxrs-json-provider-2.7.5.jar
jackson-module-jaxb-annotations-2.7.5.jar
```
By the look of it the 2.4.5 dependencies have come from `io.swagger:swagger-core` and `swagger-jaxrs` via `brooklyn-utils-rest-swagger`. 

So sounds like we're good to go but keeping this at the back of our minds. @ahgittin do you think we need to maintain `brooklyn-tosca-dist` any more?

ahgittin:
Worth testing that the apidocs (Swagger REST docs) page works with these versions?

+1 to dropping `brooklyn-tosca-dist` in favour of the karaf way.  `brooklyn-tosca` docs and `README` probably will need updating also.

Graeme-Miller:
I have just tested the swagger page works as expected with this PR.

neykov:
That's a good spot @sjcorbett. The jackson library version mix seems important. I think it's better to stick to the same version (even if higher than required for swagger), than to risk incompatibility. It's higher risk to use two parts of the library with different versions than upgrade all of them.

Another reason to upgrade everything is that in Karaf we are explicitly adding the dependencies and pinning them to  the `fasterxml.jackson.version`. This causes the classic distribution to have a different mix of jars than the karaf one - best to keep them consistent.

Graeme-Miller:
replacing with:
https://github.com/apache/brooklyn-client/pull/28
https://github.com/apache/brooklyn-server/pull/339

PR #318, opened 2016-09-02
Delete most deprecated methods from PortForwardManager

sjcorbett:
They've been deprecated since 0.7.0.


core/src/main/java/org/apache/brooklyn/core/location/access/PortForwardManagerImpl.java
```diff
@@ -356,117 +343,18 @@ private AssociationMetadata associationMetadataFromPortMapping(PortMapping portM
...
-            PortMapping oldMapping = getPortMappingWithPublicSide(publicIpId, publicPort);
-            log.debug("associating public port "+publicPort+" on "+publicIpId+" with private port "+privatePort+" at "+l+" ("+mapping+")"
-                    +(oldMapping == null ? "" : " (overwriting "+oldMapping+" )"));
-            mappings.put(makeKey(publicIpId, publicPort), mapping);
-        }
-    }
-
-    ///////////////////////////////////////////////////////////////////////////////////
     // Internal only; make protected when deprecated interface method removed
     ///////////////////////////////////////////////////////////////////////////////////
```
neykov: Can remove TODO now.

core/src/main/java/org/apache/brooklyn/core/location/access/PortForwardManagerImpl.java
```diff
@@ -83,9 +83,6 @@
 
     private final Map<AssociationListener, Predicate<? super AssociationMetadata>> associationListeners = new ConcurrentHashMap<AssociationListener, Predicate<? super AssociationMetadata>>();
 
-    @Deprecated
-    protected final Map<String,String> publicIpIdToHostname = new LinkedHashMap<String,String>();
```
neykov: Wonder how this will affect rebinding. I believe PFMs are persisted.

sjcorbett: Is there a strategy for removing deprecated fields short of making them private and leaving them forever? Not sure we can do anything sensible to transform its data during rebind.

ahgittin: @sjcorbett my strategy when deprecating fields is to mark it deprecated and transient for one major release, make sure all accessors transform it, and make sure at least one accessor gets exercised.

this should ensure that if state from one version is carried through subsequent major versions it will migrate.

however there is not currently any way to force exercising in all cases i don't think (we could introduce a method `postRebind()` for that purpose?).  for now i suggest keep it in perpetuity with a comment `// kept for persistence from previous versions; see https://github.com/apache/brooklyn-server/pull/318`

bostko:
@sjcorbett did you test your changes with https://github.com/cloudsoft/advanced-networking project

PR #320, opened 2016-09-06
Package name change due to collision when used inside OSGI

ygy:


neykov:
Tests passing for me, merging.

PR #321, opened 2016-09-06
Added configuration for template substitution variable map

grkvlt:
Adds a new `ConfigKey<Map<?,?>>` to allow setting further template variable substitutions in YAML.


neykov:
I believe we can already do that with the existing functionality? Something like:
```
services:
- type: SoftwareProcess
  brooklyn.config:
    template.substitutions:
      key1: value1
      key2: value2
```
```
key1: ${config['template.substitutions.key1']}
key2: ${config['template.substitutions.key2']}
```

grkvlt:
Yes, I used it like `${config['template.substitutions]['key1']}` which is equivalent, I think.

But if you add in the map explicitly, then you could just say `${key1}` in the template file, which is _much_ neater.

neykov:
Would be nice to have test coverage for this, but could be added in a future PR. Tests passing locally,  merging.

PR #322, opened 2016-09-07
Adds a TargetServerEntity for improved command sequencing in YAML

grkvlt:
Circumstances often arise where a controlling server must start up, then some clients or slaves elsewhere that connect to it, finally some process on the main server should be executed that relies on the availability of the client machines. This cannot be done using only `pre.launch.command` and `post.launch.command` or setting latches, since all commands will execute before `service.isUp` is set. This enables another entity to be created, that will execute its own lifecycle and commands, but using the same location (i.e. machine) as some other entity.

### DIFF
--- a/software/base/src/main/java/org/apache/brooklyn/entity/software/base/TargetServerEntityImpl.java
+++ b/software/base/src/main/java/org/apache/brooklyn/entity/software/base/TargetServerEntityImpl.java
@@ -1,0 +1,38 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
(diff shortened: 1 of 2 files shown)


neykov:
@grkvlt Can you add tests covering the new functionality.
Do you think it would be more re-usable to create an `EntityLocation` which proxies to the location of another entity? It could be used by any `SoftwareProcess` entity then.

software/base/src/main/java/org/apache/brooklyn/entity/software/base/TargetServerEntityImpl.java
```diff
@@ -0,0 +1,38 @@
...
+import org.apache.brooklyn.core.entity.Entities;
+import org.apache.brooklyn.core.entity.EntityInternal;
+
+public class TargetServerEntityImpl extends VanillaSoftwareProcessImpl implements TargetServerEntity {
+
+    protected void preStart() {
+        Entity target = config().get(TARGET_ENTITY);
+        Entities.waitForServiceUp(target);
+        Collection<Location> locations = target.getLocations();
+        clearLocations();
```
neykov: Would be surprising to have locations defined on the entity. Could log a warning if there are any.

grkvlt:
This can actually be implemented by adding a child to an entity with a latch that prevents startup until some other entity is available, so closing

PR #323, opened 2016-09-07
Minor tweaks to utils and new TypeRegistry process

ahgittin:
I've been working on a new way of reading types from yaml -- PR coming in the next couple weeks -- but found these small improvements to the existing codebase which seemed appropriate and would be good to get in sooner.

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/typereg/RegisteredTypes.java
+++ b/core/src/main/java/org/apache/brooklyn/core/typereg/RegisteredTypes.java
@@ -24,6 +24,7 @@ import java.util.Iterator;
 import java.util.Map;
 import java.util.Set;
 
+import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 
 import org.apache.brooklyn.api.catalog.CatalogItem;
@@ -45,6 +46,8 @@ import org.apache.brooklyn.util.guava.Maybe.Absent;
 import org.apache.brooklyn.util.text.NaturalOrderComparator;
 import org.apache.brooklyn.util.text.VersionComparator;
 import org.apache.brooklyn.util.yaml.Yamls;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import com.google.common.annotations.Beta;
 import com.google.common.base.Function;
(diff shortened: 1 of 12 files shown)


ahgittin:
this is pretty simple - @neykov @grkvlt could one of you cast a quick eye over it?

core/src/main/java/org/apache/brooklyn/core/typereg/RegisteredTypes.java
```diff
@@ -151,7 +181,7 @@ public static RegisteredType addSuperType(RegisteredType type, @Nullable Registe
         return type;
     }
     @Beta
-    public static RegisteredType addSuperTypes(RegisteredType type, Iterable<Object> superTypesAsClassOrRegisteredType) {
+    public static RegisteredType addSuperTypes(RegisteredType type, Iterable<? extends Object> superTypesAsClassOrRegisteredType) {
```
aledsage: Just use `Iterable<?>` instead of `Iterable<? extends Object>`.

aledsage:
@ahgittin (cc @neykov) I haven't looked thoroughly enough at the `BrooklynTypeRegistry` and `RegisteredType` before, to give a proper review of this.

It would be useful for me (and other contributors I'm sure) to get an overview of the type registry, and of your new way of reading types (we don't need to wait until the PR to get an overview - in fact posting to the dev@brooklyn mailing list for such big changes would be good).

utils/common/src/main/java/org/apache/brooklyn/util/exceptions/Exceptions.java
```diff
@@ -461,4 +461,14 @@ public static String getMessageWithAppropriatePrefix(Throwable t) {
         return appendSeparator(getPrefixText(t), t.getMessage());
     }
 
+    /** Returns true if the root cause is a "boring" CNF, ie a straightforward declaration that the clazz is not found;
+     * this permits callers to include the cause only when it is interesting, ie caused by a dependent class not loadable.
+     */
+    public static boolean isRootBoringClassNotFound(Exception e, String clazz) {
+        Throwable root = Throwables.getRootCause(e);
+        if (!(root instanceof ClassNotFoundException)) return false;
```
neykov: Include `NoClassDefFoundError`?

neykov:
LGTM. Merging - can address changes in next PRs.

utils/common/src/main/java/org/apache/brooklyn/util/exceptions/Exceptions.java
ahgittin: possibly, at least if it has no useful message.  i think maybe sometimes that one can have an interesting message?  wait until we see it occurring without clutter.

ahgittin:
@aledsage yes, i'd like to talk about that.  nearly have something cool to show.

PR #327, opened 2016-09-13
Update geoip to latest version

neykov:
Removes the dependency on google-http-client, depending directly on apache-httpcomponents


sjcorbett:
Looks good.

PR #330, opened 2016-09-16
TypeCoerce string to byte[]

aledsage:
An alternative impl would have been to treat the given string as base64 encoded. But I don't think that is what people would expect as the default. However, by using a simple `string.getBytes()` there are two downsides:
* can't pass in some byte array values, e.g. in UI effector parameter (would require base64 for that)
* it's using the default of the JVM (e.g. rather than explicitly using UTF-8)

I think that's good enough though - thoughts?

My use-case is in a test case I wrote for a HTTP client entity, where an effector takes a parameter of type `byte[]` and uses that as the payload of a POST request. The most common situation I think is where one wants to just pass a "normal" string body.

### DIFF
--- a/utils/common/src/main/java/org/apache/brooklyn/util/javalang/coerce/CommonAdaptorTypeCoercions.java
+++ b/utils/common/src/main/java/org/apache/brooklyn/util/javalang/coerce/CommonAdaptorTypeCoercions.java
@@ -92,6 +92,12 @@ public class CommonAdaptorTypeCoercions {
                 return new String(input);
             }
         });
+        registerAdapter(String.class, byte[].class, new Function<String,byte[]>() {
+            @Override
+            public byte[] apply(String input) {
+                return input.getBytes();
+            }
+        });
         registerAdapter(Collection.class, Set.class, new Function<Collection,Set>() {
             @SuppressWarnings("unchecked")
             @Override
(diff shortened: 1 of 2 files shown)


neykov:
@aledsage I suggest that we follow the same approach as for the [data protocol](https://github.com/apache/brooklyn-server/blob/master/core/src/main/java/org/apache/brooklyn/util/core/ResourceUtils.java#L247). The implementation is so "laxed" that it's a superset of what you are currently doing and can be used as a direct replacement. This will accommodate  both string.getBytes() and base64 use cases and will even allow you to specify an encoding.

geomacy:
I've seen problem in previous lives when code wasn't explicit about encodings, so I think it would be good to be explicit in this case.

ahgittin:
@neykov that impl isn't a true superset is it?  it looks at ";" if supplied and does magic with that according to the spec of the `data` protocol.  i don't think that would be appropriate here.

@aledsage the code is fine but i question the use case.  why would an effector take `byte[]` if a string is a sensible thing for a person to supply.  if that's pressing we could introduce a new type `CoercibleFromByteArrayOrString` or give multiple input parameters on the effector.

this type of low-level coercion feels risky as @geomacy says, even by my standards!

aledsage:
Ok, I'll just change the effector parameter to type `String`. That's good enough in my use-case because it's just for a test. (As for the more general use-case, it's for when invoking the effector from the web-console: I can supply a String, but can't supply a `byte[]`. So one could invoke some http POST requests through the web-console if we supported `String->byte[]` conversion, but would need to drop into the REST api for values that are not simple strings).

I will close this PR as it's not worth the hassle, particularly given it's a bit controversial what the right thing is.

PR #331, opened 2016-09-16
Fix some memory leaks and better memory usage logging

ahgittin:


ahgittin:
@neykov @aledsage the two of you have done the most memory analysis; i'd appreciate a quick look over this.  it's pretty simple and should help with memory analysis.

camp/camp-base/src/main/java/org/apache/brooklyn/camp/spi/collection/BasicResourceLookup.java
```diff
@@ -43,6 +47,11 @@ public synchronized void add(T item) {
         if (old!=null) {
             items.put(old.getId(), old);
             throw new IllegalStateException("Already contains item for "+item.getId()+": "+old+" (adding "+item+")");
```
aledsage: Why does this override the previous value and then throw the exception? I'd have thought we'd either just throw, or we'd override and just log.warn.

camp/camp-base/src/main/java/org/apache/brooklyn/camp/spi/pdp/AssemblyTemplateConstructor.java
```diff
@@ -40,15 +40,22 @@ public AssemblyTemplateConstructor(CampPlatform campPlatform) {
         this.transaction = this.campPlatform.transaction();
     }
     
-    /** records all the templates to the underlying platform */
+    /** records all the templates to the underlying platform 
+     * @deprecated since 0.10.0 use {@link #construct(boolean)} */
     public AssemblyTemplate commit() {
+        return construct(false);
```
aledsage: I see this is now deprecated and not called, but why change the semantics to not actually commit (given that it will now do `transaction.commit()`)? If deprecating, why not leave the semantics as before? Or delete the method entirely?

camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/BrooklynCampPlatform.java
```diff
@@ -61,6 +63,11 @@ public BrooklynCampPlatform(PlatformRootSummary root, ManagementContext manageme
         });
     }
 
+    /** finds and returns the {@link CampPlatform} registered for the management context, or null if none */
+    @Nullable public static CampPlatform findPlatform(ManagementContext mgmt) {
+        return mgmt.getConfig().getConfig(BrooklynCampConstants.CAMP_PLATFORM);
```
aledsage: Duplicate of `CampInternalUtils.getCampPlatform(mgmt)`, except this returns nullable rather than throwing an exception.

Feels like we should only have this in one place, rather than in utils and here.

aledsage: Ah, I see you've changed `CampInternalUtils` to delegate to this! Given it's internal, then no strong feelings.

utils/common/src/main/java/org/apache/brooklyn/util/javalang/MemoryUsageTracker.java
```diff
@@ -68,5 +87,76 @@ public long getBytesUsed() {
         memoryTrackedReferences.cleanUp();
         return bytesUsed.get();
     }
+
+    /** forces all soft references to be cleared by trying to allocate an enormous chunk of memory,
+     * returns a description of what was done 
+     * (tune with with {@link #forceClearSoftReferences(long, int)}
```
aledsage: typo: "with with"

core/src/main/java/org/apache/brooklyn/core/mgmt/internal/BrooklynGarbageCollector.java
```diff
@@ -196,7 +210,13 @@ public void gcIteration() {
             gcTasks();
             logUsage("brooklyn gc (after)");
             
-            if (doSystemGc) {
+            double memUsage = 1.0 - 1.0*Runtime.getRuntime().freeMemory() / Runtime.getRuntime().totalMemory();
```
aledsage: Should we instead be using `Runtime.getRuntime().maxMemory()` (for if `-Xms` and `-Xmx` are different)?

utils/common/src/main/java/org/apache/brooklyn/util/javalang/MemoryUsageTracker.java
```diff
@@ -68,5 +87,76 @@ public long getBytesUsed() {
...
+    public static class SoftUsageTracker {
+        private Cache<Object,SoftReference<?>> cache = null;
+        public synchronized void enable() {
+            cache = CacheBuilder.newBuilder().weakKeys().build();
+        }
+        public synchronized void disable() {
+            cache = null;
+        }
+        public long getTotalEntries() {
+            return cache.size();
```
aledsage: `return (cache == null ? -1 : cache.size())` (and hope no-one calls `disable()` concurrently, or that the java threading model messes things up for you by making the `cache` field not fully initialised from the view point of this calling thread!). Maybe you should `synchronize` as well.

aledsage:
The idea behind `MemoryUsageTracker. forceClearSoftReferences()` makes sense, but feels a little scary! The code looks fine though. It will never be enabled in production, only when profiling, so I'm happy to merge it.

The other additions to the regular output of `BrooklynGarbageCollection` look sensible.

---
What about the removal of `MemoryUsageTracker.getBytesUsed()`? Should we either delete that entirely, or leave it in the output of the `BrooklynGarbageCollection`? It's useful to know how big the soft references are (as long as we call it consistently enough).

It currently is a bit confusing having `MemoryUsageTracker.SOFT_REFERENCES` and `MemoryUsageTracker.SoftUsageTracker` (which are unrelated to each other).

You're right to mark that as `@Beta`!

---
If investigating a memory leak, then `jmap -histo:live <pid>` and `jmap -heap <pid>` will both include soft references. A subsequent analysis of the output file from `jmap -dump:live,format=b,file=heap.bin <pid>` will allow you to filter out the soft references, but that's fiddlier.

Looking at `jmap -histo:live` and manually excluding the stdout/stderr of tasks (so streams / byte arrays) will give a good indication of all non-soft references (but you may accidentally ignore things that are not soft-references).

Much longer term, we should consider how best to store historic task details (e.g. in riak) and a better deletion policy for them. But that is a much bigger undertaking.

camp/camp-base/src/main/java/org/apache/brooklyn/camp/spi/collection/BasicResourceLookup.java
ahgittin: agree doesn't make much sense -- have added a TODO comment (as wasn't introduced here and don't want to change, even though i'm pretty sure it's safe)

ahgittin: actually i get it, it puts the old value back before failing.  bit messy, have tidied.

camp/camp-base/src/main/java/org/apache/brooklyn/camp/spi/pdp/AssemblyTemplateConstructor.java
ahgittin: good spot, my mistake

camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/EntitiesYamlTest.java
```diff
@@ -1067,4 +1069,35 @@ protected String getEntityTypeName() {
...
+        String yaml =
+            "services:\n"+
+                "- type: "+TestEntity.class.getName()+"\n"+
+                "- type: "+TestEntity.class.getName()+"\n"+
+                "";
+        doTestLeak(yaml);
+    }
+
+    @Test
+    public void testLeakWithPct() throws Exception {
```
ahgittin: platform component template -- have spelt it out

core/src/main/java/org/apache/brooklyn/core/mgmt/internal/BrooklynGarbageCollector.java
ahgittin: good idea, also report on max/total.  done.

utils/common/src/main/java/org/apache/brooklyn/util/javalang/MemoryUsageTracker.java
ahgittin: yes

utils/common/src/test/java/org/apache/brooklyn/util/javalang/MemoryUsageTrackerTest.java
```diff
@@ -81,7 +81,39 @@ private long sizeOfActiveReferences(List<Maybe<byte[]>> references) {
...
+        Assert.assertTrue(1.0*freeMemory/totalMemory < 0.10, 
+            "Should have had less than 10% free memory before clearance, had "+Strings.makeSizeString(freeMemory)+" / "+Strings.makeSizeString(totalMemory));
+        
+        LOG.info("Forcing memory eviction: "+
+            MemoryUsageTracker.forceClearSoftReferences(100*1000, 10*1000*1000));
+        
+        System.gc(); System.gc();
+        totalMemory = Runtime.getRuntime().totalMemory();
+        freeMemory = Runtime.getRuntime().freeMemory();
+        Assert.assertTrue(1.0*freeMemory/totalMemory > 0.90,
```
ahgittin: if it fails because something else has leaked memory that would also be good to know -- let's watch and see

camp/camp-base/src/main/java/org/apache/brooklyn/camp/spi/pdp/AssemblyTemplateConstructor.java
ahgittin: good spot (thought i added this comment but it seems to have been lost)

ahgittin:
aha i hadn't noticed softs in jmap histo.  that's good info.

agree `MemoryUsageTracker` will need more attention in time...

i'll merge as soon as tests pass

camp/camp-base/src/main/java/org/apache/brooklyn/camp/spi/pdp/AssemblyTemplateConstructor.java
```diff
@@ -40,15 +40,22 @@ public AssemblyTemplateConstructor(CampPlatform campPlatform) {
...
         checkState();
         AssemblyTemplate at = builder.build();
-        transaction.add(at).commit();
+        if (!save) transaction.clear();
+        transaction.commit();
         transaction = null;
         return at;
     }
-
+
```
neykov: Can you just not call commit/construct (for the lack of rollback) and let the objects GC instead of doing the API changes in `camp-base`?

core/src/main/java/org/apache/brooklyn/core/mgmt/internal/BrooklynGarbageCollector.java
```diff
@@ -92,16 +93,28 @@
...
+        "brooklyn.gc.doSystemGc", "whether to periodically call System.gc()", false);
+
+    public static final ConfigKey<Double> FORCE_CLEAR_SOFT_REFERENCES_ON_MEMORY_USAGE_LEVEL = 
+        ConfigKeys.newDoubleConfigKey("brooklyn.gc.clearSoftReferencesOnMemoryUsageLevel", 
+            "force clearance of soft references (by generating a deliberate OOME) "
+            + "if memory usage gets higher than this percentage of available memory; "
+            + "Brooklyn will use up to the max, or this percentage, with soft references,"
+            + "so if using any high-memory-usage alerts they should be pegged quite a bit"
+            + "higher than this threshhold "
+            + "(default >1 means never)", 2.0);
```
neykov: The default an unusual way of saying never. `null` is the non-surprising value for this case, even if it complicates the code a bit.

utils/common/src/main/java/org/apache/brooklyn/util/javalang/MemoryUsageTracker.java
```diff
@@ -68,5 +87,76 @@ public long getBytesUsed() {
...
+    public static String forceClearSoftReferences() {
+        return forceClearSoftReferences(1000*1000, Integer.MAX_VALUE);
+    }
+    /** as {@link #forceClearSoftReferences()} but gives control over headroom and max chunk size.
+     * it tries to undershoot by headroom as it approaches maximum (and then overshoot)
+     * to minimize the chance we take exactly all the memory and starve another thread;
+     * and it uses the given max chunk size in cases where the caller wants more precision
+     * (the available memory will be fragmented so the smaller the chunks the more it can
+     * fill in, but at the expense of time and actual memory provisioned) */
+    public static String forceClearSoftReferences(long headroom, int maxChunk) {
```
neykov: `headroom` not used.

neykov:
+1 Keep in mind that there are other heavy users of soft references out of Brooklyn's control. For example groovy's runtime (which knows about every entity created).

PR #332, opened 2016-09-18
JcloudsLocation: Ability to supply external implementation for the socket

bostko:
Ability to supply external implementation for the socket tester predicate.

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/config/ConfigUtils.java
+++ b/core/src/main/java/org/apache/brooklyn/core/config/ConfigUtils.java
@@ -33,7 +33,9 @@ import org.apache.brooklyn.config.ConfigKey.HasConfigKey;
 import org.apache.brooklyn.core.config.ConfigUtils;
 import org.apache.brooklyn.core.config.WrappedConfigKey;
 import org.apache.brooklyn.core.internal.BrooklynProperties;
+import org.apache.brooklyn.util.core.config.ConfigBag;
 import org.apache.brooklyn.util.exceptions.Exceptions;
+import org.apache.brooklyn.util.text.Strings;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -125,5 +127,21 @@ public class ConfigUtils {
             }
         }
         return Collections.unmodifiableSet(result);
+    }
+
+    /**
+     * Look for keys with <code>configPrefix</code> in <code>configBag</code> and
+     * those which have <code>configPrefix</code> are added in <code>destinationBucket</code> but without <code>configPrefix</code>.
+     * @param configPrefix prefix to look for
+     * @param configBag keys to look in
+     * @param destinationBucket should not be an ImmutableMap
+     */
+    public static void addUnprefixedConfigKeyInConfigBack(String configPrefix, ConfigBag configBag, Map<String, Object> destinationBucket) {
+        for (Map.Entry<ConfigKey<?>, ?> entry : configBag.getAllConfigAsConfigKeyMap().entrySet()) {
+            String keyName = entry.getKey().getName();
+            if (keyName.startsWith(configPrefix)) {
+                destinationBucket.put(Strings.removeFromStart(keyName, configPrefix), entry.getValue());
+            }
+        }
     }
 }
--- a/core/src/main/java/org/apache/brooklyn/core/location/cloud/CloudLocationConfig.java
+++ b/core/src/main/java/org/apache/brooklyn/core/location/cloud/CloudLocationConfig.java
@@ -21,6 +21,8 @@ package org.apache.brooklyn.core.location.cloud;
 import java.util.Collection;
 
 import com.google.common.annotations.Beta;
+import com.google.common.base.Predicate;
+import com.google.common.net.HostAndPort;
 import com.google.common.reflect.TypeToken;
 
 import org.apache.brooklyn.api.location.MachineLocationCustomizer;
@@ -30,6 +32,7 @@ import org.apache.brooklyn.core.config.ConfigKeys;
 import org.apache.brooklyn.core.config.MapConfigKey;
 import org.apache.brooklyn.core.location.LocationConfigKeys;
 import org.apache.brooklyn.util.core.flags.SetFromFlag;
+import org.apache.brooklyn.util.net.Networking;
 
 public interface CloudLocationConfig {
 
@@ -79,7 +82,15 @@ public interface CloudLocationConfig {
             + "if 'false', will default to the node's first public IP (or privae if no public IPs); "
             + "if 'true' uses default duration; otherwise accepts a time string e.g. '5m' (the default) or a number of milliseconds", "5m");
 
-    public static final ConfigKey<String> WAIT_FOR_SSHABLE = ConfigKeys.newStringConfigKey("waitForSshable", 
+    ConfigKey<Predicate<HostAndPort>> POLL_FOR_FIRST_REACHABLE_ADDRESS_PREDICATE = ConfigKeys.newConfigKey(new TypeToken<Predicate<HostAndPort>>(){}, "pollForFirstReachableAddress.predicate",
+            "Predicate<HostAndPort> implementation which checks whether machine is up or not.");
+
+    ConfigKey<Class<? extends Predicate<HostAndPort>>> POLL_FOR_FIRST_REACHABLE_ADDRESS_PREDICATE_TYPE = ConfigKeys.newConfigKey(new TypeToken<Class<? extends Predicate<HostAndPort>>>(){}, "pollForFirstReachableAddress.predicate.type",
+            "Predicate<HostAndPort> class. " +
+            "Other keys prefixed with pollForFirstReachableAddress.predicate.<property> will be passed to the Map constructor of the Predicate<HostAndPort> implementation.",
+            Networking.IsReachablePredicate.class);
+
+    public static final ConfigKey<String> WAIT_FOR_SSHABLE = ConfigKeys.newStringConfigKey("waitForSshable",
             "Whether and how long to wait for a newly provisioned VM to be accessible via ssh; " +
             "if 'false', won't check; if 'true' uses default duration; otherwise accepts a time string e.g. '5m' (the default) or a number of milliseconds", "5m");
 
--- a/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
+++ b/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
@@ -28,6 +28,7 @@ import static org.jclouds.util.Throwables2.getFirstThrowableOfType;
 import java.io.ByteArrayOutputStream;
 import java.io.File;
 import java.io.IOException;
+import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.security.KeyPair;
@@ -2862,7 +2863,39 @@ public class JcloudsLocation extends AbstractCloudMachineProvisioningLocation im
         String result;
         if (enabled) {
             Duration timeout = "true".equals(pollForFirstReachable) ? Duration.FIVE_MINUTES : Duration.of(pollForFirstReachable);
-            result = JcloudsUtil.getFirstReachableAddress(node, timeout);
+
+            Predicate<HostAndPort> pollForFirstReachableHostAndPortPredicate;
+            if (setup.get(POLL_FOR_FIRST_REACHABLE_ADDRESS_PREDICATE) != null) {
+                LOG.debug("Using pollForFirstReachableAddress.predicate supplied from config for location " + this + " "
+                        + POLL_FOR_FIRST_REACHABLE_ADDRESS_PREDICATE.getName() + " : " + setup.get(POLL_FOR_FIRST_REACHABLE_ADDRESS_PREDICATE));
+                pollForFirstReachableHostAndPortPredicate = setup.get(POLL_FOR_FIRST_REACHABLE_ADDRESS_PREDICATE);
+            } else {
+                LOG.debug("Using pollForFirstReachableAddress.predicate.type supplied from config for location " + this + " " + POLL_FOR_FIRST_REACHABLE_ADDRESS_PREDICATE_TYPE.getName()
+                        + " : " + setup.get(POLL_FOR_FIRST_REACHABLE_ADDRESS_PREDICATE_TYPE));
+
+                Class<? extends Predicate<HostAndPort>> predicateType = setup.get(POLL_FOR_FIRST_REACHABLE_ADDRESS_PREDICATE_TYPE);
+
+                Map<String, Object> args = MutableMap.of();
+                ConfigUtils.addUnprefixedConfigKeyInConfigBack(POLL_FOR_FIRST_REACHABLE_ADDRESS_PREDICATE.getName() + ".", setup, args);
+                try {
+                    pollForFirstReachableHostAndPortPredicate = predicateType.getConstructor(Map.class).newInstance(args);
+                } catch (NoSuchMethodException|IllegalAccessException e) {
+                    try {
+                        pollForFirstReachableHostAndPortPredicate = predicateType.newInstance();
+                    } catch (IllegalAccessException|InstantiationException newInstanceException) {
+                        throw Exceptions.propagate("Instantiating " + predicateType + " failed.", newInstanceException);
+                    }
+                } catch (InvocationTargetException|InstantiationException e) {
+                    throw Exceptions.propagate("Problem trying to instantiate " + predicateType + " with Map constructor.", e);
+                }
+            }
+
+            try {
+                result = JcloudsUtil.getFirstReachableAddress(node, timeout, pollForFirstReachableHostAndPortPredicate);
+            } catch (Exception e) {
+                throw Exceptions.propagate("Problem instantiating reachability checker for " + this
+                        + " pollForFirstReachableHostAndPortPredicate: " + pollForFirstReachableHostAndPortPredicate, e);
+            }
             LOG.debug("Using first-reachable address "+result+" for node "+node+" in "+this);
         } else {
             result = Iterables.getFirst(Iterables.concat(node.getPublicAddresses(), node.getPrivateAddresses()), null);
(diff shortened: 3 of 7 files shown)


locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -2862,7 +2863,37 @@ protected String getFirstReachableAddress(NodeMetadata node, ConfigBag setup) {
...
         if (enabled) {
             Duration timeout = "true".equals(pollForFirstReachable) ? Duration.FIVE_MINUTES : Duration.of(pollForFirstReachable);
-            result = JcloudsUtil.getFirstReachableAddress(node, timeout);
+
+            Predicate<HostAndPort> pollForFirstReachableHostAndPortPredicate;
+            if (setup.get(POLL_FOR_FIRST_REACHABLE_ADDRESS_PREDICATE) != null) {
+                LOG.debug("Using pollForFirstReachableAddress.predicate supplied from config for location " + this);
+                pollForFirstReachableHostAndPortPredicate = setup.get(POLL_FOR_FIRST_REACHABLE_ADDRESS_PREDICATE);
+            } else if (setup.get(POLL_FOR_FIRST_REACHABLE_ADDRESS_PREDICATE_TYPE) != null) {
+                LOG.debug("Using pollForFirstReachableAddress.predicate.type supplied from config for location " + this);
```
neykov: Can you log the object as well, here and above.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -2862,7 +2863,37 @@ protected String getFirstReachableAddress(NodeMetadata node, ConfigBag setup) {
...
+                ConfigBag args = new ConfigBag();
+                final String reachablePredicatePrefix = POLL_FOR_FIRST_REACHABLE_ADDRESS_PREDICATE.getName() + ".";
+                for (Map.Entry<ConfigKey<?>, ?> entry : setup.getAllConfigAsConfigKeyMap().entrySet()) {
+                    String keyName = entry.getKey().getName();
+                    if (keyName.startsWith(reachablePredicatePrefix)) {
+                        args.putStringKey(Strings.removeFromStart(keyName, reachablePredicatePrefix), entry.getValue());
+                    }
+                }
+                pollForFirstReachableHostAndPortPredicate = Reflections.invokeConstructorFromArgs(setup.get(POLL_FOR_FIRST_REACHABLE_ADDRESS_PREDICATE_TYPE), new Object[] {args.getAllConfig()}).get();
```
neykov: No need to use `Reflections`. Change to `class.getConstructor(Map.class).newInstance(flags)`.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -2862,7 +2863,37 @@ protected String getFirstReachableAddress(NodeMetadata node, ConfigBag setup) {
...
+            } else {
+                LOG.debug("Using default predicate for pollForFirstReachableAddress for location " + this);
+                pollForFirstReachableHostAndPortPredicate = new Predicate<HostAndPort>() {
```
neykov: Avoid using anonymous classes (mostly because persistence)

neykov: Could also set is the default value on the `.type` key.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -2862,7 +2863,37 @@ protected String getFirstReachableAddress(NodeMetadata node, ConfigBag setup) {
...
+                    @Override public boolean apply(HostAndPort input) {
+                        return Networking.isReachable(input);
+                    }};
+            }
+
+            try {
+                LOG.error("Using pollForFirstReachableAddress.predicate supplied from config for " + this);
+                result = JcloudsUtil.getFirstReachableAddress(node, timeout, pollForFirstReachableHostAndPortPredicate);
+            } catch (Exception e) {
+                throw Exceptions.propagate("Problem instantiating reachability checker for " + this, e);
```
neykov: Log the predicate.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsUtil.java
```diff
@@ -354,6 +355,13 @@ public static String getFirstReachableAddress(ComputeServiceContext context, Nod
     }
     
     public static String getFirstReachableAddress(NodeMetadata node, Duration timeout) {
+        return getFirstReachableAddress(node, timeout, new Predicate<HostAndPort>() {
+            @Override public boolean apply(HostAndPort input) {
+                return Networking.isReachable(input);
+            }});
+    }
```
neykov: This is the same predicate as in the other file. Reuse, avoid anonymous classes.

utils/common/src/test/java/org/apache/brooklyn/util/net/ReachabilityCheckerTest.java
```diff
@@ -0,0 +1,89 @@
...
+    public void testReachibilityChecker() {
+        ReachableSocketFinder finder = new ReachableSocketFinder(connectionChecker(), executor);
+        HostAndPort result = finder.findOpenSocketOnNode(ImmutableList.of(HostAndPort.fromParts(server.getHostName(), server.getPort())), Duration.ONE_MINUTE);
+        Asserts.assertEquals(result, HostAndPort.fromParts(server.getHostName(), server.getPort()));
+    }
+
+    @Test(expectedExceptions = NoSuchElementException.class, expectedExceptionsMessageRegExp = "could not connect to any socket in.*")
+    public void testReachibilityCheckerFails() {
+        ReachableSocketFinder finder = new ReachableSocketFinder(connectionChecker(), executor);
+        finder.findOpenSocketOnNode(ImmutableList.of(HostAndPort.fromParts("localhost", server.getPort() + 1)), Duration.FIVE_SECONDS);
```
neykov: Not necessarily closed. The test is indeterministic.

utils/common/src/test/java/org/apache/brooklyn/util/net/ReachabilityCheckerTest.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    @Test
+    public void testReachibilityCheckerChoosesLaunched() {
+        ReachableSocketFinder finder = new ReachableSocketFinder(connectionChecker(), executor);
+        HostAndPort result = finder.findOpenSocketOnNode(ImmutableList.of(HostAndPort.fromParts("localhost", server.getPort() + 1),HostAndPort.fromParts(server.getHostName(), server.getPort())),
```
neykov: Same here

utils/common/src/test/java/org/apache/brooklyn/util/net/ReachabilityCheckerTest.java
```diff
@@ -0,0 +1,89 @@
...
+    @AfterMethod
+    public void afterMethod() throws IOException {
+        server.shutdown();
+    }
+
+    @AfterClass
+    public void afterClass() {
+        executor.shutdownNow();
+    }
+}
```
neykov: You can split the test into two:
 * how `ReachableSocketFinder` behaves - can be used with a mock/recording predicate. Doesn't need network connectivity and relying on specific ports being opened/closed, also quicker to execute.
 * The predicate - this needs network connectivity, but is a very thin wrapper over existing functionality so not too  importnant.

neykov: Need to add tests covering the location logic - whether keys are used correctly to instantiate and use the predicate. Whether configuration from the location is passed correctly to the predicate.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -2862,7 +2863,37 @@ protected String getFirstReachableAddress(NodeMetadata node, ConfigBag setup) {
...
+            } else if (setup.get(POLL_FOR_FIRST_REACHABLE_ADDRESS_PREDICATE_TYPE) != null) {
+                LOG.debug("Using pollForFirstReachableAddress.predicate.type supplied from config for location " + this);
+                ConfigBag args = new ConfigBag();
+                final String reachablePredicatePrefix = POLL_FOR_FIRST_REACHABLE_ADDRESS_PREDICATE.getName() + ".";
+                for (Map.Entry<ConfigKey<?>, ?> entry : setup.getAllConfigAsConfigKeyMap().entrySet()) {
+                    String keyName = entry.getKey().getName();
+                    if (keyName.startsWith(reachablePredicatePrefix)) {
+                        args.putStringKey(Strings.removeFromStart(keyName, reachablePredicatePrefix), entry.getValue());
+                    }
+                }
```
neykov: This is a common pattern - can it be extracted in a util method/class?

core/src/main/java/org/apache/brooklyn/core/location/cloud/CloudLocationConfig.java
```diff
@@ -79,7 +81,13 @@
             + "if 'false', will default to the node's first public IP (or privae if no public IPs); "
             + "if 'true' uses default duration; otherwise accepts a time string e.g. '5m' (the default) or a number of milliseconds", "5m");
 
-    public static final ConfigKey<String> WAIT_FOR_SSHABLE = ConfigKeys.newStringConfigKey("waitForSshable", 
+    ConfigKey<Predicate<HostAndPort>> POLL_FOR_FIRST_REACHABLE_ADDRESS_PREDICATE = ConfigKeys.newConfigKey(new TypeToken<Predicate<HostAndPort>>(){}, "pollForFirstReachableAddress.predicate",
+            "Predicate<HostAndPort> implementation which checks whether machine is up or not.");
+
+    ConfigKey<Class<Predicate<HostAndPort>>> POLL_FOR_FIRST_REACHABLE_ADDRESS_PREDICATE_TYPE = ConfigKeys.newConfigKey(new TypeToken<Class<Predicate<HostAndPort>>>(){}, "pollForFirstReachableAddress.predicate.type",
```
bostko: compiler doesn't accept it that way.

neykov: See [here](https://github.com/apache/brooklyn-library/blob/master/software/nosql/src/main/java/org/apache/brooklyn/entity/nosql/cassandra/CassandraDatacenter.java#L77) for an working example.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
bostko: Previously this predicate was not persisted and I think it should stay like that, wdyt?

neykov: That's right, but you can never be too careful with persistence. Also the same code is used at two other places so makes sense to reuse it.

bostko: Some of the comments were addressed. The new UI doesn't reflect that :(

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
bostko: @neykov could you check the method I added in ConfigUtils

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -2862,15 +2863,61 @@ protected String getFirstReachableAddress(NodeMetadata node, ConfigBag setup) {
...
+                });
+                Optional<Constructor<?>> mapConstructor = Iterables.tryFind(constructors, new Predicate<Constructor<?>>() {
+                    @Override public boolean apply(Constructor<?> constructor) {
+                        if (constructor.getParameterTypes().length == 1) {
+                            return constructor.getParameterTypes()[0].isAssignableFrom(Map.class);
+                        } else {
+                            return false;
+                        }
+                    }
+                });
```
neykov: I'd use `type.getConstructor(Map.class)` instead with the no arguments path in the `catch(NoSuchMethodException)` block.

core/src/main/java/org/apache/brooklyn/core/config/ConfigUtils.java
```diff
@@ -126,4 +128,20 @@ else if (HasConfigKey.class.isAssignableFrom(f.getType()))
...
+     * @param destinationBucket should not be an ImmutableMap
+     */
+    public static void addUnprefixedConfigKeyInConfigBack(String configPrefix, ConfigBag configBag, Map<String, Object> destinationBucket) {
+        for (Map.Entry<ConfigKey<?>, ?> entry : configBag.getAllConfigAsConfigKeyMap().entrySet()) {
+            String keyName = entry.getKey().getName();
+            if (keyName.startsWith(configPrefix)) {
+                destinationBucket.put(Strings.removeFromStart(keyName, configPrefix), entry.getValue());
+            }
+        }
+    }
```
neykov: Make `destinationBucket` a return value instantiated in the method, remove it from arguments.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -2862,15 +2863,61 @@ protected String getFirstReachableAddress(NodeMetadata node, ConfigBag setup) {
...
+                    } catch (ReflectiveOperationException e) {
+                        throw Exceptions.propagate(e);
+                    }
+                } else {
+                    try {
+                        pollForFirstReachableHostAndPortPredicate = (Predicate<HostAndPort>)emptyConstructor.get().newInstance();
+                    } catch (ReflectiveOperationException e) {
+                        throw Exceptions.propagate(e);
+                    }
+                }
```
neykov: Could log a user-friendly message for the case where both map and empty constructors are missing.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -2862,15 +2863,61 @@ protected String getFirstReachableAddress(NodeMetadata node, ConfigBag setup) {
...
+            } catch (Exception e) {
+                throw Exceptions.propagate("Problem instantiating reachability checker for " + this
+                        + " pollForFirstReachableHostAndPortPredicate: " + pollForFirstReachableHostAndPortPredicate, e);
+            }
         } else {
             result = Iterables.getFirst(Iterables.concat(node.getPublicAddresses(), node.getPrivateAddresses()), null);
             if (result == null) {
                 throw new IllegalStateException("No addresses available for node "+node+" in "+this);
             }
-            LOG.debug("Using first address "+result+" for node "+node+" in "+this);
```
neykov: This is an important distinction for troubleshooting. Leave both the first and first-reachable log statements.

locations/jclouds/src/test/java/org/apache/brooklyn/location/jclouds/JcloudsLocationReachabilityPredicateInstantiation.java
```diff
@@ -0,0 +1,147 @@
+/*
```
neykov: Can you end the name in `Test` to follow existing practices.

utils/common/src/main/java/org/apache/brooklyn/util/net/ReachableSocketFinder.java
```diff
@@ -91,7 +91,7 @@ public HostAndPort findOpenSocketOnNode(final Collection<? extends HostAndPort>
                 .backoffTo(Duration.FIVE_SECONDS)
                 .until(new Callable<Boolean>() {
                         public Boolean call() {
-                            Optional<HostAndPort> reachableSocket = tryReachable(sockets, Duration.seconds(2));
+                            Optional<HostAndPort> reachableSocket = tryReachable(sockets, Duration.FIVE_SECONDS);
```
neykov: I think this should be set in the predicate or configurable. It's implementation dependant could vary.

locations/jclouds/src/test/java/org/apache/brooklyn/location/jclouds/JcloudsLocationReachabilityPredicateInstantiation.java
```diff
@@ -0,0 +1,147 @@
...
+        public BailOutJcloudsLocation() {
+            super();
+        }
+
+        @Override
+        public String getFirstReachableAddress(NodeMetadata nodeMetada, ConfigBag setup) {
+            return super.getFirstReachableAddress(nodeMetada, setup);
+        }
+    }
+}
```
neykov: Well done! I like how you were able to unit test the location code.
I think you should add negative tests as well - for the case where the node is not reachable - either with minimal timeouts or as integration tests.

core/src/main/java/org/apache/brooklyn/core/config/ConfigUtils.java
bostko: I made it this way because it is convenient to use this way in https://github.com/apache/brooklyn-server/blob/master/core/src/main/java/org/apache/brooklyn/location/ssh/SshMachineLocation.java#L636-L645

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
bostko: NoSuchMethod exception is still being logged.
Adding in the inner catch code like
```java
if (NoSuchMethodException.class.isAssignableFrom(e.getCause().getClass())) {
  LOG.debug("No matching constructor", e);
} 
```
is too much I think.
WDYT?

neykov:
Thanks @bostko, merging.

PR #335, opened 2016-09-20
Update geoip to latest version

neykov:
Removes the dependency on google-http-client, depending directly on apache-httpcomponents

Identical to https://github.com/apache/brooklyn-server/pull/327

Depends on https://github.com/apache/brooklyn-server/pull/317


ahgittin:
To clarify #327 was reverted hence the need to replay it.

#317 is replaced by #339 so this depends on #339

sjcorbett:
Looks good. Will merge.

PR #337, opened 2016-09-20
Add config inheritance to customizers key

grkvlt:
Enables merging of customizsers for locations.


ahgittin:
failure is unrelated

intention makes sense

however this will do nothing as deep_merge only applies to maps.

with #340 you could relatively easily introduce a new `COLLECTION_ANCESTORS_FIRST` or `COLLECTION_ANCESTORS_LAST` into `BasicConfigInheritance` which would make sense for sets or lists (but with different semantics).  or maybe simpler change the `deepMerge` method in that class to support collections, probably adding ancestors last.

that change plus a test for this commit which tests that change would be welcome.

grkvlt:
@ahgittin understood; i will work on the change you suggest instead. closing this for now

PR #338, opened 2016-09-21
Support nested catalog item definitions - testCatalogItemIdInReferencedItems

geomacy:
For [testCatalogItemIdInReferencedItems](https://github.com/apache/brooklyn-server/blob/master/camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/catalog/CatalogYamlEntityTest.java#L955). 

The issue as described on that test is where one type is defined in a catalog in terms of another, i.e. with a structure like:

```
brooklyn.catalog:
  items:

  - id: inner
    icon_url: classpath://path/to/myicon.jpg
    brooklyn.libraries:
    - url: http://example.com/jar-with-myicon.jar
    item: 
      type: org.apache.brooklyn.entity.stock.BasicEntityImpl

  - id: outer
    item: inner
```

In this case the entity will get assigned a catalogItemId of the outer-most catalog
item, which doesn't have the necessary libraries with visibility to the entity's classpath.
When loading resources such as the icon above from inside the entity then we will use the wrong class loading context (BrooklynClassLoadingContextSequential). 
Note, a [workaround](https://github.com/apache/brooklyn-server/blob/master/core/src/main/java/org/apache/brooklyn/core/mgmt/internal/AbstractManagementContext.java#L135-L139) has been implemented which explicitly adds the entity's class loader to the fallbacks, but it does not cover the general case.

This PR addresses the above by updating BrooklynObject and specifications to support a stack of 'nested' catalogItemIds, so that the above case would record both inner and outer. This allows all of the nested catalog definitions to be taken into account when searching for resources.  In some more detail:
- Update BrooklynObject and BrooklynObjectSpec to support a list of nested catalogItemIds. 
  This lets testCatalogItemIdInReferencedItems work (in  modified form). 
- Modify resource loading to use the nested catalog item ids when creating the class loader context for entities, see new tests  `testIndirectCatalogItemCanLoadResources` and `testDeepCatalogItemCanLoadResources`.
- Update persistence to support the nested ids. See the  'post-rebind' equivalent of the
  `testDeepCatalogItemCanLoadResources` test, called `testReboundDeepCatalogItemCanLoadResources`.
- Support rebind of previously persisted state (with single catalogItemId), and test this in
  `testRebindWithCatalogAndAppRebindCatalogItemIds`.

I want to look into adding some test framework tests for this but I think the PR can already be reviewed in its present state. At present the manual tests I have carried out cover the following use-cases:
- Deploy applications (classic launcher + Karaf)
- Stop Brooklyn and restart, verify that rebind succeeds (classic + Karaf)
- Deploy app on Brooklyn built from current master branch; stop it; start Brooklyn built from this PR and verify that the persisted state is correctly rebound, and effectors can be successfully called etc. (classic + Karaf)

### DIFF
--- a/api/src/main/java/org/apache/brooklyn/api/internal/AbstractBrooklynObjectSpec.java
+++ b/api/src/main/java/org/apache/brooklyn/api/internal/AbstractBrooklynObjectSpec.java
@@ -68,7 +68,9 @@ public abstract class AbstractBrooklynObjectSpec<T, SpecT extends AbstractBrookl
 
     private final Class<? extends T> type;
     private String displayName;
-    private Deque<String> catalogItemIdStack;
+    private String catalogItemId;
+    private Deque<String> catalogItemIdSearchPath = new ArrayDeque<>();
+
     private Set<Object> tags = MutableSet.of();
     private List<SpecParameter<?>> parameters = ImmutableList.of();
 
@@ -108,13 +110,16 @@ public abstract class AbstractBrooklynObjectSpec<T, SpecT extends AbstractBrookl
      * (or TODO we add a separate field to record other catalog item IDs that could be applied for searching, see below)
      */
     public SpecT catalogItemId(String val) {
-        getCatalogItemIdStack().clear();
-        return stackCatalogItemId(val);
+        catalogItemId = val;
+        return self();
     }
 
-    protected SpecT catalogItemIdStack(Collection<String> catalogItemIdStack) {
-        this.catalogItemIdStack = null;
-        getCatalogItemIdStack().addAll(catalogItemIdStack);
+    public synchronized SpecT catalogItemIdAndSearchPath(String catalogItemId, Collection<String> searchPath) {
+        if (catalogItemId != null) {
+            catalogItemId(catalogItemId);
+            catalogItemIdSearchPath.clear();
+            catalogItemIdSearchPath.addAll(searchPath);
+        }
         return self();
     }
 
@@ -125,30 +130,37 @@ public abstract class AbstractBrooklynObjectSpec<T, SpecT extends AbstractBrookl
     @Beta
     public SpecT catalogItemIdIfNotNull(String val) {
         if (val!=null) {
-            stackCatalogItemId(val);
+            catalogItemId(val);
         }
         return self();
     }
 
-    private Deque<String> getCatalogItemIdStack() {
-        if (catalogItemIdStack == null) {
-            catalogItemIdStack = new ArrayDeque<>();
+    protected Object readResolve() {
+        if (catalogItemIdSearchPath == null) {
+            catalogItemIdSearchPath = new ArrayDeque<>();
         }
-        return catalogItemIdStack;
+        return this;
     }
 
     /**
      * Adds (stacks) the catalog item id of a wrapping specification.
      * Does nothing if the value is null.
+     * If the value is not null, and is not the same as the current
+     * catalogItemId, then the *current* catalogItemId will
+     * be moved to the start of the search path, and the value supplied
+     * as the parameter here will be set as the new catalogItemId.
      * <p>
-     * Used when we want to collect nested item ID's so that *all* can be searched.
+     * Used when we want to collect IDs of items that extend other items, so that *all* can be searched.
      * e.g. if R3 extends R2 which extends R1 any one of these might supply config keys
      * referencing resources or types in their local bundles.
      */
     @Beta
     public SpecT stackCatalogItemId(String val) {
-        if (null != val && (getCatalogItemIdStack().isEmpty() || !getCatalogItemIdStack().element().equals(val))) {
-            getCatalogItemIdStack().addFirst(val);
+        if (null != val) {
+            if (null != catalogItemId && !catalogItemId.equals(val)) {
+                catalogItemIdSearchPath.addFirst(catalogItemId);
+            }
+            catalogItemId(val);
         }
         return self();
     }
@@ -239,35 +251,26 @@ public abstract class AbstractBrooklynObjectSpec<T, SpecT extends AbstractBrookl
         return displayName;
     }
 
-    /**
-     * @deprecated since 0.11.0, use getOuterCatalogItemId or getInnerCatalogItemIds as appropriate
-     */
-    @Deprecated
     public final String getCatalogItemId() {
-        return getOuterCatalogItemId();
+        return catalogItemId;
     }
 
-    public final String getOuterCatalogItemId() {
-        if (getCatalogItemIdStack().size() != 0) {
-            return getCatalogItemIdStack().getFirst();
-        }
-        return null;
-    }
-
-
-        /**
-         * An immutable list of ids of this object's catalog item and its defining catalog items.
-         * Wrapping items are first in the list (i.e. wrapping items precede wrapped items),
-         * for example, if the catalog item is defined as
-         * <pre>
-         *     items:
-         *     - id: X
-         *       item: Y
-         * </pre>
-         * then the list will contain X, Y.
-         */
-    public final List<String> getCatalogItemIdHierarchy() {
-        return ImmutableList.copyOf(catalogItemIdStack);
+    /**
+     * An immutable list of ids of catalog items that define this item.
+     * Wrapping items are first in the list (i.e. wrapping items precede wrapped items),
+     * for example, for a spec of an item of type Z, where the catalog is:
+     * <pre>
+     *     items:
+     *     - id: X
+     *     - id: Y
+     *       item: X
+     *     - id: Z
+     *       item: Y
+     * </pre>
+     * then the spec will have getCatalogId() of Z and getCatalogItemIdSearchPath()  of Y, X.
+     */
+    public final List<String> getCatalogItemIdSearchPath() {
+        return ImmutableList.copyOf(catalogItemIdSearchPath);
     }
 
     public final Set<Object> getTags() {
@@ -317,7 +320,7 @@ public abstract class AbstractBrooklynObjectSpec<T, SpecT extends AbstractBrookl
             .configure(otherSpec.getConfig())
             .configure(otherSpec.getFlags())
             .tags(otherSpec.getTags())
-            .catalogItemIdStack(otherSpec.getCatalogItemIdHierarchy())
+            .catalogItemIdAndSearchPath(otherSpec.getCatalogItemId(), otherSpec.getCatalogItemIdSearchPath())
             .parameters(otherSpec.getParameters());
     }
 
@@ -327,7 +330,8 @@ public abstract class AbstractBrooklynObjectSpec<T, SpecT extends AbstractBrookl
         if (!obj.getClass().equals(getClass())) return false;
         AbstractBrooklynObjectSpec<?, ?> other = (AbstractBrooklynObjectSpec<?, ?>) obj;
         if (!Objects.equal(getDisplayName(), other.getDisplayName())) return false;
-        if (!Objects.equal(getCatalogItemIdHierarchy(), other.getCatalogItemIdHierarchy())) return false;
+        if (!Objects.equal(getCatalogItemId(), other.getCatalogItemId())) return false;
+        if (!Objects.equal(getCatalogItemIdSearchPath(), other.getCatalogItemIdSearchPath())) return false;
         if (!Objects.equal(getType(), other.getType())) return false;
         if (!Objects.equal(getTags(), other.getTags())) return false;
         if (!Objects.equal(getConfig(), other.getConfig())) return false;
@@ -338,7 +342,7 @@ public abstract class AbstractBrooklynObjectSpec<T, SpecT extends AbstractBrookl
 
     @Override
     public int hashCode() {
-        return Objects.hashCode(getCatalogItemIdHierarchy(), getDisplayName(), getType(), getTags());
+        return Objects.hashCode(getCatalogItemIdSearchPath(), getDisplayName(), getType(), getTags());
     }
 
     /**
--- a/api/src/main/java/org/apache/brooklyn/api/internal/AbstractBrooklynObjectSpec.java
+++ b/api/src/main/java/org/apache/brooklyn/api/internal/AbstractBrooklynObjectSpec.java
@@ -218,7 +218,7 @@ public abstract class AbstractBrooklynObjectSpec<T,SpecT extends AbstractBrookly
     }
 
     /**
-     * Get immutable list of ids of this object's catalog item and its nested catalog items.
+     * An immutable list of ids of this object's catalog item and its defining catalog items.
      * e.g. if the catalog item is defined as
      * <pre>
      *     items:
@@ -227,7 +227,7 @@ public abstract class AbstractBrooklynObjectSpec<T,SpecT extends AbstractBrookly
      * </pre>
      * then the list will contain X, Y.
      */
-    public final List<String> getNestedCatalogItemIds() {
+    public final List<String> getCatalogItemSuperIds() {
         return ImmutableList.copyOf(catalogItemIdStack);
     }
 
--- a/api/src/main/java/org/apache/brooklyn/api/internal/AbstractBrooklynObjectSpec.java
+++ b/api/src/main/java/org/apache/brooklyn/api/internal/AbstractBrooklynObjectSpec.java
@@ -90,9 +90,9 @@ public abstract class AbstractBrooklynObjectSpec<T, SpecT extends AbstractBrookl
     @Override
     public String toString() {
         return MoreObjects.toStringHelper(this).omitNullValues()
-            .add("type", type)
-            .add("displayName", displayName)
-            .toString();
+                .add("type", type)
+                .add("displayName", displayName)
+                .toString();
     }
 
     protected abstract void checkValidType(Class<? extends T> type);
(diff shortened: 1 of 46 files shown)


geomacy:
Latest update - adds resource lookup from nested catalog definitions.

Nested resources already have a test, [`testIndirectCatalogItemCanLoadResources`](https://github.com/geomacy/brooklyn-server/blob/master/camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/catalog/CatalogYamlEntityTest.java#L919), which succeeds, however, this is because of the workaround @neykov mentions [here](https://github.com/geomacy/brooklyn-server/blob/master/camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/catalog/CatalogYamlEntityTest.java#L952-L953).  

The latest couple of commits above first introduce a test, `testDeepCatalogItemCanLoadResources`, which illustrates that the workaround doesn't cover the more general case, and then updates the resource loading code to use the nested catalog item ids to load resources, which fixes the test.

core/src/main/java/org/apache/brooklyn/core/catalog/internal/CatalogItemDtoAbstract.java
```diff
@@ -45,6 +45,8 @@
 import com.google.common.collect.Iterables;
 import com.google.common.collect.Sets;
 
+// TODO add support for nested catalog items, implement nestCatalogItemId in terms of symbolicName/Version
+// TODO also getCatalogItemSuperIds.
```
geomacy: Comment from @neykov ([here](https://github.com/neykov/brooklyn-server/pull/3#discussion_r79803420)):

I think we should either override with NOPs, or just keep the default behaviour from AbstractBrooklynObject. The catalog items don't know what's the meaning of the plan in them.

core/src/main/java/org/apache/brooklyn/core/catalog/internal/CatalogUtils.java
```diff
@@ -170,6 +173,13 @@ public static void installLibraries(ManagementContext managementContext, @Nullab
     public static String getCatalogItemIdFromLoader(BrooklynClassLoadingContext loader) {
         if (loader instanceof OsgiBrooklynClassLoadingContext) {
             return ((OsgiBrooklynClassLoadingContext)loader).getCatalogItemId();
+        } else if (loader instanceof BrooklynClassLoadingContextSequential) {
+            final Iterator<BrooklynClassLoadingContext> iterator = ((BrooklynClassLoadingContextSequential) loader).getPrimaries().iterator();
+            if (iterator.hasNext()) {
+                BrooklynClassLoadingContext osgiLoader = iterator.next();
+                return getCatalogItemIdFromLoader(osgiLoader);
+            }
+            else return null;
```
geomacy: Comment from @neykov ([here](https://github.com/neykov/brooklyn-server/pull/3#discussion_r79803301)):

Need to revisit this logic in light of latest updates.

core/src/main/java/org/apache/brooklyn/core/mgmt/internal/AbstractManagementContext.java
```diff
@@ -128,20 +128,18 @@ private static DataGridFactory loadDataGridFactory(BrooklynProperties properties
...
-                            BrooklynClassLoadingContext seqLoader = new BrooklynClassLoadingContextSequential(internal.getManagementContext(), itemLoader, entityLoader);
-                            return seqLoader;
-                        } else {
-                            log.error("Can't find catalog item " + internal.getCatalogItemId() +
-                                    " used for instantiating entity " + internal +
-                                    ". Falling back to application classpath.");
+                    final List<String> catalogItemSuperIds = internal.getCatalogItemSuperIds();
+                    if (catalogItemSuperIds.size() > 0) {
+                        BrooklynClassLoadingContextSequential seqLoader = new BrooklynClassLoadingContextSequential(internal.getManagementContext());
+                        for (String catalogItemId : catalogItemSuperIds) {
```
geomacy: check - probably as well as the catalog items we *should* still also use the entity class loader?

neykov: Agree, the entity could be coming from another place like the white list (included in the distribution). Makes sense to include the entity's class loader as well.

geomacy:
Caught up and rebased against master.

geomacy:
Caught up and rebased against master.

core/src/main/java/org/apache/brooklyn/core/objs/AbstractBrooklynObject.java
```diff
@@ -51,7 +55,7 @@
     @SetFromFlag("id")
     private String id = Identifiers.makeRandomLowercaseId(10);
 
-    private String catalogItemId;
+    private Deque<String> catalogItemIdStack = new ArrayDeque<>();
```
neykov: `BrooklynObject`s get created by calling their constructors so removing `catalogItemId` and initializing `catalogItemIdStack`  should be safe.

core/src/main/java/org/apache/brooklyn/core/objs/AbstractBrooklynObject.java
```diff
@@ -80,7 +84,7 @@ public AbstractBrooklynObject(Map<?, ?> properties) {
             _legacyConstruction = true;
         }
 
-        catalogItemId = ApiObjectsFactory.get().getCatalogItemIdFromContext();
+        setCatalogItemId(ApiObjectsFactory.get().getCatalogItemIdFromContext());
```
neykov: Previously `catalogItemId` would be overwritten by any code that knows better. Now most locations should be calling `nestCatalogItemId`. Could this introduce a non-completely-correct inner-most catalog item?

geomacy:
Thanks for the detailed review comments @neykov, will work through them as soon as I get a chance.

geomacy:
Caught up and rebased against master. Going to make a start on the review comments from Svet.

geomacy:
@neykov, have addressed some of the review comments, see the links in the commit messages.  Will add a couple more remarks above.  

The main outstanding changes are those from the comments on CatalogUtils from 5th January https://github.com/apache/brooklyn-server/pull/338#discussion_r94623208, and on AbstractBrooklynObject https://github.com/apache/brooklyn-server/pull/338#discussion_r94763625.  I can just get on with doing the former but the latter will need a bit of thought and more discussion.

api/src/main/java/org/apache/brooklyn/api/internal/AbstractBrooklynObjectSpec.java
```diff
@@ -65,7 +68,7 @@
     
     private final Class<? extends T> type;
     private String displayName;
-    private String catalogItemId;
```
geomacy: surely I take this into account already? See testRebindWithCatalogAndAppRebindCatalogItemIds,
and BrooklynMementoPersisterToObjectStore.getCatalogItemIds().

Could add it back but maybe don't need to?

api/src/main/java/org/apache/brooklyn/api/internal/AbstractBrooklynObjectSpec.java
```diff
@@ -187,7 +202,24 @@ public final String getDisplayName() {
     }
     
     public final String getCatalogItemId() {
-        return catalogItemId;
+        if (catalogItemIdStack.size() != 0) {
+            return catalogItemIdStack.getFirst();
+        }
+        return null;
```
geomacy: Seems preferable not to retain catalogItemId if we can get away without it, see
comment https://github.com/apache/brooklyn-server/pull/338#discussion_r102679601 above.

ahgittin:
Very cool.  Haven't reviewed in depth but looks on point at a high level.  A few specific comments:

* this will conflict with https://github.com/apache/brooklyn-server/pull/573 -- should be easy to resolve, and this is a better solution to the underlying problem

* I think we want _both_ `catalogItemId` with the semantics of #573 and the `stack` introduced here for searching, on the spec (that's what we have on entity)

* there may be persistence errors migrating from older version if we persist specs and remove the `catalogItemId` here (?); the previous suggestion to keep both should resolve this (perhaps with an additional check when initializing an entity that if `catalogItemId` is set then `stack.get(0)` equals that id

And one general comment -- there is an issue if e.g. `my-cluster` and `my-node` are set in the same BOM, both at version `1` with the former referring to the latter omitting the version.  If we then install a version `2` of that bom, then `my-cluster:1` creates a cluster of `my-node:2` which is surprising.  It would be nice to solve that (and also be able to retrieve the `bom` file and bundle list where an entity is defined, for subsequent editting).  It's a separate problem to this, but the solutions may be interrelated.  Very happy to have this merged (once someone else does a detailed review and we resolve #573 conflicts) without solving the version and catalog-group / bom questions, expecting a bit more change afterwards, but wanted to flag that.  cc @geomacy @neykov @aledsage

geomacy:
Thanks @ahgittin and @neykov for comments; I will try to address the remaining items as noted above (https://github.com/apache/brooklyn-server/pull/338#issuecomment-281950519) soon.

I will have a look at #573 to see how this can be merged with it.

Just on the point of omitting or retaining the `catalogItemId` member (https://github.com/apache/brooklyn-server/pull/338#discussion_r102679601),  for rebind purposes, what do you think of my changes to `BrooklynMementoPersisterToObjectStore` which as far as I can see allow me to avoid keeping that member.  I am able to re-run all the tests from `testRebindWithCatalogAndApp` with persisted content with the format prior to this change (see https://github.com/apache/brooklyn-server/pull/338/files#diff-3804f33dfb6d89e123adcff13f9c114dR173).   It looked to me as if we don't need to retain that member, hence the changes above, but if you think there's another case to cover then I can put it back (and will probably need to think about adding more tests to catch that case).

ahgittin:
Hi @geomacy - have added another commit to #573 ; hadn't realized other inheritance semantics re catalog item id.  It now has tests and several explicit TODO comments which this PR should look at.  If you've done this right all those TODOs should be resolved or trivially resolvable, but there were quite a few so I marked anything that might be able to fall through the cracks.

ahgittin:
As for `catalogItemId`,  I think we should continue to populate and persist this.  It is useful to know from whence something came, and for audit purposes this will become even more important:  e.g. which version is it running.  I realize for persistence purposes we can get the exact same class back using the search paths, but we want to be able to find instances of `item:1.0` e.g. so we can upgrade them to `item:2.0`, and I think without the `catalogItemId` field we won't be able to do that.  WDYT?

geomacy:
Thanks @ahgittin I will have a look at #573 hopefully today or very soon.  Nice to have extra tests!

Regarding the persistence,  basically in the persisted XML my change replaces `<catalogItemId>` with the hierarchy of the catalog ids that defined the item (so if `foo` is defined as an extension of `bam` which is in turn a `bar`, you get
```
<catalogItemIdHierarchy>
  <string>foo</string>
  <string>bam</string>
  <string>bar</string>
</catalogItemIdHierarchy>
```
so I am continuing to populate and persist catalogItemId, just implemented as this list. Also I have code in place to handle previously persisted state (still using `<catalogItemId>foo</catalogItemId>`) and load it into this new model. 

Re. the question of upgrades from 1.0 to 2.0 I will need to have a look at it, I can't remember what the test coverage is like for this scenario (I think there are some tests).

ahgittin:
@geomacy I think we want `catalogItemId` explicit as the catalog item ID of the item, in addition to the "hierarchy" (which is maybe better called `catalogItemsSearchPath` ?).  It is a separate concept, and feels messy assuming the first in the hierarchy is the ID of the item (would be cleaner IMO to always put the ID of the item at the top of the hierarchy and drop the hierarchy if superfluous).  We also need to be able to tell if an item does not have a catalog item ID (but possibly things in a search path from its creation context) eg it was created straight from a java class.

ahgittin:
I expect the code comments introduced in #573 will clear up https://github.com/apache/brooklyn-server/pull/338#discussion_r94763625 .

geomacy:
@ahgittin, @neykov I've made the changes to rework the previous commits as suggested by Alex, so  instead of a single stack of catalogItemIds, the original catalogItemId is preserved and a searchPath is added.

I haven't included updates to address the TODO items in #573 here.  I did make a start on that but found some issues, so I have included those changes in a separate PR, https://github.com/apache/brooklyn-server/pull/604. Perhaps we can treat that as a separate issue and review this one in its own right.

api/src/main/java/org/apache/brooklyn/api/internal/AbstractBrooklynObjectSpec.java
```diff
@@ -51,21 +54,23 @@
...
 
     private static final long serialVersionUID = 3010955277740333030L;
 
     private static final Logger log = LoggerFactory.getLogger(AbstractBrooklynObjectSpec.class);
-    
+
     private final Class<? extends T> type;
     private String displayName;
     private String catalogItemId;
+    private Deque<String> catalogItemIdSearchPath = new ArrayDeque<>();
```
ahgittin: minor - would be nice to avoid duplicates

api/src/main/java/org/apache/brooklyn/api/internal/AbstractBrooklynObjectSpec.java
```diff
@@ -91,26 +96,39 @@ public String toString() {
...
-    // that is right, because the wrapper's defn will refer to the wrapped,
-    // but we might need also to collect the item ID's so that *all* can be searched, see #catalogItemId.
-    // e.g. if R3 references R2 which references R1 any one of these might supply config keys 
-    // referencing resources or types in their local bundles. 
+
+    /**
+     * Set the immediate catalog item ID of this object, and the search path of other catalog items used to define it.
+     */
+    public synchronized SpecT catalogItemIdAndSearchPath(String catalogItemId, Collection<String> searchPath) {
+        if (catalogItemId != null) {
```
ahgittin: what is the reason for this behaviour on null?  ignoring searchPath seems weird.

api/src/main/java/org/apache/brooklyn/api/mgmt/rebind/mementos/Memento.java
```diff
@@ -43,11 +44,19 @@
...
-    
-    public String getCatalogItemId();
-    
-    public String getDisplayName();
+    String getType();
+
+    /**
+     * The principal catalog item id.
+     */
+    String getCatalogItemId();
```
ahgittin: it's for more than persistence i think -- javadoc is accurate

ahgittin:
Still going through it, but so far looks really good.

The test `CatalogYamlRebindTest.testRebindWithCatalogAndAppRebindCatalogItemIds` seems to take a long time (28s).  I'll look at it as part of this, and if it's quick to fix I'll just do it.

core/src/main/java/org/apache/brooklyn/core/mgmt/rebind/RebindIteration.java
```diff
@@ -1076,7 +1151,7 @@ protected Policy newPolicy(PolicyMemento memento) {
                 policy = invokeConstructor(null, policyClazz, new Object[] {flags});
             }
             
-            setCatalogItemId(policy, transformedCatalogItemId);
+            setCatalogItemIds(policy, memento.getCatalogItemId(), memento.getCatalogItemIdSearchPath());
```
ahgittin: i think this should be:

> setCatalogItemIds(policy, loaded.catalogItemId, loaded.searchPath);

in case there were any transformations that were applied (by `load(...)`)

core/src/main/java/org/apache/brooklyn/core/objs/AbstractBrooklynObject.java
```diff
@@ -83,11 +90,18 @@ public AbstractBrooklynObject(Map<?, ?> properties) {
         // correct behaviour should be to inherit context's search path, perhaps, though maybe that's better done as spec?
         // in any case, should not define it as _the_ catalog item ID; also see assignment based on parent
         // in CatalogUtils.setCatalogItemIdOnAddition
-        catalogItemId = ApiObjectsFactory.get().getCatalogItemIdFromContext();
+        setCatalogItemId(ApiObjectsFactory.get().getCatalogItemIdFromContext());
```
ahgittin: context ID should be added to the _search path_ not used to set the ID of the item itself -- this was the problem spot that prompted keeping the IDs separate

and can remove the TODO now!  :)

ahgittin:
been through it.  looks good.  may be some subtleties which will only come out as we use it but feels like the right idea and rightly done.  minor changes.  i'll look at the failing test -- seems like a real problem, maybe due to not deduping the list? -- then merge with the fix, unless there is anything non-obvious.

ahgittin:
In maven, this tests fails eventually with an OOME:

```
2017-04-21 15:11:29,630 INFO  TESTNG INVOKING: "Surefire test" - org.apache.brooklyn.camp.brooklyn.catalog.CatalogYamlRebindTest.testRebindWithCatalogAndAppRebindCatalogItemIds(org.apache.brooklyn.camp.brooklyn.catalog.CatalogYamlRebindTest$RebindWithCatalogTestMode, org.apache.brooklyn.camp.brooklyn.catalog.CatalogYamlRebindTest$OsgiMode)(value(s): DISABLE_CATALOG, LIBRARY)
```

i see lots of GC at around 700MB, and failed bundle loads beforehand.  but it runs fine in the IDE. 
 curious.

ahgittin:
i think #646 adds what is missing for those tests to run better - TBC tho!

PR #340, opened 2016-09-21
ConfigMap refactor, respecting inheritance

ahgittin:
As discussed on #281, there were some severe limitations in the way we handled config and inheritance.  This addresses many of those, and others, including:

* A distinction between `NOT_REINHERITED` and `NEVER_INHERITED`
* Lift evaluation of inheritance so callers don't repeat code
* Bringing `Location` config in line with `Entity` and `EntityAdjunct`
* Clarification of what is returned by "get all" methods in an inherited, merging, resolving, coercing world
* REST API and other methods no longer show dis-inherited config
* DSL components are resolved against the ancestor who declares the value
* API supports accessing which ancestor declares a value, for contexts which need e.g. to resolve a URL in the bundle context of a catalog item

This fixes some broken tests and removes a whole lot of TODO comments (whilst only adding a few!).  It does come at a cost of complexity (especially in the evaluation of inherited config; however since *all* tests pass it doesn't need super-close study at this point, I think).  I'm convinced it's necessary complexity unless we move away from this style of inheritance (which in time we might well be able to do!).

There are a few areas where backwards compatibility is impacted which will need to go into the release notes:

* Config is now resolved against the ancestor where it is defined, e.g. attributeWhenReady;
    TODO do we need a way to specify that config should be resolved against the resolver?
* `ConfigMap.submap` is no longer supported for entity config; findKeys has been introduced, and the method is available on the main sub-interfaces where it was used
* Location config is treated like Entities and Adjuncts; it respects inheritance
    * as a consequence of this, some persisted usage information is no longer available

(The biggest of these is Location config changes now needing to go through the right API; you can no longer just edit the bag.  I think I've caught most of these but test with care!)


ahgittin:
I think there are just two things discussed last week which I've not addressed:

* For classpath URL resolution from runtime ancestors, the caller still needs to look up the declaring entity (but now they can do so, at least)
* For classpath URL resolution from type def ancestors, we need to make sure ancestor bundles are part of the search path (pretty much completely separate)

@neykov I've not worked through your 3 scenarios in detail but they are starting to feel more tractable.

/cc @aledsage @sjcorbett @grkvlt @geomacy

api/src/main/java/org/apache/brooklyn/api/objs/Configurable.java
```diff
@@ -97,5 +97,7 @@
          * @see {@link #setConfig(ConfigKey, Task)}
          */
         <T> T set(HasConfigKey<T> key, Task<T> val);
+        
+        Set<ConfigKey<?>> findKeys(Predicate<ConfigKey<?>> predicate);
```
aledsage: Should use `findKeys(Predicate<? super ConfigKey<?>> filter)`.

core/src/main/java/org/apache/brooklyn/core/config/internal/LazyContainerAndKeyValue.java
```diff
@@ -0,0 +1,84 @@
...
+    private final TContainer container;
+    private final ConfigKey<TValue> key;
+    private final Function<TContainer,Maybe<Object>> lookupResolutionFunction;
+    private final Function<Maybe<Object>,Maybe<TValue>> conversionFunction;
+    private Maybe<TValue> resolved;
+    
+    public LazyContainerAndKeyValue(ConfigKey<TValue> key, TContainer container, 
+            Function<TContainer, Maybe<Object>> lookupResolutionFunction,
+            Function<Maybe<Object>, Maybe<TValue>> conversionFunction) {
+        this.key = key;
```
aledsage: We should use `this.key = checkNotNull(key, "key")` (or if it really is allowed to be null then use `@Nullable` on parameter, and on `getKey` method.

Same for the other args.

camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/ConfigInheritanceYamlTest.java
```diff
@@ -72,6 +72,7 @@ public void setUp() throws Exception {
         
         emptyFile = Files.createTempFile("ConfigInheritanceYamlTest", ".txt");
         emptyFile2 = Files.createTempFile("ConfigInheritanceYamlTest2", ".txt");
+        emptyFile3 = Files.createTempFile("ConfigInheritanceYamlTest3", ".txt");
```
aledsage: Need to add deletion of this file to the `tearDown` method.

camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/ConfigInheritanceYamlTest.java
```diff
@@ -326,53 +320,72 @@ public void testOverridesParentConfig() throws Exception {
...
+                "        myOptionsKey3: myOptionsVal", 
                 "  brooklyn.children:",
                 "  - type: org.apache.brooklyn.entity.software.base.EmptySoftwareProcess",
                 "    brooklyn.config:",
                 "      shell.env:",
                 "        ENV2: myEnv2",
+                "        ENV3: myEnv2",
                 "      templates.preinstall:",
                 "        "+emptyFile2.toUri()+": myfile2",
+                "        "+emptyFile3.toUri()+": myfile2",
```
aledsage: Minor...

Why are you reusing the name "myfile2" for the target filename? Is that because you want to check it doesn't complain if two files are being uploaded to the same place? Or just because you don't care about the value?

If you don't care about the value, I'd prefer it to say "myfile3" as that makes the test slightly clearer (and ensures we haven't mixed up the key-values).

Same goes for "myEnv2" etc.

core/src/main/java/org/apache/brooklyn/core/internal/BrooklynProperties.java
```diff
@@ -298,15 +298,20 @@ private static void addPropertiesFromMapSupplier(BrooklynProperties p, Supplier<
...
+     * unless the default value is null in which case it returns the default.
+     */
     public <T> T getConfig(ConfigKey<T> key, T defaultValue);
 
-    @Override
-    public Object getRawConfig(ConfigKey<?> key);
-    
+    public Maybe<Object> getConfigRaw(ConfigKey<?> key);
+
+    /** Inheritance is ignored here. Preferable to use {@link #getConfigRaw(ConfigKey)}. */
```
aledsage: Perhaps we should deprecate this method?

core/src/main/java/org/apache/brooklyn/core/internal/BrooklynPropertiesImpl.java
```diff
@@ -439,24 +442,45 @@ public void putAll(Map vals) {
...
+    
+    @Override
+    public Map<ConfigKey<?>,Object> getAllConfigLocalRaw() {
         Map<ConfigKey<?>, Object> result = new LinkedHashMap<ConfigKey<?>, Object>();
         for (Object entry: entrySet())
             result.put(new BasicConfigKey<Object>(Object.class, ""+((Map.Entry)entry).getKey()), ((Map.Entry)entry).getValue());
         return result;
     }
 
+    @Override @Deprecated
```
aledsage: When deprecating, can we include javadoc to say when and to direct people to the correct method:

```
/**
 * @deprecated since 0.10.0; instead use {@link #getAllConfigLocalRaw()}
 */
```

core/src/main/java/org/apache/brooklyn/core/internal/BrooklynPropertiesImpl.java
```diff
@@ -439,24 +442,45 @@ public void putAll(Map vals) {
...
+    public Map<ConfigKey<?>, Object> getAllConfig() {
+        return getAllConfigLocalRaw();
+    }
+
+    @Override
+    public Set<ConfigKey<?>> findKeys(Predicate<ConfigKey<?>> filter) {
+        Set<ConfigKey<?>> result = new LinkedHashSet<ConfigKey<?>>();
+        for (Object entry: entrySet()) {
```
aledsage: Yuck that we have to use generics like this! Would we nice to change `BrooklynProperties` interface to not just extend `Map`, but definitely not in this huge PR!

core/src/main/java/org/apache/brooklyn/core/config/BasicConfigInheritance.java
```diff
@@ -0,0 +1,166 @@
...
+    public static BasicConfigInheritance NEVER_INHERITED = new BasicConfigInheritance(false,CONFLICT_RESOLUTION_STRATEGY_OVERWRITE,true);
+    /** Indicates that if a key has a value at both an ancestor and a descendant, the descendant and his descendants
+     * will prefer the value at the descendant. */
+    public static BasicConfigInheritance OVERWRITE = new BasicConfigInheritance(true,CONFLICT_RESOLUTION_STRATEGY_OVERWRITE,false);
+    /** Indicates that if a key has a value at both an ancestor and a descendant, the descendant and his descendants
+     * should attempt to merge the values. If the values are not mergable behaviour is undefined
+     * (and often the descendant's value will simply overwrite). */
+    public static BasicConfigInheritance DEEP_MERGE = new BasicConfigInheritance(true,CONFLICT_RESOLUTION_STRATEGY_DEEP_MERGE,false);
+    
+    // reinheritable? true/false; if false, children/descendants/inheritors will never see it; default true
```
aledsage: Use javadoc (given that it's protected, rather than private).

core/src/main/java/org/apache/brooklyn/core/config/BasicConfigInheritance.java
```diff
@@ -0,0 +1,166 @@
...
+    protected final boolean isReinherited;
+    // conflict-resolution-strategy? overwrite or deep_merge or null; default null meaning caller supplies
+    // (overriding resolveConflict)
+    protected final String conflictResolutionStrategy;
+    // use-local-default-value? true/false; if true, overwrite above means "always ignore (even if null)"; default false
+    // whereas merge means "use local default (if non-null)"
+    protected final boolean useLocalDefaultValue;
+
+    protected BasicConfigInheritance(boolean isReinherited, String conflictResolutionStrategy, boolean useLocalDefaultValue) {
```
aledsage: I'd include `@Nullable` for the parameter `conflictResolutionStrategy`.

utils/common/src/main/java/org/apache/brooklyn/config/ConfigInheritance.java
```diff
@@ -19,63 +19,198 @@
...
-    }
+    /** @deprecated since 0.10.0 see implementations of this interface */ @Deprecated
+    public static final ConfigInheritance NONE = new Legacy.None();
+    /** @deprecated since 0.10.0 see implementations of this interface */ @Deprecated
+    public static final ConfigInheritance ALWAYS = new Legacy.Always();
+    /** @deprecated since 0.10.0 see implementations of this interface */ @Deprecated
+    public static final ConfigInheritance DEEP_MERGE = new Legacy.Merged();
     
-    private ConfigInheritance() {}
+    @Deprecated
```
aledsage: Include javadoc:

```
/**
 * @deprecated since 0.10.0; use the other more specific methods, or consider using {@link ConfigInheritances}
 */
```

core/src/main/java/org/apache/brooklyn/core/config/BasicConfigInheritance.java
```diff
@@ -0,0 +1,166 @@
...
+    @Override
+    public InheritanceMode isInherited(ConfigKey<?> key, Object from, Object to) {
+        return null;
+    }
+    
+
+    @Override
+    public <TContainer, TValue> boolean isReinheritable(ConfigValueAtContainer<TContainer, TValue> parent, ConfigInheritanceContext context) {
+        if (!equals(ConfigInheritances.findInheritance(parent, context, this))) 
+            throw new IllegalStateException("Method can only be invoked on inheritance at "+parent);
```
aledsage: Can we give more info in the message about the entity/context that it was called on, versus what was expected?

core/src/main/java/org/apache/brooklyn/core/config/BasicConfigInheritance.java
```diff
@@ -0,0 +1,166 @@
...
+        return isReinherited();
+    }
+    
+    @Override
+    public <TContainer,TValue> boolean considerParent(
+            ConfigValueAtContainer<TContainer,TValue> local,
+            ConfigValueAtContainer<TContainer,TValue> parent,
+            ConfigInheritanceContext context) {
+        if (!equals(ConfigInheritances.findInheritance(local, context, this))) 
+            throw new IllegalStateException("Method can only be invoked on inheritance at "+local);
```
aledsage: As above: can we give more context in the exception message?

core/src/main/java/org/apache/brooklyn/core/config/BasicConfigInheritance.java
```diff
@@ -0,0 +1,166 @@
...
+        if (parent==null) return false;
+        if (CONFLICT_RESOLUTION_STRATEGY_OVERWRITE.equals(conflictResolutionStrategy)) {
+            // overwrite means ignore if there's an explicit value, or we're using the local default
+            return !local.isValueExplicitlySet() && !getUseLocalDefaultValue();
+        }
+        return true;
+    }
+
+    @Override
+    public <TContainer,TValue> ReferenceWithError<ConfigValueAtContainer<TContainer,TValue>> resolveWithParent(
```
aledsage: Do we need some (trace) logging in here to say why it returned what it did?

core/src/main/java/org/apache/brooklyn/core/config/internal/AbstractConfigMapImpl.java
```diff
@@ -107,4 +252,227 @@ public int size() {
...
+
+        for (ConfigKey<?> k: local.keySet()) {
+            Object v = local.get(k);
+            ReferenceWithError<ConfigValueAtContainer<TContainer, ?>> vpr = parents.remove(k);
+            @SuppressWarnings("unchecked")
+            ConfigValueAtContainer<TContainer, Object> vp = vpr==null ? null : (ConfigValueAtContainer<TContainer,Object>) vpr.getWithoutError();
+            ConfigInheritance inh = ConfigInheritances.findInheritance(k, InheritanceContext.RUNTIME_MANAGEMENT, getDefaultRuntimeInheritance());
+            ConfigValueAtContainer<TContainer,Object> vl = new BasicConfigValueAtContainer<TContainer,Object>(getContainer(), k, Maybe.ofAllowingNull(v));
+            ReferenceWithError<ConfigValueAtContainer<TContainer, Object>> vlr = null;
+            if (inh.considerParent(vl, vp, InheritanceContext.RUNTIME_MANAGEMENT)) {
```
aledsage: It feels strange to insist that the caller always calls `considerParent` before calling `resolveWithParent` (and to likely have an incorrect result if `resolveWithParent` is called without first checking `considerParent`).

Would it make sense to have a single method call that does this: a resolve method which would do the internal considerParent check (i.e. the logic here, but inside the `BasicConfigInheritance`)?

core/src/main/java/org/apache/brooklyn/core/config/internal/AbstractConfigMapImpl.java
```diff
@@ -107,4 +252,227 @@ public int size() {
...
+        } else {
+            String message = "Config key "+ownKey+" of "+getBrooklynObject()+" is not a ConfigKeySelfExtracting; cannot retrieve value; returning default";
+            LOG.warn(message);
+            return ReferenceWithError.newInstanceThrowingError(new BasicConfigValueAtContainer<TContainer,T>(getContainer(), ownKey, null, false,
+                    defaultValue),
+                new IllegalStateException(message));
+        }
+    }
+    
+    public List<ConfigValueAtContainer<TContainer,?>> getConfigAllInheritedRaw(ConfigKey<?> queryKey) {
```
aledsage: What is this for? It looks like it is never used (it is just mentioned in some javadoc) and is never tested.

core/src/main/java/org/apache/brooklyn/core/config/ConfigKeys.java
```diff
@@ -46,6 +47,9 @@
 
     private static final Logger log = LoggerFactory.getLogger(ConfigKeys.class);
     
+    public enum InheritanceContext implements ConfigInheritanceContext
+    { TYPE_DEFINITION, RUNTIME_MANAGEMENT }
```
aledsage: Personal preference: strange code layout! Can we follow a common coding convention (like that of Oracle, with our agreed variations like line length) when there's not a really good reason to deviate?

core/src/main/java/org/apache/brooklyn/core/entity/AbstractEntity.java
```diff
@@ -1209,59 +1208,18 @@ public void remove(AttributeSensor<?> attribute) {
...
-        @Override
-        public void addToLocalBag(Map<String, ?> vals) {
-            configsInternal.addToLocalBag(vals);
-        }
-
-        @Override
-        public void removeFromLocalBag(String key) {
-            configsInternal.removeFromLocalBag(key);
+        
+        protected AbstractConfigMapImpl<?> getConfigsInternal() {
```
aledsage: I don't like us exposing implementation classes. Is this really needed?!

core/src/main/java/org/apache/brooklyn/core/config/internal/AbstractConfigMapImpl.java
```diff
@@ -20,65 +20,212 @@
...
+    /*
+     * Changed Sept 2016 so that keys can determine inheritance strategy at every level in the hierarchy,
+     * and signifcantly refactored to share code among subclasses, adding Location as a subclass.
+     */
 
     private static final Logger LOG = LoggerFactory.getLogger(AbstractConfigMapImpl.class);
     
-    protected final ConfigMapViewWithStringKeys mapViewWithStringKeys = new ConfigMapViewWithStringKeys(this);
+    @SuppressWarnings("deprecation")
+    protected final transient org.apache.brooklyn.core.entity.internal.ConfigMapViewWithStringKeys mapViewWithStringKeys = new org.apache.brooklyn.core.entity.internal.ConfigMapViewWithStringKeys(this);
```
aledsage: Why `transient`? This class should never be serialised - if it is, then it's an indication that something has gone very wrong. If it ever makes it into persisted state, then it means our persisted state has become even more brittle.

`ConfigMapWithInheritance` is not `serializable`.

core/src/main/java/org/apache/brooklyn/core/config/internal/AbstractConfigMapImpl.java
```diff
@@ -20,65 +20,212 @@
...
-    protected Map<ConfigKey<?>,Object> ownConfig = Collections.synchronizedMap(new LinkedHashMap<ConfigKey<?>, Object>());
+    protected final Map<ConfigKey<?>,Object> ownConfig;
+    
+    protected AbstractConfigMapImpl(TContainer bo) {
+        // Not using ConcurrentMap, because want to (continue to) allow null values.
+        // Could use ConcurrentMapAcceptingNullVals (with the associated performance hit on entrySet() etc).
+        this(bo, Collections.synchronizedMap(new LinkedHashMap<ConfigKey<?>, Object>()));
+    }
+    protected AbstractConfigMapImpl(TContainer bo, Map<ConfigKey<?>, Object> storage) {
+        this.bo = bo;
```
aledsage: Can we make the `bo` field final, and check here that it is not null (and that storage is not null).

core/src/main/java/org/apache/brooklyn/core/config/internal/AbstractConfigMapImpl.java
```diff
@@ -20,65 +20,212 @@
...
+        return getConfigRaw(key, false);
+    }
+
+    protected abstract ExecutionContext getExecutionContext(BrooklynObject bo);
+    protected abstract void postLocalEvaluate(ConfigKey<?> key, BrooklynObject bo, Maybe<?> rawValue, Maybe<?> resolvedValue);
+    
+    @Override
+    public Map<ConfigKey<?>,Object> getAllConfigLocalRaw() {
+        Map<ConfigKey<?>,Object> result = new LinkedHashMap<ConfigKey<?>,Object>();
+        result.putAll(ownConfig);
```
aledsage: If `ownConfig` is a synchronized map, then need to synchronize on it when copying it (as that will iterate). Same for all the other places we do `putAll(ownConfig)`.

core/src/main/java/org/apache/brooklyn/core/config/internal/AbstractConfigMapImpl.java
```diff
@@ -20,65 +20,212 @@
...
+        return Collections.unmodifiableMap(result);
     }
     
+    /** an immutable copy of the config visible at this entity, local and inherited (preferring local) */
```
aledsage: Given this doesn't respect reinheritable etc, worth saying that in the javadoc (given you're not just relying on the interface's javadoc).

core/src/main/java/org/apache/brooklyn/core/entity/AbstractEntity.java
aledsage: For example, the pattern we've used elsewhere is to have an interface with suffix "Internal" for the internal methods, so the implementation class is hidden away as much as possible. We could add a `ConfigMapInternal` for the methods that really have to be visible to the rest of the Brooklyn framework.

core/src/main/java/org/apache/brooklyn/core/config/internal/AbstractConfigMapImpl.java
```diff
@@ -107,4 +252,227 @@ public int size() {
...
+        ConfigKey<T> candidate = (ConfigKey<T>) getKeyAtContainerImpl(container, queryKey);
+        return candidate;
+    }
+    
+    @Nullable protected abstract <T> ConfigKey<?> getKeyAtContainerImpl(@Nonnull TContainer container, ConfigKey<T> queryKey);
+
+    protected Maybe<Object> getRawValueAtContainer(TContainer container, ConfigKey<? extends Object> configKey) {
+        return ((BrooklynObjectInternal)container).config().getInternalConfigMap().getConfigLocalRaw(configKey);
+    }
+    protected Maybe<Object> resolveRawValueFromContainer(TContainer container, ConfigKey<?> key, Object value) {
```
aledsage: Worth some javadoc to say about the relationship between this method/impl and that in `ConfigInheritances` / `BasicConfigInheritance`. Will this respect the "reinheritable" etc, or is that the responsibility of the caller?

api/src/main/java/org/apache/brooklyn/api/objs/Configurable.java
ahgittin: +1

core/src/main/java/org/apache/brooklyn/core/config/internal/LazyContainerAndKeyValue.java
ahgittin: I've added the nullables, including at the interface.

camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/ConfigInheritanceYamlTest.java
ahgittin: now using `Os.newTempFile` instead which does deletion (i've also added the explicit deletion here)

camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/ConfigInheritanceYamlTest.java
ahgittin: fixed

core/src/main/java/org/apache/brooklyn/core/internal/BrooklynProperties.java
ahgittin: agree

core/src/main/java/org/apache/brooklyn/core/internal/BrooklynPropertiesImpl.java
ahgittin: not worth doing in cases like this where it overrides a deprecated method and there is such a comment on the super

core/src/main/java/org/apache/brooklyn/core/internal/BrooklynPropertiesImpl.java
ahgittin: yeah i was surprised there wasn't a cleaner way to say this

core/src/main/java/org/apache/brooklyn/core/config/BasicConfigInheritance.java
ahgittin: made it a `check...` method invoking added `toString()` methods on `ConfigValueAtContainer` and `BasicConfigInheritance`

core/src/main/java/org/apache/brooklyn/core/config/internal/AbstractConfigMapImpl.java
ahgittin: as per javadoc methods in `ConfigInheritances` do that; the methods here are to reflect the component logic which needs to be defined for a consistent inheritance strategy

core/src/main/java/org/apache/brooklyn/core/config/internal/AbstractConfigMapImpl.java
ahgittin: it's on the interface, useful for debugging and potentially for a config explorer, to know how a resolved value is derived.  agree we need test, will add.

core/src/main/java/org/apache/brooklyn/core/entity/AbstractEntity.java
ahgittin: I like that pattern most places.  However if it is only protected usage and the return type is in an internal package I don't think there is any serious risk of confusion and it is not worth the burden of internal interfaces.  (While not too hard to write it does make navigating the code much harder.)

core/src/main/java/org/apache/brooklyn/core/config/internal/AbstractConfigMapImpl.java
ahgittin: it's deprecated (i've added an annotation for this) ... just in case the MapImpl was written anywhere i think marking transient for one release is a good practice

core/src/main/java/org/apache/brooklyn/core/config/internal/AbstractConfigMapImpl.java
ahgittin: agree, have added note to that effect, once we've removed deprecated fields eg `entity` which may be needed to populate `bo`

core/src/main/java/org/apache/brooklyn/core/config/internal/AbstractConfigMapImpl.java
```diff
@@ -20,65 +20,212 @@
...
     
-    public <T> T getConfig(HasConfigKey<T> key, T defaultValue) {
-        return getConfig(key.getConfigKey(), defaultValue);
+    @Override
+    public Maybe<Object> getConfigLocalRaw(ConfigKey<?> key) {
+        return getConfigRaw(key, false);
+    }
+
+    protected abstract ExecutionContext getExecutionContext(BrooklynObject bo);
+    protected abstract void postLocalEvaluate(ConfigKey<?> key, BrooklynObject bo, Maybe<?> rawValue, Maybe<?> resolvedValue);
```
ahgittin: pretty obvious from usages -- remember this is in an internal package

core/src/main/java/org/apache/brooklyn/core/config/internal/AbstractConfigMapImpl.java
ahgittin: good spot

core/src/main/java/org/apache/brooklyn/core/config/internal/AbstractConfigMapImpl.java
ahgittin: deprecated and never used so not worth perfect javadoc; the interface javadoc makes clear the issues

core/src/main/java/org/apache/brooklyn/core/config/internal/AbstractConfigMapImpl.java
ahgittin: yes, javadoc quite useful to explain why this is here, it's messy but it has always been so, at least now it's put into one place instead of multiple, and now explained.

ahgittin:
@aledsage thanks for detailed review so far; all comments addressed.  i know it's a big PR but all the changes are intimately related -- sorry for the pain.  hopefully you like what it has done.

i was hoping that the removed lines would exceed the added lines but i failed on that.  maybe if we don't count tests and comments it would show up as a net decrease in LOC.  :)  but the fact several previously-failing tests were fixed as a side effect of these changes and we've cut down TODO comments hopefully indicates this is a big step in the right direction.

core/src/main/java/org/apache/brooklyn/core/config/internal/AbstractConfigMapImpl.java
```diff
@@ -107,4 +264,264 @@ public int size() {
...
+            
+            if (next.isValueExplicitlySet()) result.add(0, next);
+
+            last = next;
+            c = getParentOfContainer(c);
+            count++;
+        }
+        
+        return result;
+    }
```
aledsage: There feels like a scary looking number of methods on here now. I'm not suggesting any specific method(s) that should be removed, but it's something that we should be aware of - gut feel is that we have too many methods (even ignoring the deprecated ones).

If we really need all these different ways of access it, then another would would be to present this as "views" - e.g. a raw view, a resolved view, a local-only view, etc.

But any such change are much longer term, rather than part of this PR!

core/src/main/java/org/apache/brooklyn/core/config/internal/LazyContainerAndKeyValue.java
```diff
@@ -0,0 +1,93 @@
...
+        this.container = container;
+        this.lookupResolutionFunction = Preconditions.checkNotNull(lookupResolutionFunction);
+        this.conversionFunction = Preconditions.checkNotNull(conversionFunction);
+    }
+
+    protected synchronized Maybe<TValue> resolve() {
+        if (resolved==null) { 
+            resolved = conversionFunction.apply(
```
aledsage: Slightly uncomfortable about calling out to alien code while holding the `synchronized` lock on this; but not sure off-hand what a better pattern would be.

utils/common/src/main/java/org/apache/brooklyn/config/StringConfigMap.java
```diff
@@ -32,4 +34,12 @@
      * and 'defaultIfNone' (a default value to return if there is no such property);
      * defaults to no warning and null default value */
     public String getFirst(@SuppressWarnings("rawtypes") Map flags, String... keys);
+    
+    /** returns submap matching the given filter predicate; see ConfigPredicates for common predicates */
+    public StringConfigMap submap(Predicate<ConfigKey<?>> filter);
```
aledsage: Should be `submap(Predicate<? super ConfigKey<?>> filter)`

ahgittin: not permitted due to constraint on the (deprecated) parent interface

ahgittin:
(failure is the damn `org.apache.brooklyn.feed.jmx.RebindJmxFeedTest.testJmxFeedIsPersisted` issue, not related)

aledsage:
Is there a test of `NOT_REINHERITABLE` for runtime inheritance with app->child->grandchild, where the value is set on `app`, the `child` consumes it, and the `grandchild` does not see it?

I couldn't see that test, but then it's hard to understand/find everything that the tests are testing.

camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/ConfigInheritanceYamlTest.java
aledsage: I'd still prefer deleting immediately in tearDown, as `Os.newTempFile` only sets `deleteOnExit()` (so will not delete it if the process is killed).

core/src/main/java/org/apache/brooklyn/core/config/internal/AbstractConfigMapImpl.java
aledsage: Yikes, `EntityConfigMap.getContainer()` is scary. Its `synchronized (this)` is wrong (nothing else synchronizes on `this`). And nothing sets `EntityConfigMap.entity`, so unless there is crazy groovy or crazy reflection, then that method hopefully does nothing!

I'm fine with ignoring it for this PR, which we certainly don't want to get any bigger!

core/src/main/java/org/apache/brooklyn/core/config/internal/AbstractConfigMapImpl.java
aledsage: @ahgittin It wasn't obvious to me because nothing calls this method. I see from `EntityConfigMap.postLocalEvaluate` what the purpose is.

Is there a missing call to this method?

aledsage:
Thanks @ahgittin - finished reviewing.

Overall I'm happy with this. I agree that it feels very complicated (and some of the code is quite subtle, so could do with a more thorough review!).

Are there situations (such as the `NOT_REINHERITABLE` app->child->grandchild) that are not tested?

I'm fine with this being merged when you're ready.

core/src/main/java/org/apache/brooklyn/core/config/internal/AbstractConfigMapImpl.java
ahgittin: oops that is a bug, must have deleted the call!  great spot

ahgittin:
Big thank you @aledsage !  I'll fix that last thing you spotted and merged.

I expect there are untested corner cases -- there are so many -- but things are far better tested and implemented than before, at least!

PR #341, opened 2016-09-21
BROOKLYN-349: fix DSL resolution in location

aledsage:
There are two parts to fixing this:

1. Fix for NPE in `JcloudsLocation.getComputeService`: we are more careful when copying the `ResolvingConfigBag`, so we don't resolve all the values too early when copying it from one `ResolvingConfigBag` to another.

2. Fix for creation of DSL tasks (e.g. in `DslComponent.newTask()`) so that we pass through the current task's context entity when we create the DSL task.

For (2), it's not clear to me if that is the right fix. The alternative would be to ensure that it is the entity's `ExecutionContext` that is used to execute the created task (which would thus automatically set the entity-context as a tag on the task). Presumably we are currently sometimes using the vanilla management context's `ExecutionContext` instead.

### DIFF
--- a/camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/methods/DslComponent.java
+++ b/camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/methods/DslComponent.java
@@ -89,8 +89,12 @@ public class DslComponent extends BrooklynDslDeferredSupplier<Entity> {
     
     @Override
     public Task<Entity> newTask() {
-        return TaskBuilder.<Entity>builder().displayName(toString()).tag(BrooklynTaskTags.TRANSIENT_TASK_TAG)
-            .body(new EntityInScopeFinder(scopeComponent, scope, componentId)).build();
+        return TaskBuilder.<Entity>builder()
+                .displayName(toString())
+                .tag(BrooklynTaskTags.TRANSIENT_TASK_TAG)
+                .tagIfNotNull(BrooklynTaskTags.getTargetOrContextEntityTag(Tasks.current()))
+                .body(new EntityInScopeFinder(scopeComponent, scope, componentId))
+                .build();
     }
     
     protected static class EntityInScopeFinder implements Callable<Entity> {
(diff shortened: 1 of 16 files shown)


aledsage:
See #279 for an alternative fix (though no test - we could copy the test from here into that PR).

We'll only want one of this PR or #279 to be merged, I think. @grkvlt @ahgittin @neykov I'm keen to hear your thoughts on it.

ahgittin:
Can we merge #340 and then look at this?  It gives clearer methods for dealing with semantics of copying resolved / unresolved semantics, and it moves `Location`'s config to be treated like others.  Many of your changes here can still work but some may need changing to fit with how it works.

In particular note that with #340 the `ConfigBag` exposed by `AbstractLocation` is an RO view.  (Not sure if that matters.)

aledsage:
@ahgittin I've rebased this against master (now that #340 is merged), and confirmed `JcloudsRebindStubTest` still passes.

Can you review please?

neykov:
The changes make sense. I think they will work even if the DSL is set on the location object, nothing specific to `provisioning.properties` here. Worth a test case to confirm? 
The result still depends on the context where evaluation happens, not where it's set. Or may be not, not sure how #340 interacts with this and locations specifically. 
Good to merge.

aledsage:
@neykov can you review these additional commits please? Do you agree that we should merge this, and not @grkvlt 's #279?

When refactoring the tests to remove duplication, I realised that if one uses:

```
provisioning.properties:
  password: $brooklyn:config("password")
```
And then subsequently calls `jcloudsSshMachineLocation.getHostname()` outside of the context of an entity, then it fails to resolve (because it doesn't know which entity to look at the config on).

I discussed the short term fix with @neykov, and applied it here: it avoids propagating the exception if the credentials can't be obtained, and it only look up the credentials if we really need them (e.g. for hostname on AWS, where we ask the VM for it).

---
I also discussed the longer term solution with @neykov. Conclusion is that it's hard. And that we should resolve the DSL config's context at the time the entity is being instantiated (e.g. storing it as a field on the `DslConfigSupplier` or whatever). This would mean we'd always use the entity on which the config was declared, rather than relying on who is looking it up. However, there is (probably?!) an exception to that rule: if you declare config with an `@brooklyn:entitySpec`, such as for a `memberSpec`, then you probably want it to refer to the context of the entity to be created rather than the entity where you are declaring the entitySpec (and therefore if setting the entity context on the DSL object, you'd need to ensure it is a different DSL object instance for each member instantiated using the entitySpec!).

That deserves its own thread on dev@brooklyn.

aledsage:
Tested by @duncangrant (for his use-case with kubernetes in clocker, which also motivated his https://github.com/apache/brooklyn-server/pull/361). Merging now.

neykov:
+1 for  merging.

PR #342, opened 2016-09-22
Load usage listeners from bundles

sjcorbett:
And JcloudsLocation's image choosers.


core/src/test/java/org/apache/brooklyn/core/mgmt/internal/LocalUsageManagerTest.java
```diff
@@ -16,67 +16,77 @@
...
     @Test
     public void testAddUsageListenerViaProperties() throws Exception {
         BrooklynProperties brooklynProperties = BrooklynProperties.Factory.newEmpty();
         brooklynProperties.put(UsageManager.USAGE_LISTENERS, RecordingStaticUsageListener.class.getName());
         mgmt = LocalManagementContextForTests.newInstance(brooklynProperties);
-        
+        assertUsageListenerCalledWhenApplicationStarted();
+    }
+
+    @Test(expectedExceptions = ClassCastException   .class)
```
Graeme-Miller: what's with the space between ClassCastException and .class

core/src/main/java/org/apache/brooklyn/core/mgmt/internal/LocalUsageManager.java
```diff
@@ -173,20 +180,23 @@
...
-                if (listener instanceof UsageListener) {
-                    addUsageListener((UsageListener)listener);
-                } else if (listener == null) {
-                    throw new NullPointerException("null listener in config "+UsageManager.USAGE_LISTENERS);
+            for (Object obj : listeners) {
+                if (obj == null) {
+                    throw new NullPointerException("null listener in config " + UsageManager.USAGE_LISTENERS);
+                } else if (!(obj instanceof UsageListener)) {
+                    throw new ClassCastException("Configured object is not a UsageListener. This probably means coercion failed: " + obj);
                 } else {
```
Graeme-Miller: doesn't really need to be an else anymore

core/src/test/java/org/apache/brooklyn/core/mgmt/internal/LocalUsageManagerTest.java
sjcorbett: Accidental tab. Why doesn't GitHub show I've fixed this any more?

neykov:
LGTM, thanks for fixing `JCloudsLocation` as well. Merging.

PR #343, opened 2016-09-22
Add TestWinrmCommand for test-framework

aledsage:


neykov:
LGTM, merging.

We continue expanding the ssh<->winrm duality even further. They don't look so different - would be nice to fold them into one case.

PR #329, opened 2016-09-13
fix long-running unit test

ahgittin:
/cc @aledsage is this the best fix?  (it stalls for 2m most the time when doing `mvn clean install` without this.  haven't thought too much about it though!)

PR #344, opened 2016-09-22
JmxService tests marked as integration

aledsage:
These tests expect a specific port to be available, which may not be true
in a shared jenkins environment. Therefore marking them as "Integration".

Also moves newJmxServiceRetrying into JmxService.

---
Longer term, we could rewrite the JmxService tests in one of two ways (but I suggest we don't tackle those right now):

1. First (and simplest), we could use the `newJmxServiceRetrying` so that it tries different ports until one starts. We'd need to rewrite the test so that the entity didn't hard-code an expected port. We'd first start the `JmxService`, and then set the config/attribute on the entity.

2. We could try to pass in port `0` to the `JmxService` so that it allocates any available port. That looks a bit harder (the port is passed inside the `JMXServiceURL`. A very quick try at missing the port out from that led to other issues with `connectorServer.start()` (and with needing to get the real address that includes the port only after having started the connectorServer).

### DIFF
--- a/software/base/src/test/java/org/apache/brooklyn/entity/software/base/test/jmx/JmxService.java
+++ b/software/base/src/test/java/org/apache/brooklyn/entity/software/base/test/jmx/JmxService.java
@@ -21,6 +21,7 @@ package org.apache.brooklyn.entity.software.base.test.jmx;
 import java.io.IOException;
 import java.util.List;
 import java.util.Map;
+import java.util.NoSuchElementException;
 
 import javax.management.InstanceAlreadyExistsException;
 import javax.management.MBeanNotificationInfo;
@@ -43,7 +44,9 @@ import mx4j.tools.naming.NamingService;
 import mx4j.tools.naming.NamingServiceMBean;
 
 import org.apache.brooklyn.api.entity.Entity;
+import org.apache.brooklyn.api.location.PortRange;
 import org.apache.brooklyn.core.entity.Attributes;
+import org.apache.brooklyn.core.location.PortRanges;
 import org.apache.brooklyn.entity.java.UsesJmx;
 import org.apache.brooklyn.feed.jmx.JmxHelper;
 import org.apache.brooklyn.test.NetworkingTestUtils;
@@ -52,6 +55,8 @@ import org.slf4j.LoggerFactory;
 
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Iterables;
+import com.google.common.collect.Lists;
 
 /**
  * Set up a JMX service ready for clients to connect. This consists of an MBean server, a connector server and a naming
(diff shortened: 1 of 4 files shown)


aledsage:
@ahgittin can you take a look at this please?

ahgittin:
This is a bad cop out.  The tests aren't relying on a specific port.  I think the problem is that the `Networking.isPortAvailable()` check isn't releasing the port fast enough and probably the subsequent bind (deep in mbean code) isn't doing `setReuseAddr`.

i've snuck a nasty hack in #340 at e3dd40957867d8083e3e1b2d6d278125009048f6 .  if that works i'd prefer that to this.  if not we should take a quick look (stack trace in commit above) to see if we can turn on reuse_addr for mbean.  if that fails then we could do this until we support your long-term 2 which is the right way, let jmx service pick the port and then we plug that sensor in everywhere else it's needed (but as you say quite a lot of work and not worth it as this code is pretty solid and unchanging)

aledsage:
Re sneaking the change into #340 - next time can you do it in a separate PR, even if you merge that PR yourself immediately!

---
Maybe it is just the problem of socket `time_wait` status. But I'd have thought it would be more deterministic, and would fail on other environments (rather than just the shared jenkins) if that is the case.

I guess the alternative (that the port is snapped up by another process in the few millisecond window) is pretty unlikely, particularly with unusual port numbers like this!

I wonder if it could be more subtle: the socket might already be in a time_wait state; we call `Networking.isPortAvailable` using `setReuseAddress` so it works; we then call the `JmxService` without `setReuseAddress` and it fails because the original socket is still there in `time_wait` state. Or is that nonsense: does opening a socket with `setReuseAddress` cause the previous time_wait socket to close immediately?

---
I don't particularly like us waiting up to 3 minutes in our test (as per your observation in #329 that we want tests to be fast). I'd prefer (medium-term) if we could retry with different port numbers instead.

But happy to see if your fix works!

neykov:
Marking tests integration is same as deleting them right now so I'd be wary of such changes (integration tests don't get run, let alone on regular basis).
I was able to reproduce this by running two tests in parallel and slowing them down so that while one of them has the JMX port already open the other goes through the port selection process.
I think it's related to this comment - https://github.com/apache/brooklyn-server/blob/master/utils/common/src/main/java/org/apache/brooklyn/util/net/Networking.java#L96-L98. The JMX service binds to `0.0.0.0:40123` while `LocalhostMachineProvisioningLocation.obtainPort` checks `127.0.0.1` for available ports.

ahgittin:
@neykov running two tests in parallel would understandably generate this error.  but our test machines don't do that do they ... actually if the boxes do that could explain the problem and a better fix would be to tell it to use port `40000+Math.random(10000)+"+"` (still a 0.01% chance of collision but that's a big improvement!)

neykov:
They do actually - there are multiple jobs running the tests (server-master, build-master, server-master-pull-requests).
Still I think fixing the free port checking logic is better than picking a random port. There's nothing stopping another project's builds to allocate a block of ports for it's own use and fail our tests.

aledsage:
Closing this. We'll see if the

aledsage:
Agreed that marking tests as integration is not a good idea right now; hence closing this. If it bites us a lot before we find the time to change it to either use port `0` (so it picks a free port), or to wrap in a retry, or to inject a random numbered port, then we can mark them as `Broken` so that builds don't give us false-negatives.

PR #345, opened 2016-09-23
Fixed compilation issue in LocalUsageManagerTest

Graeme-Miller:


neykov:
Two PRs collided leading to a compile error. Thanks for the fix @Graeme-Miller. Merging.

PR #346, opened 2016-09-23
Fix LocalUsageManagerTest

aledsage:
`shouldSkipOnBoxBaseDirResolution` return type was changed from boolean to
Boolean in super-class.


aledsage:
Merging now as it's causing a compilation error for me.

neykov:
Already fixed in https://github.com/apache/brooklyn-server/pull/345 @aledsage

aledsage:
@neykov ah, thanks! Unfortunately I only spotted your comment after merging this PR. Bizarrely, git has allowed both PRs to be merged without complaining about a conflict, and without marking the second commit as being a no-op. Weird, but has hopefully done no harm!

PR #348, opened 2016-09-23
Fix SharedLocationSecurityGroupCustomizer.udpPortRanges

aledsage:
Previously it was ImmutableList so could not be used from YAML; changed
to a normal List.


aledsage:
@duncangrant or @bostko - can you review this trivial one-line fix please.

bostko:
Looks good.
Thanks!

PR #353, opened 2016-09-27
Fix security group customizer for azure strike2

aledsage:


aledsage:

PR #354, opened 2016-09-27
Config keys and parameters can now be marked as "pinned"

tbouron:
The new behaviour is as follow:
- By default, config keys and parameters are `pinned = false`
- A blueprint can specify a config key or parameter as pinned by using `@CatalogConfig(pinned = true)` or in yaml `pinned: true`
- In case of config keys or parameters pinned in YAML, the priority is auto incremented to keep the order (code lost in [this commit](https://github.com/apache/brooklyn-server/commit/6f624c78b1e7fe72c6df1ecd297b922721b2c023#diff-5310fbeb1f42a3a4cf0508ea441440baL82))

Tests have been updated. Changes in docs and JSGUI coming soon.


tbouron:
Close this in favor of #358

PR #355, opened 2016-09-27
Reduce decisions for HostAndPort which will be used in JcloudsLocation

bostko:
Use managementHostAndPort for all customizations made in
- obtainOnce(ConfigBag)
- registerMachineLocation(ConfigBag, NodeMetadata)

### DIFF
--- a/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
+++ b/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
@@ -133,6 +133,7 @@ import org.apache.brooklyn.util.text.Strings;
 import org.apache.brooklyn.util.time.Duration;
 import org.apache.brooklyn.util.time.Time;
 import org.apache.commons.lang3.ArrayUtils;
+import org.apache.commons.lang3.tuple.Pair;
 import org.jclouds.aws.ec2.compute.AWSEC2TemplateOptions;
 import org.jclouds.cloudstack.compute.options.CloudStackTemplateOptions;
 import org.jclouds.compute.ComputeService;
@@ -771,17 +772,19 @@ public class JcloudsLocation extends AbstractCloudMachineProvisioningLocation im
                 sshHostAndPortOverride = Optional.absent();
             }
 
+            String vmHostname = getPublicHostname(node, sshHostAndPortOverride, userCredentials, setup);
             LoginCredentials initialCredentials = node.getCredentials();
+            final HostAndPort managementHostAndPort = resolveManagementHostAndPort(computeService, node, vmHostname, sshHostAndPortOverride, setup);
             if (skipJcloudsSshing) {
                 boolean waitForConnectable = (windows) ? waitForWinRmable : waitForSshable;
                 if (waitForConnectable) {
                     if (windows) {
                         // TODO Does jclouds support any windows user setup?
-                        initialCredentials = waitForWinRmAvailable(computeService, node, sshHostAndPortOverride, setup);
+                        initialCredentials = waitForWinRmAvailable(initialCredentials, managementHostAndPort, setup);
                     } else {
-                        initialCredentials = waitForSshable(computeService, node, sshHostAndPortOverride, setup);
+                        initialCredentials = waitForSshable(computeService, node, managementHostAndPort, setup);
                     }
-                    userCredentials = createUser(computeService, node, sshHostAndPortOverride, initialCredentials, setup);
+                    userCredentials = createUser(computeService, node, managementHostAndPort, initialCredentials, setup);
                 }
             }
 
@@ -816,7 +819,7 @@ public class JcloudsLocation extends AbstractCloudMachineProvisioningLocation im
 
             // Wait for the VM to be reachable over SSH
             if (waitForSshable && !windows) {
-                waitForSshable(computeService, node, sshHostAndPortOverride, ImmutableList.of(userCredentials), setup);
+                waitForSshable(computeService, node, managementHostAndPort, ImmutableList.of(userCredentials), setup);
             } else {
                 LOG.debug("Skipping ssh check for {} ({}) due to config waitForSshable=false", node, setup.getDescription());
             }
@@ -829,9 +832,9 @@ public class JcloudsLocation extends AbstractCloudMachineProvisioningLocation im
 
             // Create a JcloudsSshMachineLocation, and register it
             if (windows) {
-                machineLocation = registerWinRmMachineLocation(computeService, node, userCredentials, sshHostAndPortOverride, setup);
+                machineLocation = registerWinRmMachineLocation(computeService, node, managementHostAndPort, sshHostAndPortOverride, setup);
             } else {
-                machineLocation = registerJcloudsSshMachineLocation(computeService, node, Optional.fromNullable(template), userCredentials, sshHostAndPortOverride, setup);
+                machineLocation = registerJcloudsSshMachineLocation(computeService, node, Optional.fromNullable(template), userCredentials, managementHostAndPort, sshHostAndPortOverride, vmHostname, setup);
             }
 
             PortForwardManager portForwardManager = setup.get(PORT_FORWARDING_MANAGER);
@@ -1893,7 +1896,7 @@ public class JcloudsLocation extends AbstractCloudMachineProvisioningLocation im
     /**
      * Create the user immediately - executing ssh commands as required.
      */
-    protected LoginCredentials createUser(ComputeService computeService, NodeMetadata node, Optional<HostAndPort> hostAndPortOverride, LoginCredentials initialCredentials, ConfigBag config) {
+    protected LoginCredentials createUser(ComputeService computeService, NodeMetadata node, HostAndPort managementHostAndPort, LoginCredentials initialCredentials, ConfigBag config) {
         Image image = (node.getImageId() != null) ? computeService.getImage(node.getImageId()) : null;
         UserCreation userCreation = createUserStatements(image, config);
 
@@ -1910,7 +1913,6 @@ public class JcloudsLocation extends AbstractCloudMachineProvisioningLocation im
 
             if (windows) {
                 LOG.warn("Unable to execute statements on WinRM in JcloudsLocation; skipping for "+node+": "+userCreation.statements);
-                
             } else {
                 List<String> commands = Lists.newArrayList();
                 for (Statement statement : userCreation.statements) {
@@ -1920,9 +1922,7 @@ public class JcloudsLocation extends AbstractCloudMachineProvisioningLocation im
                 }
 
                 String initialUser = initialCredentials.getUser();
-                String address = hostAndPortOverride.isPresent() ? hostAndPortOverride.get().getHostText() : getFirstReachableAddress(node, config);
-                int port = hostAndPortOverride.isPresent() ? hostAndPortOverride.get().getPort() : node.getLoginPort();
-                
+
                 // TODO Retrying lots of times as workaround for vcloud-director. There the guest customizations
                 // can cause the VM to reboot shortly after it was ssh'able.
                 Map<String,Object> execProps = Maps.newLinkedHashMap();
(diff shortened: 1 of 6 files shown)


locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -3314,15 +3314,12 @@ private String getPublicHostnameGeneric(NodeMetadata node, @Nullable ConfigBag s
...
-        // TODO If POLL_FOR_FIRST_REACHABLE_ADDRESS=false, then won't have checked if any node is reachable.
-        // TODO Some of the private addresses might not be reachable, should check connectivity before
-        // making a choice.
         // TODO Choose an IP once and stick to it - multiple places call JcloudsUtil.getFirstReachableAddress(),
         // could even get different IP on each call.
         if (groovyTruth(node.getPublicAddresses())) {
             return node.getPublicAddresses().iterator().next();
         } else if (groovyTruth(node.getPrivateAddresses())) {
-            return node.getPrivateAddresses().iterator().next();
+            return getFirstReachableAddress(node, setup);
```
aledsage: TL;DR: Tricky. I don't think we should solve the problem like this. I need to think more about this!

The comment above says `JcloudsUtil.getFirstReachableAddress() (probably) already succeeded so at least one of the provided public and private IPs is reachable`.

The "probably" is particularly scary. Also, we may need to worry about this being called some time after provisioning, where the VM might have been terminated behind Brooklyn's back. In that case, none of the addresses would be reachable. The `getFirstReachableAddress` throws an exception (and might take quite some time to do that, depending on the firewalls used within a given cloud). We really don't want `getPublicHostname()` to throw an exception.

Also, the code now looks strange: why is it calling `getFirstReachableAddress` for the private addresses but not the public addresses?

The comment you deleted is also very worrying: if `POLL_FOR_FIRST_REACHABLE_ADDRESS=false` then we shouldn't be calling `getFirstReachableAddress` here. Reasons one might set that include if the VM being provisioned requires special DNAT access (so none of the IP addresses are directly accessible), or if the VM being provisioned is locked down so Brooklyn won't be able to reach it ever.

aledsage:
I presume that the problem you're hitting is that there is a VM being provisioned that gets multiple (private) IPs, but that the corresponding entity advertises its sensor for "host.name" with the wrong IP.

I think we should look more at the call-chain for where we're calling `getPublicHostname()` from, and figure out if there is a better way of doing this. For example, if we're already checked reachability to decide on an IP then we could prefer that (rather than repeating the check).

Also, the semantics of subnet hostname and public hostname are unclear when dealing with private networks: it's unclear whether the IP should be the one reachable from Brooklyn. I guess if we're talking "public" then that's probably as good an indication as any.

aledsage:
The jenkins builds failure is probably unrelated - I've tried to fix it in https://github.com/apache/brooklyn-server/pull/357.

bostko:
Code looks more sensible this way and will be very easy to isolate parts of it related with choosing the IP.
However testing those is best to be done live with the current code.
I think any further refactoring could lead to mistakes.
Thoughts?

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -770,17 +771,20 @@ protected MachineLocation obtainOnce(ConfigBag setup) throws NoMachinesAvailable
                 sshHostAndPortOverride = Optional.absent();
             }
 
-            LoginCredentials initialCredentials = node.getCredentials();
+            String managementAddress = sshHostAndPortOverride.isPresent() ? sshHostAndPortOverride.get().getHostText() : getFirstReachableAddress(node, setup);
```
aledsage: Unfortunately this code is very subtle and delicate. There are so many permutations of different behaviour in different clouds, different OS distros, and different networking configurations.

Therefore be very cautious about changing it! There are a bunch of live tests that are useful (but very slow). But those are not sufficient, as downstream projects like clocker and other advanced-networking test cases (e.g. with DNAT etc) will require specific behaviour.

Long term, I'd love to rewrite the entire class so that it is properly unit-testable. But that is a huge undertaking, so we haven't scheduled the time for that.

For example, you can't call `getFirstReachableAddress()` here. There is a config key (see `waitForWInRmable` and `waitForSsable` below) that we also use to turn off all attempts to reach the VM's IPs. This is used, for example, in clocker when provisioning a container that is not ssh'able and whose IP is entirely private - I believe that no IP reported by docker for that container would be reachable by Brooklyn.

aledsage:
@bostko can you say specifically what you're trying to fix here?

bostko:
I am fixing

```
                machineLocation = registerWinRmMachineLocation(computeService, node, managementHostAndPort, sshHostAndPortOverride, setup);
                machineLocation = registerJcloudsSshMachineLocation(computeService, node, Optional.fromNullable(template), userCredentials, managementHostAndPort, sshHostAndPortOverride, vmHostname, setup);
```
 to use the initial address.

I also trying to make choosing `managementHostAndPort` code simpler and easier to maintain.

@grkvlt @neykov could you also take a look at the docker part related to this PR?

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -1956,6 +1955,33 @@ protected LoginCredentials createUser(ComputeService computeService, NodeMetadat
...
+
+    protected HostAndPort resolveManagementHostAndPort(ComputeService computeService, NodeMetadata node, String vmHostname, Optional<HostAndPort> hostAndPortOverride, ConfigBag config, boolean useOnlyVmHostNameAndHostAndPortOverride) {
+        String pollForFirstReachable = config.get(POLL_FOR_FIRST_REACHABLE_ADDRESS);
+        boolean pollForFirstReachableEnabled = !useOnlyVmHostNameAndHostAndPortOverride && !"false".equalsIgnoreCase(pollForFirstReachable);
+
+        String managementAddress = hostAndPortOverride.isPresent() ? hostAndPortOverride.get().getHostText() :
+                (pollForFirstReachableEnabled ? getFirstReachableAddress(node, config) : vmHostname);
+        int managementPort = hostAndPortOverride.isPresent() ? hostAndPortOverride.get().getPort() : node.getLoginPort();
+        try {
+            Networking.getInetAddressWithFixedName(managementAddress);
```
bostko: This was always used before.

Networking.getInetAddressWithFixedName is expecting two kinds of arguments:
- IPv4 or IPv6 address - when this is used it doesn't resolve the address or trying to reach it. It just checks whether it is formatted correctly.
- hostname - it tries to `private static InetAddress[] getAllByName0 (String host, InetAddress reqAddr, boolean check)` which uses `InetAddress.lookupAllHostAddr` which may do some resolving

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -771,17 +772,19 @@ protected MachineLocation obtainOnce(ConfigBag setup) throws NoMachinesAvailable
...
             if (skipJcloudsSshing) {
                 boolean waitForConnectable = (windows) ? waitForWinRmable : waitForSshable;
                 if (waitForConnectable) {
                     if (windows) {
                         // TODO Does jclouds support any windows user setup?
-                        initialCredentials = waitForWinRmAvailable(computeService, node, sshHostAndPortOverride, setup);
+                        initialCredentials = waitForWinRmAvailable(initialCredentials, managementHostAndPort, setup);
                     } else {
-                        initialCredentials = waitForSshable(computeService, node, sshHostAndPortOverride, setup);
+                        initialCredentials = waitForSshable(computeService, node, managementHostAndPort, setup);
```
bostko: Should I call this `managementHostAndPort` or `loginHostAndPort`?

aledsage:
This has caused `JcloudsByonLocationResolverStubbedRebindTest.testRebind` to be very slow (5 minutes). It is trying to find the first reachable IP (with a 5 minute timeout). After the 5 minutes, it does do everything else fast and then the test passes! This isn't ideal: we won't just stop early if the VM you are asking for is not reachable (e.g. is not running).

```
"main" prio=5 tid=0x00007fe6c1800000 nid=0x1703 waiting on condition [0x0000700000217000]
   java.lang.Thread.State: TIMED_WAITING (parking)
        at sun.misc.Unsafe.park(Native Method)
        - parking to wait for  <0x00000007f0047248> (a java.util.concurrent.CountDownLatch$Sync)
        at java.util.concurrent.locks.LockSupport.parkNanos(LockSupport.java:226)
        at java.util.concurrent.locks.AbstractQueuedSynchronizer.doAcquireSharedNanos(AbstractQueuedSynchronizer.java:1033)
        at java.util.concurrent.locks.AbstractQueuedSynchronizer.tryAcquireSharedNanos(AbstractQueuedSynchronizer.java:1326)
        at java.util.concurrent.CountDownLatch.await(CountDownLatch.java:282)
        at org.apache.brooklyn.util.net.ReachableSocketFinder.tryReachable(ReachableSocketFinder.java:141)
        at org.apache.brooklyn.util.net.ReachableSocketFinder.access$000(ReachableSocketFinder.java:50)
        at org.apache.brooklyn.util.net.ReachableSocketFinder$2.call(ReachableSocketFinder.java:93)
        at org.apache.brooklyn.util.net.ReachableSocketFinder$2.call(ReachableSocketFinder.java:91)
        at org.apache.brooklyn.util.repeat.Repeater.runKeepingError(Repeater.java:336)
        at org.apache.brooklyn.util.repeat.Repeater.run(Repeater.java:302)
        at org.apache.brooklyn.util.net.ReachableSocketFinder.findOpenSocketOnNode(ReachableSocketFinder.java:88)
        at org.apache.brooklyn.location.jclouds.JcloudsUtil.getFirstReachableAddress(JcloudsUtil.java:374)
        at org.apache.brooklyn.location.jclouds.JcloudsLocation.getFirstReachableAddress(JcloudsLocation.java:2998)
        at org.apache.brooklyn.location.jclouds.JcloudsLocation.resolveManagementHostAndPort(JcloudsLocation.java:2031)
        at org.apache.brooklyn.location.jclouds.JcloudsLocation.registerMachineLocation(JcloudsLocation.java:2378)
        at org.apache.brooklyn.location.jclouds.JcloudsLocation.registerMachine(JcloudsLocation.java:2371)
        at org.apache.brooklyn.location.jclouds.JcloudsByonLocationResolver$1.get(JcloudsByonLocationResolver.java:158)
        at org.apache.brooklyn.location.jclouds.JcloudsByonLocationResolver$1.get(JcloudsByonLocationResolver.java:1)
        at org.apache.brooklyn.location.byon.FixedListMachineProvisioningLocation.init(FixedListMachineProvisioningLocation.java:173)
        at org.apache.brooklyn.core.objs.proxy.InternalLocationFactory.createLocation(InternalLocationFactory.java:137)
        at org.apache.brooklyn.core.mgmt.internal.LocalLocationManager.createLocation(LocalLocationManager.java:94)
        at org.apache.brooklyn.core.location.BasicLocationRegistry.getLocationManaged(BasicLocationRegistry.java:336)
        at org.apache.brooklyn.location.jclouds.JcloudsByonLocationResolverStubbedRebindTest.getLocationManaged(JcloudsByonLocationResolverStubbedRebindTest.java:154)
        at org.apache.brooklyn.location.jclouds.JcloudsByonLocationResolverStubbedRebindTest.testRebind(JcloudsByonLocationResolverStubbedRebindTest.java:141)
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:57)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.lang.reflect.Method.invoke(Method.java:606)
        at org.testng.internal.MethodInvocationHelper.invokeMethod(MethodInvocationHelper.java:84)
        at org.testng.internal.Invoker.invokeMethod(Invoker.java:714)
        at org.testng.internal.Invoker.invokeTestMethod(Invoker.java:901)
        at org.testng.internal.Invoker.invokeTestMethods(Invoker.java:1231)
        at org.testng.internal.TestMethodWorker.invokeTestMethods(TestMethodWorker.java:127)
        at org.testng.internal.TestMethodWorker.run(TestMethodWorker.java:111)
        at org.testng.TestRunner.privateRun(TestRunner.java:767)
        at org.testng.TestRunner.run(TestRunner.java:617)
        at org.testng.SuiteRunner.runTest(SuiteRunner.java:348)
        at org.testng.SuiteRunner.runSequentially(SuiteRunner.java:343)
        at org.testng.SuiteRunner.privateRun(SuiteRunner.java:305)
        at org.testng.SuiteRunner.run(SuiteRunner.java:254)
        at org.testng.SuiteRunnerWorker.runSuite(SuiteRunnerWorker.java:52)
        at org.testng.SuiteRunnerWorker.run(SuiteRunnerWorker.java:86)
        at org.testng.TestNG.runSuitesSequentially(TestNG.java:1224)
        at org.testng.TestNG.runSuitesLocally(TestNG.java:1149)
        at org.testng.TestNG.run(TestNG.java:1057)
        at org.testng.remote.AbstractRemoteTestNG.run(AbstractRemoteTestNG.java:126)
        at org.testng.remote.RemoteTestNG.initAndRun(RemoteTestNG.java:152)
        at org.testng.remote.RemoteTestNG.main(RemoteTestNG.java:57)
```

aledsage:
Merging this now - it's better to have the fix for the provisioning case (and live with the 5 mins delay for jcloudsByon when the VM is unreachable), than to delay this PR longer.

PR #356, opened 2016-09-28
 BROOKLYN-325 Rebind entity starting 

iyovcheva:
This fixes BROOKLYN-325 - when server restarts while entity is starting or stopping, on rebind it will now go on fire instead of reporting the entity's sensors as starting/stopping.


software/base/src/main/java/org/apache/brooklyn/entity/software/base/SoftwareProcessImpl.java
```diff
@@ -440,12 +441,20 @@ public void rebind() {
         //Only if the expected state is ON_FIRE then the entity has permanently failed.
         Transition expectedState = getAttribute(SERVICE_STATE_EXPECTED);
         if (expectedState == null || expectedState.getState() != Lifecycle.RUNNING) {
+            //FIXME does not set on fire root application
+            if (expectedState.getState() == Lifecycle.STARTING || expectedState.getState() == Lifecycle.STOPPING) {
+                setAttribute(SERVICE_STATE_EXPECTED, new Lifecycle.Transition(Lifecycle.ON_FIRE, new Date()));
+                setAttribute(SERVICE_STATE_ACTUAL, Lifecycle.ON_FIRE);
```
neykov: Use `ServiceStateLogic.setExpectedState` instead of setting the two attributes yourself.

software/base/src/main/java/org/apache/brooklyn/entity/software/base/SoftwareProcessImpl.java
```diff
@@ -440,12 +441,20 @@ public void rebind() {
...
+            }
             LOG.warn("On rebind of {}, not calling software process rebind hooks because expected state is {}", this, expectedState);
             return;
         }
 
         Lifecycle actualState = getAttribute(SERVICE_STATE_ACTUAL);
         if (actualState == null || actualState != Lifecycle.RUNNING) {
+            if (expectedState.getState() == Lifecycle.STARTING || expectedState.getState() == Lifecycle.STOPPING) {
+                setAttribute(SERVICE_STATE_ACTUAL, Lifecycle.ON_FIRE);
+            }
```
neykov: Not sure about this one, wonder what conditions would cause it (excluding setting it on fire from above). The message below won't be correct any more if you change the state, can you update it (or log a separate msg for this case).

iyovcheva: This is fixed in a separate commit, so the comment will not show as outdated.

software/base/src/main/java/org/apache/brooklyn/entity/software/base/SoftwareProcessImpl.java
iyovcheva: This is fixed in a separate commit, so the comment will not show as outdated.

core/src/main/java/org/apache/brooklyn/core/entity/lifecycle/ServiceStateLogic.java
```diff
@@ -153,6 +153,26 @@ private ServiceStateLogic() {}
...
+            ((ComputeServiceState)enricher.get()).onEvent(null);
+        }
+    }
+    public static Lifecycle getExpectedState(Entity entity) {
+        Transition expected = entity.getAttribute(Attributes.SERVICE_STATE_EXPECTED);
+        if (expected==null) return null;
+        return expected.getState();
+    }
+
+    public static void setActualState(Entity entity, Lifecycle state) {
```
aledsage: Let's delete `setActualState`: it's not used. The pattern used is to set the expected state, and then for the actual state to be inferred by an enricher.

core/src/main/java/org/apache/brooklyn/core/mgmt/rebind/RebindIteration.java
```diff
@@ -687,8 +689,13 @@ protected void manageTheObjects() {
         Set<String> oldEntities = Sets.newLinkedHashSet(entityManager.getEntityIds());
         for (Entity entity: rebindContext.getEntities()) {
             ManagementTransitionMode oldMode = updateTransitionMode(entityManager, entity);
-            if (oldMode!=null)
+            if (oldMode!=null) {
                 oldEntities.remove(entity.getId());
+            }
+
+            if (ServiceStateLogic.getExpectedState(entity) == Lifecycle.STARTING || ServiceStateLogic.getExpectedState(entity) == Lifecycle.STOPPING) {
```
aledsage: I think this call needs to be guarded so it's not called by a HOT_STANDBY Brooklyn server (i.e. a read-only view) etc. We only want to set this if this entity is in `BrooklynObjectManagementMode.MANAGED_PRIMARY`.

(Note that is not the `oldMode`; it is the new mode that we care about).

I therefore wonder if there is a better place for this code. I'll dig around and see.

aledsage: See https://github.com/iyovcheva/brooklyn-server/pull/5 for an alternative (and for additional tests, including for hotStandby).

core/src/main/java/org/apache/brooklyn/core/mgmt/rebind/BasicEntityRebindSupport.java
```diff
@@ -237,13 +239,21 @@ protected void addLocations(RebindContext rebindContext, EntityMemento memento)
...
-    
+
+    @Override
     protected void instanceRebind(AbstractBrooklynObject instance) {
         Preconditions.checkState(instance == entity, "Expected %s and %s to match, but different objects", instance, entity);
         Lifecycle expectedState = ServiceStateLogic.getExpectedState(entity);
         if (expectedState == Lifecycle.STARTING || expectedState == Lifecycle.STOPPING) {
-            LOG.warn("Entity "+entity);
+            LOG.warn("Entity {} goes on-fire because it was in state {} on rebind",
+                    entity, ServiceStateLogic.getExpectedState(entity));
```
aledsage: Just use `expectedState` rather than calling `ServiceStateLogic.getExecptedState(entity)` again.

core/src/main/java/org/apache/brooklyn/core/mgmt/rebind/BasicEntityRebindSupport.java
```diff
@@ -237,13 +239,21 @@ protected void addLocations(RebindContext rebindContext, EntityMemento memento)
...
+            LOG.warn("not-up-indicators={}", entity.getAttribute(Attributes.SERVICE_NOT_UP_INDICATORS));
             ServiceStateLogic.setExpectedState(entity, Lifecycle.ON_FIRE);
+
+            if (!(entity instanceof BasicApplication) && entity.getLocations().isEmpty()) {
+                LOG.warn("Rebind happened during provisioning for entity {} with parent {}" +
```
aledsage: Unfortunately this is too general. There are some entity types that will not necessarily have locations (e.g. that are not software processes, but instead bind to an API etc. For example, the `GistGenerator` docs example in http://docs.cloudsoft.io/blueprints/java/defining-and-deploying.html.

Let's take this out, and come back to it in a separate PR.

aledsage:
@iyovcheva I've added a couple more minor comments. Can you please address those, and then I think this is good to merge!

aledsage:
@neykov any comments you have are much appreciated as well.

core/src/main/java/org/apache/brooklyn/core/mgmt/rebind/BasicEntityRebindSupport.java
iyovcheva: For some reason this is not showing outdated.

aledsage:
Thanks @iyovcheva - LGTM; merging. We can look at the other TODOs in SoftwareProcessRebindNotRunningEntityTest.java (which were removed in the final commit?!).

PR #357, opened 2016-09-28
Fix LocalhostMachineProvisioningLocationTest

aledsage:
Use @BeforeMethod rather than @BeforeClass for test setup/teardown

The PR build for https://github.com/apache/brooklyn-server/pull/355 (see https://builds.apache.org/job/brooklyn-server-pull-requests/org.apache.brooklyn$brooklyn-core/1138/consoleFull) shows all but the first test in `LocalhostMachineProvisioningLocationTest` failing with the exception:

```
2016-09-27 21:49:31,219 INFO  TESTNG FAILED: "Surefire test" - org.apache.brooklyn.location.localhost.LocalhostMachineProvisioningLocationTest.obtainLowNumberedPortsAutomatically() finished in 0 ms
java.lang.IllegalStateException: Management context no longer running
	at org.apache.brooklyn.core.mgmt.internal.LocalManagementContext.getLocationManager(LocalManagementContext.java:270)
	at org.apache.brooklyn.location.localhost.LocalhostMachineProvisioningLocationTest.newLocalhostProvisioner(LocalhostMachineProvisioningLocationTest.java:71)
	at org.apache.brooklyn.location.localhost.LocalhostMachineProvisioningLocationTest.obtainLowNumberedPortsAutomatically(LocalhostMachineProvisioningLocationTest.java:161)
```

Looking at the console output, there are tests from other classes run between the first and subsequent tests in `LocalhostMachineProvisioningLocationTest`. Presumably one of these spotted that there was a management context left running, and terminated it?!

The fix should be simple: to change the test to use `@BeforeMethod` and `@AfterMethod` like our other tests do.

Not sure why jenkins/surefire ran the tests in a strange order. Hopefully it's not running the tests concurrently, but was instead just interleaving tests from different classes?

### DIFF
--- a/core/src/test/java/org/apache/brooklyn/location/localhost/LocalhostMachineProvisioningLocationTest.java
+++ b/core/src/test/java/org/apache/brooklyn/location/localhost/LocalhostMachineProvisioningLocationTest.java
@@ -24,47 +24,42 @@ import static org.testng.Assert.fail;
 
 import java.net.ServerSocket;
 
+import org.apache.brooklyn.api.location.LocationSpec;
+import org.apache.brooklyn.api.location.MachineProvisioningLocation;
+import org.apache.brooklyn.api.location.NoMachinesAvailableException;
+import org.apache.brooklyn.api.location.PortRange;
+import org.apache.brooklyn.core.location.PortRanges;
+import org.apache.brooklyn.core.location.geo.HostGeoInfo;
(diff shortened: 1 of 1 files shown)


neykov:
> Not sure why jenkins/surefire ran the tests in a strange order. Hopefully it's not running the tests concurrently, but was instead just interleaving tests from different classes?

Yes, testng will interleave tests in the same suite. Which for us is a single maven module in most cases. The tests are single threaded, it's not running them in parallel.

PR #358, opened 2016-09-28
Allow REST API to return the pinned config key and their constraints for catalog apps 

tbouron:
Based on https://github.com/apache/brooklyn-server/pull/354

This expose a new `pinned` flag and `constraints` key to the REST API. It allows the following:
- A blueprint can specify a config key or parameter as pinned by using `@CatalogConfig(pinned = true)` or in yaml `pinned: true`
- In case of config keys or parameters pinned in YAML, the priority is auto incremented to keep the order (code lost in [this commit](https://github.com/apache/brooklyn-server/commit/6f624c78b1e7fe72c6df1ecd297b922721b2c023#diff-5310fbeb1f42a3a4cf0508ea441440baL82))

### DIFF
--- a/api/src/main/java/org/apache/brooklyn/api/catalog/CatalogConfig.java
+++ b/api/src/main/java/org/apache/brooklyn/api/catalog/CatalogConfig.java
@@ -30,9 +30,11 @@ public @interface CatalogConfig {
     /** a label to be displayed when a config key is exposed as editable in the catalog */ 
     String label();
     
-    /** a priority used to determine the order in which config keys are displayed when presenting as editable in the catalog;
-     * a higher value appears higher in the list. the default is 1.
-     * (negative values may be used to indicate advanced config which might not be shown unless requested.) */ 
+    /** a priority used to determine the order in which config keys are displayed when presenting as editable in the
+     * catalog. The default is 1. */
     double priority() default 1;
-    
+
+     /** a pinned configuration means that the config key will always be displayed when presenting as editable in the
+      * catalog. The default is true. */
+    boolean pinned() default true;
 }
--- a/camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/catalog/SpecParameterParsingTest.java
+++ b/camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/catalog/SpecParameterParsingTest.java
@@ -66,7 +66,8 @@ public class SpecParameterParsingTest  extends AbstractYamlTest {
                 "    - simple",
                 "    - name: explicit_name",
                 "    - name: third_input",
-                "      type: integer");
+                "      type: integer",
+                "      pinned: false");
         EntitySpec<?> item = mgmt().getTypeRegistry().createSpec(mgmt().getTypeRegistry().get(itemId), null, EntitySpec.class);
         List<SpecParameter<?>> inputs = item.getParameters();
         assertEquals(inputs.size(), 6);
@@ -84,7 +85,7 @@ public class SpecParameterParsingTest  extends AbstractYamlTest {
         
         SpecParameter<?> thirdInput = inputs.get(2);
         assertEquals(thirdInput.getLabel(), "third_input");
-        assertEquals(thirdInput.isPinned(), true);
+        assertEquals(thirdInput.isPinned(), false);
         assertEquals(thirdInput.getConfigKey().getName(), "third_input");
         assertEquals(thirdInput.getConfigKey().getTypeToken(), TypeToken.of(Integer.class));
     }
--- a/core/src/main/java/org/apache/brooklyn/core/objs/BasicSpecParameter.java
+++ b/core/src/main/java/org/apache/brooklyn/core/objs/BasicSpecParameter.java
@@ -26,16 +26,6 @@ import java.util.Date;
 import java.util.List;
 import java.util.Map;
 
-import com.google.common.annotations.Beta;
-import com.google.common.base.Function;
-import com.google.common.base.Objects;
-import com.google.common.base.Predicate;
-import com.google.common.base.Predicates;
-import com.google.common.collect.FluentIterable;
-import com.google.common.collect.ImmutableList;
-import com.google.common.collect.ImmutableMap;
-import com.google.common.reflect.TypeToken;
-
 import org.apache.brooklyn.api.catalog.CatalogConfig;
 import org.apache.brooklyn.api.entity.Entity;
 import org.apache.brooklyn.api.entity.EntitySpec;
@@ -57,6 +47,16 @@ import org.apache.brooklyn.util.collections.MutableList;
 import org.apache.brooklyn.util.guava.Maybe;
 import org.apache.brooklyn.util.text.StringPredicates;
 import org.apache.brooklyn.util.time.Duration;
+
+import com.google.common.annotations.Beta;
+import com.google.common.base.Function;
+import com.google.common.base.Objects;
+import com.google.common.base.Predicate;
+import com.google.common.base.Predicates;
+import com.google.common.collect.FluentIterable;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.reflect.TypeToken;
 
 public class BasicSpecParameter<T> implements SpecParameter<T>{
     private static final long serialVersionUID = -4728186276307619778L;
@@ -83,12 +83,12 @@ public class BasicSpecParameter<T> implements SpecParameter<T>{
         }
     }
     
-    @Beta // TBD whether "pinned" stays
+    @Beta
     public BasicSpecParameter(String label, boolean pinned, ConfigKey<T> config) {
         this(label, pinned, config, null);
     }
 
-    @Beta // TBD whether "pinned" and "sensor" stay
+    @Beta // TBD whether "sensor" stay
     public <SensorType> BasicSpecParameter(String label, boolean pinned, ConfigKey<T> config, AttributeSensor<SensorType> sensor) {
         this.label = label;
         this.pinned = pinned;
@@ -228,10 +228,11 @@ public class BasicSpecParameter<T> implements SpecParameter<T>{
             String description = (String)inputDef.get("description");
             String type = (String)inputDef.get("type");
             Object defaultValue = inputDef.get("default");
+            Boolean pinned = (Boolean)inputDef.get("pinned");
             if (specialFlagTransformer != null) {
                 defaultValue = specialFlagTransformer.apply(defaultValue);
             }
-            Predicate<?> constraints = parseConstraints(inputDef.get("constraints"), loader);
+            Predicate<?> constraint = parseConstraints(inputDef.get("constraints"), loader);
             ConfigInheritance parentInheritance = parseInheritance(inputDef.get("inheritance.parent"), loader);
             ConfigInheritance typeInheritance = parseInheritance(inputDef.get("inheritance.type"), loader);
 
@@ -247,17 +248,17 @@ public class BasicSpecParameter<T> implements SpecParameter<T>{
                 .name(name)
                 .description(description)
                 .defaultValue(defaultValue)
-                .constraint(constraints)
+                .constraint(constraint)
                 .runtimeInheritance(parentInheritance)
                 .typeInheritance(typeInheritance);
-            
+
             if (PortRange.class.equals(typeToken.getRawType())) {
                 sensorType = new PortAttributeSensorAndConfigKey(builder);
                 configType = ((HasConfigKey)sensorType).getConfigKey();
             } else {
                 configType = builder.build();
             }
-            return new BasicSpecParameter(Objects.firstNonNull(label, name), true, configType, sensorType);
+            return new BasicSpecParameter(Objects.firstNonNull(label, name), !Boolean.FALSE.equals(pinned), configType, sensorType);
         }
 
         @SuppressWarnings({ "rawtypes" })
@@ -276,7 +277,7 @@ public class BasicSpecParameter<T> implements SpecParameter<T>{
                 }
             }
         }
-    
+
         @SuppressWarnings({ "unchecked", "rawtypes" })
         private static Predicate parseConstraints(Object obj, BrooklynClassLoadingContext loader) {
             List constraintsRaw;
@@ -375,12 +376,14 @@ public class BasicSpecParameter<T> implements SpecParameter<T>{
             CatalogConfig catalogConfig = configKeyField.getAnnotation(CatalogConfig.class);
             String label = config.getName();
             Double priority = null;
+            Boolean pinned = Boolean.FALSE;
             if (catalogConfig != null) {
                 label = Maybe.fromNullable(catalogConfig.label()).or(config.getName());
                 priority = catalogConfig.priority();
+                pinned = catalogConfig.pinned();
             }
             @SuppressWarnings({ "unchecked", "rawtypes" })
-            SpecParameter<?> param = new BasicSpecParameter(label, priority != null, config);
+            SpecParameter<?> param = new BasicSpecParameter(label, pinned, config);
             return new WeightedParameter(priority, param);
         }
 
--- a/rest/rest-api/src/main/java/org/apache/brooklyn/rest/domain/EntityConfigSummary.java
+++ b/rest/rest-api/src/main/java/org/apache/brooklyn/rest/domain/EntityConfigSummary.java
@@ -18,20 +18,29 @@
  */
 package org.apache.brooklyn.rest.domain;
 
-import com.fasterxml.jackson.annotation.JsonProperty;
-import com.fasterxml.jackson.databind.annotation.JsonSerialize;
-import com.google.common.collect.ImmutableMap;
-
-import org.apache.brooklyn.config.ConfigKey;
-
 import java.net.URI;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 
+import org.apache.brooklyn.config.ConfigKey;
+import org.apache.brooklyn.util.text.StringPredicates;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import com.google.common.base.Predicates;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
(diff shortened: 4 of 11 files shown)


tbouron:
FYI, I tested before pushing. I also just retested locally and all tests pass. Not sure what is going on here

api/src/main/java/org/apache/brooklyn/api/catalog/CatalogConfig.java
```diff
@@ -34,5 +34,7 @@
      * a higher value appears higher in the list. the default is 1.
      * (negative values may be used to indicate advanced config which might not be shown unless requested.) */ 
     double priority() default 1;
-    
+
+     /** a pinned configuration means that it should be displayed, regardless of its priority */
+    boolean pinned() default false;
```
neykov: The `@CatalogConfig` annotation is set on items that we want visible in the catalog. To prevent an item from being seen by default one can set a negative priority above. This allows for having an order of the keys and at the same time declares whether they should be visible or not. Could keep the existing behaviour and not expose `pinned` here (derive it instead from priority). But having `pinned` might be more user friendly so I'm fine keeping it as is.

Can you update the javadoc above - doesn't make sense to support negative priorities any more. They make documenting and understanding the behaviour more difficult when paired with `pinned`. Let's keep the behaviour for backwards compatibility, but not advertise it.

core/src/main/java/org/apache/brooklyn/core/config/BasicConfigKey.java
```diff
@@ -331,12 +337,19 @@ public ConfigInheritance getParentInheritance() {
...
     @Override @Nonnull
     public Predicate<? super T> getConstraint() {
         // Could be null after rebinding
-        if (constraint != null) {
-            return constraint;
+        if (constraints.size() > 0) {
+            return Predicates.and(constraints.values());
         } else {
             return Predicates.alwaysTrue();
         }
```
neykov: Having a name to describe the predicate duplicates the information coming from the predicate itself. Suggest deciding on whether the predicate means "required" based on its type name in the REST API.

core/src/main/java/org/apache/brooklyn/core/objs/BasicSpecParameter.java
```diff
@@ -247,17 +248,17 @@ public String toString() {
...
-            
+
             if (PortRange.class.equals(typeToken.getRawType())) {
                 sensorType = new PortAttributeSensorAndConfigKey(builder);
                 configType = ((HasConfigKey)sensorType).getConfigKey();
             } else {
                 configType = builder.build();
             }
-            return new BasicSpecParameter(Objects.firstNonNull(label, name), true, configType, sensorType);
+            return new BasicSpecParameter(Objects.firstNonNull(label, name), Objects.firstNonNull(pinned, Boolean.FALSE), configType, sensorType);
```
neykov: The usual way to express this in the codebase is: `Boolean.TRUE.equals(pinned)`, but this works as well.

api/src/main/java/org/apache/brooklyn/api/catalog/CatalogConfig.java
tbouron: Yeah, make sense. I'll update the Javadoc

core/src/main/java/org/apache/brooklyn/core/objs/BasicSpecParameter.java
tbouron: I prefer to comply with the "standard". I'll update it

api/src/main/java/org/apache/brooklyn/api/catalog/CatalogConfig.java
neykov: This will break backwards compatibility by hiding all already annotated fields.

tbouron: I would rather modify https://github.com/apache/brooklyn-library/pull/65 to add `pinned = true` to all annotated field, than changing this to `true`. Mark by default annotated fields as pinned feels wrong

core/src/main/java/org/apache/brooklyn/core/config/BasicConfigKey.java
tbouron: Thought about that but there is issue as you can have more than one implementation of `Predicate` to do the same thing. In the case of `required`, I have seen implemented as:
- `StringPredicates.isNonBlank()`
- `Predicates.notNull()`
And we could potential have others

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/transform/EntityTransformer.java
```diff
@@ -162,7 +163,7 @@ public static EntityConfigSummary entityConfigSummary(Entity entity, ConfigKey<?
             SensorTransformer.addNamedAction(lb, na, entity.getConfig(config), config, entity);
         }
     
-        return entityConfigSummary(config, label, priority, lb.build());
+        return entityConfigSummary(config, label, priority, false, lb.build());
```
neykov: This will mark all config as invisible in `v1/applications/xxxxxxxx/entities/yyyyyyyy/config`. That's fine as long as the UI know about this caveat. Might be better to make this field nullable and set `null` here?

core/src/main/java/org/apache/brooklyn/core/config/BasicConfigKey.java
tbouron: Another solution would be to create a specific interface for constraints but it felt over-engineered. That's why I kept the basic `Map<String, Predicate>`

neykov: What we care about is the "required" as implemented in yaml (`isNonBlank`). It's enought to support that, but could easily add some more to a white-list. Also alias to it as a "required" predicate to be used in Java if anyone wants to integrate with UI better from Java land.

Also keep in mind backwards compatibility (always :) ). Instead of changing the API straight away we'd have to go through a deprecation cycle where the old signatures are still supported.  That's one more reason why stick to the existing non-labeled predicates.

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/transform/EntityTransformer.java
tbouron: Well, yes and no. I mean, the UI does not check a `pinned` flag for managed entities but only for catalog templates. So technically, config will never be invisible. 
However, I do agree that the REST API shouldn't return a `pinned: false` for this case for I'll update for a nullable field.

api/src/main/java/org/apache/brooklyn/api/catalog/CatalogConfig.java
neykov: It's not only `brooklyn-library`. Users of Brooklyn writing their own entities will have the same problem. The code will compile cleanly without warnings (even will be binary compatible for existing jars). The fields will just be invisible in UI.

Same applies (even stronger) to yaml. Making all fields invisible by default will break all existing blueprints.

For Java applying a `@CatalogConfig` annotation in 99% of the cases is done to show parameters in UI. In YAML adding top-level parameters is done for the same reason. That's why I think existing behaviour makes sense and we should keep it.

core/src/main/java/org/apache/brooklyn/core/config/BasicConfigKey.java
tbouron: OK, that makes sense. So to recap:
- `StringPredicates.isNonBlank()` becomes `required` for the REST API
- Any other predicate becomes the FQDN of the predicate? Sounds sensible @neykov ?

neykov: agree

api/src/main/java/org/apache/brooklyn/api/catalog/CatalogConfig.java
tbouron: Ok, I'll update

tbouron:
@neykov Comment addressed

camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/catalog/SpecParameterParsingTest.java
```diff
@@ -66,19 +66,20 @@ public void testYamlInputsParsed() {
...
-                "      type: integer");
+                "      type: integer",
+                "      pinned: true");
         EntitySpec<?> item = mgmt().getTypeRegistry().createSpec(mgmt().getTypeRegistry().get(itemId), null, EntitySpec.class);
         List<SpecParameter<?>> inputs = item.getParameters();
         assertEquals(inputs.size(), 6);
         SpecParameter<?> firstInput = inputs.get(0);
         assertEquals(firstInput.getLabel(), "simple");
-        assertEquals(firstInput.isPinned(), true);
+        assertEquals(firstInput.isPinned(), false);
```
neykov: How come the tests still pass with this change? Am I missing something?
The default is flipped to `true` now, so the change shouldn't be needed?

tbouron: I changed the annotation `@CatalogConfig`, forgot about the YAML parsing... Updating now

neykov: Oh I see, the default when coming parsing yaml is still `false`, can you change it to `true` as discussed above.

tbouron: @neykov addressed

core/src/main/java/org/apache/brooklyn/core/config/BasicConfigKey.java
```diff
@@ -184,7 +192,7 @@ public String getDescription() {
     protected boolean reconfigurable;
     protected ConfigInheritance typeInheritance;
     protected ConfigInheritance runtimeInheritance;
-    protected Predicate<? super T> constraint;
+    protected List<Predicate<? super T>> constraints;
```
neykov: This is persisted so should be extra careful with renames. When passing constraints from java the caller can wrap them in `and` if multiple, so this change is not needed. Suggest reverting.

rest/rest-api/src/main/java/org/apache/brooklyn/rest/domain/EntityConfigSummary.java
```diff
@@ -18,43 +18,82 @@
...
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import com.google.common.base.Function;
+import com.google.common.base.Predicate;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Iterables;
+
 public class EntityConfigSummary extends ConfigSummary {
 
     private static final long serialVersionUID = -1336134336883426030L;
```
neykov: Regenerate after adding members.

rest/rest-api/src/main/java/org/apache/brooklyn/rest/domain/EntityConfigSummary.java
```diff
@@ -18,43 +18,82 @@
...
+    @JsonSerialize(include = JsonSerialize.Inclusion.NON_NULL)
     private final Map<String, URI> links;
 
     public EntityConfigSummary(
             @JsonProperty("name") String name,
             @JsonProperty("type") String type,
             @JsonProperty("description") String description,
             @JsonProperty("defaultValue") Object defaultValue,
-            @JsonProperty("reconfigurable") boolean reconfigurable,
+            @JsonProperty("reconfigurable") Boolean reconfigurable,
```
neykov: Why change this?

rest/rest-api/src/main/java/org/apache/brooklyn/rest/domain/EntityConfigSummary.java
```diff
@@ -18,43 +18,82 @@
...
+    public EntityConfigSummary(ConfigKey<?> config, String label, Double priority, Boolean pinned, Map<String, URI> links) {
         super(config, label, priority);
+        this.pinned = pinned;
+        this.constraints = ImmutableList.copyOf(Iterables.transform(config.getConstraints(), new Function<Predicate, String>() {
+            @Nullable
+            @Override
+            public String apply(@Nullable Predicate predicate) {
+                if (predicate == null) {
+                    return null;
+                } else if (predicate.getClass().getName().equals(StringPredicates.isNonBlank().getClass().getName())) {
```
neykov: Minor,  could assign `StringPredicates.isNonBlank().getClass().getName()`  to a static field of the anonymous class.

neykov: Could even compare the classes without going through the name, but doesn't really matter in the end.

neykov:
In summary - looks good. Only concerned about introducing `constraints` which breaks persistence compatibility. Suggest reverting all related changes and making the predicate -> string conversion smarter to look inside `and` predicates.

core/src/main/java/org/apache/brooklyn/core/config/BasicConfigKey.java
tbouron: But this is the whole point of this change, expose the constraints as a list of string on the REST API. I thought about doing the change you are suggesting @neykov (looking into the `Predicates.and()`) but there is no way to get back the underlying predicates.
Reverting would mean to support only one predicate at a time which is a contradiction with the current YAML implementation (can specify more than one)

rest/rest-api/src/main/java/org/apache/brooklyn/rest/domain/EntityConfigSummary.java
tbouron: That's a mistake, I'll revert back

core/src/main/java/org/apache/brooklyn/core/config/BasicConfigKey.java
tbouron: Reverted. I'll try to address the multiple predicate issue in another PR

tbouron:
@neykov Comments addressed

core/src/main/java/org/apache/brooklyn/core/config/BasicConfigKey.java
```diff
@@ -422,15 +422,15 @@ public BasicConfigKeyOverwriting(Builder<T,?> builder, ConfigKey<T> parent) {
...
         public BasicConfigKeyOverwriting(ConfigKey<T> key, T defaultValue) {
             this(builder(key).defaultValue(defaultValue), key);
         }
-        
+
         public BasicConfigKeyOverwriting(ConfigKey<T> key, String newDescription, T defaultValue) {
             this(builder(key).description(newDescription).defaultValue(defaultValue), key);
         }
-        
+
```
neykov: No non-white space changes in the file, can you remove from the change set.

core/src/main/java/org/apache/brooklyn/core/config/MapConfigKey.java
```diff
@@ -126,7 +126,7 @@ protected MapConfigKey(Builder<V> builder) {
         // Note: it's intentionally possible to have default values that are not valid
         // per the configured constraint. If validity were checked here any class that
         // contained a weirdly-defined config key would fail to initialise.
-        this.constraint = checkNotNull(builder.constraint, "constraint");
+        this.constraint = checkNotNull(builder.constraint, "constraints");
```
neykov: Probably a leftover from reverting - the field name is still called `constraint`.

core/src/main/java/org/apache/brooklyn/core/objs/BasicSpecParameter.java
```diff
@@ -375,12 +376,14 @@ public static WeightedParameter getFieldConfig(ConfigKey<?> config, Field config
             CatalogConfig catalogConfig = configKeyField.getAnnotation(CatalogConfig.class);
             String label = config.getName();
             Double priority = null;
+            Boolean pinned = Boolean.FALSE;
```
neykov: Shouldn't the default be `true`?

neykov: I take it back - non-anotated items should be unpinned.

rest/rest-api/src/main/java/org/apache/brooklyn/rest/domain/EntityConfigSummary.java
```diff
@@ -45,16 +54,32 @@ public EntityConfigSummary(
...
+        this.pinned = pinned;
+        this.constraints = (constraints == null) ? ImmutableList.<String>of() : ImmutableList.copyOf(constraints);
         this.links = (links == null) ? ImmutableMap.<String, URI>of() : ImmutableMap.copyOf(links);
     }
 
-    public EntityConfigSummary(ConfigKey<?> config, String label, Double priority, Map<String, URI> links) {
+    public EntityConfigSummary(ConfigKey<?> config, String label, Double priority, Boolean pinned, Map<String, URI> links) {
         super(config, label, priority);
+        this.pinned = pinned;
+        this.constraints = !config.getConstraint().equals(Predicates.alwaysTrue())
```
neykov: It's unexpected to me but predicates don't override the `equals` method so they follow the default behaviour. Why this comparison works is because the implementation of `alwaysTrue` is an enum. Worth a comment.

What they usually do is override the `toString` method so the most reliable means for comparing predicates is comparing their `toString` values. We are dealing with a well defined predicate here so no need to change.

neykov:
LGTM - very minor comments only.

tbouron:
@neykov updated

neykov:
Looks great @tbouron. Can you squash then good to merge.

tbouron:
@neykov You can do it from the [Github UI directly](https://help.github.com/articles/merging-a-pull-request/#merging-a-pull-request-using-the-github-web-interface) ;)

neykov:
Unfortunately it doesn't work for Apache projects. The github repo is a read-only copy of the Apache repo where this will get merged. Also can't use the "share branch with committers" feature.

tbouron:
@neykov Ah ok, I didn't know. Squashed

neykov:
Thanks, merging.

PR #359, opened 2016-09-28
Utils tidy

ahgittin:
tiny tidies to utils, including deprecation making `Maybe` a bit easier to work with in a couple places

### DIFF
--- a/utils/common/src/main/java/org/apache/brooklyn/test/Asserts.java
+++ b/utils/common/src/main/java/org/apache/brooklyn/test/Asserts.java
@@ -18,8 +18,6 @@
  */
 package org.apache.brooklyn.test;
 
-import groovy.lang.Closure;
-
 import java.lang.reflect.Array;
 import java.util.Arrays;
 import java.util.Collection;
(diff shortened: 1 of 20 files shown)


utils/common/src/main/java/org/apache/brooklyn/util/text/Identifiers.java
```diff
@@ -168,9 +168,11 @@ public static String makeRandomPassword(final int length, String... passwordVali
         return Joiner.on("").join(list);
     }
 
+    @SuppressWarnings({ "rawtypes", "unchecked" })  
     protected static String mergeCharacterSets(String... s) {
         Set characters = new HashSet<Character>();
         for (String characterSet : s) {
+            // more efficient to lose the generics here
```
aledsage: What does this comment mean? Given that Java uses erasure (except in very special cases!) then how is it more efficient to lose the generics?

It looks like the generics are wrong: it should be `Set<String>`.

utils/common/src/main/java/org/apache/brooklyn/test/Asserts.java
```diff
@@ -1381,4 +1381,18 @@ public static void assertSize(Iterable<?> list, int expectedSize) {
         if (Iterables.size(list)!=expectedSize) fail("List has wrong size "+Iterables.size(list)+" (expected "+expectedSize+"): "+list);
     }
 
+    public static void assertInstanceOf(Object obj, Class<?> type) {
+        assertThat(obj, Predicates.instanceOf(type));
```
aledsage: Not convinced that this is useful (thought agree it can be made so, to make the test more explicit). It gives less information about the problem than a good old fasioned `ClassCastException`: it doesn't tell you what type `obj` is (just its `toString`).

aledsage:
Build failure is real: missing import for `Maybe` in `Asserts.java`:

```
[ERROR] /home/jenkins/jenkins-slave/workspace/brooklyn-server-pull-requests/utils/common/src/main/java/org/apache/brooklyn/test/Asserts.java:[1388,42] cannot find symbol
[ERROR] symbol:   class Maybe
[ERROR] location: class org.apache.brooklyn.test.Asserts
```

aledsage:
LGTM. Build failure must be fixed before merging. Other two comments are extremely minor.
Happy for you to merge when you're ready.

ahgittin:
good spots @aledsage, i've tidied both those points (give better asserts message, and improve identifiers code)

neykov:
Thanks @ahgittin. Tests are passing locally, merging.

PR #364, opened 2016-09-30
Rebind when entity starting: report notUpIndicator for reason

aledsage:
Entity is reported as on-fire if the starting/stopping task was
aborted due to a Brooklyn rebind. This will set the
SERVICE_NOT_IP_INDICATORS to say what the problem is.


iyovcheva:
Looks good

neykov:
Useful addition, merging.

PR #366, opened 2016-10-04
Test jclouds location choose reachable address

bostko:
Used changes from https://github.com/aledsage/brooklyn-server/tree/test-JcloudsLocation-chooseReachableAddress
and https://github.com/bostko/brooklyn-server/tree/test-JcloudsLocation-chooseReachableAddress


aledsage:
Looks great - love the way you have made `JcloudsReachableAddressStubbedTest` a true unit test by stubbing out the `TemplateBuilder` with `JcloudsStubTemplateBuilder`.

I may take a look at that separately, for how much we hardcode in `JcloudsStubTemplateBuilder`. And look at using that in more of our tests that don't actually require real responses from the cloud provider!

Merging now.

PR #369, opened 2016-10-05
DynamicGroup filters entities in its app, not the management context

sjcorbett:
I think this would be the expected behaviour.

The current behaviour would cause entities like [brooklyn-dns-etc-hosts-generator](https://github.com/brooklyncentral/brooklyn-dns/blob/9898438dd2c8ddbe258ab93b1b66b42818b8121a/brooklyn-dns-etc-hosts-generator.bom#L43-L49) to be overenthusiastic at best and a security hole/information leak at worst.


sjcorbett:
The build timed out building the Camp API:
```
[INFO] Brooklyn CAMP REST API ............................. FAILURE [33:24 min]
```

geomacy:
Looks good to me, 👍

sjcorbett:
@geomacy thanks. Merging.

PR #370, opened 2016-10-05
ProxyEffector

sjcorbett:
Forwards effector invocations from one entity to another.

### DIFF
--- a/camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/ProxyEffectorYamlTest.java
+++ b/camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/ProxyEffectorYamlTest.java
@@ -1,0 +1,93 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.brooklyn.camp.brooklyn;
+
+import static org.apache.brooklyn.test.Asserts.assertEquals;
+
+import java.util.Iterator;
+
+import org.apache.brooklyn.api.effector.Effector;
+import org.apache.brooklyn.api.entity.Entity;
+import org.apache.brooklyn.core.test.entity.TestEntity;
+import org.apache.brooklyn.entity.stock.BasicEntity;
+import org.apache.brooklyn.test.Asserts;
+import org.testng.annotations.Test;
+
+import com.google.common.collect.ImmutableMap;
+
+public class ProxyEffectorYamlTest extends AbstractYamlTest {
+
+    @Test
+    public void testProxyEffector() throws Exception {
+        final String effectorName = "proxy-effector";
+        Entity app = createAndStartApplication(
+                "location: localhost",
+                "services:",
+                "- id: target",
+                "  type: " + TestEntity.class.getName(),
+                "- type: " + BasicEntity.class.getName(),
+                "  brooklyn.initializers:",
+                "  - type: org.apache.brooklyn.core.effector.ProxyEffector",
+                "    brooklyn.config:",
+                "      name: " + effectorName,
+                "      targetEntity: $brooklyn:entity(\"target\")",
+                "      targetEffector: identityEffector"
+        );
+        waitForApplicationTasks(app);
+
+        assertEquals(app.getChildren().size(), 2, "expected two elements in " + app.getChildren());
+        Entity basicEntity;
+        Iterator<Entity> ei = app.getChildren().iterator();
+        basicEntity = ei.next();
+        if (!BasicEntity.class.isAssignableFrom(basicEntity.getClass())) {
+            basicEntity = ei.next();
+        }
+
+        Effector<?> effector = basicEntity.getEntityType().getEffectorByName(effectorName).get();
+        Object result = basicEntity.invoke(effector, ImmutableMap.of("arg", "hello, world")).get();
+        assertEquals(((String) result).trim(), "hello, world");
+}
+
+    @Test
+    public void testThrowsIfTargetDoesNotResolve() throws Exception {
+        final String effectorName = "proxy-effector";
+        Entity app = createAndStartApplication(
+                "location: localhost",
+                "services:",
+                "- type: " + BasicEntity.class.getName(),
+                "  brooklyn.initializers:",
+                "  - type: org.apache.brooklyn.core.effector.ProxyEffector",
+                "    brooklyn.config:",
+                "      name: " + effectorName,
+                "      targetEntity: $brooklyn:entity(\"skhbfskdbf\")",
+                "      targetEffector: identityEffector"
+        );
+        waitForApplicationTasks(app);
+        Entity basicEntity = app.getChildren().iterator().next();
+        Effector<?> effector = basicEntity.getEntityType().getEffectorByName(effectorName).get();
+        try {
+            basicEntity.invoke(effector, ImmutableMap.<String, Object>of()).get();
+            Asserts.shouldHaveFailedPreviously("expected exception when invoking effector that does not exist");
+        } catch (Exception e) {
+            Asserts.expectedFailure(e);
+        }
+    }
+
+}
--- a/core/src/main/java/org/apache/brooklyn/core/effector/ProxyEffector.java
+++ b/core/src/main/java/org/apache/brooklyn/core/effector/ProxyEffector.java
@@ -1,0 +1,96 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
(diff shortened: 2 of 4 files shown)


core/src/main/java/org/apache/brooklyn/core/effector/ProxyEffector.java
```diff
@@ -0,0 +1,99 @@
...
+            return Tasks.resolving(proxy, Entity.class)
+                    .context(entity())
+                    .swallowExceptions()
+                    .getMaybe();
+        }
+
+        private Object invokeEffectorNamed(Entity target, String effectorName, ConfigBag params) {
+            LOG.debug("{} forwarding effector invocation on {} to entity={}, effector={}, paramters={}",
+                    new Object[]{this, entity(), target, effectorName, params});
+            Effector<?> effector = EffectorUtils.findEffectorDeclared(target, effectorName).get();
```
geomacy: definitely worth checking the Maybe returned by findEffectorDeclared, and logging a meaningful exception if isAbsent() - I guess one of the common scenarios we will face will be typos in effector names.

core/src/test/java/org/apache/brooklyn/core/effector/ProxyEffectorTest.java
```diff
@@ -0,0 +1,88 @@
...
+                ProxyEffector.TARGET_ENTITY, a,
+                ProxyEffector.TARGET_EFFECTOR_NAME, "kajnfksjdnfkjsdnf"));
+        EntityInternal b = Entities.deproxy(app.createAndManageChild(EntitySpec.create(BasicEntity.class)
+                .addInitializer(proxyEffector)));
+        try {
+            b.invoke(b.getEffector("proxy-effector"), ImmutableMap.of("arg", "value"))
+                    .getUnchecked(Duration.ONE_MINUTE);
+            Asserts.shouldHaveFailedPreviously("expected exception when invoking effector that does not exist");
+        } catch (Exception e) {
+            // expected.
```
geomacy: Need to add checks in here that we get the right exception, whatever that is exactly, with info about the name of the effector at least. (`Asserts.expectedFailureContains()`).

core/src/test/java/org/apache/brooklyn/core/effector/ProxyEffectorTest.java
```diff
@@ -0,0 +1,88 @@
...
+    public void testHappyPath() {
+        TestEntity a = app.createAndManageChild(EntitySpec.create(TestEntity.class));
+        ProxyEffector proxyEffector = new ProxyEffector(ImmutableMap.of(
+                AddEffector.EFFECTOR_NAME, "proxy-effector",
+                ProxyEffector.TARGET_ENTITY, a,
+                ProxyEffector.TARGET_EFFECTOR_NAME, "identityEffector"));
+        // BasicEntity doesn't have an identityEffector.
+        EntityInternal b = Entities.deproxy(app.createAndManageChild(EntitySpec.create(BasicEntity.class)
+                .addInitializer(proxyEffector)));
+        Object output = b.invoke(b.getEffector("proxy-effector"), ImmutableMap.of("arg", "value"))
```
geomacy: Interesting - do we want to let the effector call a target effector on an entity from a different application?  Not sure of the answer to this - it might be useful?  and/or it might be risky?

camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/ProxyEffectorYamlTest.java
```diff
@@ -0,0 +1,93 @@
...
+            basicEntity = ei.next();
+        }
+
+        Effector<?> effector = basicEntity.getEntityType().getEffectorByName(effectorName).get();
+        Object result = basicEntity.invoke(effector, ImmutableMap.of("arg", "hello, world")).get();
+        assertEquals(((String) result).trim(), "hello, world");
+}
+
+    @Test
+    public void tesThrowsIfTargetDoesNotResolve() throws Exception {
```
geomacy: typo 'tesThrows'

core/src/main/java/org/apache/brooklyn/core/effector/ProxyEffector.java
```diff
@@ -0,0 +1,99 @@
...
+            return target.invoke(effector, params.getAllConfig()).getUnchecked();
```
geomacy: again here, could we wrap this in something that adds useful information about exceptions, and logs something here before it re-throws?

core/src/test/java/org/apache/brooklyn/core/effector/ProxyEffectorTest.java
```diff
@@ -0,0 +1,88 @@
...
+        try {
+            b.invoke(b.getEffector("proxy-effector"), ImmutableMap.of("arg", "value"))
+                    .getUnchecked(Duration.ONE_MINUTE);
+            Asserts.shouldHaveFailedPreviously("expected exception when invoking effector that does not exist");
+        } catch (Exception e) {
+            // expected.
+        }
+    }
+
+    @Test(expectedExceptions = NullPointerException.class)
```
geomacy: Worth adding Asserts.expected... and shouldHaveFailedPreviously.

core/src/test/java/org/apache/brooklyn/core/effector/ProxyEffectorTest.java
```diff
@@ -0,0 +1,88 @@
...
+    public void testThrowsIfEntityUnset() {
+        new ProxyEffector(MutableMap.of(
+                AddEffector.EFFECTOR_NAME, "proxy-effector",
+                ProxyEffector.TARGET_ENTITY, null,
+                ProxyEffector.TARGET_EFFECTOR_NAME, "kajnfksjdnfkjsdnf"));
+    }
+
+    @Test(expectedExceptions = NullPointerException.class)
+    public void testThrowsIfTargetEffectorNameUnset() {
```
geomacy: Worth adding Asserts.expected... and shouldHaveFailedPreviously.

camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/ProxyEffectorYamlTest.java
```diff
@@ -0,0 +1,93 @@
...
+                "      targetEffector: identityEffector"
+        );
+        waitForApplicationTasks(app);
+        Entity basicEntity = app.getChildren().iterator().next();
+        Effector<?> effector = basicEntity.getEntityType().getEffectorByName(effectorName).get();
+        try {
+            basicEntity.invoke(effector, ImmutableMap.<String, Object>of()).get();
+            Asserts.shouldHaveFailedPreviously("expected exception when invoking effector that does not exist");
+        } catch (Exception e) {
+            // expected.
```
geomacy: use Asserts.expectedFailure...

core/src/main/java/org/apache/brooklyn/core/effector/ProxyEffector.java
sjcorbett: Calling `get` on an absent value will throw a `NoSuchElementException`: https://github.com/apache/brooklyn-server/blob/6f6763c300f828cc3270f1c9db254b6eeb3f7c15/core/src/main/java/org/apache/brooklyn/core/mgmt/internal/EffectorUtils.java#L384

core/src/test/java/org/apache/brooklyn/core/effector/ProxyEffectorTest.java
sjcorbett: Don't follow this comment - `a` and `b` are in the same application. When configuring yaml you can inject any component you can find with the dsl. When using it from Java I think we have to assume whoever is supplying the target entity is doing the right thing.

core/src/test/java/org/apache/brooklyn/core/effector/ProxyEffectorTest.java
sjcorbett: What's the benefit?

sjcorbett:
@geomacy thanks for your comments. I've resolved the comments that I didn't reply to. Is there anything you'd like to follow up on?

core/src/main/java/org/apache/brooklyn/core/effector/ProxyEffector.java
geomacy: Yes, but will the exception have enough clear info in the message to explain what's gone wrong?  Maybe so, which is fine, but I was thinking just that if we construct the message in this routine we can add info about effector name, target, perhaps parameters.

core/src/test/java/org/apache/brooklyn/core/effector/ProxyEffectorTest.java
geomacy: Sorry yes made a bit of a mental leap and didn't explain it too well - so at first glance I'd thought these were entities in two separate apps; now ok they are not, but it did immediately prompt the question of what if they were?  I could imagine what I think are useful use-cases, e.g. have one app with a Management Information System of some sort that has to collect information about all the parts of your overall infrastructure. Then other apps that are created separately, perhaps much later, can still be configured to call a proxy effector pointing at the MIS in the completely different application. On the other hand allowing apps to invoke effectors on other apps' entities may be a security risk - maybe for isolation purposes it's better to forbid this?

core/src/test/java/org/apache/brooklyn/core/effector/ProxyEffectorTest.java
geomacy: Not much perhaps  - maybe more digestible failure messages. Not important really.

geomacy:
That's my comments explained a bit better.  I'm happy for this to be merged as-is though, I don't think any of the comments are blockers.   👍

core/src/main/java/org/apache/brooklyn/core/effector/ProxyEffector.java
```diff
@@ -0,0 +1,99 @@
...
+
+        private Maybe<Entity> resolveTarget() {
+            return Tasks.resolving(proxy, Entity.class)
+                    .context(entity())
+                    .swallowExceptions()
+                    .getMaybe();
+        }
+
+        private Object invokeEffectorNamed(Entity target, String effectorName, ConfigBag params) {
+            LOG.debug("{} forwarding effector invocation on {} to entity={}, effector={}, paramters={}",
```
neykov: Still present

sjcorbett:
@geomacy thanks for the thorough review. I've fixed that typo. I'll merge now.

PR #371, opened 2016-10-05
One instance of customizer at each customize step

duncangrant:
Change so that getCustomizers is only called once during provisioning.
This ensures that only a single instance of a customizer configured
with DSL will be called.  This means that state can be kept between
steps.

### DIFF
--- a/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
+++ b/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
@@ -692,6 +692,8 @@ public class JcloudsLocation extends AbstractCloudMachineProvisioningLocation im
             Set<? extends NodeMetadata> nodes;
             Template template;
             Collection<JcloudsLocationCustomizer> customizers = getCustomizers(setup);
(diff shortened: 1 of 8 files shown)


locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -409,7 +409,7 @@ protected CloudMachineNamer getCloudMachineNamer(ConfigBag config) {
         }
     }
 
-    protected Collection<JcloudsLocationCustomizer> getCustomizers(ConfigBag setup) {
+    public Collection<JcloudsLocationCustomizer> getCustomizers(ConfigBag setup) {
```
aledsage: I'm slightly uncomfortable about this becoming public - but I see why you had to do it, to let it be called by `CloudExplorerSupport.java`.

aledsage:
LGTM; I've submitted https://github.com/duncangrant/brooklyn-server/pull/1 to add some additional changes in this PR.

@duncangrant can you review that - if you merge it into your branch, then it will appear automatically in this PR.

aledsage:
LGTM; once jenkins has confirmed the tests pass, this is good to merge.

PR #372, opened 2016-10-05
Fix JcloudsByonLocationResolverStubbedTest

aledsage:
Was taking a very long time, because trying to find a reachable address


aledsage:
As a consequence of merging #355, this test is now taking 5 minutes per test method. Therefore merging now (so that our jenkins builds don't risk timing out, and so it doesn't annoy people!)

PR #373, opened 2016-10-07
Attach configurable effector to SoftwareProcess for AWS

iyovcheva:
Attaches an effector set in SoftwareProcess .BROOKLYN_LOCATION_EXTRA_HDD_EFFECTOR if the key is configured.

### DIFF
--- a/software/base/src/main/java/org/apache/brooklyn/entity/software/base/SoftwareProcessImpl.java
+++ b/software/base/src/main/java/org/apache/brooklyn/entity/software/base/SoftwareProcessImpl.java
@@ -25,11 +25,13 @@ import java.util.Map;
 import java.util.Set;
 import java.util.Timer;
 import java.util.TimerTask;
-import java.util.Date;
 import java.util.concurrent.Callable;
 import java.util.concurrent.TimeUnit;
 
+import org.apache.brooklyn.core.effector.EffectorBody;
 import org.apache.brooklyn.location.jclouds.networking.NetworkingEffectors;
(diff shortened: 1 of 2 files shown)


neykov:
This looks like a very specific use case. Can you use `AddEffector` instead?

software/base/src/main/java/org/apache/brooklyn/entity/software/base/SoftwareProcessImpl.java
```diff
@@ -268,6 +272,36 @@ protected void onUpdated() {
      * Called before driver.start; guarantees the driver will exist, and locations will have been set.
      */
     protected void preStart() {
+        if (!getLocations().isEmpty() && getLocations().iterator().next() instanceof JcloudsMachineLocation) {
+            JcloudsMachineLocation location = (JcloudsMachineLocation) getLocations().iterator().next();
+            if (location.getParent().getProvider().equals("aws-ec2")
```
aledsage: @iyovcheva I agree with @neykov - this looks like a very specific use-case. It's a code-smell when we have cloud-specific code (e.g. checking for "aws-ec2"), particularly in generic entities like this.

Also, because the `EffectorBody` class is instantiated reflectively (with classname supplied via the config key), this code really isn't doing anything specific about "extra HDD" other than wiring in a given class to be an effector with that name.

We should be able to use a "mixin". We can use `AddEffector`, which is an `EntityInitializer`, which people can optionally include.

PR #375, opened 2016-10-07
Tidy tests to always terminate ManagementContext

aledsage:
If we leave the management context running, it will not be garbage collected (because its threads reference it, so it's not just about reachability from the test class). This memory leak causes our tests to get slower and slower on the jenkins build server.

### DIFF
--- a/launcher/src/test/java/org/apache/brooklyn/launcher/CleanOrphanedLocationsIntegrationTest.java
+++ b/launcher/src/test/java/org/apache/brooklyn/launcher/CleanOrphanedLocationsIntegrationTest.java
@@ -18,6 +18,7 @@
  */
 package org.apache.brooklyn.launcher;
 
+import java.io.File;
 import java.util.Set;
 
 import org.apache.brooklyn.api.mgmt.ManagementContext;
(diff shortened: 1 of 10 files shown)


aledsage:
This fix is important because our apache jenkins builds sometimes timeout (seemingly with it consuming a lot of memory, based on the GC activity).

launcher/src/test/java/org/apache/brooklyn/launcher/CleanOrphanedLocationsIntegrationTest.java
```diff
@@ -37,54 +38,85 @@
...
+        
+        // contents of method copied from BrooklynMgmtUnitTestSupport
+        for (ManagementContext mgmt : mgmts) {
+            try {
+                if (mgmt != null) Entities.destroyAll(mgmt);
+            } catch (Throwable t) {
+                LOG.error("Caught exception in tearDown method", t);
+                // we should fail here, except almost always that masks a primary failure in the test itself,
+                // so it would be extremely unhelpful to do so. if we could check if test has not already failed,
+                // that would be ideal, but i'm not sure if that's possible with TestNG. ?
```
sjcorbett: TestNG documentation on this: http://testng.org/doc/documentation-main.html#native-dependency-injection

sjcorbett:
Looks good to me. Are there any new guidelines test writers should follow? If yes it would be helpful to record them in documentation or on the mailing list.

aledsage:
@sjcorbett thanks for review, and for link to `ITestResult`. I'll take a look at some point (but not just now).

As for test guidelines, agree it would be good to have some of those. I don't think we have much written up about that. For example, we don't document the super-classes that one can extend, or the existence of `RecordingSshTool` etc.

Merging now.

PR #379, opened 2016-10-11
OnPublicNetworkEnricher config with sensor

aledsage:
I'm not entirely sure what the best way is to wire in and use the `OnPublicNetworkEnricher`. Two of the use-cases are:

1. in an environment with port-mappings (e.g. cloud's DNAT, or docker), look up the registered port mapping in the `PortForwardManager`. This is the existing behaviour.

2. Where a sensor (e.g. URI or hostAndPort) has been published with the subnet-address, publish an equivalent sensor using the public IP (so using the same port).

It is this second use-case that this PR enables.

For "normal clouds" (i.e. without DNAT), then option 2 is an obvious way to go. However:

* what if there isn't a public address? The "host.address" sensor would normally be populated with the private ip instead, so that would be used for the public url - but is that too misleading for a value of `main.uri.mapped.public`?

* what if there really is a DNAT rule set up for accessing this VM's port? Should our `OnPublicNetworkEnricher` prefer the DNAT rule (i.e. the entry in `PortForwardManager`) if it exists, but fall back to the "host.address" if it's not?


neykov:
Nice, merging.

Build failure unrelated and local tests pass.

PR #381, opened 2016-10-12
BROOKLYN-356: workaround, by increasing timeout

aledsage:
@neykov can you take a look please?


neykov:
Looks good. Nice fix for the long running test.
Will merge after tests complete.

PR #382, opened 2016-10-13
InvokeEffectorOnCollectionSensorChange

sjcorbett:
Tracks a collection and invokes effectors when elements are added and removed.

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/entity/EntityTypeSnapshot.java
+++ b/core/src/main/java/org/apache/brooklyn/core/entity/EntityTypeSnapshot.java
@@ -64,11 +64,13 @@ public class EntityTypeSnapshot extends BrooklynTypeSnapshot implements EntityTy
 
     @Override
     public Maybe<Effector<?>> getEffectorByName(String name) {
-        for (Effector<?> contender : effectors) {
-            if (name.equals(contender.getName()))
-                return Maybe.<Effector<?>>of(contender);
+        if (name != null) {
+            for (Effector<?> contender : effectors) {
+                if (name.equals(contender.getName()))
+                    return Maybe.<Effector<?>>of(contender);
+            }
         }
-        return Maybe.<Effector<?>>absent("No effector matching '"+name+"'");        
+        return Maybe.<Effector<?>>absent("No effector matching '" + name + "'");
     }
     
     @Override
--- a/core/src/main/java/org/apache/brooklyn/core/sensor/Sensors.java
+++ b/core/src/main/java/org/apache/brooklyn/core/sensor/Sensors.java
@@ -84,6 +84,10 @@ public class Sensors {
         return new BasicAttributeSensor<T>(type, name);
     }
 
+    public static <T> AttributeSensor<T> newSensor(TypeToken<T> type, String name) {
+        return new BasicAttributeSensor<T>(type, name);
+    }
+
     public static <T> AttributeSensor<T> newSensor(Class<T> type, String name, String description) {
         return new BasicAttributeSensor<T>(type, name, description);
     }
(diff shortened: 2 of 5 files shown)


core/src/main/java/org/apache/brooklyn/policy/InvokeEffectorOnSetChange.java
```diff
@@ -0,0 +1,203 @@
...
+    /** The previous version of the set against which events will be compared. */
+    private Set<Object> previous = Collections.emptySet();
+
+    /** Guards accesses of previous. */
+    private final Object[] updateLock = new Object[0];
+
+    @Override
+    public void setEntity(EntityLocal entity) {
+        super.setEntity(entity);
+        Sensor<? extends Collection<?>> sensor =
```
neykov: Worth a comment that the method will be executed on rebind. In case of adding additional validation should be backwards compatible with existing persistence state.

core/src/main/java/org/apache/brooklyn/policy/InvokeEffectorOnSetChange.java
```diff
@@ -0,0 +1,203 @@
...
+    }
+
+    @Override
+    public void onEvent(SensorEvent<Collection<?>> event) {
+        final Set<Object> added = new HashSet<>(), removed = new HashSet<>();
+        Set<Object> newValue = new HashSet<>(event.getValue());
+        synchronized (updateLock) {
+            // Not using .immutableCopy() in case either set contains `null`.
+            Sets.difference(newValue, previous).copyInto(added);
+            Sets.difference(previous, newValue).copyInto(removed);
```
sjcorbett: We have to take the difference with the lock so that we can determine whether the collection is changed and thus should set the `previous` sensor. I'll move the `onAdded` and `onRemoved` calls out of the critical section.

core/src/main/java/org/apache/brooklyn/policy/InvokeEffectorOnSetChange.java
```diff
@@ -0,0 +1,203 @@
...
+                for (Object o : added) {
+                    onAdded(o);
+                }
+                for (Object o : removed) {
+                    onRemoved(o);
+                }
+                entity.sensors().set(getPreviousSensor(), previous);
+                requestPersist();
+                // In advance of the next event.
+                previous = Collections.unmodifiableSet(newValue);
```
sjcorbett: Just to be clear - do you mean to ask whether we really need a separate _field_? Rather than sensor. Assuming so, yes, the previous sensor is just for users to look at. I thought we'd decided that was desirable previously. I don't have any strong feelings about the `previous` sensor - am happy enough to get rid of it but can see that it might be useful.

sjcorbett:
@neykov thank you for your comments. I've addressed everything I didn't comment on. I've also renamed the policy to `InvokeEffectorOnCollectionSensorChange` to be consistent with the existing `InvokeEffectorOnSensorChange`, but no strong feelings there really.

core/src/main/java/org/apache/brooklyn/policy/InvokeEffectorOnSetChange.java
```diff
@@ -0,0 +1,203 @@
...
+        synchronized (updateLock) {
+            previous = (current != null) ? new HashSet<>(current) : Collections.emptySet();
+        }
+        subscriptions().subscribe(entity, sensor, this);
+    }
+
+    @Override
+    public void onEvent(SensorEvent<Collection<?>> event) {
+        final Set<Object> added = new HashSet<>(), removed = new HashSet<>();
+        Set<Object> newValue = new HashSet<>(event.getValue());
```
sjcorbett: Is there a way to get GitHub to hide out of date diffs now?

core/src/main/java/org/apache/brooklyn/policy/InvokeEffectorOnSetChange.java
sjcorbett: Though to be honest since effectors are asynchronous this will just be a problem further down the line, won't it?

sjcorbett:
I will merge this.

PR #383, opened 2016-10-13
Remove lock on DSL objects - they are stateless

neykov:
The same DSL object could be used in mutliple entities created from the same spec, so one entity resolving would block the rest even if they already have the values. The objects are stateless so need for a lock.


aledsage:
LGTM; merging.

aledsage:
This fixes https://issues.apache.org/jira/browse/BROOKLYN-357

PR #384, opened 2016-10-14
Fix calculation of currently used memory by the JVM

neykov:
Before the change I was getting readings like `brooklyn gc (after) - using 2.59 GB / 2.86 GB (1268 MB real) memory with `-DXmx=3G`. `2.86` is `maxMemory` same as `Xmx`. `1268` is `totalMemory` which is the currently allocated memory by the JVM from the operating system. The used memory can't be more than the `totalMemory`.

Confirmed the new readings by comparing to jconsole.

The PR reverts to the previous version of the expression (cc @ahgittin).


sjcorbett:
Looks good. Will merge.

PR #386, opened 2016-10-18
Add AbstractInvokeEffectorPolicy

sjcorbett:
Supports tracking the number of effector tasks that are ongoing and publishing that as an "is busy" sensor on the entity the policy is attached to. Tracking is enabled by providing a value for `isBusySensor` when configuring the policy.

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/policy/AbstractInvokeEffectorPolicy.java
+++ b/core/src/main/java/org/apache/brooklyn/policy/AbstractInvokeEffectorPolicy.java
@@ -1,0 +1,145 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.brooklyn.policy;
+
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicInteger;
+import javax.annotation.Nonnull;
+
+import org.apache.brooklyn.api.effector.Effector;
+import org.apache.brooklyn.api.entity.EntityLocal;
+import org.apache.brooklyn.api.mgmt.Task;
+import org.apache.brooklyn.api.sensor.AttributeSensor;
+import org.apache.brooklyn.config.ConfigKey;
+import org.apache.brooklyn.core.config.ConfigKeys;
+import org.apache.brooklyn.core.policy.AbstractPolicy;
+import org.apache.brooklyn.core.sensor.Sensors;
+import org.apache.brooklyn.util.guava.Maybe;
+import org.apache.brooklyn.util.text.Strings;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.common.base.Objects;
+import com.google.common.util.concurrent.MoreExecutors;
+
+public abstract class AbstractInvokeEffectorPolicy extends AbstractPolicy {
+
+    private static final Logger LOG = LoggerFactory.getLogger(AbstractInvokeEffectorPolicy.class);
+
+    public static final ConfigKey<String> IS_BUSY_SENSOR_NAME = ConfigKeys.newStringConfigKey(
+            "isBusySensor",
+            "Name of the sensor to publish on the entity that indicates that this policy has incomplete effectors. " +
+                    "If unset running tasks will not be tracked.");
+
+    private final AtomicInteger taskCounter = new AtomicInteger();
+
+    /**
+     * Indicates that onEvent was notified of an value of is not the latest sensor value.
+     */
+    private boolean moreUpdatesComing;
+    /**
+     * The timestamp of the event that informed moreUpdatesComing. Subsequent notifications
+     * of earlier events will not cause updates of moreUpdatesComing.
+     */
+    private long mostRecentUpdate = 0;
+    /**
+     * Guards {@link #moreUpdatesComing} and {@link #mostRecentUpdate}.
+     */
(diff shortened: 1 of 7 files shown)


core/src/main/java/org/apache/brooklyn/policy/AbstractInvokeEffectorPolicy.java
```diff
@@ -0,0 +1,120 @@
...
+            "isBusySensor",
+            "Name of the sensor to publish on the entity that indicates that this policy has incomplete effectors. " +
+                    "If unset running tasks will not be tracked.");
+
+    @Override
+    public void setEntity(EntityLocal entity) {
+        super.setEntity(entity);
+        if (isBusySensorEnabled()) {
+            // Overwrites persisted values and republishes when the entity rebinds.
+            config().set(TASK_COUNTER, new AtomicInteger());
```
aledsage: Should we be storing this as config? On rebind, our effector will not resume so the counter should be reset. However, because we've persisted it as a config key, then it will keep its old value.

Maybe it's better for now to just have this as a field (without `@SetFromFlag` obviously).

What should we do on rebind when we were executing the effector? Currently it will keep the count of 1 and publish busy "true". If we changed to a field, then it would publish "false".

aledsage:
Looks good. As well as the task-counter config comment, my only other comment is... should we add to the tests for `InvokeEffectorOnSensorChange` and `InvokeEffectorOnCollectionSensorChange`, that the busy-sensor is set?

Except that highlights the absence of tests for `InvokeEffectorOnSensorChange`.

I think this PR is fine without adding to the tests for `InvokeEffectorOnCollectionSensorChangeTest`, and should not be blocked adding tests for `InvokeEffectorOnSensorChange` (but we should add tests for that separately!).

sjcorbett:
@aledsage thanks for your comments. I've replaced `TASK_COUNTER` with a field and have added a test for the sensor to `InvokeEffectorOnCollectionSensorChangeTest`.

core/src/main/java/org/apache/brooklyn/policy/InvokeEffectorOnCollectionSensorChange.java
```diff
@@ -128,10 +132,25 @@ public void setEntity(EntityLocal entity) {
...
+        if (isBusySensorEnabled()) {
+            // There are more events coming that this policy hasn't been notified of if the
+            // value received in the event does not match the current value of the sensor.
+            final Collection<?> sensorVal = entity.sensors().get(getTriggerSensor());
+            final Set<Object> sensorValSet = sensorVal != null
+                ? new LinkedHashSet<>(sensorVal)
+                : ImmutableSet.of();
+            final boolean moreComing = !Objects.equals(newValue, sensorValSet);
+            moreUpdatesComing.set(moreComing);
+        }
```
neykov: Why not do this in the parent class? It would be useful in "InvokeEffectorOnSensorChange" as well.

neykov:
Good to merge as is but will wait for response on above comments.

core/src/main/java/org/apache/brooklyn/policy/AbstractInvokeEffectorPolicy.java
```diff
@@ -0,0 +1,110 @@
...
+                    "If unset running tasks will not be tracked.");
+
+    private final AtomicInteger taskCounter = new AtomicInteger();
+
+    @Override
+    public void setEntity(EntityLocal entity) {
+        super.setEntity(entity);
+        if (isBusySensorEnabled()) {
+            // Republishes when the entity rebinds.
+            publishIsBusy();
```
sjcorbett: It relies on `taskCounter` _not_ being persisted. When Brooklyn rebinds `taskCounter` is a fresh instance whose count is zero so `isBusy` will be `false`.

core/src/main/java/org/apache/brooklyn/policy/InvokeEffectorOnCollectionSensorChange.java
sjcorbett: Good point.

core/src/main/java/org/apache/brooklyn/policy/AbstractInvokeEffectorPolicy.java
geomacy: ah, thanks @sjcorbett that's great

sjcorbett:
@neykov I've pushed some updates that extend the 'more updates coming' notion to `InvokeEffectorOnSensorChange`. I'd appreciate you casting your eye over `AbstractInvokeEffectorPolicy` once more.

core/src/main/java/org/apache/brooklyn/policy/AbstractInvokeEffectorPolicy.java
```diff
@@ -0,0 +1,145 @@
...
+        }
+        Task<T> task = entity.invoke(effector, parameters);
+        if (isBusySensorEnabled()) {
+            task.addListener(new EffectorListener(), MoreExecutors.sameThreadExecutor());
+        }
+        return task;
+    }
+
+    protected boolean isBusy() {
+        synchronized (moreUpdatesLock) {
```
neykov: No need to lock here - `taskCounter` & `moreUpdatesComing` are independent.

core/src/main/java/org/apache/brooklyn/policy/InvokeEffectorOnSensorChange.java
```diff
@@ -40,28 +44,56 @@
...
+        final AttributeSensor<Object> sensor;
+        if (configVal == null) {
+            throw new NullPointerException("Value for " + SENSOR.getName() + " is null");
+        } else if (configVal instanceof String) {
+            sensor = Sensors.newSensor(Object.class, (String) configVal);
+        } else if (configVal instanceof AttributeSensor) {
+            sensor = (AttributeSensor<Object>) configVal;
+        } else {
+            sensor = TypeCoercions.tryCoerce(configVal, new TypeToken<AttributeSensor<Object>>() {}).get();
+        }
```
neykov: No need for `if`s, could replace it all with `coerce`.

core/src/main/java/org/apache/brooklyn/policy/AbstractInvokeEffectorPolicy.java
sjcorbett: I included the synchronisation to make sure that `moreUpdatesComing` is visible to whatever thread invokes the effector's completion listener (which calls `publishIsBusy` which calls `isBusy`). At a minimum the field should be volatile then. Is it a big one for you?

neykov: Good point - good as is.

neykov:
Thanks @sjcorbett, merging.

PR #388, opened 2016-10-19
Add a stop latch to SoftwareProcess.

geomacy:
Adds the ability to delay stop of a software process until an appropriate sensor is published on some entity that depends on that process.

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/entity/BrooklynConfigKeys.java
+++ b/core/src/main/java/org/apache/brooklyn/core/entity/BrooklynConfigKeys.java
@@ -23,6 +23,7 @@ import static org.apache.brooklyn.core.config.ConfigKeys.newConfigKey;
 import static org.apache.brooklyn.core.config.ConfigKeys.newConfigKeyWithPrefix;
 import static org.apache.brooklyn.core.config.ConfigKeys.newStringConfigKey;
 
+import com.google.common.annotations.Beta;
 import org.apache.brooklyn.api.location.Location;
 import org.apache.brooklyn.config.ConfigKey;
 import org.apache.brooklyn.core.config.BasicConfigInheritance;
@@ -179,6 +180,10 @@ public class BrooklynConfigKeys {
 
     public static final ConfigKey<Boolean> PROVISION_LATCH = newBooleanConfigKey("provision.latch", "Latch for blocking location provision until ready");
     public static final ConfigKey<Boolean> START_LATCH = newBooleanConfigKey("start.latch", "Latch for blocking start until ready");
+
+    @Beta // on stop DSLs time out after a minute and unblock; may be easier to fix after https://github.com/apache/brooklyn-server/pull/390
+    public static final ConfigKey<Boolean> STOP_LATCH = newBooleanConfigKey("stop.latch", "Latch for blocking stop until a condition is met; will block for at most 1 minute and then time out");
+
     public static final ConfigKey<Boolean> SETUP_LATCH = newBooleanConfigKey("setup.latch", "Latch for blocking setup until ready");
     public static final ConfigKey<Boolean> PRE_INSTALL_RESOURCES_LATCH = newBooleanConfigKey("resources.preInstall.latch", "Latch for blocking pre-install resources until ready");
     public static final ConfigKey<Boolean> INSTALL_RESOURCES_LATCH = newBooleanConfigKey("resources.install.latch", "Latch for blocking install resources until ready");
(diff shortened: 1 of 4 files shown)


neykov:
Changes look good on first pass. One thing that comes to mind is that on stop DSLs time out after a minute and unblock. Does this rule apply here as well, can we do something to avoid it?

geomacy:
will have a think about that and see if I can add test conditions to cover that case

justinThompson:
So does the stop latch just wait for the presence of the `stop.now` sensor?
Does it care what the value is?

geomacy:
@justinThompson that's right, it just cares if the sensor has been published or not, following the same pattern as the start latch.  This is probably good enough unless we can think of a use case we want to  support where something might want to publish the sensor with a value of `false`.

software/base/src/test/java/org/apache/brooklyn/entity/software/base/SoftwareProcessEntityLatchTest.java
```diff
@@ -116,14 +118,36 @@ public void testLaunchLatchBlocks() throws Exception {
...
+        final Task<Void> startTask = Entities.invokeEffector(app, app, MyService.START, ImmutableMap.of("locations", ImmutableList.of(loc)));
+        ((EntityLocal)triggerEntity).sensors().set(Attributes.SERVICE_UP, true);
+        startTask.get(Duration.THIRTY_SECONDS);
+
+        final Task<Void> stopTask = Entities.invokeEffector(app, app, MyService.STOP);
+
+        assertEffectorBlockingDetailsEventually(entity, MyService.STOP.getName(), "Waiting for config " + SoftwareProcess.STOP_LATCH.getName());
+
+        app.sensors().set(stopper, true);
+        stopTask.get(Duration.THIRTY_SECONDS);
```
aledsage: Very minor: I have a personal preference to use `Asserts.DEFAULT_LONG_TIMEOUT` rather than hard-coding 30 seconds here (it comes to the same thing, because that default timeout is 30 seconds).

aledsage:
@geomacy Looks good to me. Feel free to completely ignore my trivial comment - it's not worth the effort of changing it here ;-)

What is the situation with @neykov's comment "on stop DSLs time out after a minute and unblock". Any thoughts on that?

geomacy:
@aledsage yes I tested this yesterday, and sure enough it bites us - I commented out the line 136 above, and increased the timeout to a large value, and after a minute the test fails with an exception
```
Caused by: org.apache.brooklyn.util.exceptions.RuntimeTimeoutException: Unsatisfied after 1m
    at org.apache.brooklyn.core.sensor.DependentConfiguration$WaitInTaskForAttributeReady.call(DependentConfiguration.java:318)
```
This will need a bit of thought about how best to handle it - what is the _desired_ behaviour here? Do we want it to block indefinitely in this case, or for a longer timeout than one minute, or for a timeout that we can specify in the YAML?.. And how can we achieve this anyway?  @neykov points out the very relevant discussion earlier - http://mail-archives.apache.org/mod_mbox/brooklyn-dev/201608.mbox/%3c57AD01EC.5010400@CloudsoftCorp.com%3e.

core/src/main/java/org/apache/brooklyn/core/entity/BrooklynConfigKeys.java
```diff
@@ -179,6 +179,7 @@
 
     public static final ConfigKey<Boolean> PROVISION_LATCH = newBooleanConfigKey("provision.latch", "Latch for blocking location provision until ready");
     public static final ConfigKey<Boolean> START_LATCH = newBooleanConfigKey("start.latch", "Latch for blocking start until ready");
+    public static final ConfigKey<Boolean> STOP_LATCH = newBooleanConfigKey("stop.latch", "Latch for blocking stop until a condition is met");
```
geomacy: 👍 good idea

neykov:
Good to merge after addressing above and Aled's comments.

geomacy:
Comments addressed in 0559ff0.

neykov:
Thanks @geomacy, merging.

geomacy:
Changes squashed and re-pushed.

neykov:
@geomacy Can you close - looks like I didn't use the correct keyword in the message.

geomacy:
@neykov surely this isn't merged yet, can we merge it first?

neykov:
Turns out I merged it but not in the master branch. Thanks for spotting this.

geomacy:
thanks @neykov

PR #389, opened 2016-10-19
Break version range dependency on jackson-databind

neykov:
`com.maxmind.db:maxmind-db:jar:1.2.1` depends on `jackson-databind [2.7.0,)`
which causes excessive download of dependencies (all pom versions in the range,
including snapshots). We already include it explicitly anyway so can save on
the extra http requests. It's also forces maven to check frequently whether
new versions are available so it can use them.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -145,6 +145,19 @@
         <dependency>
             <groupId>com.maxmind.geoip2</groupId>
             <artifactId>geoip2</artifactId>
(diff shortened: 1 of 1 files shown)


aledsage:
Good spot, LGTM. It's taught me a new thing about versions in maven pom dependencies! 
http://stackoverflow.com/a/1172371/1393883

Merging now.

PR #390, opened 2016-10-20
BROOKLYN-356: fix race in transformer by adding getImmediately

aledsage:


core/src/main/java/org/apache/brooklyn/util/core/task/ValueResolver.java
```diff
@@ -300,6 +313,18 @@ public T get() {
             return Maybe.of((T) v);
         
         try {
+            if (immediately && v instanceof ImmediateSupplier) {
+                final ImmediateSupplier<?> supplier = (ImmediateSupplier<?>) v;
```
geomacy: Does the comment at line 383 need updated now that this is added?

core/src/test/java/org/apache/brooklyn/core/entity/EntitySubscriptionTest.java
```diff
@@ -280,4 +279,55 @@ public void testSubscriptionForInitialValueWhenNotValid() {
...
+        // as should subsequent events
+        observedEntity.sensors().set(TestEntity.NAME, "myval2");
+        assertListenerCalledOnceWithContextEntityEventually(listener, entity);
+        listener.clearEvents();
+
+        // same for subscribing to children: context should be the subscriber
+        entity.subscriptions().subscribeToChildren(observedEntity, TestEntity.SEQUENCE, listener);
+        observedChildEntity.sensors().set(TestEntity.SEQUENCE, 123);
+        assertListenerCalledOnceWithContextEntityEventually(listener, entity);
+        listener.clearEvents();
```
geomacy: should this be in a `tearDown`?

core/src/test/java/org/apache/brooklyn/util/core/task/ValueResolverTest.java
```diff
@@ -41,6 +49,125 @@ public void setUp() throws Exception {
...
+        Maybe<String> result = Tasks.resolving(newSleepTask(Duration.TEN_SECONDS, "foo")).as(String.class).context(app).timeout(Duration.ZERO).getMaybe();
+        Assert.assertFalse(result.isPresent());
+    }
+    
+    public void testTimeoutBig() {
+        Maybe<String> result = Tasks.resolving(newSleepTask(Duration.ZERO, "foo")).as(String.class).context(app).timeout(Duration.TEN_SECONDS).getMaybe();
+        Assert.assertEquals(result.get(), "foo");
+    }
+
+    public void testNoExecutionContextOnCompleted() {
```
geomacy: Don't follow what this is testing - how is it testing something about 'no execution context'?

camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/methods/DslComponent.java
```diff
@@ -353,6 +386,33 @@ public DslSensorSupplier(DslComponent component, Object sensorIndicator) {
...
+            if (si instanceof Sensor) {
+                return Maybe.<Sensor<?>>of((Sensor<?>)si);
+            } else if (si instanceof String) {
+                Entity targetEntity = component.get();
+                Sensor<?> result = null;
+                if (targetEntity!=null) {
+                    result = targetEntity.getEntityType().getSensor((String)si);
+                }
+                if (result!=null) return Maybe.<Sensor<?>>of(result);
+                return Maybe.<Sensor<?>>of(Sensors.newSensor(Object.class, (String)si));
```
aledsage: Not sure I follow. In `attributeWhenReady`, if we're given a sensor name then we will instantiate a new sensor object if it's not defined on the entity type; and then we look up the sensor's value (returning Maybe.absent if it's null). Here, we just want it to return the sensor object (rather than a sensor value) so we do the first half: instantiate a new sensor object if it's not defined on the entity type.

This is consistent with the blocking `get()` method. We return `Maybe.absent()` if the `get()` method would have blocked, and otherwise we return the same value as `get()` would have returned.

camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/dsl/DslTest.java
```diff
@@ -118,39 +160,64 @@ public void runConcurrentWorker(Supplier<Runnable> taskSupplier) {
...
+            return this;
+        }
+        
         @Override
         public void run() {
-            TestEntity entity = parent.createAndManageChild(childSpec);
-            for (int i = 0; i < RESOLVER_ITERATIONS; i++) {
+            TestEntity entity = parent.addChild(childSpec);
+            for (int i = 0; i < resolverIterations; i++) {
                 preResolve(entity);
```
aledsage: I'll rename this from "eventually" to "satisfiedAsynchronously".

What "eventually" meant was that the condition for getting the value immediately is not yet satisfied; it will become satisfied asynchronously. For example, when testing `attributeWhenReady` then another thread sets the sensor's value after we have called `execDslEventually` - we expect that method to block.

I'll also change the assertions so that we confirm we get a Maybe or the real result when calling getImmediately in this situation.

aledsage:
Thanks @geomacy for the great comments - I've addressed those in an additional commit, and rebased against master.

Once jenkins confirms the tests pass, I'll merge. If you do get a chance to review the changes in the last commit before that, it would be great though.

neykov:
@aledsage I'm currently reviewing the PR, please hold on with the merge.

camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/methods/BrooklynDslCommon.java
```diff
@@ -525,15 +583,20 @@ public DslExternal(String providerName, String key) {
         }
 
         @Override
+        public final Maybe<Object> getImmediately() {
+            ManagementContextInternal managementContext = DslExternal.managementContext();
+            return Maybe.<Object>of(managementContext.getExternalConfigProviderRegistry().getConfig(providerName, key));
```
neykov: This is not reusable since `getConfig` will block indefinitely, resolving the value.

core/src/test/java/org/apache/brooklyn/util/core/task/ValueResolverTest.java
```diff
@@ -69,46 +59,65 @@ public void testTimeoutBig() {
...
-            Assert.fail("should have thrown");
-            return null;
-        } catch (Exception e) { return e; }
+    public void testUnsubmittedTaskWhenNoExecutionContextFails() {
+        // ValueResolver.getMaybe() is called with no execution context. Therefore it will not execute the task.
+        Task<String> t = newSleepTask(Duration.ZERO, "foo");
+        Maybe<String> result = Tasks.resolving(t).as(String.class).timeout(Duration.ZERO).getMaybe();
+        
+        Assert.assertTrue(result.isAbsent(), "result="+result);
+        Exception exception = ((Maybe.Absent<?>)result).getException();
```
neykov: Can use `Maybe.getException(result)`.

neykov:
There's pretty big overlap between the `getImmediate` & `newTask` implementations - would be nice to reuse as much as possible, but I'm fine merging those as is.

core/src/test/java/org/apache/brooklyn/util/core/task/ValueResolverTest.java
```diff
@@ -69,46 +59,65 @@ public void testTimeoutBig() {
         Assert.assertEquals(result.get(), "foo");
     }
 
-    public void testNoExecutionContextOnCompleted() {
+    public void testCompletedTaskReturnsResultImmediately() {
+        // Call ValueResolver.getMaybe() from this thread, which has no execution context.
```
geomacy: I still struggled with understanding what was meant here until I realised that this line 63 refers to the `Tasks.resolving(t)` being executed in this thread (I'd been put off by concentrating on the line `app.getExecutionContext()...`).  It would maybe make this clearer if these lines of comment were moved to just above the `Tasks.resolving(t)` on 68 below.

core/src/test/java/org/apache/brooklyn/util/core/task/ValueResolverTest.java
```diff
@@ -124,4 +133,141 @@ public void testDefaultBeforeDelayAndError() {
...
+        CallInfo callInfo = Tasks.resolving(supplier).as(CallInfo.class).context(app).immediately(true).get();
+        assertNull(callInfo.task);
+        assertContainsCallingMethod(callInfo.stackTrace, "testGetImmediately");
+    }
+    
+    public void testGetImmediateSupplierWithTimeoutUsesBlocking() {
+        MyImmediateAndDeferredSupplier supplier = new MyImmediateAndDeferredSupplier();
+        CallInfo callInfo = Tasks.resolving(supplier).as(CallInfo.class).context(app).timeout(Asserts.DEFAULT_LONG_TIMEOUT).get();
+        assertNotNull(callInfo.task);
+        assertNotContainsCallingMethod(callInfo.stackTrace, "testGetImmediately");
```
geomacy: The test is called `testGetImmediateXXXX` not `testGetImmediatelyXXXX`.  In any case it might be worth supplying the full method name here to avoid possible confusion with line 140. Same with line 170 below.

geomacy:
Couple more comments but otherwise looks good along with Svet's comments.  👍

camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/methods/DslComponent.java
```diff
@@ -303,6 +330,12 @@ public DslConfigSupplier(DslComponent component, String keyName) {
         }
 
         @Override
+        public final Maybe<Object> getImmediately() {
+            Entity targetEntity = component.get();
+            return Maybe.of(targetEntity.getConfig(ConfigKeys.newConfigKey(Object.class, keyName)));
```
aledsage: Great spot! I'm writing a bunch more tests for this.

camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/methods/BrooklynDslCommon.java
aledsage: @neykov This is an interesting one! The `managementContext.getExternalConfigProviderRegistry().getConfig(...)` is calling a different getConfig whose javadoc simply says:

```
    /**
     * Searches the named {@link ExternalConfigSupplier} for the config value associated with the specified key.
     * Quietly returns <code>null</code> if no config exists for the specified key.
     * Throws {@link IllegalArgumentException} if no {@link ExternalConfigSupplier} exists for the passed name.
     */
```

The existing implementation delegates to `ExternalConfigSupplier.get(key)`.

There is no guidance on that whether the call could be blocking, or should return reasonably promptly.

I'd assume it would call out to things like Vault, and would return the value that currently exists or null. Such calls will not be super-fast, but should not be blocked waiting for other entities either.

I therefore think this is good enough, to just call getConfig without wrapping it in another task for timeout-purposes.

If it's not good enough, then we should really change `DslExternal` back so that it doesn't even try to support `getImmediately()`. Or we could try to change the interface of `ExternalConfigSupplierRegistry` etc.

aledsage:
@neykov @geomacy latest comments addresses in the last two commits. Can you take another look please?

camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/methods/BrooklynDslCommon.java
neykov: Agree - it's fine calling this in `getImmediately`.

camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/methods/DslComponent.java
```diff
@@ -340,9 +360,9 @@ public DslConfigSupplier(DslComponent component, String keyName) {
         public final Maybe<Object> getImmediately() {
             Maybe<Entity> targetEntityMaybe = component.getImmediately();
             if (targetEntityMaybe.isAbsent()) return Maybe.absent("Target entity not available");
-            Entity targetEntity = targetEntityMaybe.get();
+            EntityInternal targetEntity = (EntityInternal) targetEntityMaybe.get();
             
-            return Maybe.of(targetEntity.getConfig(ConfigKeys.newConfigKey(Object.class, keyName)));
+            return targetEntity.config().getNonBlocking(ConfigKeys.newConfigKey(Object.class, keyName));
```
neykov: Nice, forgot we have this method. Now take a guess at how it's implemented :)
The [simple](https://github.com/apache/brooklyn-server/blob/master/core/src/main/java/org/apache/brooklyn/core/objs/AbstractConfigurationSupportInternal.java#L133) case looks easy enough to update. The [structure key](https://github.com/apache/brooklyn-server/blob/master/core/src/main/java/org/apache/brooklyn/core/objs/AbstractConfigurationSupportInternal.java#L91) case looks like a rabbit hole though (with an all too familiar comment).
I'm fine with leaving the structured key case as is for this PR, may be log a warning.

aledsage: Yes, I'd updated `AbstractConfigurationSupportInternal` line 142 to use `.immediately(true)`, so I hope that covers the case of `getNonBlockingResolvingSimple`. I didn't want to touch the "structure key" code. We should probably change that at some point to go through ValueResolver, even if that still ends up scheduling it in a task.

I therefore think this comment has been addressed. Am I missing something?

neykov: Somehow missed that. All good then.

neykov:
LGTM, merging.

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/resources/AbstractBrooklynRestResource.java
```diff
@@ -143,11 +145,12 @@ public Object resolve() {
         
         private static Object UNRESOLVED = "UNRESOLVED".toCharArray();
         
-        private static Object getImmediateValue(Object value, @Nullable Entity context, @Nullable Duration timeout) {
+        private static Object getImmediateValue(Object value, @Nullable Entity context, @Nullable Boolean immediately, @Nullable Duration timeout) {
             return Tasks.resolving(value)
                     .as(Object.class)
                     .defaultValue(UNRESOLVED)
                     .timeout(timeout)
```
geomacy: What happens with a timeout _and_ an immediately? Should this be `.timeout( immediately ? 0 : timeout)`? (with adjustments for Boolean)

geomacy:
One minor comment in AbstractBrooklynRestResource.java but overall LGTM 👍 . I had a look at the code @neykov referred to and I suppose there is more to do as noted in the comment in `getNonBlockingResolvingStructuredKey` but I agree that that should not be part of this change.

PR #387, opened 2016-10-19
Convert jclouds yaml tests to unit tests

aledsage:

PR #391, opened 2016-10-20
Enable BasicExecutionManager logging for specific exceptions

drigodwin:
A NullPointerException in ServiceReplacer was being swallowed in BasicExecutionManager, this makes exemptions for logging specific types of Exceptions.

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/util/core/task/BasicExecutionManager.java
+++ b/core/src/main/java/org/apache/brooklyn/util/core/task/BasicExecutionManager.java
@@ -536,7 +536,11 @@ public class BasicExecutionManager implements ExecutionManager {
                         // debug only here, because most submitters will handle failures
                         if (error instanceof InterruptedException || error instanceof RuntimeInterruptedException) {
                             log.debug("Detected interruption on task "+task+" (rethrowing)" +
-                                (Strings.isNonBlank(error.getMessage()) ? ": "+error.getMessage() : ""));
+                                    (Strings.isNonBlank(error.getMessage()) ? ": "+error.getMessage() : ""));
+                        } else if (error instanceof NullPointerException ||
+                                error instanceof IndexOutOfBoundsException ||
+                                error instanceof ClassCastException) {
+                            log.debug("Exception running task "+task+" (rethrowing): "+error, error);
                         } else {
                             log.debug("Exception running task "+task+" (rethrowing): "+error);
                         }


aledsage:
@drigodwin the jenkins build failed because it hung while executing `org.apache.brooklyn.camp.brooklyn.DslAndRebindYamlTest.testDslAttributeWhenReady()`. 

I believe I fixed that in https://github.com/apache/brooklyn-server/pull/387, in this commit:

```
Fix cancelling of BrooklynDslDeferredSupplier
DslAndRebindYamlTest was hanging - we were waiting for all tasks of that
entity to complete. However, the ValueResolver had kicked off a sub-task
which was then left running.
```
Unfortunately that PR is not merged yet. I'll take a look at https://github.com/apache/brooklyn-server/pull/387, for what it takes to get that merged.

core/src/main/java/org/apache/brooklyn/util/core/task/BasicExecutionManager.java
```diff
@@ -536,7 +536,11 @@ public T call() {
                         // debug only here, because most submitters will handle failures
                         if (error instanceof InterruptedException || error instanceof RuntimeInterruptedException) {
                             log.debug("Detected interruption on task "+task+" (rethrowing)" +
-                                (Strings.isNonBlank(error.getMessage()) ? ": "+error.getMessage() : ""));
+                                    (Strings.isNonBlank(error.getMessage()) ? ": "+error.getMessage() : ""));
+                        } else if (error instanceof NullPointerException ||
```
aledsage: I agree with this, even though the code looks surprising! Here's the reasoning...

We really do not want to swallow the stacktrace of exceptions. Doing so can make it extremely hard to investigate/fix, particularly if that problem is reported by a user who hit it non-deterministically (rather than by a Brooklyn developer during normal dev work). On the other hand, we don't want to log exceptions repeatedly (particularly if that is from something like an SshFeed that fails to reach a VM every 1 second).

At this level in the code, we expect the common-case to be for the exception to be logged by the person who invoked the task (they decide what level to log it at, etc). But for certain errors that should never ever happen in a production use-case, it's better to risk logging it repeatedly than never logging it at all (for example, a `NullPointerException`, etc). Such errors suggest a configuration or programming error.

The options considered were:
* always log these special cases, accepting the duplication (done here)
* always log all exceptions at debug (risking filling the log with repeated problems, such as polling a dead VM every second)
* leave as-was (but then the example of the underlying configuration issue causing the NullPointerException would be really hard (or impossible for many) to fix!)
* do something more complicated to guarantee the exception gets retrieved or logged. For example, when the task is finalised, if no-one has called get() on the task in such a way as would return the exception, then log it. But then the log message might have a completely different time, so that sounds like a bad idea!

neykov:
Tests passing locally, merging.

PR #393, opened 2016-10-24
Lifecycle fixes

sjcorbett:
Fixes a handful of surprising behaviours in AbstractApplication, DynamicGroup and BasicStartable.

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/entity/stock/BasicStartableImpl.java
+++ b/core/src/main/java/org/apache/brooklyn/entity/stock/BasicStartableImpl.java
@@ -38,9 +38,11 @@ import org.apache.brooklyn.core.entity.Entities;
 import org.apache.brooklyn.core.entity.EntityPredicates;
 import org.apache.brooklyn.core.entity.lifecycle.Lifecycle;
 import org.apache.brooklyn.core.entity.lifecycle.ServiceStateLogic;
+import org.apache.brooklyn.core.entity.lifecycle.ServiceStateLogic.ServiceProblemsLogic;
 import org.apache.brooklyn.core.entity.trait.Startable;
 import org.apache.brooklyn.core.entity.trait.StartableMethods;
 import org.apache.brooklyn.core.location.Locations;
+import org.apache.brooklyn.util.collections.QuorumCheck;
 import org.apache.brooklyn.util.exceptions.Exceptions;
 
 public class BasicStartableImpl extends AbstractEntity implements BasicStartable {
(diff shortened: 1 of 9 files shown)


sjcorbett:
One possibly surprising new behaviour from 1e44522 involves the status of a `BasicStartable` when its children restart. I'll use this blueprint as an example:
```yaml
location: localhost
services:
- type: org.apache.brooklyn.entity.stock.BasicStartableImpl
  brooklyn.children:
    brooklyn.config:
      postLaunchCommand: "exit 1"
      startTimeout: 2s
```
Previously, once the ESP entity has failed once the `BasicStartable` is _always_ on fire, even if `postLaunchCommand` were modified to something sensible. With the changes in this pull request when the ESP restarts the BasicStartable's state transitions to RUNNING, then back to `ON_FIRE` when the post-launch command fails for the second time. I discussed this with @justin.thompson. We felt that though the new behaviour isn't entirely satisfactory it is less wrong than before.

core/src/main/java/org/apache/brooklyn/entity/stock/BasicStartableImpl.java
```diff
@@ -51,6 +53,7 @@
     public void start(Collection<? extends Location> locations) {
         try {
             ServiceStateLogic.setExpectedState(this, Lifecycle.STARTING);
+            ServiceProblemsLogic.clearProblemsIndicator(this, START);
```
neykov: Any specific reason to add this? Don't see it being set elsewhere.

software/base/src/main/java/org/apache/brooklyn/entity/software/base/AbstractSoftwareProcessDriver.java
```diff
@@ -312,16 +312,36 @@ public void run() {
...
+                }
             }});
             DynamicTasks.queue("post-launch", new Runnable() { public void run() {
-                postLaunch();
+                try {
+                    postLaunch();
+                } catch (Exception e) {
+                    ServiceStateLogic.setExpectedState(getEntity(), Lifecycle.ON_FIRE);
+                    throw Exceptions.propagate(e);
+                }
```
neykov: Still need to do the same for the `doFullStartOnRestart()` case. Suggest instead wrapping the `restart` effector as done in [`start`](https://github.com/apache/brooklyn-server/blob/master/software/base/src/main/java/org/apache/brooklyn/entity/software/base/lifecycle/MachineLifecycleEffectorTasks.java#L353-L363).

sjcorbett:
@neykov I've updated per your comments.

neykov:
Thanks @sjcorbett, merging.

PR #394, opened 2016-10-26
add exist check before getting the blob

andreaturli:


ahgittin:
from andrea:

| Blob blob = blobStore.getBlob(containerName, blobName); is retryable while blobExists is not. When you disable wire logging, jclouds uses a different code path which ends up in a IOException rather than KeyNotFoundException causing problems in Brooklyn persistence

| as jclouds convert KeyNotFoundException into null, and propagates IOException for that getBlob method

looks reasonable to me, merging

PR #395, opened 2016-10-26
modify jclouds logging categories to fix certain OOME's

ahgittin:


andreaturli:
Lgtm although the extra comment for wire logger (suboptimal..) will be not needed as soon as I fix jclouds

geomacy:
Do we need to add the same config in the karaf logging config - brooklyn-dist/karaf/apache-brooklyn/src/main/resources/etc/org.ops4j.pax.logging.cfg?

ahgittin:
yes @geomacy - good spot, will add

ahgittin:
@geomacy done in https://github.com/apache/brooklyn-dist/pull/55

PR #396, opened 2016-10-31
[WIP] A test to check rebind when bundle prefixes are used to load classes

neykov:
Do not merge until fixed as will lead to test failure.


aledsage:
@neykov thanks for the pointers. The problem is that the entity's memento is written with the plain type (i.e. not having the OSGi bundle name). The `BasicEntityMemento.type` is a string of the implementation type, set in `MementoGenerators.populateBrooklynObjectMementoBuilder`.

When we go to load that, we therefore don't know which bundle it came from. And we can't tell the bundle from the catalog item either (because we're trying to load from an arbitrary bundle, rather than a bundle that was in this catalog item's `libraries` section).

Note that for "normal" class names, we write the OSGi prefix inside `XmlMementoSerializer.OsgiClassnameMapper.serializedClass`.

---
It's not obvious where the best place is to fix this. The `MementoGenerators` currently know nothing about OSGi etc.

And the `XmlMementoSerializer` does not really want to poke around inside `BasicEntityMemento` etc. It could register a `Converter` for it (and for all the other subtypes of `AbstractMemento`), but that doesn't feel ideal.

On balance, I think putting something in the `MementoGenerators` is the best bet. I'll explore that more.

aledsage:
@neykov see https://github.com/neykov/brooklyn-server/pull/4 for the changes that should fix this test. Can you review and merge that into your branch, which should automatically add the commit to your pull request.

aledsage:
@neykov This is replaced by https://github.com/apache/brooklyn-server/pull/442 - do you want to close this PR?

neykov:
Superseded by https://github.com/apache/brooklyn-server/pull/442, closing.

PR #385, opened 2016-10-14
Don't allow multiple context entities in a task's tags

neykov:
When executing a task which is already tagged with a context entity in the execution context of another task, the task ends up with two context entities in its tags. It's indeterministic which one will be available through BrooklynTaskTags.getContextEntity. 

The context entity should be set by internal code at a single place. Remove all places where we are setting the context additionally and add a warning at runtime if detected. To be changed to an exception after a release or two, currently a warning in the logs.

Merge together with https://github.com/apache/brooklyn-library/pull/71.

PR #398, opened 2016-11-02
BROOKLYN-377: adds testConfigOnParentUsesConfigKeyDeclaredOnParent

aledsage:
This PR contains a failing test, which demonstrates the underlying problem in https://issues.apache.org/jira/browse/BROOKLYN-377.

I believe that https://github.com/apache/brooklyn-server/pull/385 will fix this. We should rebase + retest at that point. In particular, if I change `ValueResolver` to delete `.tagIfNotNull(BrooklynTaskTags.getTargetOrContextEntityTag(Tasks.current()))` then the test passes.

@neykov @drigodwin you'll both likely be interested in this.


aledsage:
#385 is now merged; I've rebased against master. The "many times" now passes for me. Assuming jenkins passes, this will be good to merge I think.

neykov:
Looks good and build is successful, merging.

PR #399, opened 2016-11-02
Adds TestCaseYamlTest

aledsage:


test-framework/src/test/java/org/apache/brooklyn/test/framework/yaml/TestCaseYamlTest.java
```diff
@@ -0,0 +1,105 @@
...
+import org.testng.annotations.AfterMethod;
+import org.testng.annotations.BeforeMethod;
+import org.testng.annotations.Test;
+
+//Checks that the test cases work in YAML
+@Test
+public class TestCaseYamlTest extends AbstractYamlRebindTest {
+    
+    // TODO Would like this test to be in brooklyn-camp, but that would create a circular dependency.
+    // brooklyn-test-framework depends on brooklyn-camp.
```
neykov: I assume it depends on `brooklyn-camp` just for the tests? That's the problem with other modules as well - we want the functionality tested in YAML, but the code has nothing to do with CAMP otherwise. That's the reason why `brooklyn-camp`'s tests are huge - they are testing other's modules code through YAML.

I think the way to go is to create a camp-itest module just for tests like this.

On the other hand the YAML is just an alternative spec description so could we just use in-java specs to achieve the same goal?

PR #402, opened 2016-11-03
Bump version of snakeyaml to 1.16

googlielmo:
- align to version pulled in by the `kubernetes-client` feature


aledsage:
I talked to @googlielmo offline about this. He said he'd confirmed that `mvn dependency:tree` shows no other version of snakyaml coming from any other transitive dependencies; and that he'd tested this change by building a Brooklyn distro and deploying some apps.

Therefore it looks good to me (once jenkins confirms that the build works).

PR #403, opened 2016-11-04
BROOKLYN-380: Adds DynamicFabric.includeInitialChildren config

aledsage:
And adds tests for existing behaviour.

### DIFF
--- a/core/src/test/java/org/apache/brooklyn/entity/group/DynamicRegionsFabricTest.java
+++ b/core/src/test/java/org/apache/brooklyn/entity/group/DynamicRegionsFabricTest.java
@@ -19,6 +19,8 @@
 package org.apache.brooklyn.entity.group;
 
 import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertFalse;
+import static org.testng.Assert.assertTrue;
 
 import java.util.Collection;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


core/src/test/java/org/apache/brooklyn/entity/group/DynamicRegionsFabricTest.java
```diff
@@ -60,16 +64,98 @@ public void setUp() throws Exception {
...
+            .locations(ImmutableList.of(loc1, loc2)));
+        DynamicRegionsFabric fabric2 = (DynamicRegionsFabric) Iterables.getOnlyElement(app2.getChildren());
+        app2.start(ImmutableList.<Location>of());
+        
+        assertFabricChildren(fabric2, 2, ImmutableList.of(loc1, loc2));
     }
     
     @Test
+    @SuppressWarnings("deprecation")
+    public void testUsesInitialLocationsFromEntitySpec() throws Exception {
```
mikezaccardo: This test is identical to `testUsesInitialLocationsFromAppSpec()`

mikezaccardo:
Reviewed, all tests pass

core/src/test/java/org/apache/brooklyn/entity/group/DynamicRegionsFabricTest.java
aledsage: Not quite the same. In `testUsesInitialLocationsFromAppSpec` we set the locations on the app's spec. In `testUsesInitialLocationsFromEntitySpec` we set the locations on the fabric's spec.

Could refactor to remove duplication, but I think that it's probably simpler to have the 7 lines duplicated rather than a delegate method with an if-else block (given setting the location is buried in the 5 lines required to build the spec).

aledsage:
@mikezaccardo thanks, I've incorporated the review comments (see second commit). I've also created a jira ticket to track this, and referred to it in the PR title / commit message.

If jenkins confirms this builds ok, I'll merge.

core/src/test/java/org/apache/brooklyn/entity/group/DynamicRegionsFabricTest.java
mikezaccardo: Ah yes, I see it now -- line 80 vs. line 94. Thanks.

PR #404, opened 2016-11-04
Fix Transformer (for when registered before entity managed)

aledsage:
The transformer can be added before the entity is managed (e.g. if
specified in the entity spec, as is often done via yaml). In that case,
the entity’s management context is still a `NonDeploymentManagementContext`.
The subscriptions are stored in a queue. When the entity is managed,
the subscriptions are replayed.

However, the queued subscriptions did not store the entity context.
Therefore when the transformer received its events, it didn’t have
an entity context so could not determine which entity it was talking
about.

This caused brooklyn-library’s `EnrichersSlightlySimplerYamlTest.testWithAppEnricher`
to fail.


neykov:
Makes sense, merging.

PR #406, opened 2016-11-04
Fix app.stop so doesn’t go on-fire briefly

aledsage:
Previously in app.stop, we set the serviceUp=false before setting
expectedState=stopping. This meant there was a race where we could
briefly infer the actual state as on_fire.

This caused failure of `SoftwareProcessEntityLatchTest.testStopLatchBlocks`:
the attributeWhenReady latch would abort because it received an
event saying that the app was on_fire, so the stop_latch completely
immediately.

The fix is simple: in app.stop, set the expectedSate=stopping before
setting serviceUp=false.

@neykov @geomacy you'll both be interested in this.

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/entity/AbstractApplication.java
+++ b/core/src/main/java/org/apache/brooklyn/core/entity/AbstractApplication.java
@@ -200,9 +200,9 @@ public abstract class AbstractApplication extends AbstractEntity implements Star
     public void stop() {
         logApplicationLifecycle("Stopping");
 
+        setExpectedStateAndRecordLifecycleEvent(Lifecycle.STOPPING);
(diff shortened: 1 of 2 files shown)


geomacy:
👍  LGTM. The change makes sense according to the description above.  

On the question of reproducing this, I wasn't able to reproduce the test failure prior with master at date of 26th October (34eab17).  I caught up with master today and was able to get a couple of test failures by running 500 iterations of `testStopLatchBlocks`.   It's possible that recent changes have affected the visibility of the race that @aledsage mentions above.   

Having pulled and merged this PR, however, I was able to run 1000 iterations without any failures.

aledsage:
Thanks @geomacy - merging now.

PR #408, opened 2016-11-06
Use actual config key if available in DSL

grkvlt:
Fixes issue where config defined in `brooklyn.parameters` is resolved as `null` instead of using the default value, when used inside a transformer.

### DIFF
--- a/camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/methods/DslComponent.java
+++ b/camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/methods/DslComponent.java
@@ -29,10 +29,13 @@ import org.apache.brooklyn.api.sensor.AttributeSensor;
 import org.apache.brooklyn.api.sensor.Sensor;
 import org.apache.brooklyn.camp.brooklyn.BrooklynCampConstants;
 import org.apache.brooklyn.camp.brooklyn.spi.dsl.BrooklynDslDeferredSupplier;
+import org.apache.brooklyn.config.ConfigKey;
 import org.apache.brooklyn.core.config.ConfigKeys;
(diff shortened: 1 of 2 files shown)


andreaturli:
Failures look related, I can see the same problems as https://builds.apache.org/job/brooklyn-server-pull-requests/1283/org.apache.brooklyn$brooklyn-test-framework/testReport/ when running `mvn clean install` locally.

camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/methods/DslComponent.java
```diff
@@ -257,7 +260,7 @@ public EntityId(DslComponent component) {
         @Override
         public Maybe<Object> getImmediately() {
             Maybe<Entity> targetEntityMaybe = component.getImmediately();
-            if (targetEntityMaybe.isAbsent()) return Maybe.absent("Target entity not available");
+            if (targetEntityMaybe.isAbsentOrNull()) return Maybe.absent("Target entity not available");
```
aledsage: As above: this is changing the semantics of `getImmediately` compared to `get`.

aledsage:
@grkvlt thanks for finding and fixing this!

See https://github.com/apache/brooklyn-server/pull/410 - I've extracted your fix for the problem from this PR, and added an additonal test. Happy to merge that as soon as jenkins confirms it works.

For the changes to `isAbsentOrNull`, see my comments. I believe that is what is making `RelativeEntityTestCaseTest` fail in jenkins (it fails for me locally as well with your branch, but does not fail with PR #410).

grkvlt:
@aledsage your #410 looks good, so I will close this. I had not thought about the semantics of `get()` vs. `getImmediately()` and `getRaw()` etc. but the rationale for just checking for `Maybe.absent()` seems sound.

PR #410, opened 2016-11-07
transformer config uses brooklyn param defaults

aledsage:
This extracts part of https://github.com/apache/brooklyn-server/pull/408, adding an additional test.


aledsage:
Merging this now - the only non-test code changes were extracted form @grkvlt's #408. My changes just add more tests.

ahgittin:
Was seeing some NPE regression in enrichers, confirmed it's fixed with this merged.  Nice one.

PR #411, opened 2016-11-07
Test change to coercion of default config values

ahgittin:
Several minor fixes/changes, plus one significant test in https://github.com/apache/brooklyn-server/commit/e9c1f4cbb67d1bd89d1e7d84f8730225047ea337 which highlights a change in behaviour since early Oct

Also see related PR adding this to release notes

### DIFF
--- a/camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/AbstractYamlTest.java
+++ b/camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/AbstractYamlTest.java
@@ -40,6 +40,7 @@ import org.apache.brooklyn.core.test.entity.LocalManagementContextForTests.Build
 import org.apache.brooklyn.core.typereg.RegisteredTypeLoadingContexts;
 import org.apache.brooklyn.util.collections.MutableMap;
 import org.apache.brooklyn.util.core.ResourceUtils;
+import org.apache.brooklyn.util.net.Urls;
 import org.apache.brooklyn.util.stream.Streams;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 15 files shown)


camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/AbstractYamlTest.java
```diff
@@ -112,7 +113,14 @@ protected void waitForApplicationTasks(Entity app) {
     }
 
     protected String loadYaml(String yamlFileName, String ...extraLines) throws Exception {
-        String input = new ResourceUtils(this).getResourceAsString(yamlFileName).trim();
+        ResourceUtils ru = new ResourceUtils(this);
+        if (!ru.doesUrlExist(yamlFileName)) {
+            if (ru.doesUrlExist(Urls.mergePaths(getClass().getPackage().getName().replace('.', '/'), yamlFileName))) {
+                // look in package-specific folder if not found at root
+                yamlFileName = Urls.mergePaths(getClass().getPackage().getName().replace('.', '/'), yamlFileName);
```
aledsage: Very minor (feel free to ignore). If it still can't be found, I'd prefer the final exception to give the original `yamlFileName` rather than the package-specific name. With this change, it will throw the exception in `run.getResourceAsString` with the transformed value.

camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/ConfigTypeCoercionYamlTest.java
```diff
@@ -0,0 +1,77 @@
...
+    public void testSshConfigFromDefault() throws Exception {
+        RecordingSshTool.setCustomResponse(".*myCommand.*", new RecordingSshTool.CustomResponse(0, "myResponse", null));
+        
+        String bp = loadYaml("config-type-coercion-test.yaml",
+            "location:",
+            "  localhost:",
+            "    sshToolClass: "+RecordingSshTool.class.getName());
+        // remove all lines referring to "exact" -- that's useful for expository and running in UI
+        // but it will fail (timeout) if the port isn't available so not good in tests
+        bp = Strings.removeLines(bp, StringPredicates.containsLiteralIgnoreCase("exact"));
```
aledsage: Worth adding a comment at the top of the file `config-type-coercion-test.yaml`, to say that the unit test `ConfigTypeCoercionYamlTest.testSshConfigFromDefault ` will transform the file to remove the lines containing "exact". Otherwise it will likely surprise people in the future who don't read through the entire test method.

aledsage:
@ahgittin LGTM; only one very minor comment for you to take a look at.

camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/AbstractYamlTest.java
ahgittin: the package-specific name is only used if the original path **isn't** found *and* the package-specific one **is** -- in which case i think it's correct to use the package-specific name in the error

ahgittin:
thanks @aledsage, fixed, plus RAT fix, retesting then will merge

camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/AbstractYamlTest.java
aledsage: Ah, you're absolutely right - I missed that second `ru.doesUrlExist` check.

aledsage:
jenkins confirmed this passes - merging now.

PR #414, opened 2016-11-08
Improvement of TestHttpCall entity

tbouron:
This adds the possibility to:
- do GET, POST, PUT, DELETE and OPTIONS requests
- specify headers
- specify request body for POST and PUT requests

### DIFF
--- a/test-framework/src/main/java/org/apache/brooklyn/test/framework/TestHttpCall.java
+++ b/test-framework/src/main/java/org/apache/brooklyn/test/framework/TestHttpCall.java
@@ -18,10 +18,14 @@
  */
 package org.apache.brooklyn.test.framework;
 
+import java.util.Map;
+
 import org.apache.brooklyn.api.entity.ImplementedBy;
 import org.apache.brooklyn.config.ConfigKey;
 import org.apache.brooklyn.core.config.ConfigKeys;
 import org.apache.brooklyn.util.core.flags.SetFromFlag;
+
+import com.google.common.reflect.TypeToken;
 
 /**
  * Entity that makes a HTTP Request and tests the response
@@ -32,8 +36,33 @@ public interface TestHttpCall extends BaseTest {
     @SetFromFlag(nullable = false)
     ConfigKey<String> TARGET_URL = ConfigKeys.newStringConfigKey("url", "URL to test");
 
+    @SetFromFlag(nullable = false)
+    ConfigKey<HttpMethod> TARGET_METHOD = ConfigKeys.builder(HttpMethod.class)
+            .name("method")
+            .description("Method to request the URL: GET, POST, PUT, DELETE, etc")
+            .defaultValue(HttpMethod.GET)
+            .build();
+
+    @SetFromFlag
+    ConfigKey<Map<String, String>> TARGET_HEADERS = ConfigKeys.builder(new TypeToken<Map<String, String>>() {})
+            .name("headers")
+            .description("Headers to add to the request")
+            .build();
+
+    @SetFromFlag
+    ConfigKey<String> TARGET_BODY = ConfigKeys.newStringConfigKey("body", "The request body to send (only for POST and PUT requests)");
+
     ConfigKey<HttpAssertionTarget> ASSERTION_TARGET = ConfigKeys.newConfigKey(HttpAssertionTarget.class, "applyAssertionTo",
         "The HTTP field to apply the assertion to [body,status]", HttpAssertionTarget.body);
+
+    enum HttpMethod {
+        GET, POST, PUT, DELETE, HEAD;
+
+        @Override
+        public String toString() {
+            return this.name();
+        }
+    }
 
     enum HttpAssertionTarget {
         body("body"), status("status");
--- a/test-framework/src/main/java/org/apache/brooklyn/test/framework/TestHttpCall.java
+++ b/test-framework/src/main/java/org/apache/brooklyn/test/framework/TestHttpCall.java
@@ -24,6 +24,14 @@ import org.apache.brooklyn.api.entity.ImplementedBy;
 import org.apache.brooklyn.config.ConfigKey;
 import org.apache.brooklyn.core.config.ConfigKeys;
 import org.apache.brooklyn.util.core.flags.SetFromFlag;
+import org.apache.http.client.methods.HttpDelete;
+import org.apache.http.client.methods.HttpGet;
+import org.apache.http.client.methods.HttpHead;
+import org.apache.http.client.methods.HttpOptions;
+import org.apache.http.client.methods.HttpPost;
+import org.apache.http.client.methods.HttpPut;
+import org.apache.http.client.methods.HttpRequestBase;
+import org.apache.http.client.methods.HttpTrace;
 
 import com.google.common.reflect.TypeToken;
 
@@ -43,20 +51,30 @@ public interface TestHttpCall extends BaseTest {
             .defaultValue(HttpMethod.GET)
             .build();
 
-    @SetFromFlag
     ConfigKey<Map<String, String>> TARGET_HEADERS = ConfigKeys.builder(new TypeToken<Map<String, String>>() {})
             .name("headers")
             .description("Headers to add to the request")
             .build();
 
-    @SetFromFlag
     ConfigKey<String> TARGET_BODY = ConfigKeys.newStringConfigKey("body", "The request body to send (only for POST and PUT requests)");
 
     ConfigKey<HttpAssertionTarget> ASSERTION_TARGET = ConfigKeys.newConfigKey(HttpAssertionTarget.class, "applyAssertionTo",
         "The HTTP field to apply the assertion to [body,status]", HttpAssertionTarget.body);
 
     enum HttpMethod {
-        GET, POST, PUT, DELETE, HEAD;
+        GET(HttpGet.class),
+        POST(HttpPost.class),
+        PUT(HttpPut.class),
+        DELETE(HttpDelete.class),
+        HEAD(HttpHead.class),
+        OPTIONS(HttpOptions.class),
+        TRACE(HttpTrace.class);
+
+        public final Class<? extends HttpRequestBase> requestClass;
+
+        HttpMethod(Class<? extends HttpRequestBase> requestClass) {
+            this.requestClass = requestClass;
+        }
 
         @Override
         public String toString() {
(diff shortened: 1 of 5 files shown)


aledsage:
@tbouron can you add tests to `org.apache.brooklyn.test.framework.TestHttpCallTest`?

test-framework/src/main/java/org/apache/brooklyn/test/framework/TestHttpCall.java
```diff
@@ -32,9 +36,34 @@
...
     ConfigKey<String> TARGET_URL = ConfigKeys.newStringConfigKey("url", "URL to test");
 
+    @SetFromFlag(nullable = false)
+    ConfigKey<HttpMethod> TARGET_METHOD = ConfigKeys.builder(HttpMethod.class)
+            .name("method")
+            .description("Method to request the URL: GET, POST, PUT, DELETE, etc")
+            .defaultValue(HttpMethod.GET)
+            .build();
+
+    @SetFromFlag
```
aledsage: Remove the `@SetFromFlag` - we're just relying on the config key name, which is the right thing to do!
Same below for TARGET_BODY.

test-framework/src/main/java/org/apache/brooklyn/test/framework/TestHttpCallImpl.java
```diff
@@ -50,19 +55,25 @@
      */
     public void start(Collection<? extends Location> locations) {
         String url = null;
+        HttpMethod method = null;
```
aledsage: Minor; feel free to ignore :-)
We only declare `url` above because the variable is needed in the catch block for a nicer error message. I'd declare these variables in-line, rather than outside of the try block, unless we also want to log any of them in the catch error message.

test-framework/src/main/java/org/apache/brooklyn/test/framework/TestHttpCallImpl.java
```diff
@@ -77,13 +88,18 @@ public void start(Collection<? extends Location> locations) {
...
-                                             HttpAssertionTarget target, final String url) {
+                                             HttpAssertionTarget target, final HttpMethod method, final String url, final Map<String, String> headers, final String body) {
         switch (target) {
             case body:
                 Supplier<String> getBody = new Supplier<String>() {
                     @Override
                     public String get() {
-                        return HttpTool.getContent(url);
+                        try {
+                            return HttpTool.execAndConsume(HttpTool.httpClientBuilder().build(), createHttpMethod(method, url, headers, body)).getContentAsString();
```
aledsage: Minor: a couple of very long lines here. Probably worth splitting them up for readability (particularly as there is another method call at the end of this one, which can easily be missed if someone doesn't scroll).

test-framework/src/main/java/org/apache/brooklyn/test/framework/TestHttpCallImpl.java
```diff
@@ -94,7 +110,12 @@ public String get() {
                     @Override
                     public Integer get() {
                         try {
-                            return HttpTool.getHttpStatusCode(url);
+                            final Maybe<HttpResponse> response = HttpTool.execAndConsume(HttpTool.httpClientBuilder().build(), createHttpMethod(method, url, headers, body)).getResponse();
+                            if (response.isPresentAndNonNull()) {
+                                return response.get().getStatusLine().getStatusCode();
+                            } else {
+                                throw new Exception("HTTP call did not return any reponse");
```
aledsage: Type: `reponse`.

test-framework/src/main/java/org/apache/brooklyn/test/framework/TestHttpCall.java
tbouron: But if I remove the `@SetFromFlag`, we won't be able to set this config key outside the `brooklyn.config` block, isn't it @aledsage ?

I also keep them as it was the "standard" across the test entities

test-framework/src/main/java/org/apache/brooklyn/test/framework/TestHttpCallImpl.java
tbouron: Good catch, I'll update

test-framework/src/main/java/org/apache/brooklyn/test/framework/TestHttpCallImpl.java
```diff
@@ -109,6 +130,49 @@ public Integer get() {
         }
     }
 
+    private HttpRequestBase createHttpMethod(HttpMethod method, String url, Map<String, String> headers, String body) throws Exception {
+        switch (method) {
+            case GET:
+                HttpTool.HttpGetBuilder httpGetBuilder = new HttpTool.HttpGetBuilder(new URI(url));
```
aledsage: This is a code-smell that we need to do lots of identical code for each case, rather than being able to use the `HttpTool` in a simpler way (e.g. passing in the method to a generic request builder, perhaps).

tbouron: Agree but the the `HttpTool.HttpRequestBuilder` is based on the request class which is why there is this conversion enum -> request. Do you think the `HttpMethod` and this private method `createHttpMethod` should move into `HttpTool`?

test-framework/src/main/java/org/apache/brooklyn/test/framework/TestHttpCall.java
aledsage: I believe you would still be able to set it outside of the `brooklyn.config:` block in yaml. See https://github.com/apache/brooklyn-server/pull/418 for a new test that demonstrates this.

We should get rid of all the `@SetFromFlag`s, where it's not adding additional info (e.g. to say non-null or such like).

test-framework/src/main/java/org/apache/brooklyn/test/framework/TestHttpCallImpl.java
aledsage: I wondered about moving the `HttpMethod` into `HttpTool`. No strong feelings. If we did try to add a generic builder, then it would allow setting a body for all methods, and then would have to validate on `build()`. That's not ideal, but maybe worth it to avoid duplication.

Also, see https://en.wikipedia.org/wiki/Hypertext_Transfer_Protocol#Request_methods. There are other methods we're not supporting (e.g. `TRACE`, `OPTIONS`, etc). Are we right to use an enum with a subset of the methods? I think we are: the enum lists all the methods that are supported. And there is no need to try to support the less common methods until there's a real use-case.

Let's leave it as-is, and think about it some more :-)
But if you feel it would be a lot cleaner to have a generic builder, and/or to move `HttpMethod` into `HttpTool` then go for it.

test-framework/src/test/java/org/apache/brooklyn/test/framework/TestHttpCallTest.java
```diff
@@ -87,6 +96,11 @@ public void testHttpBodyAssertions() {
                 .configure(TestHttpCall.TARGET_URL, server.getUrl() + "/body.json")
                 .configure(TestHttpCall.TIMEOUT, new Duration(10L, TimeUnit.SECONDS))
                 .configure(TestSensor.ASSERTIONS, newAssertion("matches", ".*123.*")));
+        app.createAndManageChild(EntitySpec.create(TestHttpCall.class)
+                .configure(TestHttpCall.TARGET_URL, server.getUrl() + "/foo/bar")
+                .configure(TestHttpCall.TARGET_METHOD, TestHttpCall.HttpMethod.POST)
+                .configure(TestHttpCall.TIMEOUT, new Duration(10L, TimeUnit.SECONDS))
```
aledsage: Personal preference: we should use `Asserts.DEFAULT_LONG_TIMEOUT` rather than our code being littered with arbitrary timeouts. In my opinion, the only time one should set different timeouts for something we expect to not timeout is if it's a performance test.

test-framework/src/test/java/org/apache/brooklyn/test/framework/TestHttpCallTest.java
```diff
@@ -107,6 +121,47 @@ public void testHttpStatusAssertions() {
...
+                .configure(TestHttpCall.ASSERTION_TARGET, TestHttpCall.HttpAssertionTarget.status)
+                .configure(TestSensor.ASSERTIONS, newAssertion("isEqualTo", HttpStatus.SC_CREATED)));
+        app.createAndManageChild(EntitySpec.create(TestHttpCall.class)
+                .configure(TestHttpCall.TARGET_URL, server.getUrl() + "/foo/bar")
+                .configure(TestHttpCall.TARGET_METHOD, TestHttpCall.HttpMethod.PUT)
+                .configure(TestHttpCall.TIMEOUT, new Duration(10L, TimeUnit.SECONDS))
+                .configure(TestHttpCall.ASSERTION_TARGET, TestHttpCall.HttpAssertionTarget.status)
+                .configure(TestSensor.ASSERTIONS, newAssertion("isEqualTo", HttpStatus.SC_METHOD_NOT_ALLOWED)));
```
aledsage: Would be nice to have something that tests it really did do a `PUT` correctly, but this is good enough to start with.

aledsage:
@tbouron this now has merge conflicts. Can you resolve those, take a look at the additional minor comments to see what you think should be done, and then we can merge this.

test-framework/src/main/java/org/apache/brooklyn/test/framework/TestHttpCall.java
tbouron: Fair enough, I'll remove the unnecessary ones

test-framework/src/test/java/org/apache/brooklyn/test/framework/TestHttpCallTest.java
tbouron: It does implicitly. The server will return a 405 if the method is different than `PUT`

tbouron:
@aledsage Comment addressed. I created a generic builder for the `HttpRequest`. Hope this is fine now and can be merged

test-framework/src/test/java/org/apache/brooklyn/test/framework/TestHttpCallTest.java
aledsage: But it would return a `405` if we used `PUT` or `DELETE`, so the test isn't checking that we really did create a `PUT` request. I'm fine with it as-is though.

aledsage:
LGTM; merging.

test-framework/src/test/java/org/apache/brooklyn/test/framework/TestHttpCallTest.java
tbouron: No it wont. By default, the return status if 200 [thanks to this](https://github.com/apache/brooklyn-server/pull/414/files#diff-5fd62a21e2e690469afbe6fef8f2013dR191). Then, if the method is different than [what we registered](https://github.com/apache/brooklyn-server/pull/414/files#diff-5fd62a21e2e690469afbe6fef8f2013dR70), it will update the status to 405

PR #409, opened 2016-11-06
bump sshj to same jclouds version [merge after jclouds 2.0 upgrade]

andreaturli:
please use `https://github.com/apache/brooklyn-server/compare/master...andreaturli:upgrade-sshj?w=1` to review this pr

PR #415, opened 2016-11-08
[NOT MERGE] Jclouds2.0.0 snapshot

andreaturli:
it is built on top of #409 

It uses jclouds-2.0.0-SNAPSHOT just to verify what needs to be changed in brooklyn-server.

I think it would be good to set up a jenkins job that builds this branch continuously, but I can't set it up myself at https://builds.apache.org/view/Brooklyn/

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -58,7 +58,7 @@
             <version>${project.version}</version>
         </dependency>
         <dependency>
-            <groupId>net.schmizz</groupId>
+            <groupId>com.hierynomus</groupId>
             <artifactId>sshj</artifactId>
         </dependency>
         <dependency>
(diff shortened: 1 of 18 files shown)


aledsage:
@andreaturli can you rebase this against master (and same for https://github.com/apache/brooklyn-server/pull/409)? Now that we've bumped the master version to `0.11.0-SNAPSHOT`, I think we can merge this.

neykov:
@andreaturli Can you change the jclouds version to `2.0.0` (from `2.0.0-SNAPSHOT`).
I tried merging the PR against latest master yesterday and had to resolve a few compile-time problems in the tests. Now I end up with a bunch of `java.lang.IllegalArgumentException: an instance of the ImageCacheSupplier is needed` errors, so I didn't do it correctly I assume.

andreaturli:
@aledsage @neykov I've rebased the branch and update jclouds version from `2.0.0-SNAPSHOT` to `2.0.0`

PR #416, opened 2016-11-08
Refactor and add to DSL tests

aledsage:


camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/DslAndRebindYamlTest.java
```diff
@@ -61,58 +56,41 @@
 
 import com.google.common.collect.Iterables;
 import com.google.common.collect.Lists;
-import com.google.common.io.Files;
 
 @Test
-public class DslAndRebindYamlTest extends AbstractYamlTest {
+public class DslAndRebindYamlTest extends AbstractYamlRebindTest {
```
aledsage: The `AbstractYamlRebindTest` does the rebind correctly - it makes sure the `mgmt` field is updated on rebind, so subsequent provisioning uses the new application context. Before this, `testDslSensorFromClass` was not doing what you'd expect - the second call returned the app from the first call to `doTestOnEntityWithSensor`, so the subsequent assertions weren't useful.

googlielmo:
Builds correctly. LGTM

aledsage:
Thanks @googlielmo - I think the previous jenkins failure was because an assertion timeout of one second failed. I've increased the timeout to use `Asserts.DEFAULT_LONG_TIMEOUT`. The apache jenkins box sometimes does starve the thread for a short (!) period of time, which can be greater than one second. For example, this could be during a full GC or due to contention on the build server.

If jenkins passes this time, I'll merge.

PR #418, opened 2016-11-08
Adds ConfigYamlTest.testConfigAtTopLevel

aledsage:


aledsage:
Merging this now (even though not reviewed), as it only adds tests.
@tbouron you might be interested in taking a look - it was motivated by something you said, about whether we needed to use `@SetFromFlag` on config fields.

tbouron:
Thanks @aledsage! Indeed, it's great to have this test, it confirm what you said, i.e we don't need to mark a ConfigKey with `@SetConfigFlag` annotation to be able to use it as top level config

PR #419, opened 2016-11-09
Fix key lookup

ahgittin:
fixes subtle bug which prevents some evaluation strategies for map subkeys

### DIFF
--- a/camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/ConfigNestedYamlTest.java
+++ b/camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/ConfigNestedYamlTest.java
@@ -1,0 +1,148 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.brooklyn.camp.brooklyn;
+
+import java.util.Map;
+
+import org.apache.brooklyn.api.entity.Entity;
+import org.apache.brooklyn.core.config.ConfigKeys;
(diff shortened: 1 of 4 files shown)


ahgittin:
BTW i'm not enamoured of the ambiguity in map config keys dot syntax ... it was useful in the properties file days but not anymore imo ... but we do need to be careful and deprecate it, not just chop it off at the knees which the change to config inheritance lookup did inadvertently

camp/camp-brooklyn/src/test/resources/org/apache/brooklyn/camp/brooklyn/config-nested-test.bom
```diff
@@ -0,0 +1,76 @@
...
+      brooklyn.config:
+        template.substitutions:
+          field: val
+  - id: base-map-parameter
+    itemType: entity
+    item:
+      type: org.apache.brooklyn.entity.stock.BasicEntity
+      brooklyn.parameters:
+        - name: template.substitutions
+          type: java.util.Map
```
neykov: We should probably create a `MapConfigKey` if we see this in parameters.

core/src/main/java/org/apache/brooklyn/core/config/internal/AbstractConfigMapImpl.java
```diff
@@ -356,8 +357,9 @@ protected ConfigInheritance getDefaultRuntimeInheritance() {
             
             Function<TContainer, Maybe<Object>> lookupFn = new Function<TContainer, Maybe<Object>>() {
                 @Override public Maybe<Object> apply(TContainer input) {
-                    Maybe<Object> result = getRawValueAtContainer(input, queryKey);
-                    if (!raw) result = resolveRawValueFromContainer(input, queryKey, result);
+                    // lookup against ownKey as it may do extra resolution (eg grab *.* subkeys if a map)
```
neykov: Makes sense. Also good testo coverage. Merging.

camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/ConfigNestedYamlTest.java
```diff
@@ -0,0 +1,148 @@
...
+        Entity app = createAndStartApplication( bp );
+        waitForApplicationTasks(app);
+        return Iterables.getOnlyElement( app.getChildren() );
+    }
+    
+    protected Entity checkEntity(Entity ent, boolean expectDotSyntaxToWork) throws Exception {
+        String dotSyntax = ent.getConfig( ConfigKeys.newStringConfigKey("template.substitutions.field") );
+        Assert.assertEquals(dotSyntax, expectDotSyntaxToWork ? "val" : null, "wrong value for dot syntax: "+dotSyntax);
+        // observed in some situations we get the dot syntax instead!
+        // Assert.assertEquals(dotSyntax, "val", "wrong value for dot syntax: "+dotSyntax);
```
neykov: I'd have expected this to always be the case, for any `Map` value, not just for `MapConfigKey`. The logic doing the lookup (if I remember correctly) is trying to get for increasingly longer parts of the full key until it has a hit.

PR #421, opened 2016-11-09
Show info after application start

ahgittin:
handy with the docker one-line deployer

### DIFF
--- a/launcher-common/src/main/java/org/apache/brooklyn/launcher/common/BasicLauncher.java
+++ b/launcher-common/src/main/java/org/apache/brooklyn/launcher/common/BasicLauncher.java
@@ -43,6 +43,7 @@ import org.apache.brooklyn.camp.CampPlatform;
 import org.apache.brooklyn.camp.brooklyn.BrooklynCampPlatformLauncherNoServer;
 import org.apache.brooklyn.config.ConfigKey;
 import org.apache.brooklyn.core.catalog.internal.CatalogInitialization;
+import org.apache.brooklyn.core.entity.Attributes;
 import org.apache.brooklyn.core.entity.Entities;
 import org.apache.brooklyn.core.entity.StartableApplication;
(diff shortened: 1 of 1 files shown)


launcher-common/src/main/java/org/apache/brooklyn/launcher/common/BasicLauncher.java
```diff
@@ -724,6 +725,15 @@ protected void startApps() {
...
                     ((Startable)app).start(locations);
+                    Entities.dumpInfo(app);
+                    String sensors = "";
+                    if (app.getAttribute(Attributes.MAIN_URI_MAPPED_PUBLIC)!=null) {
+                        sensors = ": "+app.getAttribute(Attributes.MAIN_URI_MAPPED_PUBLIC);
+                    } else if (app.getAttribute(Attributes.MAIN_URI)!=null) {
+                        sensors += ": "+app.getAttribute(Attributes.MAIN_URI);
+                    }
+                    LOG.info("Started brooklyn application {} in location{} {}{}", new Object[] { app, locations.size()!=1?"s":"", locations,
+                        sensors });
```
neykov: I thought this is executed in classic mode only, no? How is the app passed to the osgi launcher?

ahgittin:
this is classic only, as that's what the docker container uses at present (and the normal brooklyn dist), and it was easier to get going.

fully accept that when we move to osgi we'll have to change and do this all another way.

BTW i had been thinking the move to osgi would drop these args, most args in fact, in favour of config files and sending additional commands to the server afterwards, and we'd have to rebuild the `Dockerfile` of course.  however it's worth bearing in mind that that the multi-command approach is a little tedious in docker-land however, as containers prefer the "just-one-process" model ... and many shells like alpine don't have job control to be able to background.  entirely possible but maybe worth a discussion about whether to send startup instructions / duplicate this flag to brooklyn osgi or not.

neykov:
LGTM in that case, merging.

PR #422, opened 2016-11-09
SshCommandEffector + shell.env

sjcorbett:
SshCommandEffector resolves shell.env against the target entity rather than always resolving it against the entity the effector is attached to. This also fixes the resolution of the execution directory.

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/effector/ssh/SshCommandEffector.java
+++ b/core/src/main/java/org/apache/brooklyn/core/effector/ssh/SshCommandEffector.java
@@ -59,11 +59,12 @@ public final class SshCommandEffector extends AddEffector {
     public static final ConfigKey<String> EFFECTOR_EXECUTION_DIR = SshCommandSensor.SENSOR_EXECUTION_DIR;
     public static final MapConfigKey<Object> EFFECTOR_SHELL_ENVIRONMENT = BrooklynConfigKeys.SHELL_ENVIRONMENT;
 
-    public static enum ExecutionTarget {
+    public enum ExecutionTarget {
         ENTITY,
         MEMBERS,
         CHILDREN
     }
+
     public static final ConfigKey<ExecutionTarget> EXECUTION_TARGET = ConfigKeys.newConfigKey(ExecutionTarget.class, "executionTarget", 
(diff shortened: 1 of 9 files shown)


sjcorbett:
Resolution of shell environments is badly affected by the problem #385 fixes. For example when you invoke the effector created in this bluerint:

```yaml
location: localhost
services:
- type: brooklyn.entity.group.DynamicCluster
  id: cluster
  brooklyn.config:
    initialSize: 3
    memberSpec:
      $brooklyn:entitySpec:
        brooklyn.config:
          shell.env:
            MEMBER_CFG: $brooklyn:config("cluster.member.id")
            MEMBER_ID: $brooklyn:attributeWhenReady("entity.id")

  - type: brooklyn.entity.software.ssh.SshCommandEffector
    brooklyn.config:
      name: test-effector
      shell.env:
        CLUSTER_ID: $brooklyn:attributeWhenReady("entity.id")
      command: 'echo "cluster=${CLUSTER_ID}, memberId=${MEMBER_ID}, member=${MEMBER_CFG}" >> /tmp/echo'
      executionTarget: members
```
You generally observe output like:
```
cluster=hgnwhpehtq, memberId=h0ea5qzz4k, member=1
cluster=hgnwhpehtq, memberId=hgnwhpehtq, member=
cluster=hgnwhpehtq, memberId=hgnwhpehtq, member=
```
In two of the three cases the entries in the member's `shell.env` were resolved against the cluster, not the member.

geomacy:
👍 changes look good to me at a read through.

aledsage:
LGTM; merging.

@sjcorbett I've also merged #385 - do we want to add any more tests (similar to your example yaml, that it gets the right entity id)?

sjcorbett:
Thanks @geomacy and @aledsage. I'll write a test for the new behaviour. Is there any way to inject a mock SSH tool? Would be neat if the test could be a unit test rather than an integration test.

aledsage:
@sjcorbett you can use `RecordingSshTool`. For example https://github.com/apache/brooklyn-server/blob/master/camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/ConfigParametersYamlTest.java#L452-L493 for how that can be used in a YAML blueprint.

PR #423, opened 2016-11-09
BROOKLYN-382: get-catalog performance improvements

aledsage:
The `specCache` speeds up the REST api calls to `/catalog/entities` considerably.

For example, with 10 small yaml entities in the catalog, before the cache it takes about 170ms per call (and proportionately more as we increase the number of catalog items or their size):

```
test=PerformanceTestDescriptor{summary=CatalogResourcePerformanceTest.testListAllEntities, iterations=100, warmupIterations=10, job=org.apache.brooklyn.rest.resources.CatalogResourcePerformanceTest$8@246f150f, minAcceptablePerSecond=1.0}; result=PerformanceTestResult{duration=16s 981ms, warmup=3s 274ms, iterations=100, warmupIterations=10, ratePerSecond=5.888934691714269, histogram={134.2ms-268.4ms: 100}, cpuTotalFraction=1.3575868461219567, cpuSampleFractions=[1.2573016983016982, 1.9225778443113772, 1.4971834496510468, 1.5281373134328358, 1.1606623505976095, 1.2851962151394423, 1.2523649052841475, 1.2795744255744255, 1.211623880597015, 1.1832784431137724, 1.162554, 1.3302412761714855, 1.1046873126873127, 1.4044291417165669, 1.0696699900299103, 1.846793, 1.0917181274900398], minAcceptablePerSecond=1.0}
```

With the cache, it takes 24ms per call:
```
test=PerformanceTestDescriptor{summary=CatalogResourcePerformanceTest.testListAllEntities, iterations=100, warmupIterations=10, job=org.apache.brooklyn.rest.resources.CatalogResourcePerformanceTest$8@345a656a, minAcceptablePerSecond=1.0}; result=PerformanceTestResult{duration=2s 417ms, warmup=1s 393ms, iterations=100, warmupIterations=10, ratePerSecond=41.373603640877114, histogram={8.389ms-16.78ms: 20, 16.78ms-33.55ms: 74, 33.55ms-67.11ms: 4, 67.11ms-134.2ms: 1, 134.2ms-268.4ms: 1}, cpuTotalFraction=2.552031110340871, cpuSampleFractions=[2.367437, 1.722136], minAcceptablePerSecond=1.0}
```


neykov:
Need to clear the **whole** cache on any addition/removal/update. Specs will change when transient catalog item dependencies change.
There are things outside of the control of the catalog that a spec depends on - like non-catalog locations, type registry, adding bundles.
Re-using the specs will cause objects to be shared between applications - something that could lead to subtle problems (similar to the dsl memberSpeck locking).

Having said that I believe `BrooklynCatalog.createSpec` is only used by the REST API so not as bad as it sounds. But better move the logic to the REST API?

The cache will speed up listing, but not adding items or deploying which could still take considerable time. A much better fix would be to streamline our catalog parsing logic which currently is way too complicated.

aledsage:
@neykov great comments - I've addressed them. Can you take another look please?

aledsage:
The commit `EntityExecutionManagerTest: tidy` is because `org.apache.brooklyn.core.mgmt.internal.EntityExecutionManagerTest.testGcTaskAtNormalTagLimit` failed in the jenkins PR build. But I can't reproduce that, and I don't have the full stacktrace unfortunately.

core/src/main/java/org/apache/brooklyn/core/catalog/internal/BasicBrooklynCatalog.java
```diff
@@ -1115,4 +1170,24 @@ private synchronized void loadSerializer() {
             serializer = new CatalogXmlSerializer();
     }
 
+    private static class SpecCache {
+        private final Map<String, AbstractBrooklynObjectSpec<?,?>> cache = Maps.newLinkedHashMap();
```
neykov: How do you synchronise access to the cache?

core/src/test/java/org/apache/brooklyn/core/test/qa/performance/EntityPerformanceLongevityTest.java
```diff
@@ -21,7 +21,7 @@
 import org.testng.annotations.Test;
 
 
-@Test(groups={"Integration", "Acceptance"})
+@Test(groups={"Acceptance", "Performance"})
```
neykov: If a test is part of two groups it will never execute (that's how `Broken` disables tests). Obviously it's been like this before, but better choose only one of the groups.

core/src/test/java/org/apache/brooklyn/location/ssh/SshMachineLocationPerformanceTest.java
```diff
@@ -79,7 +78,7 @@ public void testConsecutiveSmallCommands() throws Exception {
     }
 
     // Mimics SshSensorAdapter's polling
-    @Test(groups = {"Integration"})
+    @Test(groups={"Integration", "Performance"})
```
neykov: Same here

parent/pom.xml
```diff
@@ -1395,42 +1395,32 @@
...
+            <properties>
+                <includedTestGroups>Performance</includedTestGroups>
+                <excludedTestGroups>Live,WIP,Broken</excludedTestGroups>
+            </properties>
+        </profile>
+        <profile>
             <id>Acceptance</id>
             <properties>
                 <includedTestGroups>Acceptance</includedTestGroups>
                 <excludedTestGroups>Integration,Live,WIP,Broken</excludedTestGroups>
```
neykov: Why exclude `Performance` in the other groups but not here?
To have the tests which are annotated with two groups to run the of of the groups should not be excluded from the profile of the other one. But then why make `Performance` special?
I think we should just exclude `Broken` and let mixed tests execute in either of the groups.

rest/rest-resources/src/test/java/org/apache/brooklyn/rest/resources/CatalogResourcePerformanceTest.java
```diff
@@ -0,0 +1,185 @@
...
+    protected boolean useLocalScannedCatalog() {
+        return true;
+    }
+    
+    protected int numIterations() {
+        return 100;
+    }
+    
+    @Override
+    protected void initClass() throws Exception {
```
neykov: Annotate it as `@BeforeClass`?

neykov:
Looks good, some comments to consider.

core/src/test/java/org/apache/brooklyn/location/ssh/SshMachineLocationPerformanceTest.java
aledsage: This does get run if you do `mvn test -Dtest=SshMachineLocationPerformanceTest -PPerformance`, and gets excluded if you do `-PIntegration`. This is because "Integration" is not excluded from the "Performance" profile, but "Performance" *is* excluded from the "Integration" profile.

I'll comment more on that against your comment in `parent/pom.xml`.

parent/pom.xml
aledsage: Let's take this to the dev@brooklyn mailing list. I'll email something out today.

aledsage: I'll take out the testng group/profile changes. That can go in a separate commit, once discussed properly on the mailing list.

aledsage:
@neykov I believe those comments are now addressed (note the "Performance" profile stuff has been removed from this PR, pending discussion on dev@brooklyn).

Are you happy with merging this now?

neykov:
Thanks @aledsage. Looks good, merging.

PR #424, opened 2016-11-10
BROOKLYN-383: fix json serialisation of tasks

aledsage:
* Also do the same for policies, enrichers and feeds.
* Adds AbstractEntityAdjunct.getEntity() to make that easier.
* Adds TestEntity.sleepEffector

And tests xml persistence of a task, while I'm at it.


sjcorbett:
Looks good to me, happy to merge as-is.

neykov:
Nice one.

PR #425, opened 2016-11-10
Factor security group code out into separate utility class

geomacy:
Factor out the functionality for security groups.

Brings the create/delete and add/remove permissions functionality
for security groups from JcloudsLocationSecurityGroupCustomizer
out into its own class. The aim is to make it more convenient
to write other customizers etc. that might need security groups.

This is intended as nearly as possible as a behaviour-preserving
change. I have permitted myself one change to the behaviour where
this seemed sensible; the loop in `runOperationWithRetry`
was effectively an infinite loop because of its 100 iteration
exponential back-off, so I have changed this to a back-off that
tries for about five minutes.


locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/JcloudsLocationSecurityGroupEditor.java
```diff
@@ -0,0 +1,223 @@
...
+     * Constructor.
+     * @param location JClouds location where security groups will be managed.
+     * @param computeService The JClouds compute service instance.
+     * @param predicate A predicate indicating whether the customiser can retry a request to add a security group
+     * or a rule after an throwable is thrown.
+     */
+    public JcloudsLocationSecurityGroupEditor(Location location, ComputeService computeService, Predicate<Exception> predicate) {
+        this.location = location;
+        this.computeService = computeService;
+        this.locationId = this.computeService.getContext().unwrap().getId();
```
neykov: Isn't the `computeService` extracted from the `location`? Why not pass only the `location` in.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/JcloudsLocationSecurityGroupCustomizer.java
```diff
@@ -456,17 +457,16 @@ public boolean apply(final SecurityGroup input) {
...
-    private SecurityGroup createBaseSecurityGroupInLocation(String groupName, Location location, SecurityGroupExtension securityApi) {
-        SecurityGroup group = addSecurityGroupInLocation(groupName, location, securityApi);
+    private SecurityGroup createBaseSecurityGroupInLocation(String groupName, JcloudsLocationSecurityGroupEditor groupEditor) {
+        SecurityGroup group = groupEditor.createSecurityGroup(groupName).get();
 
         String groupId = group.getProviderId();
         int fromPort = 0;
-        if (isOpenstackNova(location)) {
+        if (isOpenstackNova(groupEditor.getLocation())) {
             groupId = group.getId();
```
neykov: We should report these inconsistencies to jclouds.

geomacy:
TODO add handling of exceptions,  to make operations idempotent

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/JcloudsLocationSecurityGroupCustomizer.java
```diff
@@ -320,19 +319,22 @@ void removePermissionsFromLocation(Iterable<IpPermission> permissions, final Str
...
      * @param nodeId The id of the node in question
-     * @param locationId The id of the location in question
-     * @param securityApi The API to use to list security groups
+     * @param groupEditor The id of the node in question
      * @return the security group unique to the given node, or null if one could not be determined.
      */
-    private SecurityGroup getUniqueSecurityGroupForNodeCachingSharedGroupIfPreviouslyUnknown(String nodeId, String locationId, SecurityGroupExtension securityApi) {
-        Set<SecurityGroup> groupsOnNode = securityApi.listSecurityGroupsForNode(nodeId);
+    private SecurityGroup getUniqueSecurityGroupForNodeCachingSharedGroupIfPreviouslyUnknown(String nodeId,
+             final JcloudsLocationSecurityGroupEditor groupEditor) {
```
geomacy: +100!  I'd vote 120 chars max. Much nicer when reviewing code on Github if nothing else.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/JcloudsLocationSecurityGroupEditor.java
geomacy: The compute service is actually obtained direct from the ComputeServiceRegistry, [here](https://github.com/apache/brooklyn-server/blob/master/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java#L664).   This uses the `setup` created using the flags that were passed into `JcloudsLocation.obtain()`.  If, from inside the editor code, I just call `location.getComputeService()` to get the compute service, the call sequence will pass in the config bag of the `config()` of the location.    While I don't really think that the value this obtains will be different from the value obtained using the flags from `obtain()`, I couldn't rule it out, so to preserve the semantics I kept the `computeService` here as a separate argument.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/JcloudsLocationSecurityGroupCustomizer.java
```diff
@@ -127,6 +127,7 @@ protected JcloudsLocationSecurityGroupCustomizer(String applicationId) {
     }
 
     protected JcloudsLocationSecurityGroupCustomizer(String applicationId, Supplier<Cidr> sshCidrSupplier) {
+        super();
```
aledsage: This line is redundant - it happens automatically. But I'm fine with leaving it in if you think it makes it clearer (or helps with breakpoints perhaps?)

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/JcloudsLocationSecurityGroupEditor.java
```diff
@@ -0,0 +1,303 @@
...
+     * @param location JClouds location where security groups will be managed.
+     * @param computeService The JClouds compute service instance.
+     * @param predicate A predicate indicating whether the customiser can retry a request (e.g. to add a security group
+     * or a rule) if the attempted operation throws a Throwable.
+     */
+    public JcloudsLocationSecurityGroupEditor(Location location, ComputeService computeService, Predicate predicate) {
+        this.location = location;
+        this.computeService = computeService;
+        this.locationId = this.computeService.getContext().unwrap().getId();
+        this.securityApi = this.computeService.getSecurityGroupExtension(); // TODO surely check for isPresent else throw?
```
aledsage: Interesting TODO question. I believe it was previously doing this to help in the writing of portable blueprints. You can add a `JcloudsLocationSecurityGroupCustomizer` to a YAML blueprint, and then use it in clouds that do or don't have security groups. For example, if you deployed to SoftLayer which doesn't need to open the security groups then it would still work (these would be no-ops).

However, it could be argued that a silent (except for log.warn) no-op is not necessarily the right thing, particularly if the cloud requires some different mechanism to be used (e.g. the Firewall-as-a-Service that some OpenStack clouds use).

It would seem sensible to have some level of code that truly does expect the securityGroup extension to exist, and fails otherwise. I'd be happy for this code to assume that (particularly given that some methods do `securityApi.get()`, without checking).

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/JcloudsLocationSecurityGroupEditor.java
```diff
@@ -0,0 +1,303 @@
...
+     * Constructor for editor that will retry operations upon exceptions.
+     * @param location JClouds location where security groups will be managed.
+     * @param computeService The JClouds compute service instance.
+     * @param predicate A predicate indicating whether the customiser can retry a request (e.g. to add a security group
+     * or a rule) if the attempted operation throws a Throwable.
+     */
+    public JcloudsLocationSecurityGroupEditor(Location location, ComputeService computeService, Predicate predicate) {
+        this.location = location;
+        this.computeService = computeService;
+        this.locationId = this.computeService.getContext().unwrap().getId();
```
aledsage: The `computeService` is only used for logging, and for getting the initial securitygroupExension.

The `locationId` is only used for the `getLocationId()` method, and for logging.
The `locationId` is also a bit confusing in Brooklyn, given we have a location class in Brooklyn as well. Its value is compared against "aws-ec2" - perhaps we should call it the `providerId`?

I wonder if we should pass less in - and perhaps expect the caller to have done the `computeservice.getSecurityGroupExtension().get()` call for us.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/JcloudsLocationSecurityGroupEditor.java
```diff
@@ -0,0 +1,303 @@
...
+            }
+        };
+        return runOperationWithRetry(callable);
+    }
+
+    /**
+     * Removes a security group and its permissions.
+     * @return true if the group was found and removed.
+     */
+    public Boolean removeSecurityGroup(final SecurityGroup group) {
```
aledsage: Personal preference for returning `boolean` if possible - under what circumstances would it return null?

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/JcloudsLocationSecurityGroupEditor.java
```diff
@@ -0,0 +1,303 @@
...
+                return input.getName().equals(query);
+            }
+        });
+        final ImmutableList<SecurityGroup> matches = ImmutableList.copyOf(groupsMatching);
+        if (matches.size() == 0) {
+            return Optional.absent();
+        } else if (matches.size() == 1) {
+            return Optional.of(matches.get(0));
+        } else {
+            throw new AmbiguousGroupName("Unexpected result of multiple groups matching " + name);
```
aledsage: Ah, ignore me - it's an exact name match, rather than a predicate!

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/JcloudsLocationSecurityGroupCustomizer.java
geomacy: I find myself adding `super()` at times (almost unconsciously!) as a reminder-to-self when I happen to be thinking about inherited behaviour.  It's probably a bad habit!  I'll take this out again, it can be helpful for breakpoints as you say but I don't think there's any need for that here.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/JcloudsLocationSecurityGroupEditor.java
geomacy: Have given this and the previous comment further thought and discussion with @neykov. The choice here seems to be between

1. going "higher level" and trying to pass in a Brooklyn class to the constructor, from which it could extract the location and compute service.  This would make the API JClouds-unaware, and turn the class into a more widely re-usable tool for manipulating security-group-like-things however they were implemented.
2. going "lower level" and leaving it up to the caller to get the required objects - security group extension and location, with a resultingly simpler API for this class.

While the former might be nicer to have, it looks like it would require a bit of rework in the `JcloudsLocationSecurityGroupCustomizer` and/or `JCloudsLocation` classes, as there are a couple of paths through them that require this functionality, and these get the two required objects from different sources - specifically, one path gets the location from a Template, the other from the actual node's machine location (via `JcloudsMachineLocation`).  This seems too much to take on for the sake of this PR.

In short, I'll Keep it Simple and go with option 2.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/JcloudsLocationSecurityGroupEditor.java
```diff
@@ -0,0 +1,303 @@
...
+        Exception lastException = null;
+        for (int retries = 0; retries < 100; retries++) { // TODO this will try for about 2.0e+21 years; maybe not try so hard?
+            try {
+                return operation.call();
+            } catch (Exception e) {
+                lastException = e;
+                if (isExceptionRetryable.apply(e)) {
+                    LOG.debug("Attempt #{} failed to add security group: {}", retries + 1, e.getMessage());
+                    try {
+                        Thread.sleep(backoff);
```
geomacy: I had a go at this, but it's complicated by the fact that you have to return the result of the `operation`, plus `Repeater` doesn't have an exact equivalent of `isExceptionRetryable`, so you have to capture the exception in the `repeat` body and save it for the `until` to look at.  It's doable but didn't seem compellingly simpler than what's there now.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/JcloudsLocationSecurityGroupEditor.java
```diff
@@ -0,0 +1,303 @@
...
+     * @param group The group to update
+     * @param permissions The new permissions
+     * @return The updated group with the added permissions.
+     */
+    public SecurityGroup addPermissions(final SecurityGroup group, final Iterable<IpPermission> permissions) {
+        SecurityGroup lastGroup = group;
+        for (IpPermission permission : permissions) {
+            lastGroup = addPermission(group, permission);
+        }
+        // TODO any of the calls could return null because of duplicate record. Fetch the new SG state ourselves in this case
```
geomacy: Leftover comment, will remove.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/JcloudsLocationSecurityGroupCustomizer.java
```diff
@@ -223,56 +230,64 @@ private SecurityGroup getSecurityGroup(final String nodeId, final SecurityGroupE
...
+              final Iterable<IpPermission> permissions) {
+        addPermissionsToLocationAndReturnSecurityGroup(location, permissions);
         return this;
     }
 
-    public Collection<SecurityGroup> addPermissionsToLocationAndReturnSecurityGroup(ComputeService computeService, final JcloudsMachineLocation location, final Iterable<IpPermission> permissions) {
+    public Collection<SecurityGroup> addPermissionsToLocationAndReturnSecurityGroup(
+        final JcloudsMachineLocation location, final Iterable<IpPermission> permissions) {
+
         synchronized (JcloudsLocationSecurityGroupCustomizer.class) {
```
geomacy: Possibly, but I would hope to keep this as nearly a behaviour-preserving refactor as possible, and I'd be reluctant to change this in this PR.  For now I'll add a TODO comment and the question can be discussed with @grkvlt, who I think added the synchronisation.

geomacy:
Changes made in response to the above comments, squashed and rebased against master.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/JcloudsLocationSecurityGroupCustomizer.java
neykov: I'm asking because of [this](https://github.com/apache/brooklyn-server/blob/master/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/JcloudsLocationSecurityGroupCustomizer.java#L217-L221) comment. Looks like the same restrictions will apply on the synchronizer.
Agree it can be tackled in a separate PR.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/JcloudsLocationSecurityGroupCustomizer.java
```diff
@@ -219,60 +226,62 @@ private SecurityGroup getSecurityGroup(final String nodeId, final SecurityGroupE
...
      */
-    @VisibleForTesting
-    void removePermissionsFromLocation(Iterable<IpPermission> permissions, final String nodeId, ComputeService computeService) {
-        if (!computeService.getSecurityGroupExtension().isPresent()) {
+    private void removePermissionsInternal(JcloudsMachineLocation location, Iterable<IpPermission> permissions) {
+        ComputeService computeService = location.getParent().getComputeService();
+        String nodeId = location.getNode().getId();
+
+        final Optional<SecurityGroupExtension> securityApi = computeService.getSecurityGroupExtension();
+        if (!securityApi.isPresent()) {
```
neykov: Could check with the editor instead.

neykov: nvm, that's no longer editor's responsibility

neykov:
LGTM, merging.

PR #428, opened 2016-11-11
Adding new LocationEntity for creating location specific entities

grkvlt:
A new type that allows configuring different entities depending on the location being deployed to. Uses configuration like this:

```YAML
 - type: org.apache.brooklyn.entity.stock.LocationEntity
   brooklyn.config:
     location.entity.spec.default:
       $brooklyn:entitySpec:
         type: traefik-load-balancer
     location.entity.spec.aws-ec2:
       $brooklyn:entitySpec:
         type: elastic-load-balancer
     location.entity.spec.gce:
       $brooklyn:entitySpec:
         type: google-load-balancer
```

The map entries can also be class names like `JcloudsLocation` or ISO 3166 country codes like `UK`.

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/entity/stock/LocationEntity.java
+++ b/core/src/main/java/org/apache/brooklyn/entity/stock/LocationEntity.java
@@ -1,0 +1,80 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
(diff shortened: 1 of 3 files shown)


ahgittin:
@grkvlt @aledsage @neykov love this direction - i assume this `LocationEntity` proxies the child?

i was thinking a similar but more generic pattern

```
- type: context-sensitive-proxying-parent
  switch:
    01-aws:
      condition:
        regex-match: [ $brooklyn:location(), ".*aws-ec2" ]
      spec:
        type: aws-load-balancer
    05-default:
      spec:
        type: nginx
```

not exactly sure how we want to do:
* the location reference in the DSL
* expressing conditions in this entity
* the actual condition test

feels like a `subtype-of` with `jclouds:aws-ec2` and `jclouds:aws-ec2:us-west-1` recognised as types, and maybe using typed relations to find the actual location (`$brooklyn:relation("in-location")`) ..

but that's in the future, something pragmatic like what we have here makes sense for now (though haven't reviewed code)

PR #429, opened 2016-11-11
BasicJcloudsLocationCustomizer is an EntityInitializer

sjcorbett:
Repeats https://github.com/apache/brooklyn-server/pull/168 but puts the location customizers into the entity's provisioning properties rather than setting it on the entity directly. It required moving `SoftwareProcess.PROVISIONING_PROPERTIES` to `BrooklynConfigKeys` so that it could be referenced from `brooklyn-locations-jclouds`.

### DIFF
--- a/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/BasicJcloudsLocationCustomizer.java
+++ b/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/BasicJcloudsLocationCustomizer.java
@@ -18,19 +18,73 @@
  */
 package org.apache.brooklyn.location.jclouds;
 
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+
+import org.apache.brooklyn.api.entity.Entity;
+import org.apache.brooklyn.api.entity.EntityInitializer;
+import org.apache.brooklyn.api.entity.EntityLocal;
+import org.apache.brooklyn.config.ConfigKey;
+import org.apache.brooklyn.core.config.ConfigKeys;
+import org.apache.brooklyn.core.entity.BrooklynConfigKeys;
+import org.apache.brooklyn.core.location.LocationConfigKeys;
+import org.apache.brooklyn.core.objs.BasicConfigurableObject;
+import org.apache.brooklyn.util.core.config.ConfigBag;
 import org.jclouds.compute.ComputeService;
 import org.jclouds.compute.domain.Template;
 import org.jclouds.compute.domain.TemplateBuilder;
 import org.jclouds.compute.options.TemplateOptions;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
-import org.apache.brooklyn.api.entity.Entity;
-import org.apache.brooklyn.core.location.LocationConfigKeys;
-import org.apache.brooklyn.core.objs.BasicConfigurableObject;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Iterables;
+import com.google.common.collect.Lists;
 
 /**
- * A default no-op implementation, which can be extended to override the appropriate methods.
+ * A default no-op location customizer, which can be extended to override the appropriate methods.
+ * <p>
+ * When the class is used as an {@link EntityInitializer} it inserts itself into the entity's
+ * {@link JcloudsLocationConfig#JCLOUDS_LOCATION_CUSTOMIZERS} under the
+ * {@link BrooklynConfigKeys#PROVISIONING_PROPERTIES} key.
  */
-public class BasicJcloudsLocationCustomizer extends BasicConfigurableObject implements JcloudsLocationCustomizer {
+public class BasicJcloudsLocationCustomizer extends BasicConfigurableObject implements JcloudsLocationCustomizer, EntityInitializer {
+
+    private static final Logger LOG = LoggerFactory.getLogger(BasicJcloudsLocationCustomizer.class);
+
+    public BasicJcloudsLocationCustomizer() {
+        this(ImmutableMap.<String, String>of());
+    }
+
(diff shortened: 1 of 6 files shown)


locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/BasicJcloudsLocationCustomizer.java
```diff
@@ -18,19 +18,72 @@
...
+    public void apply(EntityLocal entity) {
+        ConfigKey<Object> subkey = BrooklynConfigKeys.PROVISIONING_PROPERTIES.subKey(JcloudsLocationConfig.JCLOUDS_LOCATION_CUSTOMIZERS.getName());
+        // newInstance handles the case that provisioning properties is null.
+        ConfigBag provisioningProperties = ConfigBag.newInstance(entity.config().get(BrooklynConfigKeys.PROVISIONING_PROPERTIES));
+        Collection<JcloudsLocationCustomizer> existingCustomizers = provisioningProperties.get(JcloudsLocationConfig.JCLOUDS_LOCATION_CUSTOMIZERS);
+        List<? super JcloudsLocationCustomizer> merged;
+        if (existingCustomizers == null) {
+            merged = ImmutableList.<JcloudsLocationCustomizer>of(this);
+        } else {
+            merged = Lists.newArrayListWithCapacity(1 + existingCustomizers.size());
```
neykov: Personal preference: I'd use `ImmutableList.builder()` instead (to avoid playing with sizes).

locations/jclouds/src/test/java/org/apache/brooklyn/location/jclouds/BasicJcloudsLocationCustomizerTest.java
```diff
@@ -0,0 +1,81 @@
...
+                        ImmutableList.of(new BasicJcloudsLocationCustomizer())));
+        Object object = entity.config().get(BrooklynConfigKeys.PROVISIONING_PROPERTIES.subKey(
+                JcloudsLocationConfig.JCLOUDS_LOCATION_CUSTOMIZERS.getName()));
+        assertNotNull(object, "expected value for customizers in " + entity.config().get(BrooklynConfigKeys.PROVISIONING_PROPERTIES));
+        assertTrue(object instanceof Collection, "expected collection, got " + object.getClass());
+        Collection<JcloudsLocationCustomizer> customizers = (Collection<JcloudsLocationCustomizer>) object;
+        assertEquals(customizers.size(), 2, "expected two customizers in " + Iterables.toString(customizers));
+        assertTrue(customizers.contains(testCustomizer), "expected to find testCustomizer in " + Iterables.toString(customizers));
+    }
+
```
neykov: Would be useful to test `addInitializer` on already managed entities.

neykov: Also testing `provisioning.properties` inheritance - for example attaching to a cluster.

neykov:
LGTM overall. Would like to see better initializer config integration.

locations/jclouds/src/test/java/org/apache/brooklyn/location/jclouds/BasicJcloudsLocationCustomizerTest.java
sjcorbett: How can you add initialisers to already managed entities? Either I'm missing something obvious or it's not possible.

sjcorbett:
@neykov thanks for your comments. I've removed `params` and added a test that confirms that the initialiser is applied to cluster members. Can you clarify what you meant when you said that it would be useful to test `addInitializer`?

locations/jclouds/src/test/java/org/apache/brooklyn/location/jclouds/BasicJcloudsLocationCustomizerTest.java
neykov: Thought that there's an `addInitializer` method on the entity, but that's not the case.
To add an initializer to an existing entity create the initializer and call `apply` on the entity.

neykov:
Would be useful to test applying the initializer to an existing entity (before it's started obviously). To add an initializer to an entity just call its `apply` method on the entity.
The difference is subtle - initializers in the spec are applied on the entity before they are managed. Applying it after having a managed entity could lead to different results.
That's minor though, feel free to ignore.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/BasicJcloudsLocationCustomizer.java
```diff
@@ -18,19 +18,73 @@
...
+
+    public BasicJcloudsLocationCustomizer(final ConfigBag params) {
+        for (Map.Entry<String, Object> entry : params.getAllConfig().entrySet()) {
+            config().set(ConfigKeys.newConfigKey(Object.class, entry.getKey()), entry.getValue());
+        }
+    }
+
+    @Override
+    public void apply(EntityLocal entity) {
+        ConfigKey<Object> subkey = BrooklynConfigKeys.PROVISIONING_PROPERTIES.subKey(JcloudsLocationConfig.JCLOUDS_LOCATION_CUSTOMIZERS.getName());
```
geomacy: very minor - maybe call this `customizers` rather than just `subkey`, to highlight its role

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/BasicJcloudsLocationCustomizer.java
neykov: I was referring to the code that follows as well. What I meant was:
```
merged = ImmutableList.builder()
                      .addAll(existingCustomizers)
                      .add(this)
                      .build();
```

neykov:
Looks good, merging.

PR #430, opened 2016-11-13
Use CSRF headers

ahgittin:
Adds a filter which returns and requires special cookies/headers to protect against forged cross-site requests.  Extensive documentation in `CsrfTokenFilter`.  @neykov or @m4rkmckenna or @geomacy you might be best placed to check this.

Also small tidy to logout process in https://github.com/apache/brooklyn-server/commit/e9aecbac1ccdebec1cb07f0f44c5c73f0137c64d -- would be useful if someone familiar with the thinking behind the existing process (switching to user) gives it a quick eyeball (probably @neykov or @bostko ?)


geomacy:
To the best of my knowledge this looks good, from visual inspection and from test. (A malicious web page posting to the server got the 401 response generated by the CSRF filter.)  +1

PR #431, opened 2016-11-14
SoftwareProcess integration tests: use INSTALL_UNIQUE_LABEL

aledsage:
Fixes VanillaSoftwareProcessStreamsIntegrationTest.testGetsStreams,
for if run multiple times on the same machine.

@neykov can you take a quick look please?

### DIFF
--- a/software/base/src/test/java/org/apache/brooklyn/entity/software/base/VanillaSoftwareProcessIntegrationTest.java
+++ b/software/base/src/test/java/org/apache/brooklyn/entity/software/base/VanillaSoftwareProcessIntegrationTest.java
@@ -51,7 +51,7 @@ public class VanillaSoftwareProcessIntegrationTest extends BrooklynAppLiveTestSu
         super.setUp();
         localhost = app.getManagementContext().getLocationRegistry().getLocationManaged("localhost");
         
-        runRecord = Files.createTempFile("testVanillaSoftwareProcess-runRecord", ".txt");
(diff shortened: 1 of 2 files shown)


software/base/src/test/java/org/apache/brooklyn/entity/software/base/VanillaSoftwareProcessIntegrationTest.java
```diff
@@ -51,7 +51,7 @@ public void setUp() throws Exception {
         super.setUp();
         localhost = app.getManagementContext().getLocationRegistry().getLocationManaged("localhost");
         
-        runRecord = Files.createTempFile("testVanillaSoftwareProcess-runRecord", ".txt");
+        runRecord = Files.createTempFile("testVanillaSoftwareProcess-runRecord-"+Identifiers.makeRandomId(8), ".txt");
```
neykov: `createTempFile` is supposed to ensure unique name, is this really needed?

neykov:
Good to merge, but don't really understand why randomize the temporary files?

software/base/src/test/java/org/apache/brooklyn/entity/software/base/VanillaSoftwareProcessIntegrationTest.java
aledsage: Not sure what guarantees `createTempFile` gives. If the previous temp files have been deleted, then might it reuse the name? Particularly if we run two tests within the same millisecond, which admittedly seems unlikely when they are integration tests.

PR #433, opened 2016-11-14
Fix dependency URI path for downstream projects + Karaf feature verification

tbouron:


tbouron:
For some reason, the official maven karaf plugin cannot verify this feature if the quotes are not single quotes AND properly URI encoded, hence the `%27`

karaf/features/src/main/feature/feature.xml
```diff
@@ -256,7 +256,7 @@
             https://github.com/square/okhttp/pull/2246
         -->
         <!-- TODO use ServiceMix bundle for okhttp when dependency on okio is fixed there -->
-        <bundle dependency="true">wrap:mvn:com.squareup.okhttp/okhttp/2.2.0$Import-Package=org.apache.http.*;resolution:=optional,android.util.*;resolution:=optional,okio;version="[1.2.0, 1.3.0)",*</bundle>
+        <bundle dependency="true">wrap:mvn:com.squareup.okhttp/okhttp/2.2.0$Import-Package=org.apache.http.*;resolution:=optional,android.util.*;resolution:=optional,okio;version=%27[1.2.0,1.3.0)%27,*</bundle>
```
aledsage: @tbouron Why did you go for single quotes rather than double quotes (`%22`), out of interest?

aledsage:
I understand that this has been confirmed manually with a mvn build, checking in karaf that okhttp bundle was added correctly. Therefore merging now.

karaf/features/src/main/feature/feature.xml
tbouron: Tested with encoded double quotes (`%22`) but the feature verification was still failing. Not sure why unfortunately

PR #435, opened 2016-11-14
BROOKLYN-325: alert if provisioning/termination aborted on rebind

aledsage:


core/src/main/java/org/apache/brooklyn/core/entity/internal/AttributesInternal.java
```diff
@@ -0,0 +1,52 @@
+/*
+ * or more contributor license agreements.  See the NOTICE file
```
iyovcheva: First line is missing from the Licence

neykov: Good spot, interestingly rat doesn't complain.

core/src/main/resources/org/apache/brooklyn/core/mgmt/persist/deserializingClassRenames.properties
```diff
@@ -1434,3 +1434,5 @@ brooklyn.test.entity.BlockingEntityImpl
 org.apache.brooklyn.config.ConfigInheritance$None                                : org.apache.brooklyn.config.ConfigInheritance$Legacy$None
 org.apache.brooklyn.config.ConfigInheritance$Always                              : org.apache.brooklyn.config.ConfigInheritance$Legacy$Always
 org.apache.brooklyn.config.ConfigInheritance$Merged                              : org.apache.brooklyn.config.ConfigInheritance$Legacy$Merged
+
+org.apache.brooklyn.entity.software.base.lifecycle.MachineLifecycleEffectorTasks$ProvisioningTaskState : org.apache.brooklyn.core.entity.internal.AttributesInternal$ProvisioningTaskState
```
neykov: Can you add a comment "since 0.10.0" so we can start removing entries at some point.

core/src/main/java/org/apache/brooklyn/core/mgmt/rebind/BasicEntityRebindSupport.java
```diff
@@ -245,14 +249,52 @@ protected void instanceRebind(AbstractBrooklynObject instance) {
...
+        super.instanceRebind(instance);
+    }
+    
+    protected void markTransitioningEntityOnFireOnRebind(EntityInternal entity, Lifecycle expectedState) {
+        LOG.warn("Entity {} being marked as on-fire because it was in state {} on rebind; indicators={}", new Object[] {entity, expectedState, entity.getAttribute(Attributes.SERVICE_NOT_UP_INDICATORS)});
+        ServiceStateLogic.setExpectedState(entity, Lifecycle.ON_FIRE);
+        ServiceStateLogic.ServiceNotUpLogic.updateNotUpIndicator(
+                entity, 
+                "Task aborted on rebind", 
+                "Set to on-fire (from previous expected state "+expectedState+") because tasks aborted on rebind");
```
neykov: Tasks get aborted during shutdown.

core/src/main/java/org/apache/brooklyn/core/mgmt/rebind/BasicEntityRebindSupport.java
```diff
@@ -245,14 +249,52 @@ protected void instanceRebind(AbstractBrooklynObject instance) {
...
+        ((EntityInternal)entity).sensors().remove(AttributesInternal.INTERNAL_PROVISIONING_TASK_STATE);
+        ((EntityInternal)entity).sensors().remove(AttributesInternal.INTERNAL_TERMINATION_TASK_STATE);
+        
+        super.instanceRebind(instance);
+    }
+    
+    protected void markTransitioningEntityOnFireOnRebind(EntityInternal entity, Lifecycle expectedState) {
+        LOG.warn("Entity {} being marked as on-fire because it was in state {} on rebind; indicators={}", new Object[] {entity, expectedState, entity.getAttribute(Attributes.SERVICE_NOT_UP_INDICATORS)});
+        ServiceStateLogic.setExpectedState(entity, Lifecycle.ON_FIRE);
+        ServiceStateLogic.ServiceNotUpLogic.updateNotUpIndicator(
```
neykov: Who gets to clean these entries? Does a start/restart get the entity in a healthy state?

core/src/main/java/org/apache/brooklyn/core/entity/internal/AttributesInternal.java
```diff
@@ -0,0 +1,52 @@
...
+    /**
+     * Used only internally by {@link org.apache.brooklyn.entity.software.base.lifecycle.MachineLifecycleEffectorTasks}
+     * to track provisioning/termination. This is used so the machine can be terminated if stopped while opaque provision
+     * call is being made; and is used to report if termination was prematurely aborted (e.g. during Brooklyn restart).
+     */
+    @Beta
+    @VisibleForTesting
+    public enum ProvisioningTaskState {
+        RUNNING,
+        DONE;
```
neykov: What's the point of a `DONE`? Isn't `null` the same?

neykov:
Good clean up. Some questions above.

iyovcheva:
Looks good. Tested all the cases and seems to work well.

core/src/main/resources/org/apache/brooklyn/core/mgmt/persist/deserializingClassRenames.properties
aledsage: @neykov Agree that would be useful. Do you think we should add it in this properties file, or just on the javadoc of the class in question?

For this file, there are a bunch of entries higher up that are also "since 0.10.0" so feels like we should figure out where and put the comment on that group.

It gets a little more complicated if we want to group them with finer granularity (e.g. some downstream projects like AMP have been taking timestamped early-access releases of Brooklyn. But here we should probably just focus on the Brooklyn GA releases.

I suggest we do that separately (figuring out where the changes since 0.9.0 begin!)

neykov: I think it will be useful to have a marker here as well.  Agree it can be done separately. A good time to begin will be post-0.10.0 release. Then can treat all previous entries as added in 0.10.0.

core/src/main/java/org/apache/brooklyn/core/mgmt/rebind/BasicEntityRebindSupport.java
aledsage: I've changed wording in the message.

---
I also looked again at how we set `INTERNAL_PROVISIONING_TASK_STATE`. In `org.apache.brooklyn.entity.software.base.lifecycle.MachineLifecycleEffectorTasks.ProvisionMachineTask`, there is a try-finally block, where it sets the sensor to "done" in the finally block. If that managed to execute + get persisted, then on rebind we wouldn't know that it was still executing!

We get away with this because the `managementContext.terminate()` stops the `rebindManager` and `BrooklynStorage`, before it shuts down the execution manager. Therefore our thread doesn't get interrupted (triggering the finally block) until after persistence has been disabled.

core/src/main/java/org/apache/brooklyn/core/mgmt/rebind/BasicEntityRebindSupport.java
aledsage: Good question. No, start/restart won't clear that not-up-indicator.

In `brooklyn-jsgui` (i.e. the web-console) there is a "reset problems" button in the advanced tab. This makes a couple of API calls for the entity. It sets `service.notUp.indicators` and `service.problems` to be empty maps.

I think that is probably the only way to do this currently!

We need to at least update the docs about this. What would be a better way? Should start() and restart() effectors clear all not-up-indicators? That feels fiddly - it might cause the entity to immediately report `serviceUp=true` (which is why `SoftwareProcessImpl.initEnrichers()` calls `ServiceNotUpLogic.updateNotUpIndicator(this, SERVICE_PROCESS_IS_RUNNING, "No information yet on whether this service is running")`).

Thoughts?

core/src/main/java/org/apache/brooklyn/core/entity/internal/AttributesInternal.java
aledsage: Good point; I'll delete "done" (it won't be in persisted state because `INTERNAL_PROVISIONING_TASK_STATE` used to have `AttributeSensor.SensorPersistenceMode.NONE`), and change it to use null.

aledsage:
@neykov comments addressed in the new final commit (and rebased against master).

core/src/main/resources/org/apache/brooklyn/core/mgmt/persist/deserializingClassRenames.properties
neykov: On a tangential note, we should delete all anonymous classes that already have named replacements and use this mechanism to let rebind know about them.

core/src/main/java/org/apache/brooklyn/core/mgmt/rebind/BasicEntityRebindSupport.java
neykov: I think `start` & `restart` should clear these particular values, not all. Since the errors are triggered by start, restart, stop actions - these effectors are kind of responsible. So a subsequent start will reset the state and the entity will become healthy.

neykov:
Good to merge. I'll leave it to you to decide whether to clean the indicators in start/restart in this PR. Or leave it to the user to do it from the UI.

aledsage:
Thanks, I'll not include clearing the indicators in this PR.

PR #437, opened 2016-11-15
Asserts: use DEFAULT_LONG_TIMEOUT in eventually()

aledsage:
And update javadoc to say why. Also makes the timeout configurable
via `-Dbrooklyn.test.defaultTimeout=1s`.

### DIFF
--- a/utils/common/src/main/java/org/apache/brooklyn/test/Asserts.java
+++ b/utils/common/src/main/java/org/apache/brooklyn/test/Asserts.java
@@ -74,17 +74,45 @@ import groovy.lang.Closure;
 public class Asserts {
 
     /** 
-     * Timeout for use when something should happen within several seconds,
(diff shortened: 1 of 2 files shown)


utils/common/src/main/java/org/apache/brooklyn/test/Asserts.java
```diff
@@ -74,17 +74,45 @@
...
+    public static final Duration DEFAULT_LONG_TIMEOUT;
+
+    static {
+        String defaultTimeout = System.getProperty("brooklyn.test.defaultTimeout");
+        if (defaultTimeout == null){
+            DEFAULT_LONG_TIMEOUT = Duration.THIRTY_SECONDS;
+        } else {
+            DEFAULT_LONG_TIMEOUT = Duration.of(defaultTimeout);
+        }
+    }
```
neykov: Good idea.

neykov:
LGTM, merging.

PR #440, opened 2016-11-16
BROOKLYN-345: persist brooklyn.parameters (fixes ServiceFailureDetector rebind)

aledsage:


neykov:
Compile time errors in the jenkins build:
```
[ERROR] /home/jenkins/jenkins-slave/workspace/brooklyn-server-pull-requests/camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/ServiceFailureDetectorYamlTest.java:[31,39] cannot find symbol
  symbol:   class SensorEventPredicates
  location: package org.apache.brooklyn.core.sensor
```

camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/ServiceFailureDetectorYamlTest.java
```diff
@@ -0,0 +1,189 @@
...
+            "      default: 1m",
+            "    services:",
+            "    - type: " + TestEntity.class.getName(),//FailingEntity.class.getName(),
+            "      brooklyn.enrichers:",
+            "      - type: " + ServiceFailureDetector.class.getName(),
+            "        brooklyn.config:",
+            "          serviceOnFire.stabilizationDelay: $brooklyn:config(\"custom.stabilizationDelay\")",
+            "          entityFailed.stabilizationDelay: $brooklyn:config(\"custom.stabilizationDelay\")",
+            "          entityRecovered.stabilizationDelay: $brooklyn:config(\"custom.stabilizationDelay\")",
+            "          entityFailed.republishTime: $brooklyn:config(\"custom.republishTime\")");
```
neykov: How does this even work? Default values are not inherited.
It's probably because the root parameters are merged to the top-level entity on unwrap. So in the end it's identical to the above test case.

core/src/main/java/org/apache/brooklyn/core/mgmt/rebind/dto/BasicEntityMemento.java
```diff
@@ -139,17 +142,28 @@ protected BasicEntityMemento(Builder builder) {
...
+        
         if (configByKey!=null) {
             configKeys = Maps.newLinkedHashMap();
             config = Maps.newLinkedHashMap();
+            
+            for (ConfigKey<?> key : allConfigKeys) {
+                if (!key.equals(staticConfigKeys.get(key.getName()))) {
+                    configKeys.put(key.getName(), key);
+                }
+            }
```
neykov: If there's no `configByKey` you might still want to initialize `configKeys`.

neykov: This will remove entries which override the default value. I'd use reference comparison here.

policy/src/main/java/org/apache/brooklyn/policy/ha/ServiceFailureDetector.java
```diff
@@ -230,6 +237,7 @@ protected void setActualState(Maybe<Lifecycle> state) {
                     }
                     lastPublished = LastPublished.FAILED;
                     entity.sensors().emit(HASensors.ENTITY_FAILED, new HASensors.FailureDescriptor(entity, getFailureDescription(now)));
+                    requestPersist();
```
neykov: Oh that's sneaky. Can you use config instead? (not sure if config change triggers persistence on adjuncts).

neykov: Yes it does.

utils/common/src/main/java/org/apache/brooklyn/util/time/Time.java
```diff
@@ -157,7 +157,7 @@ public static String makeTimeStringRounded(long t, TimeUnit unit) {
         return makeTimeStringNanoRounded(nanos);
     }
     public static String makeTimeStringRounded(Stopwatch timer) {
-        return makeTimeStringRounded(timer.elapsed(TimeUnit.MILLISECONDS), TimeUnit.MILLISECONDS);
+        return (timer == null) ? null : makeTimeStringRounded(timer.elapsed(TimeUnit.MILLISECONDS), TimeUnit.MILLISECONDS);
```
neykov: Can you add `@Nullable`, useful when looking at the signature from the call site.

neykov:
Finished review. Happy to merge after resolving the "lost default overrides" problem.

camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/ServiceFailureDetectorYamlTest.java
aledsage: You're right. I change it to have two children, and the test failed (because it didn't merge to the top-level). I've therefore also changed the yaml to use `$brooklyn:scopeRoot().config(...)`.

neykov:
Failing jenkins tests:
```

Failed tests: 
  ConfigNestedYamlTest.testCatalogParameterFromSuperYamlTypeAsSoftware:90->doTestWithBlueprint:101->checkEntity:122 missing 'field' in {} expected [true] but found [false]
org.apache.brooklyn.camp.brooklyn.catalog.SpecParameterUnwrappingTest.testDependantCatalogMergesParameters(org.apache.brooklyn.camp.brooklyn.catalog.SpecParameterUnwrappingTest)
  Run 1: SpecParameterUnwrappingTest.testDependantCatalogMergesParameters:247 expected [4] but found [3]
  Run 2: SpecParameterUnwrappingTest.testDependantCatalogMergesParameters:250 expected [3] but found [2]
  Run 3: SpecParameterUnwrappingTest.testDependantCatalogMergesParameters:253 expected [8] but found [7]
```

aledsage:
Thanks @neykov - `ConfigNestedYamlTest.testCatalogParameterFromSuperYamlTypeAsSoftware` is a very annoying and valid failure!

There is a `MapConfigKey` defined on the java type (`SoftwareProcess.TEMPLATE_SUBSTITUTIONS`). In the yaml (`config-nested-tests,bom`), the super-type declares a brooklyn.parameters for this: brooklyn.parameters: [ {name: template.substitutions, type: java.util.Map} ]`.

Note the difference in type for the config keys: the first is a `MapConfigKey` and the latter is a `BasicConfigKey<java.util.Map>`.

When the config values are written into the entity's spec, it uses the `MapConfigKey` so the value of `{field: val}` is actually written as `{template.substitutions.field: val}`.

When the config value is later read, it uses the `BasicConfigKey<java.util.Map>`, so it just looks up `template.substitutions` and thus returns an empty map.

---
A few things spring to mind:
1. don't try to change the type of a config key when you're overriding it in brooklyn.parameters. (However, the user doesn't think they've done this - it's of type map!)
2. write creating the spec, use the same config key as will subsequently be used for reading the value back out again.
3. I really don't like the `MapConfigKey` stuff. Particularly now that we have capabilities for defining the merge-semantics of map config keys, the need for the `MapConfigKey` is greatly reduced.

Let's not try to tackle (3) right now, but that is one for the mailing list!

(2) feels like the right thing to do. But it's fiddly. We're in the middle of the stacktrace below for (one of the places?) where we're writing the config value using the `MapConfigKey`:
```
"main" prio=5 tid=0x00007fdbab000000 nid=0x1703 at breakpoint[0x0000700000217000]
   java.lang.Thread.State: RUNNABLE
        at org.apache.brooklyn.api.internal.AbstractBrooklynObjectSpec.configure(AbstractBrooklynObjectSpec.java:293)
        at org.apache.brooklyn.camp.brooklyn.spi.creation.BrooklynComponentTemplateResolver.configureEntityConfig(BrooklynComponentTemplateResolver.java:337)
        at org.apache.brooklyn.camp.brooklyn.spi.creation.BrooklynComponentTemplateResolver.decorateSpec(BrooklynComponentTemplateResolver.java:259)
        at org.apache.brooklyn.camp.brooklyn.spi.creation.BrooklynComponentTemplateResolver.populateSpec(BrooklynComponentTemplateResolver.java:250)
        at org.apache.brooklyn.camp.brooklyn.spi.creation.BrooklynComponentTemplateResolver.resolveSpec(BrooklynComponentTemplateResolver.java:189)
        at org.apache.brooklyn.camp.brooklyn.spi.creation.BrooklynAssemblyTemplateInstantiator.buildTemplateServicesAsSpecs(BrooklynAssemblyTemplateInstantiator.java:114)
        at org.apache.brooklyn.camp.brooklyn.spi.creation.BrooklynAssemblyTemplateInstantiator.createServiceSpecs(BrooklynAssemblyTemplateInstantiator.java:73)
        at org.apache.brooklyn.camp.brooklyn.spi.creation.BrooklynAssemblyTemplateInstantiator.createApplicationSpec(BrooklynAssemblyTemplateInstantiator.java:90)
        at org.apache.brooklyn.camp.brooklyn.spi.creation.CampResolver.createEntitySpecFromServicesBlock(CampResolver.java:145)
        at org.apache.brooklyn.camp.brooklyn.spi.creation.CampResolver.createSpecFromFull(CampResolver.java:111)
        at org.apache.brooklyn.camp.brooklyn.spi.creation.CampToSpecTransformer.createCatalogSpec(CampToSpecTransformer.java:102)
        at org.apache.brooklyn.core.catalog.internal.BasicBrooklynCatalog$1.apply(BasicBrooklynCatalog.java:331)
        at org.apache.brooklyn.core.catalog.internal.BasicBrooklynCatalog$1.apply(BasicBrooklynCatalog.java:1)
        at org.apache.brooklyn.core.plan.PlanToSpecFactory.attemptWithLoaders(PlanToSpecFactory.java:126)
        at org.apache.brooklyn.core.plan.PlanToSpecFactory.attemptWithLoaders(PlanToSpecFactory.java:118)
        at org.apache.brooklyn.core.catalog.internal.BasicBrooklynCatalog.internalCreateSpecLegacy(BasicBrooklynCatalog.java:328)
        at org.apache.brooklyn.core.catalog.internal.BasicBrooklynCatalog$PlanInterpreterGuessingType.attemptType(BasicBrooklynCatalog.java:808)
        at org.apache.brooklyn.core.catalog.internal.BasicBrooklynCatalog$PlanInterpreterGuessingType.reconstruct(BasicBrooklynCatalog.java:735)
        at org.apache.brooklyn.core.catalog.internal.BasicBrooklynCatalog.collectCatalogItems(BasicBrooklynCatalog.java:508)
        at org.apache.brooklyn.core.catalog.internal.BasicBrooklynCatalog.collectCatalogItems(BasicBrooklynCatalog.java:476)
        at org.apache.brooklyn.core.catalog.internal.BasicBrooklynCatalog.collectCatalogItems(BasicBrooklynCatalog.java:383)
        at org.apache.brooklyn.core.catalog.internal.BasicBrooklynCatalog.collectCatalogItems(BasicBrooklynCatalog.java:372)
        at org.apache.brooklyn.core.catalog.internal.BasicBrooklynCatalog.addItems(BasicBrooklynCatalog.java:929)
        at org.apache.brooklyn.core.catalog.internal.BasicBrooklynCatalog.addItems(BasicBrooklynCatalog.java:1)
        at org.apache.brooklyn.camp.brooklyn.AbstractYamlTest.addCatalogItems(AbstractYamlTest.java:199)
        at org.apache.brooklyn.camp.brooklyn.ConfigNestedYamlTest.testCatalogParameterFromSuperYamlTypeAsSoftware(ConfigNestedYamlTest.java:89)
```

Maybe this is one to document and defer? Thoughts?

aledsage:
The problem that I'm looking at fixing for `ConfigNestedYamlTest.testCatalogParameterFromSuperYamlTypeAsSoftware` should also fix https://issues.apache.org/jira/browse/BROOKLYN-328.

aledsage:
The latest changes (particularly for the commit `entity.setConfig(key,v): lookup ownKey`) are motivated by fixing the tests that failed after the config-param override fix. Investigating that led to fixing https://issues.apache.org/jira/browse/BROOKLYN-328.

I talked through these changes with @neykov before making them, so hopefully it's not too controversial!

ahgittin:
i'll have a look at this ... likely some conflicts with https://github.com/apache/brooklyn-server/pull/462

camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/creation/BrooklynComponentTemplateResolver.java
```diff
@@ -374,15 +374,45 @@ protected ConfigInheritance getDefaultConfigInheritance() {
...
+        // 
+        // However, that is hard/fiddly because of the way a config key can be referenced by
+        // its real name or flag-name.
+        // 
+        // I wonder if this is fundamentally broken (and I really do dislike our informal use 
+        // of aliases). Consider a configKey with name A and alias B. The bagFlags could have 
+        // {A: val1, B: val2}. There is no formal definition of which takes precedence. We'll add 
+        // both to the entity's configBag, without any warning - it's up to the `config().get()` 
+        // method to then figure out what to do. It gets worse if there is also a ConfigKey with 
+        // name "B" the "val2" then applies to both!
```
ahgittin: I think things should be de-aliased when writing to the config map, so config map has canonical config key names _only_ -- ie it should not be up to `config().get()` to figure out if any aliases are in effect.

Is this not the case?

Agree if we have `{A: val1, B: val2 }` as flags where `A` and `B` are names/aliases for a given config key there is ambiguity -- this is possible even with `{A: val1, A: val2 }`.  I'm not too worried about either of these, fair enough to say it's not guaranteed which is taken.

BTW good code change to treat the parameter definitions of config keys as the preferred ones.  The only tweak I might suggest is that if the parameter fields are incomplete they might be merged with those defined on the type (eg description, default value) so that it isn't necessary to re-declare them as parameters.

ahgittin:
looks good, tests passing.  merging.  some follow-up to do in #462 which i'll do next.

core/src/main/java/org/apache/brooklyn/core/config/internal/AbstractConfigMapImpl.java
```diff
@@ -163,13 +163,18 @@ public ConfigBag getLocalConfigBag() {
...
 
-    public Object setConfig(ConfigKey<?> key, Object v) {
-        Object val = coerceConfigVal(key, v);
+    public Object setConfig(final ConfigKey<?> key, Object v) {
+        // Use our own key for writing, (e.g. in-case it should (or should not) be a structured key like MapConfigKey).
+        // This is same logic as for getConfig, except we only have to look at our own container.
+        ConfigKey<?> ownKey = getKeyAtContainer(getContainer(), key);
+        if (ownKey==null) ownKey = key;
+
+        Object val = coerceConfigVal(ownKey, v);
```
neykov: I guess it's useful for catching errors early. On the other hand the uncertainty of what config key (and type) will be used to get values vs setting them makes it harder to reason about the behaviour.
Say we are setting a double with an integer key - it will throw away the fractional part. Then getting it back with a `Double` key.

camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/creation/BrooklynComponentTemplateResolver.java
```diff
@@ -374,15 +374,45 @@ protected ConfigInheritance getDefaultConfigInheritance() {
...
+        //
+        // I plan to propose a change on dev@brooklyn, to replace `@SetFromFlag`!
+        
         Set<FlagConfigKeyAndValueRecord> allKeys = MutableSet.of();
```
ahgittin: agree -- applying in #462

camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/creation/BrooklynComponentTemplateResolver.java
ahgittin: added above as a code comment in #462

core/src/main/java/org/apache/brooklyn/core/config/internal/AbstractConfigMapImpl.java
ahgittin: that's true @neykov re putting with a key with a different type to what you get with.  but it would be a weird thing to do to put with an incompatible key (normally we'll put with the same key or one which uses a supertype eg `Object`) and not surprising if it's coerced to int on get then double on put.  given that this has been the behaviour, and catching errors early is useful, i definitely don't see a compelling reason to change.

PR #441, opened 2016-11-16
Fixing HttpResponse headers value

ygy:


core/src/test/java/org/apache/brooklyn/util/executor/HttpExecutorImplTest.java
```diff
@@ -71,6 +75,36 @@ public void afterMethod() throws Exception {
...
+                .method("GET")
+                .uri(baseUrl.toURI())
+                .build();
+        HttpResponse executorResponse = executor.execute(executorRequest);
+        assertTrue(executorResponse.headers().containsKey(HTTP_RESPONSE_CUSTOM_HEADER_KEY));
+        assertTrue(Iterables.getOnlyElement(executorResponse.headers().get(HTTP_RESPONSE_HEADER_KEY)).equals(HTTP_RESPONSE_HEADER_VALUE));
+
+        assertTrue(executorResponse.headers().containsKey(HTTP_RESPONSE_HEADER_KEY));
+        assertTrue(Iterables.getOnlyElement(executorResponse.headers().get(HTTP_RESPONSE_CUSTOM_HEADER_KEY)).equals(HTTP_RESPONSE_CUSTOM_HEADER_VALUE));
+    }
```
neykov: The essence of the test is line `90`. You can cut it down to verifying that `httpResponse.headers()` contains the expected values. Everything else just gets that value through a number of hoops before you check it at the end.

utils/common/src/main/java/org/apache/brooklyn/util/http/executor/HttpResponse.java
```diff
@@ -60,7 +60,7 @@ public Builder reasonPhrase(@Nullable String val) {
         }
 
         public Builder headers(Multimap<String, String> val) {
-            headers.putAll(headers);
+            headers.putAll(val);
```
neykov: Nice find :)

neykov:
Could simplify the test, otherwise good to go.

ygy:
@aledsage @neykov The test has been simplified

PR #442, opened 2016-11-16
BROOKLYN-379: Fix rebind when bundle prefixes are used to load classes

aledsage:
This replaces https://github.com/apache/brooklyn-server/pull/396

See also https://github.com/neykov/brooklyn-server/pull/4 (but I'll replay all the comments from that PR into this PR).


core/src/main/java/org/apache/brooklyn/core/mgmt/ha/OsgiManager.java
```diff
@@ -189,10 +193,11 @@ public static boolean isBundleNameEqualOrAbsent(CatalogBundle bundle, Bundle b)
                 Maybe<Bundle> bundle = findBundle(osgiBundle);
                 if (bundle.isPresent()) {
                     Bundle b = bundle.get();
-                    Class<T> clazz;
+                    Optional<String> strippedType = osgiClassPrefixer.stripMatchingPrefix(b, type);
+                    String typeToLoad = strippedType.isPresent() ? strippedType.get() : type;
```
aledsage: @aledsage wrote 9 days ago:
Not sure if this is the right place for the fix! Arguably if it has the prefix, we should only look in one bundle (or multiple versions of the bundle), rather than iterating over all bundles.

@neykov responded 7 days ago:
If a class has bundle prefix obviously it's not relying on the libraries functionality to load it. Still could be useful to have this to avoid version mismatch (half of the blueprint is coming from libraries section the other half is using prefixes because of copy&paste).
I'd structure it a bit differently - if there's a prefix short-circuit and don't try to load it from non-matching bundles, but can't get away from iterating over them. And it's not all, just those in the libraries section so it's fine.
The prefix could include version as well - would be nice to support it here as well (by skipping non-matching versions) to avoid introducing subtle inconsistencies in the prefix behaviour.

@neykov responded 7 days ago:
This makes more sense now that I saw the commit comment. This is called by persistence, not only by yaml parsing code.
Another approach would be to skip writing the prefix if it the class comes from bundles in the memento's catalog item.

core/src/main/java/org/apache/brooklyn/core/mgmt/persist/OsgiClassPrefixer.java
```diff
@@ -0,0 +1,59 @@
...
+        if (bundle.isPresent() && !whiteListRetriever.isBundleWhiteListed(bundle.get())) {
+            return Optional.of(bundle.get().getSymbolicName() + ":");
+        }
+        return Optional.absent();
+    }
+    
+    public Optional<String> stripMatchingPrefix(Bundle bundle, String type) {
+        String symbolicName = bundle.getSymbolicName();
+        if (symbolicName != null && type.startsWith(symbolicName + ":")) {
+            return Optional.of(type.substring((symbolicName + ":").length()));
```
aledsage: @neykov wrote: The prefix could include version identifier as well.

core/src/main/java/org/apache/brooklyn/core/mgmt/persist/OsgiClassPrefixer.java
```diff
@@ -0,0 +1,59 @@
...
+
+    private final ClassLoaderUtils whiteListRetriever;
+    
+    public OsgiClassPrefixer() {
+        whiteListRetriever = new ClassLoaderUtils(getClass());
+    }
+    
+    public Optional<String> getPrefix(Class<?> type) {
+        Optional<Bundle> bundle  = Osgis.getBundleOf(type);
+        if (bundle.isPresent() && !whiteListRetriever.isBundleWhiteListed(bundle.get())) {
```
aledsage: @neykov wrote 7 days:
Don't like how the white list is proliferating throughout the code base. It's supposed to go the other way :).
What do we gain by stripping the prefix for white listed bundles? I think not much because if classes move bundles then their package will change anyway. This can be handled by the renaming list.

Removing this will simplify the code and we don't lose much if anything.

core/src/main/java/org/apache/brooklyn/core/mgmt/ha/OsgiManager.java
aledsage: Good points @neykov. I'll strip the version from the prefix as well.

I'll let it match by just comparing the symbolic name (even if wrong version number) because that will better support upgrades/rollback of bundles, where the `symbolicName:version:className` has been persisted (I'd hope we miss out the version number in persisted state for that reason though).

I'd prefer not to try skipping the prefix "if the class comes from the bundles in the memento's catalog item". It makes sense at a high level, but I worry about situations where there are nested catalog items or sub-types (where different levels declare their libraries). We should handle such cases, but it feels to me like belt-and-braces if we include the bundle id here as well. If we do want to skip writing the prefix in such cases, let's look at that in a separate PR.

core/src/main/java/org/apache/brooklyn/core/mgmt/persist/OsgiClassPrefixer.java
aledsage: On the plus side, this is just moving the white list code (rather than proliferating). It is deleted from `XmlMementoSerializer.OsgiClassnameMapper`, which now delegates to this code. It does mean the white-list code is called slightly more often, but thus in a more consistent manner for handling prefixes.

Let's revisit the white list in a separate PR (potentially on the dev@brooklyn mailing list first).

aledsage:
@neykov can you review this again please?

core/src/main/java/org/apache/brooklyn/core/mgmt/ha/OsgiManager.java
```diff
@@ -189,10 +194,15 @@ public static boolean isBundleNameEqualOrAbsent(CatalogBundle bundle, Bundle b)
...
+                    if (osgiClassPrefixer.hasPrefix(typeToLoad)) {
+                        bundleProblems.put(osgiBundle, new UserFacingException("Bundle does not match prefix in type name '"+typeToLoad+"'"));
+                        continue;
+                    }
```
neykov: `getResource(s)` needs the same update. Turns out we can't load prefixed resources when combined with `brooklyn.libraries`.
Could be addressed as another PR.

neykov:
LGTM, merging.
Resource handling can be added in another PR.

PR #443, opened 2016-11-18
Adds maxConcurrentChildCommands parameter to DynamicCluster

sjcorbett:
The option configures the maximum number of simultaneous `Startable` effector invocations that will be made on members of the group.

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/entity/group/DynamicCluster.java
+++ b/core/src/main/java/org/apache/brooklyn/entity/group/DynamicCluster.java
@@ -103,7 +103,7 @@ public interface DynamicCluster extends AbstractGroup, Cluster, MemberReplaceabl
             "dynamiccluster.restartMode", 
             "How this cluster should handle restarts; "
             + "by default it is disallowed, but this key can specify a different mode. "
-            + "Modes supported by dynamic cluster are 'off', 'sequqential', or 'parallel'. "
+            + "Modes supported by dynamic cluster are 'off', 'sequential', or 'parallel'. "
             + "However subclasses can define their own modes or may ignore this.", null);
 
     @SetFromFlag("quarantineFailedEntities")
@@ -182,6 +182,15 @@ public interface DynamicCluster extends AbstractGroup, Cluster, MemberReplaceabl
     @SetFromFlag("clusterMemberId")
     ConfigKey<Integer> CLUSTER_MEMBER_ID = ConfigKeys.newIntegerConfigKey(
             "cluster.member.id", "The unique ID number (sequential) of a member of a cluster");
+
+    @Beta
+    @SetFromFlag("maxConcurrentChildCommands")
+    ConfigKey<Integer> MAX_CONCURRENT_CHILD_COMMANDS = ConfigKeys.builder(Integer.class)
+            .name("dynamiccluster.maxConcurrentChildCommands")
+            .description("[Beta] The maximum number of effector invocations that will be made on children at once " +
+                    "(e.g. start, stop, restart). Any value null or less than or equal to zero means invocations are unbounded")
+            .defaultValue(0)
+            .build();
 
     AttributeSensor<List<Location>> SUB_LOCATIONS = new BasicAttributeSensor<List<Location>>(
             new TypeToken<List<Location>>() {},
--- a/core/src/main/java/org/apache/brooklyn/entity/group/DynamicClusterImpl.java
+++ b/core/src/main/java/org/apache/brooklyn/entity/group/DynamicClusterImpl.java
@@ -787,7 +787,8 @@ public class DynamicClusterImpl extends AbstractGroupImpl implements DynamicClus
         Collection<Entity> removedEntities = pickAndRemoveMembers(delta * -1);
 
         // FIXME symmetry in order of added as child, managed, started, and added to group
-        Task<?> invoke = Entities.invokeEffector(this, (Iterable<Entity>)(Iterable<?>)Iterables.filter(removedEntities, Startable.class), Startable.STOP, Collections.<String,Object>emptyMap());
+        final Iterable<Entity> removedStartables = (Iterable<Entity>) (Iterable<?>) Iterables.filter(removedEntities, Startable.class);
+        Task<?> invoke = Entities.invokeEffector(this, removedStartables, Startable.STOP, Collections.<String,Object>emptyMap());
         try {
             invoke.get();
             return removedEntities;
--- a/core/src/main/java/org/apache/brooklyn/entity/group/DynamicClusterImpl.java
+++ b/core/src/main/java/org/apache/brooklyn/entity/group/DynamicClusterImpl.java
@@ -30,10 +30,12 @@ import java.util.Map;
 import java.util.NoSuchElementException;
 import java.util.Set;
 import java.util.concurrent.Callable;
+import java.util.concurrent.Semaphore;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
-
 import javax.annotation.Nullable;
 
+import org.apache.brooklyn.api.effector.Effector;
 import org.apache.brooklyn.api.entity.Entity;
 import org.apache.brooklyn.api.entity.EntitySpec;
 import org.apache.brooklyn.api.entity.Group;
@@ -98,6 +100,8 @@ import com.google.common.collect.Maps;
 import com.google.common.collect.Multimap;
 import com.google.common.collect.Sets;
 import com.google.common.reflect.TypeToken;
+import com.google.common.util.concurrent.ListenableFuture;
+import com.google.common.util.concurrent.MoreExecutors;
 
 /**
  * A cluster of entities that can dynamically increase or decrease the number of entities.
@@ -107,6 +111,12 @@ public class DynamicClusterImpl extends AbstractGroupImpl implements DynamicClus
     @SuppressWarnings("serial")
     private static final AttributeSensor<Supplier<Integer>> NEXT_CLUSTER_MEMBER_ID = Sensors.newSensor(new TypeToken<Supplier<Integer>>() {},
             "next.cluster.member.id", "Returns the ID number of the next member to be added");
+
+    /**
+     * Controls the maximum number of effector invocations the cluster will make on members at once.
+     * Only used if {@link #MAX_CONCURRENT_CHILD_COMMANDS} is configured.
+     */
+    private transient Semaphore childTaskSemaphore;
 
     private volatile FunctionFeed clusterOneAndAllMembersUp;
 
@@ -212,13 +222,31 @@ public class DynamicClusterImpl extends AbstractGroupImpl implements DynamicClus
     public void init() {
         super.init();
         initialiseMemberId();
+        initialiseTaskPermitSemaphore();
         connectAllMembersUp();
+    }
+
+    @Override
+    public void rebind() {
+        super.rebind();
+        initialiseTaskPermitSemaphore();
     }
 
     private void initialiseMemberId() {
         synchronized (mutex) {
             if (sensors().get(NEXT_CLUSTER_MEMBER_ID) == null) {
                 sensors().set(NEXT_CLUSTER_MEMBER_ID, Suppliers.incrementing());
+            }
+        }
+    }
+
+    private void initialiseTaskPermitSemaphore() {
+        synchronized (mutex) {
+            if (getChildTaskSemaphore() == null) {
+                Integer maxChildTasks = config().get(MAX_CONCURRENT_CHILD_COMMANDS);
+                if (maxChildTasks != null && maxChildTasks > 0) {
+                    childTaskSemaphore = new Semaphore(maxChildTasks);
+                }
             }
         }
     }
@@ -551,8 +579,9 @@ public class DynamicClusterImpl extends AbstractGroupImpl implements DynamicClus
                 Iterables.filter(getChildren(), Predicates.and(Predicates.instanceOf(Startable.class), EntityPredicates.isManaged()))));
         } else if ("parallel".equalsIgnoreCase(mode)) {
             ServiceStateLogic.setExpectedState(this, Lifecycle.STARTING);
-            DynamicTasks.queue(Effectors.invocationParallel(Startable.RESTART, null, 
-                Iterables.filter(getChildren(), Predicates.and(Predicates.instanceOf(Startable.class), EntityPredicates.isManaged()))));
+            for (Entity member : Iterables.filter(getChildren(), Predicates.and(Predicates.instanceOf(Startable.class), EntityPredicates.isManaged()))) {
+                DynamicTasks.queue(newThrottledEffectorTask(member, Startable.RESTART, Collections.emptyMap()));
+            }
         } else {
             throw new IllegalArgumentException("Unknown "+RESTART_MODE.getName()+" '"+mode+"'");
         }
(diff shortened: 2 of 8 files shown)


core/src/main/java/org/apache/brooklyn/entity/group/DynamicClusterImpl.java
```diff
@@ -825,8 +858,12 @@ protected Entity replaceMember(Entity member, @Nullable Location memberLoc, Map<
             addedEntities.add(entity);
             addedEntityLocations.put(entity, loc);
             if (entity instanceof Startable) {
+                boolean privileged = false;
+                if (Boolean.TRUE.equals(entity.sensors().get(AbstractGroup.FIRST_MEMBER))) {
```
geomacy: Remind me why not just `privileged = entity.sensors().get(AbstractGroup.FIRST_MEMBER)`?

core/src/main/java/org/apache/brooklyn/entity/group/DynamicClusterImpl.java
```diff
@@ -1040,14 +1077,80 @@ protected void discardNode(Entity entity) {
...
             Entities.unmanage(member);
         }
     }
+
+    protected <T> Task<?> newThrottledEffectorTask(Entity target, Effector<T> effector, Map<?, ?> arguments) {
+        return newThrottledEffectorTask(target, effector, arguments, false);
+    }
+
+    protected <T> Task<?> newThrottledEffectorTask(Entity target, Effector<T> effector, Map<?, ?> arguments, boolean isPrivileged) {
+        final Semaphore permit = childTaskSemaphore;
```
geomacy: why the local variable?

core/src/main/java/org/apache/brooklyn/entity/group/DynamicClusterImpl.java
```diff
@@ -1040,14 +1077,80 @@ protected void discardNode(Entity entity) {
...
+            // Acquire the permit now for the privileged task and just queue the effector invocation.
+            // If it's unprivileged then queue a task to obtain a permit first.
+            final String description = "Waiting for permit to run " + effector.getName() + " on " + target;
+            final Runnable obtain = new ObtainPermit(permit, description);
+            if (isPrivileged) {
+                obtain.run();
+                toSubmit = effectorTask;
+            } else {
+                // acquire semaphore here if privileged?
+                Task<?> obtainMutex = Tasks.builder()
```
geomacy: Is it worth moving this and `ObtainPermit` into `Tasks.java`, so that this call would become something like 
```
                Task<?> obtainMutex = Tasks.builder()
                        .description(description)
                        .throttleOn(permit)
                        .build();
```

core/src/test/java/org/apache/brooklyn/entity/group/DynamicClusterRebindTest.java
```diff
@@ -0,0 +1,54 @@
...
+import org.apache.brooklyn.core.mgmt.rebind.RebindTestFixtureWithApp;
+import org.testng.annotations.Test;
+
+import com.google.common.collect.ImmutableList;
+
+public class DynamicClusterRebindTest extends RebindTestFixtureWithApp {
+
+    @Test
+    public void testThrottleAppliesAfterRebind() throws Exception {
+        DynamicCluster cluster = origApp.createAndManageChild(EntitySpec.create(DynamicCluster.class)
```
geomacy: Worth having a repeat of the test with MAX > 1?  Either here and/or in DynamicClusterTest?

core/src/main/java/org/apache/brooklyn/entity/group/DynamicCluster.java
```diff
@@ -183,6 +183,14 @@
     ConfigKey<Integer> CLUSTER_MEMBER_ID = ConfigKeys.newIntegerConfigKey(
             "cluster.member.id", "The unique ID number (sequential) of a member of a cluster");
 
+    @SetFromFlag("maxConcurrentChildCommands")
+    ConfigKey<Integer> MAX_CONCURRENT_CHILD_COMMANDS = ConfigKeys.builder(Integer.class)
+            .name("cluster.maxConcurrentChildCommands")
```
sjcorbett: Agree. However, it's consistent with the other config and sensors in the class.

core/src/test/java/org/apache/brooklyn/entity/group/DynamicClusterRebindTest.java
sjcorbett: Probably. I'll add it in.

core/src/main/java/org/apache/brooklyn/entity/group/DynamicClusterImpl.java
sjcorbett: I'd be open to doing that later when we want to reuse this patter elsewhere.

core/src/test/java/org/apache/brooklyn/entity/group/DynamicClusterRebindTest.java
sjcorbett: I added this to `DynamicClusterTest`.

sjcorbett:
Build failure external:
```
[WARNING] Failed to notify spy hudson.maven.Maven3Builder$JenkinsEventSpy: Invalid object ID 9 iota=46
[WARNING] Failed to notify spy hudson.maven.Maven3Builder$JenkinsEventSpy: Invalid object ID 9 iota=46
[WARNING] Failed to notify spy hudson.maven.Maven3Builder$JenkinsEventSpy: Invalid object ID 25 iota=46
```

sjcorbett:
Actually I wonder if that's true. The build timed out testing DynamicCluster. I'll double-check locally.

sjcorbett:
Jenkins did its job. I had forgotten to `submit` a task so `stopAndRemoveNode` blocked forever.

core/src/test/java/org/apache/brooklyn/entity/group/DynamicClusterTest.java
```diff
@@ -1225,4 +1235,93 @@ private void assertFirstAndNonFirstCounts(Collection<Entity> members, int expect
         assertEquals(found.size(), expectedNonFirstCount);
     }
 
+    @DataProvider
+    public Object[][] maxConcurrentCommandsTestProvider() {
```
geomacy: nice!

geomacy:
Have caught up and can confirm the tests work for me.

geomacy:
Have done some more testing with this, with the blueprint:
```
name: test
location: amazon-eu-central-1
services:
- type: org.apache.brooklyn.entity.group.DynamicCluster:0.10.0-SNAPSHOT
  brooklyn.config:
    dynamiccluster.firstmemberspec:
      $brooklyn:entitySpec:
        name: first
    dynamiccluster.memberspec:
      $brooklyn:entitySpec:
        name: VM
    cluster.initial.size: 20
    cluster.maxConcurrentChildCommands: 3
```
and one issue I noticed was that, once you've deployed the cluster, if you have second thoughts and decide to stop it when only a few nodes have started, say three, the stop effector call "goes to the back of the queue", and the remaining seventeen nodes gradually start up, three at a time, before the stop finally gets its chance to do anything.   Then the nodes gradually shut down (oddly, two at a time?).

I think this would be sufficiently annoying for users that we should think about allowing priority to stop commands?

sjcorbett:
@geomacy I agree in principle but think it's outwith this pull request. Completing a resize before beginning another is currently part of `DynamicCluster's` contract. Your comment about stopping two-at-a-time is interesting and I'll look into it more. My preference is to merge this pull request now.

core/src/main/java/org/apache/brooklyn/entity/group/DynamicClusterImpl.java
aledsage: But it could be `boolean privileged = Boolean.TRUE.equals(entity.sensors().get(AbstractGroup.FIRST_MEMBER));`

I'm fine with it as-is.

core/src/main/java/org/apache/brooklyn/entity/group/DynamicClusterImpl.java
```diff
@@ -1040,14 +1077,81 @@ protected void discardNode(Entity entity) {
...
+                        .description(description)
+                        .body(new ObtainPermit(permit, description))
+                        .build();
+                toSubmit = Tasks.sequential(obtainMutex, effectorTask);
+            }
+            toSubmit.addListener(new ReleasePermit(permit), MoreExecutors.sameThreadExecutor());
```
aledsage: I worry about `ReleasePermit` being called if the task is aborted. If that happens, then the `ObtainPermit` might not have actually happened, yet we'll call `permit.release()`.

core/src/test/java/org/apache/brooklyn/entity/group/DynamicClusterTest.java
```diff
@@ -1225,4 +1235,93 @@ private void assertFirstAndNonFirstCounts(Collection<Entity> members, int expect
...
+        private static final Logger LOG = LoggerFactory.getLogger(ThrowOnAsyncStartEntityImpl.class);
+        @Override
+        public void start(Collection<? extends Location> locs) {
+            int count = config().get(COUNTER).incrementAndGet();
+            LOG.debug("{} starting (first={})", new Object[]{this, sensors().get(AbstractGroup.FIRST_MEMBER)});
+            config().get(START_LATCH);
+            // Throw if more than one entity is starting at the same time as this.
+            assertTrue(count <= config().get(MAX_CONCURRENCY), "expected " + count + " <= " + config().get(MAX_CONCURRENCY));
+            super.start(locs);
+            config().get(COUNTER).decrementAndGet();
```
aledsage: Wrap in a try-finally block, to ensure the counter is deprecated? Probably not necessary for the existing tests, as they're not failing the start or interrupting. But still worth adding I think.

core/src/test/java/org/apache/brooklyn/entity/group/DynamicClusterTest.java
```diff
@@ -1225,4 +1235,93 @@ private void assertFirstAndNonFirstCounts(Collection<Entity> members, int expect
...
+
+        EntitySpec<ThrowOnAsyncStartEntity> memberSpec = EntitySpec.create(ThrowOnAsyncStartEntity.class)
+                .configure(ThrowOnAsyncStartEntity.COUNTER, counter)
+                .configure(ThrowOnAsyncStartEntity.START_LATCH, firstMemberUp);
+
+        cluster.config().set(DynamicCluster.FIRST_MEMBER_SPEC, firstMemberSpec);
+        cluster.config().set(DynamicCluster.MEMBER_SPEC, memberSpec);
+
+        // app.start blocks so in the failure case this test would block forever.
+        Thread t = new Thread() {
```
aledsage: Very minor - you could use `Asserts.assertReturnsEventually` to do the thread interrupt for you. For example, something like:
```
Asserts.assertReturnsEventually(new Runnable() {
    public void run() {
        app.start(ImmutableList.of(app.newSimulatedLocation()));
    }});
```

Also, I don't think you need to do `Entities.unmanage(cluster)` - will that not happen automatically in `tearDown` anyway?

core/src/test/java/org/apache/brooklyn/entity/group/DynamicClusterTest.java
```diff
@@ -1225,4 +1235,93 @@ private void assertFirstAndNonFirstCounts(Collection<Entity> members, int expect
...
+    }
+
+    @ImplementedBy(ThrowOnAsyncStartEntityImpl.class)
+    public interface ThrowOnAsyncStartEntity extends TestEntity {
+        ConfigKey<Integer> MAX_CONCURRENCY = ConfigKeys.newConfigKey(Integer.class, "concurrency", "max concurrency", 1);
+        ConfigKey<AtomicInteger> COUNTER = ConfigKeys.newConfigKey(AtomicInteger.class, "counter");
+        ConfigKey<Boolean> START_LATCH = ConfigKeys.newConfigKey(Boolean.class, "startlatch");
+    }
+
+    public static class ThrowOnAsyncStartEntityImpl extends TestEntityImpl implements ThrowOnAsyncStartEntity {
```
aledsage: Could generalise to `ThrowOnAsyncEffectorEntityImpl`, given that it is not supposed to do concurrent `start`, `stop` or `restart`. All methods would presumably all share the same counter.

core/src/test/java/org/apache/brooklyn/entity/group/DynamicClusterTest.java
```diff
@@ -1225,4 +1235,93 @@ private void assertFirstAndNonFirstCounts(Collection<Entity> members, int expect
         assertEquals(found.size(), expectedNonFirstCount);
     }
 
+    @DataProvider
+    public Object[][] maxConcurrentCommandsTestProvider() {
+        return new Object[][]{{1}, {2}, {3}};
+    }
+
+    @Test(dataProvider = "maxConcurrentCommandsTestProvider")
+    public void testEntitiesStartAndStopSequentiallyWhenMaxConcurrentCommandsIsOne(int maxConcurrentCommands) {
```
aledsage: I wonder what the changes are of this failing without your semaphore. The problem is that the work done inside the `start()` method (between the increment + decrement of the counter) is pretty small. We're probably ok with 10 entities and such a small `maxConcurrentCommands` but still...

I'm loathe to suggest it, but maybe we want to configure the entity with a very short sleep in the middle of `start()`. Even just a few milliseconds would force the thread to yield, and thus make the concurrent commands much more likely to try to execute.

Up to you whether you think that's worth it.

aledsage:
@sjcorbett (cc @geomacy) LGTM. A few minor comments. Sam, if you can take a look through them and see what you think is worth changing. Then either merge it yourself or ping me if you want me to take another look.

core/src/test/java/org/apache/brooklyn/entity/group/DynamicClusterTest.java
sjcorbett: It always fails for me when I remove `.configure(DynamicCluster.MAX_CONCURRENT_CHILD_COMMANDS, maxConcurrentCommands)` from a few lines below.

sjcorbett: I tend to observe high concurrency even for this simple entity. When it fails it's typically with an error like
```
2016-11-28 13:57:33,285 ERROR Cluster DynamicClusterImpl{id=gve8830u85} failed to start entity ThrowOnAsyncStartEntityImpl{id=vofv04muyy} (removing): java.lang.AssertionError: expected 9 <= 1 expected [true] but found [false]
```
And that's with a `finally` block decrementing the counter again.

core/src/main/java/org/apache/brooklyn/entity/group/DynamicClusterImpl.java
sjcorbett: Do you think it's better to queue the release as another task?

sjcorbett:
@aledsage regarding the obtain and release of permits, I've pushed an update and test case that checks how permits are handled when tasks are cancelled. See https://github.com/apache/brooklyn-server/pull/443/files#diff-a0b8d0880424918e737424abac6a9694R1123 for the implementation and https://github.com/apache/brooklyn-server/pull/443/files#diff-f27a4972ae513473372a4703cb4526caR1301 for the test. Could you re-review this? Thanks.

sjcorbett:
I'm going to merge this now. Happy to address any further comments separately.

PR #444, opened 2016-11-18
BROOKLYN-396: test and fix

aledsage:
When rebinding a location, if a config flag corresponds to a field
with `@SetFromFlag` then just set the field, and don’t leave the 
key-value inside `config().getBag()`.

Some classes use fields to store internal state, which should not be
passed to child locations (e.g. `LocalhostMachineProvisioningLocation.origConfigs`).

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/mgmt/rebind/BasicLocationRebindSupport.java
+++ b/core/src/main/java/org/apache/brooklyn/core/mgmt/rebind/BasicLocationRebindSupport.java
@@ -70,33 +70,41 @@ public class BasicLocationRebindSupport extends AbstractBrooklynObjectRebindSupp
         // Note that the flags have been set in the constructor
         // Sept 2016 - now ignores unused and config description
         
-        location.config().putAll(memento.getLocationConfig());
-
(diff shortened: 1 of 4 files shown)


aledsage:
It's concerning that the implementation of `BasicLocationRebindSupport.addConfig()` is very different from `BasicEntityRebindSupport` and `BasicPolicyRebindSupport`.

But I don't want to try changing that stuff now. It's one to mull over, and change in the future (possibly next time we're doing major work on rebind).

sjcorbett:
@aledsage the fix looks good. The blueprint I gave in 396 resizes correctly after rebind. 

In the comments on the issue you wrote that "I presume the solution will be to improve what we store in the `origConfigs` (e.g. so that the sub-machines don't inherit it)." Is `origConfigs` doing the right thing? Or have we side-stepped it?

aledsage:
@sjcorbett this change ensures the `origConfigs` has the right thing.

The intent of `origConfigs` is to remember the state of the location from before we returned it, so if a BYON gets it back and then gives it out again, it can reset the config to what it was originally. This means that `origConfigs` is a map from `MachineLocation` instance to its config map.

aledsage:
Merging now.

PR #445, opened 2016-11-18
Use subnet IP on machine locations that dont implement HasSubnetHostname

grkvlt:
The `SshMachineLocation` interface does *NOT* implement `HasSubnetHostname` but it does include the method `getPrivateAddresses()` which means that when something creates a machine and sets the config key `SshMachineLocation.PRIVATE_ADDRESSES` with the private IP then the `machines.getSubnetHostname(Location)` helper method (called during entity startup) will happily set the `SUBNET_HOSTNAME` attribute to the public IP (via `location.getAddress().getHostAddress()`

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/location/Machines.java
+++ b/core/src/main/java/org/apache/brooklyn/core/location/Machines.java
@@ -49,8 +49,13 @@ public class Machines {
             hostname = ((HasSubnetHostname) where).getSubnetHostname();
         }
         if (hostname == null && where instanceof MachineLocation) {
-            InetAddress addr = ((MachineLocation) where).getAddress();
-            if (addr != null) hostname = addr.getHostAddress();
+            Maybe<String> subnetIp = getSubnetIp(where);
+            if (subnetIp.isPresent()) {
+                hostname = subnetIp.get();
+            } else {
+                InetAddress addr = ((MachineLocation) where).getAddress();
+                if (addr != null) hostname = addr.getHostAddress();
+            }
         }
         log.debug("computed subnet hostname {} for {}", hostname, where);
         // TODO if Maybe.absent(message) appears, could/should use that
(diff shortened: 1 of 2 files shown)


aledsage:
This change makes sense to me.

For those interested, here are a couple of contexts where it's useful:
1. if there is a custom location type that create `SshMachineLocation` instances (i.e. not using `JcloudsLocation`, which creates `JcloudsSshMachineLocation`), then those instances might have different addresses for inside and outside the subnet.
2. if using BYON, then one can have machines that have subnet and public address, so you'd want the subsequent `SshMachineLocation` to correctly advertise both.

It's important to advertise the right subnet address. For example, the nginx entity by default monitors its members' `host.subnet.hostname` (controlled via the sensor `member.sensor.hostname`).

---
@grkvlt can you add test(s) please.

grkvlt:
@aledsage I updated `testFindSubnetHostnameFromLocation` and some other tests to use the private addresses

aledsage:
Thanks @grkvlt - looks good to me; merging.

aledsage:
This is merged (see https://git-wip-us.apache.org/repos/asf?p=brooklyn-server.git;a=summary), but this github mirror is not up-to-date. I've raise this with Apache Infrastructure team (https://issues.apache.org/jira/browse/INFRA-11538).

geomacy:
A bit late perhaps with review comment but, noting that this calls `getSubnetIp()`, is [this](https://github.com/grkvlt/brooklyn-server/blob/4f3a3da979c4b0a6011bf2f2dcdfa0e3d873d1fa/core/src/main/java/org/apache/brooklyn/core/location/Machines.java#L72) line meant to read
```java
if (result != null && where instanceof HasNetworkAddresses) {
```
i.e. it goes to the trouble of setting up `result` at line 70 only to potentially overwrite it at 75, but the code looks like it's meant to be setting `result` based on three possibilities in a preference order? If that's not intended maybe some explanatory comment would be worthwhile.

PR #446, opened 2016-11-19
Disable testChildFailuresOnStartButWithQuorumCausesAppToSucceed

aledsage:
This test is failing occasionally (non-deterministically) when running `mvn clean install`. Therefore disabling until we make it more reliable!


aledsage:
Merged. (But github/apache/brooklyn-server seems to be slow in picking this up - it should automatically sync with the apache git repo).

neykov:
https://issues.apache.org/jira/browse/INFRA-12946

neykov:
Fails for me every time (when run through Eclipse).
The "at least one" quorum check is failing because the `running` events for `TestEntity` are processed **before** the `starting` events. Need to investigate further.

PR #450, opened 2016-11-21
Allow all certs in TestHttpCall

drigodwin:
This enables self signed and other certificates to be used within `TestHttpCall`. Tested on AWS with the [brooklyncentral/brooklyn-tomcat-7-server](https://github.com/brooklyncentral/brooklyn-tomcat-7-server) tests.

### DIFF
--- a/test-framework/src/main/java/org/apache/brooklyn/test/framework/TestHttpCall.java
+++ b/test-framework/src/main/java/org/apache/brooklyn/test/framework/TestHttpCall.java
@@ -51,6 +51,8 @@ public interface TestHttpCall extends BaseTest {
             .defaultValue(HttpMethod.GET)
             .build();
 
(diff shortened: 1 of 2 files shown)


test-framework/src/main/java/org/apache/brooklyn/test/framework/TestHttpCall.java
```diff
@@ -51,6 +51,8 @@
             .defaultValue(HttpMethod.GET)
             .build();
 
+    ConfigKey<Boolean> TRUST_ALL = ConfigKeys.newBooleanConfigKey("trustAll","Trust all certificates used to sign this request",true);
```
m4rkmckenna: I would prefer `sslTrustAll` but its personal preference

drigodwin: Yes, I preferred to stick to the nomenclature defined by `HttpTool` and I'm not sure there's a lot else it could potentially be confused with.

aledsage:
LGTM; merging now.

PR #451, opened 2016-11-21
Adds ConfigParametersYamlTest.testPortSetAsAttributeOnSoftwareProcess

aledsage:


googlielmo:
Tested. LGTM

neykov:
Wondered how this test is passing. Turns out that we are [creating port sensors](https://github.com/apache/brooklyn-server/blob/master/core/src/main/java/org/apache/brooklyn/core/objs/BasicSpecParameter.java#L256) in addition to the config keys from the yaml. And those get [set on the entity](https://github.com/apache/brooklyn-server/blob/master/core/src/main/java/org/apache/brooklyn/core/objs/proxy/InternalEntityFactory.java#L276) just like config keys.

The sensors aren't serialized as part of the entity type. That's fine though as the sensor sets a value for itself on start which gets serialized (along with the sensor that set it).

PR #452, opened 2016-11-22
Fix out of order events triggered by ComputeServiceState

neykov:
`ComputeServiceState.onEvent` should only be called in the event handler executor to keep the ordering of events.
Before the change what would happen is on calling `setExpectedState(state)` a "state.expected" event would be scheduled, with `ComputeServiceState.onEvent(null)` called in the same thread right after. This would create a race condition when calling `setExpectedState` in quick succession (`onEvent` called from the `setExpectedState` thread and from the event handler thread with different values).

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/entity/lifecycle/ServiceStateLogic.java
+++ b/core/src/main/java/org/apache/brooklyn/core/entity/lifecycle/ServiceStateLogic.java
@@ -23,6 +23,7 @@ import java.util.Date;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.atomic.AtomicInteger;
 
 import javax.annotation.Nullable;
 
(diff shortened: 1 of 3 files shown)


core/src/test/java/org/apache/brooklyn/core/entity/lifecycle/ServiceStateLogicTest.java
```diff
@@ -281,6 +291,35 @@ public void testQuorumWithStringStates() {
...
+            cntCalled.incrementAndGet();
+            if (event == null) {
+                cntCalledWithNull.incrementAndGet();
+            }
+            super.onEvent(event);
+        }
+    }
+
+    @Test
+    public void testServiceStateNotCalledExplicitly() throws Exception {
```
neykov: Wasn't able to come up with a deterministic test to observe the failure case. Instead I'm asserting that the cause is not happening any more.

aledsage:
LGTM - good to merge.

core/src/main/java/org/apache/brooklyn/core/entity/lifecycle/ServiceStateLogic.java
```diff
@@ -278,14 +276,18 @@ public void setEntity(EntityLocal entity) {
                 suppressDuplicates = true;
             }
             
-            subscriptions().subscribe(entity, SERVICE_PROBLEMS, this);
-            subscriptions().subscribe(entity, SERVICE_UP, this);
-            subscriptions().subscribe(entity, SERVICE_STATE_EXPECTED, this);
-            onEvent(null);
+            Map<String, ?> notifyOfInitialValue = ImmutableMap.of("notifyOfInitialValue", Boolean.TRUE);
+            subscriptions().subscribe(notifyOfInitialValue, entity, SERVICE_PROBLEMS, this);
```
neykov: It will call the handler even on null value. So that's covered.

PR #453, opened 2016-11-22
BROOKLYN-352: fix ConcurrentModification in getAllEntitiesInApplication

aledsage:


core/src/main/java/org/apache/brooklyn/core/mgmt/internal/LocalEntityManager.java
```diff
@@ -212,13 +213,41 @@ public EntityTypeRegistry getEntityTypeRegistry() {
...
         Predicate<Entity> predicate = EntityPredicates.applicationIdEqualTo(application.getId());
-        Iterable<Entity> allentities = Iterables.concat(preRegisteredEntitiesById.values(), preManagedEntitiesById.values(), entityProxiesById.values());
-        Iterable<Entity> result = Iterables.filter(allentities, predicate);
-        return ImmutableSet.copyOf(Iterables.transform(result, new Function<Entity, Entity>() {
-            @Override public Entity apply(Entity input) {
-                return Entities.proxy(input);
-            }}));
+        Set<Entity> result = Sets.newLinkedHashSet();
+        
+        synchronized (preRegisteredEntitiesById) {
```
geomacy: Maybe it would be worth making `preRegisteredEntitiesById` and `preManagedEntitiesById` private, and providing access to them with code like this loop, but put into a `protected` method.  At the moment it's not impossible that some derived class in future would again use one of the maps but forget the synchronization.

aledsage: Yeah, I wondered why it was protected. That would definitely not be my choice! I wouldn't even bother having a protected getter that provided direct access. Any such use-case would likely be making changes to core brooklyn, rather than a downstream project needing it - so we can add a getter if the need arises.

PR #454, opened 2016-11-22
LocalEntityManager: make fields private

aledsage:
As discussed with @geomacy in https://github.com/apache/brooklyn-server/pull/453#discussion_r89106124 ...

Making these fields non-private is asking for trouble - it makes it
harder to reason about what the class does and whether it’s safe,
particularly around thread safety etc.
    
It is also exposing more implementation detail than is required, making
refactoring harder in the future.
    
If there is a compelling reason to provide access to the fields,
we can consider adding accessor(s).

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/mgmt/internal/LocalEntityManager.java
+++ b/core/src/main/java/org/apache/brooklyn/core/mgmt/internal/LocalEntityManager.java
@@ -91,25 +91,25 @@ public class LocalEntityManager implements EntityManagerInternal {
     private final InternalPolicyFactory policyFactory;
     
     /** Entities that have been created, but have not yet begun to be managed */
-    protected final Map<String,Entity> preRegisteredEntitiesById = Collections.synchronizedMap(new WeakHashMap<String, Entity>());
+    private final Map<String,Entity> preRegisteredEntitiesById = Collections.synchronizedMap(new WeakHashMap<String, Entity>());
 
     /** Entities that are in the process of being managed, but where management is not yet complete */
-    protected final Map<String,Entity> preManagedEntitiesById = Collections.synchronizedMap(new WeakHashMap<String, Entity>());
+    private final Map<String,Entity> preManagedEntitiesById = Collections.synchronizedMap(new WeakHashMap<String, Entity>());
     
     /** Proxies of the managed entities */
-    protected final ConcurrentMap<String,Entity> entityProxiesById = Maps.newConcurrentMap();
+    private final ConcurrentMap<String,Entity> entityProxiesById = Maps.newConcurrentMap();
     
     /** Real managed entities */
-    protected final Map<String,Entity> entitiesById = Maps.newLinkedHashMap();
+    private final Map<String,Entity> entitiesById = Maps.newLinkedHashMap();
     
     /** Management mode for each entity */
-    protected final Map<String,ManagementTransitionMode> entityModesById = Collections.synchronizedMap(Maps.<String,ManagementTransitionMode>newLinkedHashMap());
+    private final Map<String,ManagementTransitionMode> entityModesById = Collections.synchronizedMap(Maps.<String,ManagementTransitionMode>newLinkedHashMap());
 
     /** Proxies of the managed entities */
-    protected final ObservableList entities = new ObservableList();
+    private final ObservableList entities = new ObservableList();
     
     /** Proxies of the managed entities that are applications */
-    protected final Set<Application> applications = Sets.newConcurrentHashSet();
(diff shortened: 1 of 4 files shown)


aledsage:
@neykov the failing test (`LoopOverGroupMembersTestCaseTest.testMultipleChildrenWhichPass`) is I maybe happening after merging https://github.com/apache/brooklyn-server/pull/452.

It fails because `LoopOverGroupMembersTestCase.start` calls `child.start()` and then immediately checks the child's `SERVICE_STATE_ACTUAL`. This still says "starting" so it does its log.warn, sets `allSuccesful = false`, and thus sets its own serviceState to `ON_FIRE`.

I wonder if child is still reporting `starting` for a few extra milliseconds (set async) because of not doing `onEvent(null)`. I haven't looked into it further yet though.

I wonder if the "correct" fix is for things like `TestSensorImpl.start()` to wait for actual serviceState to be running before the start() method returns (similar to `SoftwareProcessImpl.waitForServiceUp`, but that just waits for serviceUp=true rather than serviceState=running).

neykov:
So that's the reason why `onEvent` was called explicitly - to give callers of `setExpectedState` synchronous control over `state.actual`. Otherwise the change from `config().set(STATE_ACTUAL)` to `setExpectedState` would've been a disaster will the code all over making the synchronous assumption. Now it makes sense.

>I wonder if child is still reporting starting for a few extra milliseconds (set async) because of not doing onEvent(null)

That's the case. This child (`TestSensor`) is manipulating the state by calling `setExpectedState` which sets `state.actual` async so it will be evaluated **after** start completes.

Even if we fix `TestSensor` there are a ton of places assuming `state.actual` has been set synchronously. I think the right thing to do is wait in `setExpectedState` for the to-be fired event to be handled. Wonder could this lead to deadlocks.

core/src/main/java/org/apache/brooklyn/core/mgmt/internal/LocalEntityManager.java
```diff
@@ -742,7 +742,7 @@ private boolean unmanageNonRecursive(Entity e) {
             if (e instanceof Group) {
                 Collection<Entity> members = ((Group)e).getMembers();
                 for (Entity member : members) {
-                    if (!Entities.isNoLongerManaged(member)) member.removeGroup((Group)e);
+                    if (!Entities.isNoLongerManaged(member)) ((EntityInternal)member).groups().remove((Group)e);
```
geomacy: `removeGroup` is deprecated, but it's just defined as 
```java
   public void removeGroup(Group group) {
        groups().remove(group);
    }
```
so I don't see how this change is changing anything really? Is it not just doing the same thing? The comments on removeGroup are slightly ambiguous but seem to suggest that the preferred thing to be doing here is calling `Group#removeMember`, which will update the group size sensors and so on.  Same thought applies to the changes to `add` below.

aledsage: You're right that this should be in its own PR - It's unrelated to the private fields. I was just being lazy. I'll separate it out.

The change is just to get rid of deprecation warnings - to remove the calls to `removeGroup` and `addGroup` with the equivalent recommended calls.

geomacy: I guess what I am saying is that these are not the recommended calls, as far as I understand.  `.removeGroup()` is just doing the same thing under the hood as `groups().remove()`.  Should we not be using `Group#removeMember()`?

aledsage: @geomacy I don't think so. These are each very special cases:
* `LocalEntityManager.unmanageNonRecursive()` is handling the case where the group has been unmanaged - updating each of the members to ensure they know they are no longer in that group. We don't want to start calling other methods on `group` (triggering a bunch of events) because this entity is in the process of being unmanaged.
* `AbstractGroupImpl`'s calls are the implementation of `addMember`/`removeMember`, so they definitely need to call the `member.addGroup()`.
* `GroupTest. testAddingUnmanagedGroupDoesNotFailBadly()` is trying to add an unmanaged group to ensure that doesn't break things (so we couldn't call `group.addMember()`).

aledsage:
Thanks @geomacy for reviewing - merging now.

core/src/main/java/org/apache/brooklyn/core/mgmt/internal/LocalEntityManager.java
geomacy: sounds fair, thanks for the explanation @aledsage

PR #457, opened 2016-11-23
Upgrade to jclouds 1.9.3

bostko:


aledsage:
Excellent, thanks @bostko (and thanks @andreaturli for getting the jclouds 1.9.3 release done!).

Can you say what testing you have done with this PR so far?

Have any of the versions of transitive dependencies changed (which would mean we'd need to update our dependencies as well)?

andreaturli:
As far as I can tell jclouds 1.9.3 and jclouds 1.9.2 have the same transitive dependencies, but I can double-check

andreaturli:
I've compared with `mvn dependency:list -DincludeScope=runtime` 1.9.2 and 1.9.3 tags and they seem to have the same versions for transitive dependencies 

see [1.9.2 deps](https://gist.github.com/andreaturli/b7c178519ab4d029d562643426a2738d) vs [1.9.3 deps](https://gist.github.com/andreaturli/8d54e4340ef0a4c650022396b4b54b89)

geomacy:
👍  LGTM, @neykov and I have tested a variety of complex blueprints on different clouds against this and so far haven't encountered problems with it.  Will merge.

andreaturli:
YAY! glad it worked!

PR #458, opened 2016-11-23
BROOKLYN-399: aws-ec2 only looks up images in specified region

aledsage:
See https://issues.apache.org/jira/browse/BROOKLYN-399 for a description of the surprisingly large performance boost we get from this simple change!


sjcorbett:
Looks good. Will check the live tests and merge.

PR #459, opened 2016-11-23
[BROOKLYN-401] Fix failing GC tasks test

neykov:
Background tasks get in the way, messing with the logic of what to evict next.

Note that this only changes the expectations in the tests, doesn't improve the GC as suggested in BROOKLYN-401.


geomacy:
👍  LGTM and tests pass, merging

PR #460, opened 2016-11-23
Fix the Apache Brooklyn source release build

neykov:
Ignore tests relying on binary files in the source distribution.


geomacy:
👍  LGTM, merging

PR #461, opened 2016-11-24
Upgrade rat plugin version

neykov:
Version 0.12 adds the option to dump licence violations to the console.
Also removes the unused rat-check profile, the plugin is already included in the default build in build/plugins.


geomacy:
👍  LGTM, builds and RAT still works, merging.

PR #462, opened 2016-11-24
Inherit config default values

ahgittin:
resolve https://issues.apache.org/jira/browse/BROOKLYN-267

also during deserializing migrate to non-deprecated classes


ahgittin:
failure seems unrelated / intermittent ?

```
Failed tests: 
org.apache.brooklyn.core.entity.ApplicationLifecycleStateTest.(org.apache.brooklyn.core.entity.ApplicationLifecycleStateTest)
  Run 1: PASS
  Run 2: PASS
  Run 3: PASS
  Run 4: PASS
  Run 5: PASS
  Run 6: PASS
  Run 7: PASS
  Run 8: ApplicationLifecycleStateTest.testStartsThenChildFailsButWithQuorumCausesAppToSucceed:170->assertHealthContinually:196 entity=Application[csqdskg2]; attribute=Sensor: service.state (org.apache.brooklyn.core.entity.lifecycle.Lifecycle) expected [running] but found [on-fire]
  Run 9: PASS
  Run 10: PASS
```

core/src/main/java/org/apache/brooklyn/core/config/BasicConfigInheritance.java
```diff
@@ -170,12 +306,88 @@ public String getConflictResolutionStrategy() {
...
+                f.set(this, true);
+            }
+            
+        } catch (Exception e) {
+            throw Exceptions.propagate(e);
+        }
+        
+        for (ConfigInheritance knownMode: Arrays.asList(
+                NOT_REINHERITED, NOT_REINHERITED_ELSE_DEEP_MERGE, NEVER_INHERITED, OVERWRITE, DEEP_MERGE)) {
+            if (equals(knownMode)) return knownMode;
```
aledsage: I don't understand how this works. I added a simple test to `XmlMementoSerializerTest`, which fails:
```
    @Test
    public void testConfigInheritanceVals() throws Exception {
        ConfigInheritance val = BasicConfigInheritance.NEVER_INHERITED;

        ConfigInheritance newVal = assertSerializeAndDeserialize(val); // TODO this line fails
        assertSame(val, newVal);
    }
```

It fails because the deserialized object has `null` for its `delegate` field. It feels to me like the `equals(knownMode)` will always return false because the deserialized object's `delegate` field will never have been set.

core/src/main/java/org/apache/brooklyn/util/core/xstream/ClassRenamingMapper.java
```diff
@@ -42,7 +42,7 @@ public ClassRenamingMapper(Mapper wrapped, Map<String, String> nameToType) {
     
     @Override
     public Class<?> realClass(String elementName) {
-        Optional<String> elementNameOpt = Reflections.tryFindMappedName(nameToType, elementName);
+        Maybe<String> elementNameOpt = Reflections.findMappedNameMaybe(nameToType, elementName);
```
aledsage: Why rename this? It was previously following the same naming conventions as guava's `Iterables` (e.g. `tryFind` versus `find`).

Also, if changing it to a `Maybe` then should we explicitly check for it returning `Maybe.of(null)`, rather than risking a NPE later?

core/src/test/java/org/apache/brooklyn/core/mgmt/rebind/RebindConfigInheritanceTest.java
```diff
@@ -0,0 +1,145 @@
...
+    @Test
+    public void testBasicConfigInheritanceProgrammatic() throws Exception {
+        origApp.config().set(key1, "1");
+        
+        RebindOptions opts = RebindOptions.create();
+        opts.keepSameInstance = true;
+        
+        rebind(opts);
+        
+        String entityFile = Streams.readFully(new FileReader(new File(opts.mementoDir, "entities/"+origApp.getApplicationId())));
```
ahgittin: d'oh.  thought it was awfully hard to find that!  agree, will remove keepSameInstance.

core/src/main/java/org/apache/brooklyn/util/core/xstream/ClassRenamingMapper.java
aledsage: My question about renaming wasn't so much about why change to return a `Maybe`. It's why change the method name? I prefer `tryFindMappedName` because it follows the guava naming convention. We use that naming convention in quite a few other places in Brooklyn as well.

I see now why - the previous `tryFindMappedName` is deprecated in favour of `findMappedNameMaybe` so that it can have a different return type. That's really annoying (that we're creating a different method naming convention so as to change the return type). But I'm  not sure how better to tackle that while still preserving backwards compatibility!

As for null, you've changed the contract. Maybe it's fine that the super throws the `NullPointerException` rather than us checking so not a big deal. But on balance, I'd prefer to report a null present value here, rather than calling `super.realClass(null)`, which could get into completely different code before an NPE is thrown. No strong feelings though.

ahgittin:
this improves fix for BROOKLYN-328 and now also fixes many of the issues described at BROOKLYN-329, specifically

*( causes default values to be inheritable

and

* 5) the ConfigMap.findKeys method never looked at the keys declared on the container; I'm tidying that in order to test the above

* 6) if a yaml subtype reclares a supertype's parameter (from yaml) or config key (from java) only partially, the results are not merged as one would expect, but instead all parts of the super's declaration are lost

ahgittin:
rebased on master; still need to address comments above

core/src/main/java/org/apache/brooklyn/core/objs/BasicSpecParameter.java
```diff
@@ -404,7 +431,126 @@ public static void addParameters(AbstractBrooklynObjectSpec<?, ?> spec, List<? e
...
+
+        /** used if yaml specifies a parameter, but possibly incomplete, and a spec supertype has a parameter */
+        SpecParameter<?> resolveWithAncestor(SpecParameter<?> ancestor) {
+            if (ancestor==null) return new BasicSpecParameter<>(getLabel(), isPinned(), getConfigKey(), getSensor());
+
+            return new BasicSpecParameter<>(
+                    hasLabelSet ? getLabel() : ancestor.getLabel(), 
+                            hasPinnedSet ? isPinned() : ancestor.isPinned(), 
+                                    resolveWithAncestorConfigKey(ancestor.getConfigKey()), 
+                                    hasType ? getSensor() : ancestor.getSensor());
```
geomacy: ignore that; have looked at it with a fresh eye, the `type` _is_ the sensor type of the parameter.

core/src/main/java/org/apache/brooklyn/core/config/BasicConfigInheritance.java
```diff
@@ -28,51 +32,188 @@
...
+     * defines a local default value (with this field true) will have its conflict resolution strategy
+     * applied with the ancestor's default value.
+     * <p>
+     * if this is false, ancestor defaults are completely ignored; prior to 0.10.0 this was the normal behaviour,
+     * but it caused surprises where default values in parameters did not take effect.
+     * <p>
+     * null should not be used. a boxed object is taken (as opposed to a primitive boolean) only in order to support migration.    
+     */
+    @Nonnull
+    protected final Boolean ancestorDefaultInheritable;
```
geomacy: Is this really needed?   Is this condition actually an independent dimension of the behaviour of the class (i.e. it makes sense to specify it as an independently settable parameter), or is it really an "emergent" behaviour that can be inferred from other requirements?

Looking at how it's actually used in the strategies above, it turns out that it is always the logical inverse of `localDefaultResolvesWithAncestorValue`.   How about we use Occam's Razor and do without this parameter, but instead replace its use with `!getLocalDefaultResolvesWithAncestorValue()`, and then tidy up the logic? 

It's a small change to the code, but it does avoid introducing the overhead of another concept for developers to have to consider (config inheritance is complicated enough already!).   The only changes that would be needed, aside from trivial simplifications to `readResolve` and `equals`, would be a couple of lines in `resolveWithParent`, where line 268 becomes slightly simpler with the loss of one test, and at line 283 you just replace the condition.

core/src/main/java/org/apache/brooklyn/core/config/internal/AbstractConfigMapImpl.java
```diff
@@ -355,40 +355,40 @@ protected ConfigInheritance getDefaultRuntimeInheritance() {
...
-                    return result;
-                }
-            };
-            Function<TContainer, TContainer> parentFn = new Function<TContainer, TContainer>() {
-                @Override public TContainer apply(TContainer input) {
-                    return getParentOfContainer(input);
+                queryKey.hasDefaultValue() ? coerceFn.apply(Maybe.of((Object)queryKey.getDefaultValue())) :
+                    Maybe.<T>absent();
+
+                if (ownKey instanceof ConfigKeySelfExtracting) {
```
geomacy: Indentation is wrong

geomacy:
Wrote a few extra tests to try out the change. The only one that didn't perform as I'd expected was

```java
    @Test
    public void testConfigDefaultIsNotInheritedWith_LocalDefaultResolvesWithAncestorValue_SetToTrue() throws Exception {

        addCatalogItems(
            "brooklyn.catalog:",
            "  itemType: entity",
            "  items:",
            "  - id: entity-with-keys",
            "    item:",
            "      type: "+TestEntity.class.getName(),
            "      brooklyn.parameters:",
            "      - name: my.param.key",
            "        type: string",
            "        inheritance.parent: never",
            "        description: description one",
            "        default: myDefaultVal",
            "      brooklyn.config:",
            "        my.other.key: $brooklyn:config(\"my.param.key\")");

        addCatalogItems(
            "brooklyn.catalog:",
            "  itemType: entity",
            "  items:",
            "  - id: wrapper-entity",
            "    item:",
            "      brooklyn.parameters:",
            "      - name: my.param.key",
            "        description: description two",
            "      type: entity-with-keys");

        String yaml = Joiner.on("\n").join(
            "services:",
            "- type: wrapper-entity");

        Entity app = createStartWaitAndLogApplication(yaml);
        final TestEntity entity = (TestEntity) Iterables.getOnlyElement(app.getChildren());
        LOG.info("Config keys declared on "+entity+": "+entity.config().findKeysDeclared(Predicates.alwaysTrue()));
        ConfigKey<?> key = Iterables.getOnlyElement( entity.config().findKeysDeclared(ConfigPredicates.nameEqualTo("my.param.key")) );
        assertEquals(key.getDescription(), "description two");
        assertEquals(entity.config().get(key), null);
    }
```
noting that the last assertion above expects the value to be null because the "inheritance.parent" 
has been set to "never".  Actually the failure is 
```
java.lang.AssertionError: expected [null] but found [myDefaultVal]
Expected :null
Actual   :myDefaultVal
 <Click to see difference>
```

I debugged to see what goes on here, and found that the BasicConfigInheritance class is 
doing what I expected, and the default value is *not* inherited.   Instead, the default
is set higher up in the stack, in AbstractConfigurationSupportInternal#getNonBlockingResolvingSimple,
at the [line](https://github.com/apache/brooklyn-server/blob/master/core/src/main/java/org/apache/brooklyn/core/objs/AbstractConfigurationSupportInternal.java#L139)
```
        Object unresolved = getRaw(key).or(key.getDefaultValue());
```

so in this case the higher level code is ignoring and overriding the instructions of the 
way the inheritance is configured on the key.  Maybe now that the inheritance code is 
written this line and any others like it should be changed, to avoid "second-guessing" 
the specified inheritance behaviour.

geomacy:
In general I think this looks good to me and does what it says on the tin; I just had 
the few observations and questions above.

core/src/main/java/org/apache/brooklyn/core/config/BasicConfigInheritance.java
```diff
@@ -28,51 +32,188 @@
...
+        final static BasicConfigInheritance DELEGATE = new BasicConfigInheritance(false, CONFLICT_RESOLUTION_STRATEGY_OVERWRITE, false, true); 
+        @Override protected ConfigInheritance getDelegate() { return DELEGATE; }
+    }
     /** Indicates that a config key value should not be passed down from a container where it is defined.
      * Unlike {@link #NEVER_INHERITED} these values can be passed down if set as anonymous keys at a container
      * (ie the container does not expect it) to a container which does expect it, but it will not be passed down further. 
      * If the inheritor also defines a value the parent's value is ignored irrespective 
      * (as in {@link #OVERWRITE}; see {@link #NOT_REINHERITED_ELSE_DEEP_MERGE} if merging is desired). */
-    public static BasicConfigInheritance NOT_REINHERITED = new BasicConfigInheritance(false,CONFLICT_RESOLUTION_STRATEGY_OVERWRITE,false);
+    public static ConfigInheritance NOT_REINHERITED = new NotReinherited();
```
neykov: Suggestion for a less mind-bending, delegate-free implementation:
```
    private static class NotReinherited extends BasicConfigInheritance {
        public NotReinherited() {super(false, CONFLICT_RESOLUTION_STRATEGY_OVERWRITE, false, true);}
        @SuppressWarnings("unused") private ConfigInheritance readResolve() {return NOT_REINHERITED;};
    }
    public static ConfigInheritance NOT_REINHERITED = new NotReinherited();
```

core/src/main/java/org/apache/brooklyn/core/config/BasicConfigInheritance.java
```diff
@@ -28,51 +32,188 @@
...
+     * (there is normally no reason to set this false in the context of <b>type hierarchy</b> inheritance because
+     * an ancestor type cannot "consume" a value.) */
     protected final boolean isReinherited;
-    /** conflict-resolution-strategy? in {@link BasicConfigInheritance} supported values are
+    
+    /** a symbol indicating a conflict-resolution-strategy understood by the implementation.
+     * in {@link BasicConfigInheritance} supported values are
      * {@link #CONFLICT_RESOLUTION_STRATEGY_DEEP_MERGE} and {@link #CONFLICT_RESOLUTION_STRATEGY_OVERWRITE}.
-     * subclasses may pass null if they provide a custom implementaton of {@link #resolveWithParentCustomStrategy(ConfigValueAtContainer, ConfigValueAtContainer, org.apache.brooklyn.config.ConfigInheritance.ConfigInheritanceContext)} */
+     * subclasses may pass null or a different string if they provide a custom implementaton
```
neykov: typo `implementaton`

core/src/main/java/org/apache/brooklyn/core/config/BasicConfigInheritance.java
```diff
@@ -82,9 +223,17 @@ public InheritanceMode isInherited(ConfigKey<?> key, Object from, Object to) {
...
+        if (!isSameRootInstanceAs(rightInheritance)) {
             throw new IllegalStateException("Low level inheritance computation error: caller should invoke on "+rightInheritance+" "
                 + "(the inheritance at "+local+"), not "+this);
+        }
+    }
+
+    private boolean isSameRootInstanceAs(ConfigInheritance other) {
+        if (other==null) return false;
+        if (this==other) return true;
+        if (other instanceof DelegatingConfigInheritance) return isSameRootInstanceAs( ((DelegatingConfigInheritance)other).getDelegate() );
```
neykov: Another reason to avoid the delegate. Why is `equals` a bad idea?
Note that you could still end up with an anonymous `BasicConfigInheritance` coming from persistence - they are not resolved to the singletons.

core/src/main/java/org/apache/brooklyn/core/config/BasicConfigInheritance.java
```diff
@@ -170,12 +324,103 @@ public String getConflictResolutionStrategy() {
...
-        return useLocalDefaultValue;
+        return getLocalDefaultResolvesWithAncestorValue();
+    }
+
+    /** see {@link #localDefaultResolvesWithAncestorValue} */
+    public boolean getLocalDefaultResolvesWithAncestorValue() {
+        if (localDefaultResolvesWithAncestorValue==null) {
+            // in case some legacy path is using an improperly deserialized object
+            log.warn("Encountered legacy "+this+" with null localDefaultResolvesWithAncestorValue; transforming", new Throwable("stack trace for legacy "+this));
+            readResolve();
```
neykov: Are you suspecting the object could be coming from a non-xstream based deserialization? If not, then this is too much on the defensive side.

core/src/main/java/org/apache/brooklyn/core/config/BasicConfigInheritance.java
```diff
@@ -170,12 +324,103 @@ public String getConflictResolutionStrategy() {
...
+                    }
+                }
+                fOld.set(this, null);
+            }
+            
+            if (ancestorDefaultInheritable==null) {
+                Field f = getClass().getDeclaredField("ancestorDefaultInheritable");
+                f.setAccessible(true);
+                f.set(this, true);
+            }
```
neykov: Could you add a comment describing why the use of reflection - can't figure it out.

core/src/main/java/org/apache/brooklyn/core/config/BasicConfigInheritance.java
```diff
@@ -170,12 +324,103 @@ public String getConflictResolutionStrategy() {
...
+        }
+                
+        return returnEquivalentConstant(this);
     }
 
     @Override
+    public boolean equals(Object obj) {
+        if (obj==null) return false;
+        if (obj instanceof DelegatingConfigInheritance) return equals( ((DelegatingConfigInheritance)obj).getDelegate() );
+        if (obj.getClass().equals(BasicConfigInheritance.class)) {
```
neykov: This seems wrong, why not `instanceof`?
You already had to work around it in the delegate.

core/src/main/java/org/apache/brooklyn/core/objs/BrooklynObjectInternal.java
```diff
@@ -152,8 +152,8 @@
         /** This is currently the only way to get some rolled up collections and raw,
          * and also to test for the presence of a value (without any default).
          * As more accessors are added callers may be asked to migrate. 
-         * Callers may also consider using {@link #findKeys(com.google.common.base.Predicate)}
-         * if that isn't too inefficient. */
+         * Callers may also consider using {@link #findKeysDeprecated(com.google.common.base.Predicate)}
```
aledsage: Do you mean `findKeysDeclared`, rather than `findKeysDeprecated`?

core/src/main/java/org/apache/brooklyn/core/config/BasicConfigInheritance.java
```diff
@@ -28,51 +32,188 @@
...
-    public static BasicConfigInheritance OVERWRITE = new BasicConfigInheritance(true,CONFLICT_RESOLUTION_STRATEGY_OVERWRITE,false);
+    public static ConfigInheritance OVERWRITE = new Overwrite();
+    
+    private static class DeepMerge extends DelegatingConfigInheritance {
+        final static BasicConfigInheritance DELEGATE = new BasicConfigInheritance(true, CONFLICT_RESOLUTION_STRATEGY_DEEP_MERGE, false, true);
+        @Override protected ConfigInheritance getDelegate() { return DELEGATE; }
+    }
     /** Indicates that if a key has a value at both an ancestor and a descendant, the descendant and his descendants
-     * should attempt to merge the values. If the values are not mergable behaviour is undefined
+     * should attempt to merge the values. If the values are not mergeable behaviour is undefined
```
geomacy: +1 agree nothing needs done in this PR

core/src/main/java/org/apache/brooklyn/core/config/internal/AbstractConfigMapImpl.java
```diff
@@ -397,48 +397,83 @@ protected ConfigInheritance getDefaultRuntimeInheritance() {
...
+    public Set<ConfigKey<?>> findKeysDeclared(Predicate<? super ConfigKey<?>> filter) {
+        return findKeys(filter, KeyFindingMode.DECLARED_OR_PRESENT);
+    }
+
+    @Override
+    public Set<ConfigKey<?>> findKeysPresent(Predicate<? super ConfigKey<?>> filter) {
+        return findKeys(filter, KeyFindingMode.PRESENT_BUT_RESOLVED);
+    }
+
+    protected enum KeyFindingMode { DECLARED_OR_PRESENT, PRESENT_BUT_RESOLVED, PRESENT_NOT_RESOLVED }
```
ahgittin: agree

core/src/main/java/org/apache/brooklyn/core/objs/BasicSpecParameter.java
```diff
@@ -404,7 +431,126 @@ public static void addParameters(AbstractBrooklynObjectSpec<?, ?> spec, List<? e
             spec.parametersAdd(BasicSpecParameter.fromSpec(loader.getManagementContext(), spec));
         }
         if (explicitParams.size() > 0) {
-            spec.parametersAdd(explicitParams);
+            spec.parametersReplace(resolveParameters(explicitParams, spec.getParameters(), spec));
+        }
+    }
+
+    /** merge parameters against other parameters and known and type-inherited config keys */
+    static Collection<SpecParameter<?>> resolveParameters(Collection<? extends SpecParameter<?>> newParams, Collection<? extends SpecParameter<?>> existingReferenceParamsToKeep, AbstractBrooklynObjectSpec<?,?> spec) {
```
ahgittin: it would be needed for the `TODO` section

core/src/main/java/org/apache/brooklyn/core/objs/BasicSpecParameter.java
```diff
@@ -404,7 +431,126 @@ public static void addParameters(AbstractBrooklynObjectSpec<?, ?> spec, List<? e
...
+        Map<String,SpecParameter<?>> existingToKeep = MutableMap.of();
+        if (existingReferenceParamsToKeep!=null) {
+            for (SpecParameter<?> p: existingReferenceParamsToKeep) { 
+                existingToKeep.put(p.getConfigKey().getName(), p);
+            }
+        }
+
+        List<SpecParameter<?>> result = MutableList.<SpecParameter<?>>of();
+        for (SpecParameter<?> p: newParams) {
+            if (p instanceof SpecParameterForInheritance) {
```
ahgittin: that would work but i put the extra branch in to show the logic that should be applied; have used your cleanup to extract `existingP` just once, and expanded the placeholder code so it's clearer

core/src/main/java/org/apache/brooklyn/core/config/BasicConfigInheritance.java
ahgittin: have thought about this:  the two conditions are independent dimensions, and it would matter if we used `localDefaultResolvesWithAncestorValue` for any purpose other than blocking inheritance from ancestors.  have added more comments to explain, including your words about the "independent dimensions" which is a good point:  including the control dimensions lets the logic be correct, even if we only expose a subset of possibilities to the end user (which simplifies a little bit the complexity of config inheritance).

camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/creation/BrooklynComponentTemplateResolver.java
```diff
@@ -403,8 +408,10 @@ protected ConfigInheritance getDefaultConfigInheritance() {
         // name "B" the "val2" then applies to both!
         //
         // I plan to propose a change on dev@brooklyn, to replace `@SetFromFlag`!
-        
-        Set<FlagConfigKeyAndValueRecord> allKeys = MutableSet.of();
+
+        // need to de-dupe? (can't use Set bc FCKVR doesn't impl equals/hashcode)
```
ahgittin: good job we didn't call it `FlagUnifiedConfigKeyValueRecord` :)

core/src/main/java/org/apache/brooklyn/core/mgmt/EntityManagementUtils.java
```diff
@@ -255,8 +255,9 @@ private static void mergeWrapperParentSpecToChildEntity(EntitySpec<? extends App
         wrappedChild.locationSpecs(wrapperParent.getLocationSpecs());
         wrappedChild.locations(wrapperParent.getLocations());
         
-        if (!wrapperParent.getParameters().isEmpty())
-            wrappedChild.parametersReplace(wrapperParent.getParameters());
+        if (!wrapperParent.getParameters().isEmpty()) {
+            wrappedChild.parametersAdd(wrapperParent.getParameters());
```
ahgittin: good spot

core/src/main/java/org/apache/brooklyn/core/objs/BasicSpecParameter.java
```diff
@@ -227,13 +234,30 @@ public String toString() {
...
             Predicate<?> constraint = parseConstraints(inputDef.get("constraints"), loader);
-            ConfigInheritance parentInheritance = parseInheritance(inputDef.get("inheritance.parent"), loader);
+
+            ConfigInheritance runtimeInheritance;
+            boolean hasRuntimeInheritance;
+            if (inputDef.containsKey("inheritance.runtime")) {
+                hasRuntimeInheritance = true;
+                runtimeInheritance = parseInheritance(inputDef.get("inheritance.runtime"), loader);
+            } else if (inputDef.containsKey("inheritance.parent")) {
+                // this alias will be deprecated
```
ahgittin: done

core/src/main/java/org/apache/brooklyn/core/objs/BasicSpecParameter.java
```diff
@@ -227,13 +234,30 @@ public String toString() {
...
+                hasRuntimeInheritance = true;
+                runtimeInheritance = parseInheritance(inputDef.get("inheritance.parent"), loader);
+            } else {
+                hasRuntimeInheritance = false;
+                runtimeInheritance = parseInheritance(null, loader);
```
ahgittin: done, thx

utils/common/src/main/java/org/apache/brooklyn/util/guava/Maybe.java
```diff
@@ -457,6 +457,8 @@ public boolean equals(Object obj) {
         Maybe<?> other = (Maybe<?>)obj;
         if (!isPresent()) 
             return !other.isPresent();
+        if (!other.isPresent())
+            return false;
```
ahgittin: that seems harsh to me; could cause problems if we serialized a `Maybe`?

could compare the exceptions but for now I will just add javadoc

neykov: I suppose it's useful when you put it inside containers (i.e. `remove(Object)` works for sets)

core/src/main/java/org/apache/brooklyn/core/config/BasicConfigInheritance.java
```diff
@@ -28,51 +32,188 @@
...
+            return returnEquivalentConstant(this);
+        }
+    }
+
+    private static ConfigInheritance returnEquivalentConstant(ConfigInheritance candidate) {
+        for (ConfigInheritance knownMode: Arrays.asList(
+                NOT_REINHERITED, NOT_REINHERITED_ELSE_DEEP_MERGE, NEVER_INHERITED, OVERWRITE, BasicConfigInheritance.DEEP_MERGE)) {
+            if (candidate.equals(knownMode)) return knownMode;
+        }
+        if (candidate.equals(new BasicConfigInheritance(false, CONFLICT_RESOLUTION_STRATEGY_OVERWRITE, true, true))) {
```
ahgittin: agree

core/src/main/java/org/apache/brooklyn/core/config/BasicConfigInheritance.java
ahgittin: @neykov my idea was for the class to such that if serialized it just writes the class name; won't your variant write all the fields passed to `BCI` ?

core/src/main/java/org/apache/brooklyn/core/config/BasicConfigInheritance.java
ahgittin: `equals` would work but it wouldn't catch quite as many errors; it should be invoked on the same _instance_; have added comment to explain.  you're right dropping the delegate pattern would allow this line to be removed but i think losing the internal fields in persisted state is worth it.

core/src/main/java/org/apache/brooklyn/core/config/BasicConfigInheritance.java
ahgittin: i think this should never happen; logging is safety first in case it is

core/src/main/java/org/apache/brooklyn/core/config/BasicConfigInheritance.java
ahgittin: reflection because fields are final; added comment

core/src/main/java/org/apache/brooklyn/core/config/BasicConfigInheritance.java
ahgittin: we don't have any subclasses so it doesn't matter; have switched to instance, assuming if it is subclasses then the author will extend the equals if needed.  also refactored so easier to read and if subclassed easier to extend the check.  (you're right, `instanceof` is better as it permits the caller to reuse some of the logic.)

core/src/main/java/org/apache/brooklyn/core/config/BasicConfigInheritance.java
```diff
@@ -28,51 +32,188 @@
...
+        final static BasicConfigInheritance DELEGATE = new BasicConfigInheritance(false, CONFLICT_RESOLUTION_STRATEGY_DEEP_MERGE, false, true); 
+        @Override protected ConfigInheritance getDelegate() { return DELEGATE; }
+    }
     /** As {@link #NOT_REINHERITED} but in cases where a value is inherited because a parent did not recognize it,
      * if the inheritor also defines a value the two values should be merged. */
-    public static BasicConfigInheritance NOT_REINHERITED_ELSE_DEEP_MERGE = new BasicConfigInheritance(false,CONFLICT_RESOLUTION_STRATEGY_DEEP_MERGE,false);
-    /** Indicates that a key's value should never be inherited, even if defined on a container that does not know the key.
-     * Most usages will prefer {@link #NOT_REINHERITED}. */
-    public static BasicConfigInheritance NEVER_INHERITED = new BasicConfigInheritance(false,CONFLICT_RESOLUTION_STRATEGY_OVERWRITE,true);
+    public static ConfigInheritance NOT_REINHERITED_ELSE_DEEP_MERGE = new NotReinheritedElseDeepMerge();
```
ahgittin: egads yes absolutely should be `static final` !

core/src/main/java/org/apache/brooklyn/core/config/BasicConfigInheritance.java
```diff
@@ -170,12 +324,103 @@ public String getConflictResolutionStrategy() {
...
+            
+        } catch (Exception e) {
+            throw Exceptions.propagate(e);
+        }
+                
+        return returnEquivalentConstant(this);
     }
 
     @Override
+    public boolean equals(Object obj) {
```
ahgittin: good spot, also added for `DelegatingCI`

core/src/main/java/org/apache/brooklyn/core/objs/BrooklynObjectInternal.java
ahgittin: yes - TY

ahgittin:
Thanks for such great comments.  Nearly done, should finish today.  Sorry I dropped the ball on this.

ahgittin:
failure appears non-deterministic and i think unrelated

```
Tests run: 2036, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 328.296 sec <<< FAILURE! - in TestSuite
(org.apache.brooklyn.core.entity.ApplicationLifecycleStateTest)  Time elapsed: 0.183 sec  <<< FAILURE!
java.lang.AssertionError: entity=Application[1naq8bil]; attribute=Sensor: service.state (org.apache.brooklyn.core.entity.lifecycle.Lifecycle) expected [running] but found [on-fire]
	at org.apache.brooklyn.core.entity.ApplicationLifecycleStateTest.assertHealthContinually(ApplicationLifecycleStateTest.java:196)
	at org.apache.brooklyn.core.entity.ApplicationLifecycleStateTest.testStartsThenChildFailsButWithQuorumCausesAppToSucceed(ApplicationLifecycleStateTest.java:170)
```

and

```
Failed tests: 
org.apache.brooklyn.core.entity.ApplicationLifecycleStateTest.(org.apache.brooklyn.core.entity.ApplicationLifecycleStateTest)
  Run 1: PASS
  Run 2: PASS
  Run 3: PASS
  Run 4: PASS
  Run 5: PASS
  Run 6: PASS
  Run 7: PASS
  Run 8: ApplicationLifecycleStateTest.testStartsThenChildFailsButWithQuorumCausesAppToSucceed:170->assertHealthContinually:196 entity=Application[1naq8bil]; attribute=Sensor: service.state (org.apache.brooklyn.core.entity.lifecycle.Lifecycle) expected [running] but found [on-fire]
  Run 9: PASS
  Run 10: PASS
```

ahgittin:
@geomacy good find in your failing test [above](https://github.com/apache/brooklyn-server/pull/462#issuecomment-270408039) -- we do not filter out hidden ancestor config keys whenever we populate the list of declared config keys, neither in:

* `BasicSpecParameter.fromSpec(..)`, where we extend a spec from a spec; nor
* `InternalEntityFactory.loadUnitializedEntity(...)`, where we create a spec for a java class; nor 
* `BrooklynDynamicType.buildConfigKeys(..)`, where we record the entity signature for a java type

i think it is worth fixing this.  :(

api/src/main/java/org/apache/brooklyn/api/internal/AbstractBrooklynObjectSpec.java
```diff
@@ -147,27 +147,28 @@ public SpecT tagsReplace(Iterable<Object> tagsToReplace) {
...
                 .addAll(params)
                 .addAll(current)
-                .build();
-        return self();
+                .build());
     }
     /** replaces parameters with the given */
     @Beta
-    public SpecT parametersReplace(List<? extends SpecParameter<?>> parameters) {
+    public SpecT parametersReplace(Collection<? extends SpecParameter<?>> parameters) {
```
ahgittin: going with `Iterable`

ahgittin:
more updates; now just need to:

    * fix the bug geomacy found
    * rewview comments from neykov re SpecParameterUnwrappingTest.java

core/src/main/java/org/apache/brooklyn/core/objs/BasicSpecParameter.java
```diff
@@ -404,7 +431,126 @@ public static void addParameters(AbstractBrooklynObjectSpec<?, ?> spec, List<? e
...
+        }
+        result.addAll(existingToKeep.values());
+        return result;
+    }
+
+    /** instance of {@link SpecParameter} which includes information about what fields are explicitly set,
+     * for use with a subsequent merge */
+    @SuppressWarnings("serial")
+    @Beta
+    static class SpecParameterForInheritance<T> extends BasicSpecParameter<T> {
```
ahgittin: the separate class is to know when something still needs to be inherited and when inheritance has been completed.  would welcome a second opinion here @neykov because it is ugly, but i think it should never leak ie the ugliness is entirely hidden.

core/src/main/java/org/apache/brooklyn/core/objs/BasicSpecParameter.java
```diff
@@ -404,7 +431,126 @@ public static void addParameters(AbstractBrooklynObjectSpec<?, ?> spec, List<? e
...
+            this.hasConstraints = hasConstraints;
+            this.hasRuntimeInheritance = hasRuntimeInheritance;
+            this.hasTypeInheritance = hasTypeInheritance;
+        }
+
+        /** used if yaml specifies a parameter, but possibly incomplete, and a spec supertype has a parameter */
+        SpecParameter<?> resolveWithAncestor(SpecParameter<?> ancestor) {
+            if (ancestor==null) return new BasicSpecParameter<>(getLabel(), isPinned(), getConfigKey(), getSensor());
+
+            return new BasicSpecParameter<>(
```
ahgittin: yes, should always be top-down, and the `SpecParameterForInheritance` items replaced immediately.  i'll have another look.

ahgittin:
^^ and neykov's comments re `SpecParameterForInheritance`

camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/catalog/SpecParameterUnwrappingTest.java
```diff
@@ -146,8 +151,9 @@ public void testRootParametersUnwrapped() {
         CatalogItem<?, ?> item = catalog.getCatalogItem(SYMBOLIC_NAME, TEST_VERSION);
         AbstractBrooklynObjectSpec<?,?> spec = catalog.peekSpec(item);
         List<SpecParameter<?>> params = spec.getParameters();
-        assertEquals(params.size(), NUM_APP_DEFAULT_CONFIG_KEYS + 1, "params="+params);
         assertTrue(Iterables.tryFind(params, nameEqualTo("simple")).isPresent());
+        assertTrue(Iterables.tryFind(params, nameEqualTo(SHARED_CONFIG.getName())).isPresent());
+        assertEquals(params.size(), NUM_APP_DEFAULT_CONFIG_KEYS + 2, "params="+params);
```
ahgittin: nice idea to use constants for all the counts.  have also added comment about unwrapping, agree it's a bit wasteful, but doesn't seem to do any harm.

camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/catalog/SpecParameterUnwrappingTest.java
```diff
@@ -309,7 +298,7 @@ public void testCatalogConfigOverridesParameters() {
         CatalogItem<?, ?> item = catalog.getCatalogItem(SYMBOLIC_NAME, TEST_VERSION);
         AbstractBrooklynObjectSpec<?,?> spec = catalog.peekSpec(item);
         List<SpecParameter<?>> params = spec.getParameters();
-        assertEquals(params.size(), 3);
+        assertEquals(params.size(), NUM_ENTITY_DEFAULT_CONFIG_KEYS + 2);
```
ahgittin: if we deployed i think it would be; here we're just doing `peekSpec` so it doesn't have the app config keys, it just has `NUM_ENTITY_DEFAULT_CONFIG_KEYS + ConfigEntityForTest.NUM_CONFIG_KEYS_DEFINED_HERE + NUM_CONFIG_KEYS_FROM_TEST_BLUEPRINT` = `1+1+1`

core/src/main/java/org/apache/brooklyn/core/objs/BasicSpecParameter.java
```diff
@@ -404,7 +431,126 @@ public static void addParameters(AbstractBrooklynObjectSpec<?, ?> spec, List<? e
...
+            //                  .remove(null)
+            //                  .build();
+            //          // order above is important, respected below to take the first one defined 
+            //          MutableMap<String, ConfigKey<?>> result = MutableMap.copyOf(FlagUtils.findAllConfigKeys(null, types));
+
+            return new BasicSpecParameter<>(
+                    getLabel(), 
+                    isPinned(), 
+                    resolveWithAncestorConfigKey(ancestor),
+                    // TODO port sensor will be lost (see messy code above which sets the port sensor)
```
ahgittin: no, we've never done proper combining of sensors when combinining a config definition with an ancestor.  and btw this method isn't really used yet, at least this code path isn't, it's in the TODO code path at https://github.com/apache/brooklyn-server/pull/462/files/a777b36e16688ccf9ef53ad6c1d392c8ef0f85bc#diff-94206c636ff5bbab4c4251c3316c129e

ahgittin:
@neykov @geomacy @aledsage Done.  Hurrah.  I've fixed the "declared keys" exclusions -- in the end @geomacy's suggestion re cleaning up the `resolveParameters` method was on point, and the missing case @neykov asked about was guaranteed not to occur, so there are a few other cleanups also.  Apologies for the complexity of this all (I wish we could move to YOML and do away with `Entity` subclasses!).  The main new PRs to review are the last two.  The others are mostly dull fixes.

core/src/main/java/org/apache/brooklyn/core/objs/BrooklynDynamicType.java
```diff
@@ -175,20 +179,29 @@ protected static void buildConfigKeys(Class<? extends BrooklynObject> clazz, Abs
...
+                }
+                if (key instanceof BasicConfigKeyOverwriting) {                            
+                    ConfigKey<?> parent = ((BasicConfigKeyOverwriting<?>)key).getParentKey();
+                    // find and remove the parent from consideration
+                    for (FieldAndValue<ConfigKey<?>> k2: kk) {
+                        if (value(k2) == parent)
+                            configKeysAll.remove(kn, k2);
                     }
                 }
                 kk = Lists.newArrayList(configKeysAll.get(kn));
```
geomacy: You've moved this update of `kk` inside the `for( ... k : kk)` loop, which doesn't look right. Updating the list variable each time round an iterator initialised from that variable?

ahgittin:
Could cause an issue if absents are used as keys with the intention that
they all be different but that would be weird. Have checked our hashCode
and it's a constant for absents already.

ahgittin:
Good spot. Will fix and merge.

utils/common/src/main/java/org/apache/brooklyn/util/guava/Maybe.java
neykov: Don't mind me, somehow missed the previous check where two absents will be equal.
My impression was exactly the opposite, that with the change two absents are never equal, even for `this.equals(this)`.

ahgittin:
Seems there may be a pathway by which REST does not populate an entity's `EntityType.configKeys` in this PR, reported by @sjcorbett .  Looking in to this now.

PR #463, opened 2016-11-24
if AnyoneSecurityProvider is set explicitly then switch it out for Nop

ahgittin:
improves logic of #434 -- https://github.com/apache/brooklyn-server/pull/434/commits/bfdf522c452df262b3b4bb55cd567ddf45b31fa7

that change caused some tests to fail if you had a security provider set in `brooklyn.properties` locally!  this ensures that an explicit `AnyoneSecurityProvider` is respected.


ahgittin:
@neykov as you did #434 can you sanity check this?  (should be very quick)

neykov:
LGTM, merging.

PR #456, opened 2016-11-23
Revert #452: setExpectedState sets actualState synchronously

aledsage:
@neykov (cc @Graeme-Miller) can you review please?

As discussed in https://github.com/apache/brooklyn-server/pull/454 (due to the failing test in its PR build for `LoopOverGroupMembersTestCaseTest.testMultipleChildrenWhichPass`), a lot of places expect the actual state to have been calculated/set by the time that `setExpectedState()` has returned.

Short-term, it's better than we live with the race-condition fixed by https://github.com/apache/brooklyn-server/pull/452, rather than change the semantics of `setExpectedState()`.

PR #464, opened 2016-11-24
Avoid uncaught exceptions in testOnDoneCallback

geomacy:
All tasks hit the listener, including tasks for starting the application
etc, so before this change the test logged errors as below. These fail the
assertion on the name, giving the messages below.

Change the test to avoid the assertion on name and to explicitly check for
the two tasks we want to find.

Note had to add the synchronization on 'completedTasks' as without it the
test would occasionally fail (~ one time in a dozen); as the state in the
exec thread wouldn't necessarily be propagated to the main thread without
synchronization.

2016-11-24 10:02:33,097 INFO  TESTNG INVOKING: "Surefire test" - org.apache.brooklyn.core.mgmt.internal.EntityExecutionManagerTest.testOnDoneCallback()
2016-11-24 10:02:33,099 INFO  TESTNG PASSED: "Surefire test" - org.apache.brooklyn.core.mgmt.internal.EntityExecutionManagerTest.testOnDoneCallback() finished in 2 ms
2016-11-24 10:02:33,100 INFO  TESTNG INVOKING CONFIGURATION: "Surefire test" - @AfterMethod org.apache.brooklyn.core.test.BrooklynMgmtUnitTestSupport.tearDown()
2016-11-24 10:02:33,104 ERROR Uncaught exception in thread brooklyn-execmanager-AxEAJeYA-4
java.lang.AssertionError: expected [foo] but found [null]
	at org.testng.Assert.fail(Assert.java:94) ~[testng-6.9.10.jar:na]
2016-11-24 10:02:33,105 ERROR Uncaught exception in thread brooklyn-execmanager-AxEAJeYA-3
java.lang.AssertionError: expected [foo] but found [null]
	at org.testng.Assert.fail(Assert.java:94) ~[testng-6.9.10.jar:na]
2016-11-24 10:02:33,105 ERROR Uncaught exception in thread brooklyn-execmanager-AxEAJeYA-6
java.lang.AssertionError: expected [foo] but found [[]]
	at org.testng.Assert.fail(Assert.java:94) ~[testng-6.9.10.jar:na]
2016-11-24 10:02:33,105 ERROR Uncaught exception in thread brooklyn-execmanager-AxEAJeYA-0
java.lang.AssertionError: expected [foo] but found [null]
	at org.testng.Assert.fail(Assert.java:94) ~[testng-6.9.10.jar:na]
2016-11-24 10:02:33,115 ERROR Uncaught exception in thread brooklyn-execmanager-AxEAJeYA-9
java.lang.AssertionError: expected [foo] but found [null]
	at org.testng.Assert.fail(Assert.java:94) ~[testng-6.9.10.jar:na]
2016-11-24 10:02:33,115 ERROR Uncaught exception in thread brooklyn-execmanager-AxEAJeYA-2
java.lang.AssertionError: expected [foo] but found [null]
	at org.testng.Assert.fail(Assert.java:94) ~[testng-6.9.10.jar:na]
2016-11-24 10:02:33,115 ERROR Uncaught exception in thread brooklyn-execmanager-AxEAJeYA-10
java.lang.AssertionError: expected [foo] but found [null]
	at org.testng.Assert.fail(Assert.java:94) ~[testng-6.9.10.jar:na]
2016-11-24 10:02:33,116 ERROR Uncaught exception in thread brooklyn-execmanager-AxEAJeYA-7
java.lang.AssertionError: expected [foo] but found [[[]]]
	at org.testng.Assert.fail(Assert.java:94) ~[testng-6.9.10.jar:na]

### DIFF
--- a/core/src/test/java/org/apache/brooklyn/core/mgmt/internal/EntityExecutionManagerTest.java
+++ b/core/src/test/java/org/apache/brooklyn/core/mgmt/internal/EntityExecutionManagerTest.java
@@ -27,7 +27,9 @@ import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.Semaphore;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 import org.apache.brooklyn.api.entity.Entity;
@@ -69,32 +71,47 @@ import com.google.common.collect.Lists;
 import com.google.common.util.concurrent.Callables;
 
 /** Includes many tests for {@link BrooklynGarbageCollector} */
+@Test
 public class EntityExecutionManagerTest extends BrooklynAppUnitTestSupport {
     
     private static final Logger LOG = LoggerFactory.getLogger(EntityExecutionManagerTest.class);
-    
-    @Test
+
     public void testOnDoneCallback() throws InterruptedException {
         ExecutionManager em = mgmt.getExecutionManager();
         BasicExecutionManager bem = (BasicExecutionManager)em;
         final Map<Task<?>,Duration> completedTasks = MutableMap.of();
-        final Semaphore sema4 = new Semaphore(-1);
+        final CountDownLatch latch = new CountDownLatch(2);
         bem.addListener(new ExecutionListener() {
             @Override
             public void onTaskDone(Task<?> task) {
                 Assert.assertTrue(task.isDone());
-                Assert.assertEquals(task.getUnchecked(), "foo");
-                completedTasks.put(task, Duration.sinceUtc(task.getEndTimeUtc()));
-                sema4.release();
+                Object result = task.getUnchecked();
+                if (result != null && result.equals("foo")) {
+                    synchronized (completedTasks) {
+                        completedTasks.put(task, Duration.sinceUtc(task.getEndTimeUtc()));
+                    }
+                    latch.countDown();
+                }
             }
         });
-        Task<String> t1 = em.submit( Tasks.<String>builder().displayName("t1").dynamic(false).body(Callables.returning("foo")).build() );
-        t1.getUnchecked();
-        Task<String> t2 = em.submit( Tasks.<String>builder().displayName("t2").dynamic(false).body(Callables.returning("foo")).build() );
-        sema4.acquire();
-        Assert.assertEquals(completedTasks.size(), 2, "completed tasks are: "+completedTasks);
-        completedTasks.get(t1).isShorterThan(Duration.TEN_SECONDS);
-        completedTasks.get(t2).isShorterThan(Duration.TEN_SECONDS);
+        Task<String> t1 = em.submit(
+            Tasks.<String>builder()
+                .displayName("t1")
+                .dynamic(false)
+                .body(Callables.returning("foo"))
+                .build());
+        Task<String> t2 = em.submit(
+            Tasks.<String>builder()
+                .displayName("t2")
+                .dynamic(false)
+                .body(Callables.returning("foo"))
+                .build());
+        latch.await(Asserts.DEFAULT_LONG_TIMEOUT.toMilliseconds(), TimeUnit.MILLISECONDS);
+        synchronized (completedTasks) {
+            Assert.assertEquals(completedTasks.size(), 2, "completed tasks should be 2 but are: "+completedTasks);
+            completedTasks.get(t1).isShorterThan(Duration.TEN_SECONDS);
+            completedTasks.get(t2).isShorterThan(Duration.TEN_SECONDS);
+        }
     }
     
     protected void forceGc() {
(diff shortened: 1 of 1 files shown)


core/src/test/java/org/apache/brooklyn/core/mgmt/internal/EntityExecutionManagerTest.java
```diff
@@ -68,6 +68,7 @@
 import com.google.common.util.concurrent.Callables;
 
 /** Includes many tests for {@link BrooklynGarbageCollector} */
+@Test
```
sjcorbett: All the tests are annotated - does this have any effect?

core/src/test/java/org/apache/brooklyn/core/mgmt/internal/EntityExecutionManagerTest.java
```diff
@@ -82,18 +83,33 @@ public void testOnDoneCallback() throws InterruptedException {
             @Override
             public void onTaskDone(Task<?> task) {
                 Assert.assertTrue(task.isDone());
-                Assert.assertEquals(task.getUnchecked(), "foo");
-                completedTasks.put(task, Duration.sinceUtc(task.getEndTimeUtc()));
-                sema4.release();
+                Object result = task.getUnchecked();
+                if(result != null && result.equals("foo")) {
```
sjcorbett: Nitpick: add a space between the `if` and the bracket.

sjcorbett:
Looks good. Couple of minor comments but happy for this to be merged with or without the changes.

core/src/test/java/org/apache/brooklyn/core/mgmt/internal/EntityExecutionManagerTest.java
```diff
@@ -82,18 +83,33 @@ public void testOnDoneCallback() throws InterruptedException {
...
+            .dynamic(false)
+            .body(Callables.returning("foo"))
+            .build());
+        Task<String> t2 = em.submit(
+            Tasks.<String>builder()
+                .displayName("t2")
+                .dynamic(false)
+                .body(Callables.returning("foo"))
+                .build());
         sema4.acquire();
```
aledsage: I much prefer:
```
        boolean acquired = sema4.tryAcquire(Asserts.DEFAULT_LONG_TIMEOUT.toMilliseconds(), TimeUnit.MILLISECONDS);
        assertTrue(acquired);
```
The current code will hang the build if someone changes something that breaks the test.

Also (for future reference - not worth changing here unless you really want to) a `CountDownLatch` is a more appropriate mechanism in this case - doing `new CountDownLatch(2)` is much simpler/clearer in its intent, compared to `new Semaphore(-1)`.

core/src/test/java/org/apache/brooklyn/core/mgmt/internal/EntityExecutionManagerTest.java
aledsage: Agreed - seems unnecessary.

It's also a little confusing because some tests in this class are marked with `@Test(groups="Integration")`, which will (I believe) take precedence over the class' test annotation, but makes people pause and think.

aledsage:
@geomacy looks good - only a couple of very minor comments from me. Then good to merge.

geomacy:
Thanks for comments @sjcorbett and @aledsage, will update now.

core/src/test/java/org/apache/brooklyn/core/mgmt/internal/EntityExecutionManagerTest.java
geomacy: I'll take out the plain `@Test`s  - I take it, though, that it's better to leave the `@Test(groups="Integration")` as-is?

geomacy:
hi @aledsage , @sjcorbett , changes made, rebased, squashed and pushed; hope this look ok.

core/src/test/java/org/apache/brooklyn/core/mgmt/internal/EntityExecutionManagerTest.java
```diff
@@ -69,32 +71,47 @@
...
+                .dynamic(false)
+                .body(Callables.returning("foo"))
+                .build());
+        Task<String> t2 = em.submit(
+            Tasks.<String>builder()
+                .displayName("t2")
+                .dynamic(false)
+                .body(Callables.returning("foo"))
+                .build());
+        latch.await(Asserts.DEFAULT_LONG_TIMEOUT.toMilliseconds(), TimeUnit.MILLISECONDS);
```
aledsage: FYI This will return `false` if not satisfied within the given time. But that's fine. The assertion below will then fail.

aledsage:
Merged. (Note jenkins failure was for `ApplicationLifecycleStateTest.testStartsThenChildFailsButWithQuorumCausesAppToSucceed` which I believe is fixed by https://github.com/apache/brooklyn-server/pull/456, already merged into master.)

PR #465, opened 2016-11-25
Change SetHostnameCustomizer to check if /etc/sysconfig/network exist…

iyovcheva:
…s on machine and create it if not

In some rare situations it's possible `/etc/sysconfig/network` to be missing on the VM on CentOS 6.
Changed the behavior of `SetHostnameCustomizer` to check and create the file.


iyovcheva:
The PR requires more testing. Will reopen it after it's stable for merging.

PR #468, opened 2016-11-25
Added functionality to overwrite private ip/hostname as public ip/hostname

Graeme-Miller:


geomacy:
hi @Graeme-Miller, just for info, what's the use-case here, why add this capability?

Graeme-Miller:
@geomacy a number of blueprints (e.g. Clocker) rely on the ability of nodes to communicate via the subnet IP. 

This functionality is useful in a situation where a cloud you are deploying to gives you a private IP, but you cannot configure that cloud to allow nodes to talk to one another via that IP. In this case, you could use this functionality to ensure that only the public IPs are used.

geomacy:
makes sense thanks @Graeme-Miller

neykov:
@Graeme-Miller would be useful to include the explanation as a comment. Also examples of concrete clouds where to expect this behaviour if supported by open source jclouds providers.

Graeme-Miller:
@neykov the issue I saw was with Azure, which is not currently packaged as part of Brooklyn.

andreaturli:
@aledsage re "@andreaturli and @Graeme-Miller looked into this for a while, but didn't manage to get the VMs to be added to the same Azure subnet" that's not entirely true. Adding
```
templateOptions:
  virtualNetworkName: your-net
  subnetNames: 
  - your-subnet
```
one can place VM in the specified subnet. Notice also that when in the sam subnet VMs can reach one another using private ip address *without* fiddling with `Azure endpoint` aka fw rules.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocationConfig.java
```diff
@@ -293,6 +293,12 @@
     public static final ConfigKey<Map<String,Object>> TEMPLATE_OPTIONS = ConfigKeys.newConfigKey(
             new TypeToken<Map<String, Object>>() {}, "templateOptions", "Additional jclouds template options");
 
+    @Beta
+    public static final ConfigKey<Boolean> USE_PUBLIC_ENDPOINT_AS_PRIVATE_ENDPOINT = ConfigKeys.newBooleanConfigKey(
+            "usePublicEndpointAsPrivateEndpoint",
```
aledsage: @Graeme-Miller I agree with @neykov that we should have a comment here to explain why it exists, and an example of when one would use it.

aledsage:
Thanks @Graeme-Miller - merging.

andreaturli:
I think my comment was sensible although not addressed. I suppose there are some other compelling use cases - that I'm ignoring now - where this configkey will be useful.

PR #470, opened 2016-11-29
Fix https tests

drigodwin:
This fixes HTTPS YAML tests. Previously the `trustAll(trustAll)` was not working due to no URL being specified during the build method which caused it to be ignored [here](https://github.com/apache/brooklyn-server/blob/master/utils/common/src/main/java/org/apache/brooklyn/util/http/HttpTool.java#L361).


aledsage:
Thanks @drigodwin - merging now.

PR #339, opened 2016-09-21
Updated jackson version

Graeme-Miller:
This is successor to [this PR](https://github.com/apache/brooklyn-server/pull/317).
After the rest client was moved to brooklyn-client in [this PR](https://github.com/apache/brooklyn-client/pull/25) two new PRs had to be created, one for brooklyn-server and one for brooklyn-client.

Please merge [this PR](https://github.com/apache/brooklyn-client/pull/28) at the same time.

This PR updates jackson to 2.7.5.

PR #472, opened 2016-11-30
Revert 5377839: Replace MultimapSerializer with jackson-datatype-guava

bostko:
Reverts https://github.com/apache/brooklyn-server/pull/339/commits/53778394ac9eee9d62fca9535f3c49b8f54f8adf in order to preserve the output format of `/v1/applications/{applicationId}/descendants/sensor/{sensor}`
The new library uses toString() rather than object representation used for this call.

Added simple test for retrieving map sensor.


rest/rest-server/src/test/java/org/apache/brooklyn/rest/entity/SensorsApiTest.java
```diff
@@ -0,0 +1,64 @@
...
+    @Test(groups = "Integration")
+    public void testGetHostAndPortSensor() throws Exception {
+        entity.sensors().set(Attributes.SSH_ADDRESS, UserAndHostAndPort.fromParts("testHostUser", "1.2.3.4", 22));
+
+        String sensorGetPath = "/v1/applications/"+app.getId()+"/entities/"+entity.getId()+"/sensors/" + Attributes.SSH_ADDRESS.getName();
+        assertEquals(httpGet(sensorGetPath), "\"testHostUser@1.2.3.4:22\"");
+
+        String descendantSensorPath = "/v1/applications/"+app.getId()+"/descendants/sensor/" + Attributes.SSH_ADDRESS.getName();
+        assertEquals(httpGet(descendantSensorPath), "{\"" + entity.getId() + "\":{\"user\":\"testHostUser\",\"hostAndPort\":{\"host\":\"1.2.3.4\",\"port\":22,\"hasBracketlessColons\":false}}}");
+    }
```
neykov: Could follow the [`testMultiMapSerialization`](https://github.com/apache/brooklyn-server/blob/master/rest/rest-resources/src/test/java/org/apache/brooklyn/rest/util/json/BrooklynJacksonSerializerTest.java#L195-L203) example how to structure this as a unit test. Agree no need to start an http server.

neykov:
LGTM. Suggest rewriting the test as commented above. Then good to merge.

I think this is good as a short term fix. Longer term would be nice to get the package back in, resolving the problems it has:
  * Changes in `HostAndPort` serialization. We can either accept the new format (warning users beforehand) or register a custom serializer which overrides the `toString` behaviour.
  * Binds to guava 18 (used by swagger) thus not able to work with guava 16 objects (used by Brooklyn). I believe we can upgrade our guava version after migrating to jclouds 1.9.3 which properly supports guava 16-20 version ranges.

rest/rest-resources/src/test/java/org/apache/brooklyn/rest/util/json/BrooklynJacksonSerializerTest.java
```diff
@@ -203,6 +204,14 @@ public void testMultiMapSerialization() throws Exception {
     }
 
     @Test
+    public void testUserHostAndPortSerialization() throws Exception {
```
bostko: @neykov suggested to add that test as well.
Thanks.

ahgittin:
@bostko is there a reason you didn't `git revert` this?  (looks like the commit is not a strict reversion as it adds new tests.)

@neykov good summary of the problems.  agree having the same guava version is an obvious improvement, but sounds like there will continue to be issues afterwards.

unless there is an easy fix where we can continue to deserialize the old format whilst also being able to serialize and deserialize the new format (which sounds like a lot of work, unless the `GuavaModule` supports it as an option), i'd be inclined _not_ to try to get this working and focus instead on restricting serialization/deserialization to explicitly registered types, ideally using yoml, and defining our own YOML serialization rules for those cases where we really need to serialize a guava type.

bostko:
@ahgittin you are right, sorry. Changed the PR to be with `git revert`.

neykov:
Thanks @bostko. I'll wait for tests to pass and will merge it then.

neykov:
Test failure seems unrelated:
```
CleanOrphanedLocationsIntegrationTest.testCleanedCopiedPersistedState:167

CleanOrphanedLocationsIntegrationTest.testSelectionWithOrphanedLocationsInData:136->AbstractCleanOrphanedStateTest.assertTransformDeletes:91->AbstractCleanOrphanedStateTest.assertTransformDeletes:101->AbstractCleanOrphanedStateTest.assertRawData:139 expected [true] but found [false]
```

Tests are passing locally. Merging.

PR #473, opened 2016-12-01
Test framework: avoid repeated stacktrace logging

aledsage:


test-framework/src/test/java/org/apache/brooklyn/test/framework/TestSensorTest.java
```diff
@@ -332,6 +343,56 @@ public void testFailFastIfNoSensor() throws Exception {
...
+                .configure(TestSensor.ASSERTIONS, newListAssertion("matches", String.format(".*%s.*", time))));
+
+        String loggerName = Repeater.class.getName();
+        ch.qos.logback.classic.Level logLevel = ch.qos.logback.classic.Level.DEBUG;
+        Predicate<ILoggingEvent> repeatedFailureMsgMatcher = EventPredicates.containsMessage("repeated failure; excluding stacktrace");
+        Predicate<ILoggingEvent> stacktraceMatcher = EventPredicates.containsExceptionStackLine(TestFrameworkAssertions.class, "checkActualAgainstAssertions");
+        Predicate<ILoggingEvent> filter = Predicates.or(repeatedFailureMsgMatcher, stacktraceMatcher);
+        LogWatcher watcher = new LogWatcher(loggerName, logLevel, filter);
+
+        Task<Void> task;
```
grkvlt: Why is this declared outside the `try`/`catch` block? It doesn't seem to be used anywhere else?

test-framework/src/test/java/org/apache/brooklyn/test/framework/TestSensorTest.java
```diff
@@ -332,6 +343,56 @@ public void testFailFastIfNoSensor() throws Exception {
...
+        watcher.start();
+        try {
```
grkvlt: Since `LogWartcher` is  `Closeable` can do this in the `try` block and it will be closed automatically.
```Java
try (LogWatcher watcher = new LogWatcher(loggerName, logLevel, filter)) {
    // stuff
}
```

test-framework/src/test/java/org/apache/brooklyn/test/framework/TestSensorTest.java
```diff
@@ -332,6 +343,56 @@ public void testFailFastIfNoSensor() throws Exception {
...
+            // Invoke async; will let it complete after we see the log messages we expect
+            task = Entities.invokeEffector(app, app, Startable.START, ImmutableMap.of("locations", locs));
```
grkvlt: Define variable here as `Task<Void> task = Entities...`

utils/common/src/main/java/org/apache/brooklyn/util/repeat/Repeater.java
```diff
@@ -334,8 +335,14 @@ public void runRequiringTrue() {
             try {
                 lastError = null;
                 done = exitCondition.call();
+                hasLoggedTransientException = false;
             } catch (Throwable e) {
-                if (log.isDebugEnabled()) log.debug(description, e);
+                if (hasLoggedTransientException) {
+                    if (log.isDebugEnabled()) log.debug(description + " (repeated failure; excluding stacktrace): " + e);
```
grkvlt: Do  not need to check `log.isDebugEnabled()` for simple logging like this

aledsage:
Thanks @grkvlt - I've changed the declaration of `Task<?> task` as suggested.
I haven't changed to use `try (...)` for closing the `LogWatcher` (we certainly could use that java 7 idiom, but don't do it in many other places either.
I haven't changed the `log.isDebugEnabled()` - other than noise in the code, there's a slight benefit of avoiding the string concatenation/creation. Given that the `isDebugEnabled` guard was already there (rather than me adding it), I'm fine leaving it.

Merging now.

PR #475, opened 2016-12-01
BROOKLYN-405: ssh doesn't log password environment variables

aledsage:
First commit just tidies up tests (allowing more of them to run as unit tests).

Second commit actually changes non-test code.

### DIFF
--- a/core/src/test/java/org/apache/brooklyn/location/ssh/SshMachineLocationIntegrationTest.java
+++ b/core/src/test/java/org/apache/brooklyn/location/ssh/SshMachineLocationIntegrationTest.java
@@ -103,6 +103,13 @@ public class SshMachineLocationIntegrationTest extends SshMachineLocationTest {
         super.testIsSshableWhenTrue();
     }
 
+    // Overridden just to make it integration (because `newHost()` returns a real ssh'ing host)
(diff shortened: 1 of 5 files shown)


core/src/test/java/org/apache/brooklyn/location/ssh/SshMachineLocationIntegrationTest.java
```diff
@@ -19,53 +19,206 @@
...
+        }
+    }
+
+    // Note: on some (home/airport) networks, `ssh 123.123.123.123` hangs seemingly forever.
+    // Make sure we fail, waiting for longer than the 70 second TCP timeout.
+    //
+    // Times out in 2m7s on Ubuntu Vivid (syn retries set to 6)
+    @Test(groups = "Integration")
+    public void testIsSshableWhenFalse() throws Exception {
+        byte[] unreachableIp = new byte[] {123,123,123,123};
```
aledsage: Thanks @grkvlt - I've changed it to use `240.0.0.0`, but it still takes 75 seconds for the test to fail! I just moved this code from `SshMachineLocationTest` into `SshMachineLocationIntegrationTest`, so I'm not inclined to spend more time on it during this PR.

PR #476, opened 2016-12-02
Upgrade guava to v18.0

neykov:
Depends on #457. Fixes wiring problems with the new jclouds in Karaf.

`jclouds 1.9.3` depends on guava [16-20), so it will bind to the latest version it finds - which is 18. And Brooklyn is stuck on 16.0.1. Bumping Brooklyn to 18.0 moves everything to depend on the same version of guava thus solving the wiring problems.

Note that `guava-16.0.1` is still pulled in the Karaf distribution by [jclouds](https://github.com/jclouds/jclouds-karaf/blob/master/feature/src/main/resources/feature.xml#L30). But is not loaded at runtime.


geomacy:
👍  looks good, will merge

PR #478, opened 2016-12-05
Fix project build, mark test as integration

neykov:
The test requires an ssh server to be listening on port 22.


ahgittin:
also discovered this -- merging

PR #479, opened 2016-12-05
Upgrade karaf-maven-plugin to avoid double User-Agent bug

neykov:
Deploying the project as is leads to the creation of two staging repos. With modules using the `karaf-maven-plugin` in one and the rest in the other bucket.

This is caused by a bug in `wagon-http-2.8` already fixed in maven 3.3.3. `karaf-maven-plugin` has a dependency on `wagon-http-2.8` so it pulls in the version even in versions of maven that use later (and fixed) releases.

Nexus will create a unique repo based on a number of factors - the IP of the user and the User-Agent being included. If one of those changes during deployment it will lead to a new staging repo being created.

Changes already applied on 0.10.0 branch.

Merge together with https://github.com/apache/brooklyn-library/pull/81 and https://github.com/apache/brooklyn-dist/pull/66.

### DIFF
--- a/karaf/features/pom.xml
+++ b/karaf/features/pom.xml
@@ -36,7 +36,7 @@
             <plugin>
                 <groupId>org.apache.karaf.tooling</groupId>
                 <artifactId>karaf-maven-plugin</artifactId>
-                <version>${karaf.version}</version>
+                <version>${karaf.plugin.version}</version>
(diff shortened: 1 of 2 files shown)


sjcorbett:
Looks good. Will merge this and the library/dist PRs.

PR #480, opened 2016-12-07
Config self reference fix

ahgittin:
builds on #462 fixing the remaining issues discussed in BROOKLYN-329

immediate eval could be improved yet further but this fixes scope/resolution issues and in general makes immediate things much more reliable, as well as fixing recursive references in config settings, and helping to debug bad task traces


aledsage:
@ahgittin This is way too hard to review until #462 is merged (which requires rebasing and a bunch of PR comments to be addressed; as I asked there, will you have time to address those or would you like one of us to pick it up)?

This PR also looks like it has a bunch of other commits that could (and should) be extracted into their own PRs (e.g. "solve problem with map eval" perhaps? Definitely "slow tests moved to integration group"? "cleanup, and allow TaskFactory to be supplied as a config..." perhaps?). I presume they are not all related to the title "Config self reference fix"?

Including many changes in a single PR makes the reviewers job much harder, and makes it take much much longer to get things merged.

ahgittin:
Rebased, and once #462 is merged this should be easier to review.  Cherry-picking my "slow tests" meant it automatically left this PR which was nice.

ahgittin:
Re `TaskFactory` being a separate PR, that would mean massaging the tests and other places so there is a clean state where `TF` isn't supported, and a clean state where `TF` is supported.  Agree in general PRs should be split up, but where something is trivial (integration tests), or where things are entangled (`TaskFactory`) it should be a judgment call.  We should avoid hypothetical problem-spotting:  "Your PR includes things not 100% related.  That's scientifically proven to make it much harder to review."  If in the course of a review it is _actually_ hard, then ask for a tidy-up.  I might get the balance wrong sometimes but I think here my atrocities are limited to having neglected #462 for months and months.

ahgittin:
Pretty sure the failures reported [here](https://builds.apache.org/job/brooklyn-server-pull-requests/1728/) are unrelated:

```
* org.apache.brooklyn.launcher.CleanOrphanedLocationsIntegrationTest.testCleanedCopiedPersistedState
* org.apache.brooklyn.launcher.CleanOrphanedLocationsIntegrationTest.testSelectionWithOrphanedLocationsInData
```

Both pass for me locally.  (Incidentally @aledsage none of the tests in that class are marked "Integration".)

ahgittin:
Unrelated test failures in penultimate jenkins build are discussed at https://github.com/apache/brooklyn-server/pull/560

core/src/test/java/org/apache/brooklyn/core/entity/EntityConfigTest.java
```diff
@@ -244,59 +252,152 @@ public void testGetConfigMapWithSubValueAsStringNotCoerced() throws Exception {
...
+    //
     // I (Aled) question whether we want to support passing a task (rather than a 
     // DeferredSupplier or TaskFactory, for example). Our EntitySpec.configure is overloaded
     // to take a Task, but that feels wrong!?
-    @Test(groups="Broken")
-    public void testGetTaskNonBlocking() throws Exception {
-        final CountDownLatch latch = new CountDownLatch(1);
-        Task<String> task = Tasks.<String>builder().body(
+    //
+    // If starting clean I (Alex) would agree, we should use TaskFactory. However the
```
aledsage: I suspect our `DependentConfiguration` class could do with a re-write, but not sure off-hand exactly what it should look like. And doesn't feel like it's worth doing right now.

It's trick because `WaitInTaskForAttributeReady extends Callable` expects to subscribe to (multiple) sensors, and then block for a matching value. We could most likely re-implement that (see below).

Using a `TaskFactory` would be tricky, because we only want to do the subscriptions once. It is conceptually a single "task" (i.e. we don't want to do the subscriptions multiple times).

My gut-feel is therefore we could implement it as a `DeferredSupplier`. When first called (or maybe when created), it would create the subscriptions. The thread executing the `get` would just be blocking on a `CountDownLatch`, which would be set by the subscription callbacks when the value was available (or when the entity had failed). Multiple calls to `get()` would share the same instance, so would all be blocking on the same thing.

core/src/test/java/org/apache/brooklyn/core/entity/EntityConfigTest.java
```diff
@@ -244,59 +252,152 @@ public void testGetConfigMapWithSubValueAsStringNotCoerced() throws Exception {
...
+        new ConfigNonBlockingFixture().usingTask().runGetConfigNonBlockingInKey(); }
+    @Test(groups="Integration") // because takes 1s+
+    public void testGetTaskNonBlockingMap() throws Exception {
+        new ConfigNonBlockingFixture().usingTask().runGetConfigNonBlockingInMap(); }
+    
+    @Test(groups="Integration") // because takes 1s+
+    public void testGetTaskFactoryNonBlockingKey() throws Exception {
+        new ConfigNonBlockingFixture().usingTaskFactory().runGetConfigNonBlockingInKey(); }
+    @Test(groups="Integration") // because takes 1s+
+    public void testGetTaskFactoryNonBlockingMap() throws Exception {
```
aledsage: Why does this now work? We still have the call to `task.cancel(true)` in `AbstractConfigurationSupportInternal.getNonBlockingResolvingStructuredKey()`, so during the first call to `getNonBlocking` we'll cancel the task. Is it because that is a different task that does not pass the `cancel()` call down to the actual config-key-value task that we are trying to evaluate?

utils/common/src/main/java/org/apache/brooklyn/util/exceptions/Exceptions.java
```diff
@@ -263,8 +264,41 @@ public static Throwable collapseIncludingAllCausalMessages(Throwable source) {
     public static Throwable collapse(Throwable source, boolean collapseCausalChain) {
         return collapse(source, collapseCausalChain, false, ImmutableSet.<Throwable>of(), new Object[0]);
     }
+
+    /** As {@link Throwables#getCausalChain(Throwable)} but safe in the face of perverse classes which return themselves as their cause or otherwise have a recursive causal chain. */
+    public static List<Throwable> getCausalChain(Throwable t) {
```
aledsage: Add to `ExceptionsTest`:
```
    @Test
    public void testGetCausalChain() throws Exception {
        Exception e1 = new Exception("e1");
        Exception e2 = new Exception("e2", e1);
        assertEquals(Exceptions.getCausalChain(e2), ImmutableList.of(e2, e1));
    }
    
    @Test
    public void testGetCausalChainRecursive() throws Exception {
        Exception e1 = new Exception("e1") {
            public synchronized Throwable getCause() {
                return this;
            }
        };
        Exception e2 = new Exception("e2", e1);
        assertEquals(Exceptions.getCausalChain(e2), ImmutableList.of(e2, e1));
    }
```

api/src/main/java/org/apache/brooklyn/api/mgmt/ExecutionContext.java
```diff
@@ -64,4 +65,6 @@
 
     boolean isShutdown();
 
+    <T> Maybe<T> getImmediately(Object callableOrSupplier);
```
ahgittin: good idea.  we should maybe move `ImmediateSupplier` to the utils package, then we could reference its javadoc?  we might also change the return type to be `ReferenceWithError<Maybe<T>>` so the "can't immediately tell if there's a value" problem state can be detected without throwing.  have marked `@Beta` for now.

camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/methods/DslComponent.java
```diff
@@ -539,8 +550,9 @@ protected String resolveKeyName(boolean immediately) {
         @Override
         public final Maybe<Object> getImmediately() {
             Maybe<Entity> targetEntityMaybe = component.getImmediately();
-            if (targetEntityMaybe.isAbsent()) return Maybe.absent("Target entity not available");
+            if (targetEntityMaybe.isAbsent()) return Maybe.<Object>cast(targetEntityMaybe);
             EntityInternal targetEntity = (EntityInternal) targetEntityMaybe.get();
+            checkAndTagForRecursiveReference(targetEntity);
```
ahgittin: the tag is left on the task.  you're right that could cause problems if the calling code isn't in a task (maybe it always will be but safer not to assume).  TODO remove tag afterwards

camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/ConfigYamlTest.java
```diff
@@ -91,6 +96,62 @@ public void testConfigInConfigBlock() throws Exception {
...
         assertNull(entity.getMyField2()); // field with @SetFromFlag("myField2Alias"), set using alias
     }
+    
+
+    @Test
+    public void testRecursiveConfigFailsGracefully() throws Exception {
+        doTestRecursiveConfigFailsGracefully(false);
+    }
+    
+    // TODO this test fails because entities aren't available when evaluating immediately
```
ahgittin: stale comment, good catch

camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/ConfigYamlTest.java
```diff
@@ -91,6 +96,62 @@ public void testConfigInConfigBlock() throws Exception {
...
+            public void run() {
+                try {
+                    Time.sleep(Duration.FIVE_SECONDS);
+                    // error, loop wasn't interrupted or detected
+                    LOG.warn("Timeout elapsed, destroying items; usage: "+
+                            ((LocalManagementContext)mgmt()).getGarbageCollector().getUsageString());
+                    //Entities.destroy(app);
+                } catch (RuntimeInterruptedException e) {
+                    // expected on normal execution
+                    Thread.interrupted();
```
ahgittin: exactly, comment added

core/src/main/java/org/apache/brooklyn/core/objs/AbstractConfigurationSupportInternal.java
```diff
@@ -146,7 +146,6 @@ public T call() {
                 .immediately(true)
                 .deep(true)
                 .context(getContext())
-                .swallowExceptions()
```
ahgittin: yeah, the immediate stuff is quite new so i'd prefer to try this.  if it's a problem we should perhaps look at wrapping in a `ReferenceWithError` as above, then caller will be forced to deal with the failure in the appropriate way.

core/src/test/java/org/apache/brooklyn/core/entity/EntityConfigTest.java
ahgittin: Interesting idea that it is a single (shared) task.  I feel like each caller gets its own instance/subscription hence `TaskFactory`.  But for another day... (and in any case I'd love us to move to a JS promise/then model)

core/src/test/java/org/apache/brooklyn/core/entity/EntityConfigTest.java
```diff
@@ -244,59 +252,152 @@ public void testGetConfigMapWithSubValueAsStringNotCoerced() throws Exception {
...
+            assertEquals(entity.config().get(TestEntity.CONF_MAP_OBJ_THING.subKey("mysub")), expectedVal);
+            
+            assertEquals(entity.config().getNonBlocking(TestEntity.CONF_MAP_OBJ_THING).get(), ImmutableMap.of("mysub", expectedVal));
+            assertEquals(entity.config().getNonBlocking(TestEntity.CONF_MAP_OBJ_THING.subKey("mysub")).get(), expectedVal);
+        }
     }
     
+    @Test(groups="Integration") // because takes 1s+
+    public void testGetTaskNonBlockingKey() throws Exception {
+        new ConfigNonBlockingFixture().usingTask().runGetConfigNonBlockingInKey(); }
```
ahgittin: not quite sure how this happened :|

core/src/test/java/org/apache/brooklyn/core/entity/EntityConfigTest.java
ahgittin: alex TODO investigate and explain.  also there is still one map test failing, investigate it also

core/src/main/java/org/apache/brooklyn/util/core/task/InterruptingImmediateSupplier.java
```diff
@@ -0,0 +1,103 @@
...
+ * then this class concludes that there is no value available immediately and returns {@link Maybe#absent()}.
+ * If the call throws any other error, that is returned.
+ * The interruption is cleared afterwards (unless the thread was interrupted when the method was entered).
+ * <p>
+ * Note that some "immediate" methods, such as {@link Semaphore#acquire()} when a semaphore is available,
+ * will throw if the thread is interrupted.  Typically there are workarounds, for instance:
+ * <code>if (semaphore.tryAcquire()) semaphore.acquire();</code>. 
+ */
+@Beta
+public class InterruptingImmediateSupplier<T> implements ImmediateSupplier<T>, DeferredSupplier<T> {
```
ahgittin: seeing as you wrote it it would be churlish not to add this!  :)

core/src/main/java/org/apache/brooklyn/util/core/task/ValueResolver.java
```diff
@@ -352,33 +357,51 @@ public T get() {
...
+                            : result;
                 } catch (ImmediateSupplier.ImmediateUnsupportedException e) {
                     log.debug("Unable to resolve-immediately for "+description+" ("+v+"); falling back to executing with timeout", e);
                 }
             }
             
+            // TODO if evaluating immediately should use a new ExecutionContext.submitImmediate(...)
+            // and sets a timeout but which wraps a task but does not spawn a new thread
+            
+            if ((v instanceof TaskFactory<?>) && !(v instanceof DeferredSupplier)) {
```
ahgittin: TODO think about this

ahgittin:
thanks @aledsage -- have done most of them.  a few more will require more thought from me, marked them so i can find them easily.

camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/methods/DslComponent.java
ahgittin: actually i've done a better strategy -- @aledsage appreciate any thoughts on this:

```
                // don't check on ourself; only look higher in hierarchy;
                // this assumes impls always spawn new tasks (which they do, just maybe not always in new threads)
                // but it means it does not rely on tag removal to prevent weird errors, 
                // and more importantly it makes the strategy idempotent
```

ahgittin: but actually this won't work either will it --

consider P1 needs to evaluate key C1, then key C2, and C2 refers to C1.  if it isn't a dedicated subtask the second check will fail.

guess we need to ensure a dedicated subtask. :( . i'll see whether we can do that, probably with an "official" task tag type.

ahgittin: did it, but without an official task type -- easy enough to get the routines to share code and trigger a dedicated (non-thread) task

ahgittin:
All should be fixed now @aledsage 

https://github.com/apache/brooklyn-server/pull/480/commits/99ccc0f6c8703e924c5fff5197bf1a0c6e39bc81 adds comments and assertions for the cancellation behaviour.  TL;DR, `TaskFactory` config values should have their tasks cancelled, but `Task` values should _not_.  The last commit fixes one thing which had been causing a failure in integration tests.

The remaining messy/broken tests seem to be mainly around immediate not having any guarantees due to timing.  I'd like to try to switch most/all to use the interrupting-immediate pattern, but that'll be another PR :) .

ahgittin:
@aledsage Thanks for thorough comments.  I'll review this afternoon and revert with whether to merge or not.

One high level comment:  `Task` instances _should_ be left running, even if immediate execution is requested, because someone else might subsequently ask for that value in a non-immediate context.  If we cancelled after immediate execution then the subsequent request would fail.  This isn't a **leak** but an implication of supporting `Task` evaluation (and for that reason I think we want to prefer `TaskFactory` evaluation).  I _think_ this addresses many of your comments; will look closer after lunch.

core/src/main/java/org/apache/brooklyn/util/core/task/ValueResolver.java
```diff
@@ -352,33 +357,73 @@ public T get() {
...
+            if (allowImmediateExecution && isEvaluatingImmediately()) {
+                // TODO could allow for everything, when evaluating immediately -- but if the target isn't safe to run again
+                // then we have to fail if immediate didn't work; to avoid breaking semantics we only do that for a few cases;
+                // might be nice to get to the point where we can break those semantics however, 
+                // ie weakening what getImmediate supports and making it be non-blocking, so that bailOut=true is the default.
+                // if: v instanceof TaskFactory -- it is safe, it's a new API (but it is currently the only one supported);
+                //     more than safe, we have to do it -- or add code here to cancel tasks -- because it spawns new tasks
+                //     (other objects passed through here don't get cancelled, because other things might try again later;
+                //     ie a task or future passed in here might naturally be long-running so cancelling is wrong,
+                //     but with a task factory generated task it would leak if we submitted and didn't cancel!)
```
ahgittin: DST should cancel -- this looks like a bug

camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/methods/DslComponent.java
```diff
@@ -554,15 +562,55 @@ protected String resolveKeyName(boolean immediately) {
...
-                            Entity targetEntity = component.get();
-                            String keyNameS = resolveKeyName(true);
-                            ConfigKey<?> key = targetEntity.getEntityType().getConfigKey(keyNameS);
-                            return targetEntity.getConfig(key != null ? key : ConfigKeys.newConfigKey(Object.class, keyNameS));
-                        }})
-                    .build();
+                    .body(newCallable(false)).build();
+        }
+
+        private Callable<Object> newCallable(final boolean immediate) {
```
ahgittin: yeah it's ugly; at least it's private.  have renamed.

core/src/main/java/org/apache/brooklyn/util/core/task/BasicExecutionContext.java
```diff
@@ -96,7 +98,50 @@ public ExecutionManager getExecutionManager() {
...
-     
+
+    /** performs execution without spawning a new task thread, though it does temporarily set a fake task for the purpose of getting context;
+     * currently supports suppliers or callables  */
+    @SuppressWarnings("unchecked")
+    @Override
+    public <T> Maybe<T> getImmediately(Object callableOrSupplier) {
+        BasicTask<?> fakeTaskForContext;
+        if (callableOrSupplier instanceof BasicTask) {
+            fakeTaskForContext = (BasicTask<?>)callableOrSupplier;
```
ahgittin: as noted above leaving the task running in background (or possibly not submitting it) is the right thing to do for immediate execution of a task; cancelling it is risky unless we know the task is dedicated to this one usage (hence preferring `TaskFactory`); in particular `EntityConfigTest.testGetTaskNonBlockingKey()` fails if `allowImmediateExecution=true`

ahgittin: As for the particular code block marked here, it is invoked from `ValueResolver` when you supply a `TaskFactory`; `VR` generates the `Task` and passes it in with `allowImmediateExecution=true`

core/src/main/java/org/apache/brooklyn/util/core/task/BasicExecutionContext.java
```diff
@@ -96,7 +98,50 @@ public ExecutionManager getExecutionManager() {
...
+            if (fakeTaskForContext.isQueuedOrSubmitted()) {
+                if (fakeTaskForContext.isDone()) {
+                    return Maybe.of((T)fakeTaskForContext.getUnchecked());
+                } else {
+                    throw new ImmediateUnsupportedException("Task is in progress and incomplete: "+fakeTaskForContext);
+                }
+            }
+            callableOrSupplier = fakeTaskForContext.getJob();
```
ahgittin: as above

core/src/main/java/org/apache/brooklyn/util/core/task/ValueResolver.java
```diff
@@ -352,33 +357,73 @@ public T get() {
...
+                    BrooklynTaskTags.setTransient(((TaskAdaptable<?>)v).asTask());
+                    if (isEvaluatingImmediately()) {
+                        // not needed if executing immediately
+                        BrooklynTaskTags.addTagDynamically( ((TaskAdaptable<?>)v).asTask(), BrooklynTaskTags.IMMEDIATE_TASK_TAG );
+                    }
+                }
+                
+                //if it's a task or a future, we wait for the task to complete
+                if (v instanceof TaskAdaptable<?>) {
+                    v = ((TaskAdaptable<?>) v).asTask();
```
ahgittin: as above, definitely not

core/src/main/java/org/apache/brooklyn/util/core/task/ValueResolver.java
ahgittin: aha, `DST` which we mostly use is fine, but its `doCancel` method was missing from the lighter-weight `CompoundTask`; have added it there and test passes

core/src/main/java/org/apache/brooklyn/util/core/task/ValueResolver.java
```diff
@@ -512,6 +559,24 @@ public Object call() throws Exception {
         }
     }
 
+    protected Maybe<T> execImmediate(ExecutionContext exec, Object immediateSupplierOrImmediateTask) {
+        Maybe<T> result;
+        try {
+            result = exec.getImmediately(immediateSupplierOrImmediateTask);
+        } catch (ImmediateSupplier.ImmediateUnsupportedException e) {
+            return null;
```
ahgittin: +1

ahgittin:
@aledsage I think I've addressed these comments, TY; but now merge conflicts with #155 are hurting; as noted there I think we should revert that.

ahgittin:
@aledsage minor changes addressing your comments, and fixed conflicts; would like your confirmation that an immediate `Task` "leaking" is the right behaviour (and longer term moving away from tasks as these values!) and anything else on your mind, but feels like it may be good to merge now :)

aledsage:
@ahgittin agreed - we should not cancel the task, so "leaking" it is right behaviour.

The build failure is real, until we commit to switching to Java 8 - please give your opinion on that email thread :-) 
```
[ERROR] /home/jenkins/jenkins-slave/workspace/brooklyn-server-pull-requests/core/src/main/java/org/apache/brooklyn/util/core/task/BasicExecutionContext.java:[31,25] error: package java.util.function does not exist
```
That package was added in Java 8.

I think it's just the wrong import (should be guava's; it's only used for the javadoc).

Can you fix that, and then merge?

ahgittin:
thanks @aledsage - fixed, if these tests pass, then good to merge

aledsage:
Excellent, thanks @ahgittin - merging now.

PR #481, opened 2016-12-07
[WIP] [NOT MERGE] Jclouds 2.0.0

googlielmo:
Built on top of @andreaturli 's #415 

Compile with jclouds 2.0.0
 - tweak karaf dependencies [WIP]
 - small changes to accommodate new signatures
 - sshj package rename throughout

TODO: 
- Fix karaf dependencies
- Fix failing unit tests

### DIFF
--- a/karaf/features/src/main/feature/feature.xml
+++ b/karaf/features/src/main/feature/feature.xml
@@ -266,7 +266,7 @@
         <bundle dependency="true">mvn:org.apache.jclouds.provider/google-compute-engine/${jclouds.version}</bundle>
 
         <feature>jclouds-services</feature>
-        <bundle dependency='true'>mvn:org.apache.servicemix.bundles/org.apache.servicemix.bundles.jsch-agentproxy-sshj/0.0.9_1</bundle>
+        <bundle dependency="true">wrap:mvn:com.jcraft/jsch.agentproxy.sshj/0.0.9$Bundle-Name=wrap%20%3A%3A%20jsch.agentproxy.sshj</bundle>
 
         <!-- Same as allcompute -->
         <feature>jclouds-aws-ec2</feature>
--- a/karaf/features/src/main/feature/feature.xml
+++ b/karaf/features/src/main/feature/feature.xml
@@ -144,7 +144,7 @@
         <bundle dependency="true">mvn:net.minidev/asm/${minidev.asm.version}</bundle>
         <bundle dependency="true">mvn:com.thoughtworks.xstream/xstream/${xstream.version}</bundle>
         <bundle dependency="true">mvn:org.freemarker/freemarker/${freemarker.version}</bundle>
-        <bundle dependency="true">mvn:net.schmizz/sshj/${sshj.version}</bundle>
+        <bundle dependency="true">mvn:com.hierynomus/sshj/${sshj.version}</bundle>
         <bundle dependency="true">mvn:org.apache.servicemix.bundles/org.apache.servicemix.bundles.jzlib/1.1.3_2</bundle> <!-- jzlib version is 1.1.3, but bundle is 1.1.3_2 -->
         <bundle dependency="true">mvn:org.bouncycastle/bcprov-ext-jdk15on/${bouncycastle.version}</bundle>
         <bundle dependency="true">mvn:org.bouncycastle/bcpkix-jdk15on/${bouncycastle.version}</bundle>
@@ -260,13 +260,13 @@
         <bundle dependency="true">mvn:org.apache.jclouds.driver/jclouds-okhttp/${jclouds.version}</bundle>
 
         <bundle dependency="true">mvn:org.apache.jclouds.driver/jclouds-sshj/${jclouds.version}</bundle>
-        <bundle dependency="true">mvn:org.apache.jclouds.labs/abiquo/${jclouds.version}</bundle>
-        <bundle dependency="true">mvn:org.apache.jclouds.labs/docker/${jclouds.version}</bundle>
+        <!--<bundle dependency="true">mvn:org.apache.jclouds.labs/abiquo/${jclouds.version}</bundle>--> <!-- jclouds abstraction not implemented -->
+        <bundle dependency="true">mvn:org.apache.jclouds.api/docker/${jclouds.version}</bundle>
 
-        <!-- Not bundle-ized, fixed in 2.0.0-SNAPSHOT -->
-        <!-- bundle dependency="true">mvn:org.apache.jclouds.labs/google-compute-engine/${jclouds.version}</bundle -->
+        <bundle dependency="true">mvn:org.apache.jclouds.provider/google-compute-engine/${jclouds.version}</bundle>
 
         <feature>jclouds-services</feature>
+        <bundle dependency='true'>mvn:org.apache.servicemix.bundles/org.apache.servicemix.bundles.jsch-agentproxy-sshj/0.0.9_1</bundle>
 
         <!-- Same as allcompute -->
         <feature>jclouds-aws-ec2</feature>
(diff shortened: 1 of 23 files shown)


bostko:
@googlielmo wdyt about opening the PR against andreaturli:jclouds2.0.0-SNAPSHOT

aledsage:
@googlielmo (cc @drigodwin @m4rkmckenna) to make the unit tests pass, you'll want to change `locations/jclouds/src/test/java/org/apache/brooklyn/location/jclouds/JcloudsStubTemplateBuilder.java`:
```
-        AtomicReference<AuthorizationException> authException = new AtomicReference<AuthorizationException>(new AuthorizationException());
+        AtomicReference<AuthorizationException> authException = new AtomicReference<AuthorizationException>();
```

With that change, `mvn clean install` of brooklyn-server works for me.

andreaturli:
I think we still have issues with `SshjToolIntegrationTest` though

karaf/features/src/main/feature/feature.xml
```diff
@@ -260,13 +260,13 @@
...
-        <bundle dependency="true">mvn:org.apache.jclouds.labs/docker/${jclouds.version}</bundle>
+        <!--<bundle dependency="true">mvn:org.apache.jclouds.labs/abiquo/${jclouds.version}</bundle>--> <!-- jclouds abstraction not implemented -->
+        <bundle dependency="true">mvn:org.apache.jclouds.api/docker/${jclouds.version}</bundle>
 
-        <!-- Not bundle-ized, fixed in 2.0.0-SNAPSHOT -->
-        <!-- bundle dependency="true">mvn:org.apache.jclouds.labs/google-compute-engine/${jclouds.version}</bundle -->
+        <bundle dependency="true">mvn:org.apache.jclouds.provider/google-compute-engine/${jclouds.version}</bundle>
 
         <feature>jclouds-services</feature>
+        <bundle dependency='true'>mvn:org.apache.servicemix.bundles/org.apache.servicemix.bundles.jsch-agentproxy-sshj/0.0.9_1</bundle>
```
m4rkmckenna: @googlielmo I had to also add
```xml
<bundle dependency='true'>mvn:net.schmizz/sshj/0.10.0</bundle>
```

andreaturli: I think we need `0.12.0` to match the version specified in the pom.xml, right?

FYI long-term I think this should come from `jclouds-services` feature of jclouds-karaf project, what do you think? I'll try to fix `jclouds-karaf`

m4rkmckenna: Yeah i just noticed that this was removed i dont know whats trying to pull it in ... looking now

m4rkmckenna: So `net.schmizz/sshj` is being pulled in by [org.apache.servicemix.bundles.jsch-agentproxy-sshj](https://github.com/apache/servicemix-bundles/blob/master/jsch-agentproxy-sshj-0.0.9/pom.xml#L75-L79)

And that is still required as `jclouds-ssh` has `net.schmizz.sshj` in its imports

andreaturli: be careful `sshj` has now a different `groupId` 
https://github.com/hierynomus/sshj/blob/master/README.adoc#depending-on-sshj

It should be an osgi bundle

m4rkmckenna: @andreaturli @googlielmo So I dont think we can get around including `net.schmizz.sshj` as [org.apache.servicemix.bundles.jsch-agentproxy-sshj](https://github.com/apache/servicemix-bundles/blob/master/jsch-agentproxy-sshj-0.0.9/pom.xml#L75-L79) and [jsch-agentproxy-sshj](https://github.com/ymnk/jsch-agent-proxy/blob/master/jsch-agent-proxy-sshj/pom.xml#L16-L20) depend upon it and there fore will not start. (Alos `jclouds-ssh` has `net.schmizz.sshj*` in its OSGi imports even though it excludes it explicity)

The good thing with the change to `com.hierynomus/sshj` is that nothing will import / use the old bundle as the symbolic name has changed.

andreaturli:
FYI looks like `jclouds-karaf` project is already dealing with sshj problem

those are the 2 relevant discussions 
https://issues.apache.org/jira/browse/JCLOUDS-1051
https://issues.apache.org/jira/browse/JCLOUDS-1156

I think `jclouds-karaf` is ready to switch back to sshj — see https://github.com/jclouds/jclouds-karaf/pull/77#issuecomment-244902646 and this will help long-term, I think.

aledsage:
@googlielmo (cc @andreaturli) to fix `SshjToolIntegrationTest` (the `testAsync.*` tests), you'll need this fix:

```
diff --git a/core/src/main/java/org/apache/brooklyn/util/core/internal/ssh/ShellAbstractTool.java b/core/src/main/java/org/apache/brooklyn/util/core/internal/ssh/ShellAbstractTool.java
index de94c98..d641478 100644
--- a/core/src/main/java/org/apache/brooklyn/util/core/internal/ssh/ShellAbstractTool.java
+++ b/core/src/main/java/org/apache/brooklyn/util/core/internal/ssh/ShellAbstractTool.java
@@ -428,7 +428,8 @@ public abstract class ShellAbstractTool implements ShellTool {
             cmdParts.add(
                     //ignore error output for the case where there are no running processes and kill is called without arguments
                     "ps aux | grep \"tail -c\" | grep \""+stdoutPath+"\" | grep -v grep | awk '{ printf $2 }' | xargs kill 2> /dev/null",
-                    "ps aux | grep \"tail -c\" | grep \""+stderrPath+"\" | grep -v grep | awk '{ printf $2 }' | xargs kill 2> /dev/null");
+                    "ps aux | grep \"tail -c\" | grep \""+stderrPath+"\" | grep -v grep | awk '{ printf $2 }' | xargs kill 2> /dev/null",
+                    "exit 0");
```

It looks like the `ShellAction`'s behaviour has changed: if there is not an explicit exit, then the it does not return (so we don't get any exit code). This shouldn't impact any Brooklyn entities, because they all use `SshjTool.execScript()`, which uploads the commands as a file and then executes it (always including a return statement). The reason brooklyn does that is that `ShellAction` had serious problems if buffers filled up (e.g. because the command was too big).

aledsage:
I ran the live tests listed below. Based on that, I'm happy for this to be merged (we can deal with Windows @ GCE later, which might be fixed with configuration):
* `AwsEc2LocationLiveTest` (slightly modified)
* `SoftlayerLocationLiveTest` (newly added, based on `AwsEc2LocationLiveTest`)
* `GceLocationLiveTest` (newly added, based on `AwsEc2LocationLiveTest`)
* `JcloudsLoginLiveTest`
* `JcloudsWinrmingLiveTest` (extended to test in AWS and GCE)

Windows @ GCE failed:
```
Caused by: java.lang.IllegalArgumentException: Missing config password[ConfigKey:java.lang.String] in {user=Administrator, spec.original=jclouds:google-compute-engine:us-central1-a, spec.final=jclouds:google-compute-engine:us-central1-a, address=104.154.209.43/104.154.209.43, port=5985, host=104.154.209.43, winrm.useNtlm=true}
	at org.apache.brooklyn.util.core.internal.winrm.winrm4j.Winrm4jTool.getRequiredConfig(Winrm4jTool.java:217)
```

`SoftlayerLocationLiveTest` had a worrying error logged, but then the test passed - see https://gist.github.com/aledsage/816460d719bf463f03672b72dc370d4a for details.

I also saw the worrying looking AWS log message below, which deserves some attention later - but the tests passed.
```
2016-12-13 12:17:14,786 WARN  o.j.l.s.i.GetRegionIdMatchingProviderURIOrNull [main]: failed to find key for value https://ec2.us-east-1.amazonaws.com in {eu-west-1=https://ec2.eu-west-1.amazonaws.com}; choosing first: eu-west-1
```

aledsage:
LGTM - merging. Thanks @googlielmo @andreaturli @m4rkmckenna

neykov:
@aledsage I'm seeing the same message with Brooklyn 0.10.0 RC2, based on jclouds 1.9.3
```
2016-12-15 07:25:29,249 WARN  o.j.l.s.i.GetRegionIdMatchingProviderURIOrNull [user thread 1]: failed to find key for value https://ec2.us-east-1.amazonaws.com in {eu-west-1=https://ec2.eu-west-1.amazonaws.com}; choosing first: eu-west-1
```

PR #482, opened 2016-12-08
DSL deferred execution

neykov:
Allows DSL to be applied on the return value of `attributeWhenReady` and `config` calls. Before the change the DSL calls are made on the `AttributeWhenReady` and `DslConfigSupplier` objects correspondingly. After the changes the value will first be resolved and then the DSL evaluated on result. `DslComponent` is a special case where it will not be resolved before evaluating the DSL as that's the expected behaviour.

-------------------------------

Allows us to call custom DSL methods on any object we are interested in without the object knowing about DSL. Also lets us do chaining.
For example:
```
attributeWhenReady("targetSensor").attributeWhenReady("mySensor")
attributeWhenReady("anyBrooklynObject").config("myConfig")
```

Builds on https://github.com/apache/brooklyn-server/pull/482

-------------------------------

Introduces `DslAccessible` annotation used to white-list methods which are allowed to be called by DSL, replacing the current approach of expecting them to be under the `org.apache.brooklyn.camp.brooklyn.spi.dsl.methods` package.

### DIFF
--- a/camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/DslAccessible.java
+++ b/camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/DslAccessible.java
@@ -1,0 +1,28 @@
+/*
+ * Copyright 2016 The Apache Software Foundation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.brooklyn.camp.brooklyn.spi.dsl;
+
+import static java.lang.annotation.ElementType.METHOD;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+@Retention(RUNTIME)
+@Target(METHOD)
+public @interface DslAccessible {
+
+}
--- a/camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/DslCallable.java
+++ b/camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/DslCallable.java
@@ -1,0 +1,26 @@
+/*
+ * Copyright 2016 The Apache Software Foundation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.brooklyn.camp.brooklyn.spi.dsl;
+
+import org.apache.brooklyn.util.core.task.DeferredSupplier;
+
+/**
+ * Marker interface so the evaluator can tell apart objects which are {@link DeferredSupplier}
+ * but which expect DSL methods called on them instead of the value they supply.
+ */
+public interface DslCallable {
+
+}
--- a/camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/DslDeferredFunctionCall.java
+++ b/camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/DslDeferredFunctionCall.java
@@ -16,6 +16,7 @@
 package org.apache.brooklyn.camp.brooklyn.spi.dsl;
 
 import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
 import java.util.List;
 import java.util.concurrent.Callable;
 
@@ -30,6 +31,7 @@ import org.apache.brooklyn.util.javalang.Reflections;
 
 import com.google.common.base.Joiner;
 import com.google.common.base.Objects;
+import com.google.common.collect.ImmutableList;
 
 public class DslDeferredFunctionCall extends BrooklynDslDeferredSupplier<Object> {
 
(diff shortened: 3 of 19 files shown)


camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/DslCallable.java
```diff
@@ -0,0 +1,26 @@
...
+ */
+package org.apache.brooklyn.camp.brooklyn.spi.dsl;
+
+import org.apache.brooklyn.util.core.task.DeferredSupplier;
+
+/**
+ * Marker interface so the evaluator can tell apart objects which are {@link DeferredSupplier}
+ * but which expect DSL methods called on them instead of the value they supply.
+ */
+public interface DslCallable {
```
neykov: I'm considering changing this to an annotation. When applied on:
  * A class will work as now - let the DSL evaluator know to use the object as is, stop evaluation
  * On a method - white-list the method for DSL calls. Deprecate non-annotated method calls. This will make it very easy to inspect what is executable by the DSL.

sjcorbett: I like this idea. Do you want to make it part of this PR or do it subsequently?

neykov: This one. Don't want to introduce APIs that will go away in master.

neykov: After spiking the changes decided to keep `DslCallable` interface for first case and introduce a `DslAccessible` annotation for second. They have different use-cases, better not to mix them in the same annotation.

Will send an email to dev@brooklyn.apache.org discussing the `DslAccessible` change and implement it in a separate PR.

camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/DslDeferredFunctionCall.java
```diff
@@ -0,0 +1,169 @@
...
+        return resolve(object, true);
+    }
+    protected Maybe<?> resolveBlocking(Object object) {
+        return resolve(object, false);
+    }
+    protected Maybe<?> resolve(Object object, boolean immediate) {
+        if (object instanceof DslCallable || object == null) {
+            return Maybe.of(object);
+        }
+        Maybe<?> resultMaybe = Tasks.resolving(object, Object.class)
```
ahgittin: How is this code reachable?  This class is only constructed when `object` is `DslCallable` in which case the above branch applies, isn't it?

ahgittin: nvm - missed the `!`

ahgittin:
makes sense although something niggles...

quick things:
* definitely `@DslAccessible` is clearer for methods
* maybe `DslFunctionSource` is a better name than `DslCallable` ... though feels like this interface shouldn't be needed; we simply skip evaluation in `DslDeferredFunctionCall` if `object` is already a valid source for invoking `fn` eg `object.getClass().getMethod(fn)` exists and is `@DslAccessible` (or maybe we always evaluate, given my re-reading of code, `DDFC` is only created when `object` is *not* a `DslCallable` ... or in an annotation world it would only be created when `object.fn()` is `@DslAccessible`)

my gut says a cleaner solution would be methods of the form `attributeWhenReady(Entity source, String sensorName)` and we do "C++ on C" style caller injection into the first argument.  this would solve the issue you mention where e.g. `attributeWhenReady("target.cluster.entity").attributeWhenReady("size")` isn't supported.  it's a bigger change perhaps than wanted at this point so happy to defer but let's keep that in mind and any interfaces or annotations we introduce along the way probably mark beta.

neykov:
Good idea about stopping once the function is available, will think about it.
The `DDFC` chaining is started on the first non-`DslCallable` `DeferredSupplier`. But said supplier could return a `DslCallable` after an unspecified number of resolve iterations. And it's known only at resolve time. Once you start chaining you can't stop as all subsequent calls depend on the return value of the previous which is already deferred.

Re `C++ on C` see https://github.com/apache/brooklyn-server/pull/486. Still think its orthogonal to `DDFC`.

neykov:
Did the removal of `DslCallable` locally. Works fine, but turns out we are losing one important feature of the current approach - if you misspell a method on `DslComponent` you won't find until executing the blueprint. Now it will complain when you try to deploy (at least until we switch to `DDFC`). The reason is that if it doesn't find the method on the `DslComponent` it will defer evaluation to get the underlying `Entity`.
Wdyt @ahgittin, is it worth keeping `DslCallable/DslFunctionSource` to get the (now limited) early warning?

ahgittin:
nice.  and good spot re errors.  probably is worth keeping, not just for errors, but also so that we have some hope of being able to propose autocompletions / populate graphical editor.  fine by me either way.

camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/DslYamlTest.java
```diff
@@ -0,0 +1,655 @@
...
+                "    id: sourceEntity",
+                "    brooklyn.initializers:",
+                "    - type: org.apache.brooklyn.core.sensor.StaticSensor",
+                "      brooklyn.config:",
+                "        name: source",
+                "        static.value: myvalue");
+        assertEquals(getConfigEventually(app, DEST), "myvalue");
+    }
+
+    @Test(groups="WIP") // config accepts strings only, no suppliers
```
geomacy: Cut-n-paste - think this comment should be `attributeWhenReady accepts strings only, no suppliers`, also on next method

camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/DslYamlTest.java
```diff
@@ -0,0 +1,655 @@
...
+        ConfigKey<TestDslSupplier> customCallableWrapperKey = ConfigKeys.newConfigKey(TestDslSupplier.class, "customCallableWrapper");
+        app.config().set(customCallableWrapperKey, new TestDslSupplier(new DslTestSupplierWrapper(new DslTestCallable())));
+        assertEquals(getConfigEventually(app, DEST), Boolean.TRUE);
+    }
+
+    private static <T> T getConfigEventually(final Entity entity, final ConfigKey<T> configKey) throws Exception {
+        Task<T> result = ((EntityInternal)entity).getExecutionContext().submit(new Callable<T>() {
+            @Override
+            public T call() throws Exception {
+                // TODO Move the getNonBlocking call out of the task after #280 is merged.
```
geomacy: Do you mean #480?

neykov:
Decided to stick with `DslFunctionSource`. Address all other comments. Merged with #486.

camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/DslDeferredFunctionCall.java
```diff
@@ -0,0 +1,215 @@
...
+
+        // TODO white-list using brooklyn.properties (at runtime)
+
+        Class<?> clazz = m.getDeclaringClass();
+        Package whiteListPackage = BrooklynDslCommon.class.getPackage();
+        boolean isPackageAllowed = (clazz.getPackage() != null && // Proxy objects don't have a package
+                    clazz.getPackage().getName().startsWith(whiteListPackage.getName()));
+        if (isPackageAllowed) {
+            if (DEPRECATED_ACCESS_WARNINGS.add(m)) {
+                log.warn("Deprecated since 0.11.0. The method '" + m.toString() + "' called by DSL should be white listed using the " + DslAccessible.class.getSimpleName() + " annotation. Support for DSL callable methods under the " + whiteListPackage + " will be fremoved in a future release.");
```
geomacy: typo "fremoved".   Would be nice to break this line up over several (currently 297 characters wide!)

camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/DslAccessible.java
```diff
@@ -0,0 +1,28 @@
...
+
+import static java.lang.annotation.ElementType.METHOD;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+@Retention(RUNTIME)
+@Target(METHOD)
+public @interface DslAccessible {
```
geomacy: It would be nice to add some javadoc around this to explain what it's for.

camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/methods/DslComponent.java
```diff
@@ -410,6 +424,7 @@ public String toString() {
         }
     }
 
+    @DslAccessible
```
geomacy: How is it that `attributeWhenReady()` and `config()` are able to be `@DslAccessible` when a few commits ago they did not return a value that implemented `DslCallable`?

geomacy: (i.e. I thought the difference was that e.g. `attributeWhenReady()` and `config()` couldn't be invoked within `BrooklynDslInterpreter.evaluateOn()` using `DslDeferredFunctionCall.invoke()`  and so for them a `DslDeferredFunctionCall` object gets created, whereas general `DslComponents` get evaluated with an immediate `DslDeferredFunctionCall.invoke()`.   Now, however, both general `DslComponent`s like `parent()` and also e.g. `attributeWhenReady()/config()` are marked as `@DslAccessible`.

geomacy: (does the comment on the PR header, "DslComponent is a special case where it will not be resolved before evaluating the DSL as that's the expected behaviour.", still apply?)

ahgittin:
LGTM, once @geomacy 's comments addressed

geomacy:
Actually I think my comments above are not worth holding this up for, and I answered my question above myself with a bit of further inspection and running tests, so happy to merge. LGTM.

PR #483, opened 2016-12-09
Use SnakeYAML SafeConstructor by default

rdowner:
Overridable by setting a system property.


utils/common/src/main/java/org/apache/brooklyn/util/internal/BrooklynSystemProperties.java
```diff
@@ -37,4 +37,6 @@
     public static StringSystemProperty HOST_GEO_LOOKUP_IMPL_LEGACY = new StringSystemProperty("brooklyn.location.geo.HostGeoLookup");
     public static StringSystemProperty HOST_GEO_LOOKUP_IMPL = new StringSystemProperty("org.apache.brooklyn.core.location.geo.HostGeoLookup");
 
+    /** Allows the use of YAML tags to create arbitrary types known to Java. */
+    public static BooleanSystemProperty YAML_TYPE_INSTANTIATION = new BooleanSystemProperty("org.apache.brooklyn.unsafe.YamlTypeInstantiation");
```
neykov: I believe using [BrooklynFeatureEnablement](https://github.com/apache/brooklyn-server/blob/master/core/src/main/java/org/apache/brooklyn/core/BrooklynFeatureEnablement.java) is the preferred method for configuring global behaviour. It allows for configuring through system properties and brookyn.properties.

neykov:
Follow up PR in https://github.com/apache/brooklyn-library/pull/82.

PR #484, opened 2016-12-09
Small JcloudsLocation refactor

sjcorbett:
I took the opportunity to move a number of the supporting classes in `JcloudsLocation` to their own top-level classes. I particular I touched `RebindToMachinePredicate`, the logic for creating users, and all of the `CustomizeTemplateOptions` and `CustomizeTemplateBuilders`.

It's a fairly large PR but I haven't introduced much that's new or changed any existing functionality. Most of it is ASF license headers. 

Suggest reviewing it with whitespace changes disabled (append `?w=1` to GitHub's URL).

### DIFF
--- a/locations/jclouds/src/test/java/org/apache/brooklyn/location/jclouds/JcloudsLocationTest.java
+++ b/locations/jclouds/src/test/java/org/apache/brooklyn/location/jclouds/JcloudsLocationTest.java
@@ -25,7 +25,6 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
-
 import javax.annotation.Nullable;
 
 import org.apache.brooklyn.api.location.LocationSpec;
@@ -41,7 +40,6 @@ import org.apache.brooklyn.core.location.cloud.names.CustomMachineNamer;
 import org.apache.brooklyn.core.location.geo.HostGeoInfo;
 import org.apache.brooklyn.core.mgmt.internal.LocalManagementContext;
 import org.apache.brooklyn.core.test.entity.LocalManagementContextForTests;
-import org.apache.brooklyn.location.jclouds.JcloudsLocation.UserCreation;
 import org.apache.brooklyn.test.Asserts;
 import org.apache.brooklyn.util.collections.MutableMap;
 import org.apache.brooklyn.util.core.config.ConfigBag;
@@ -580,8 +578,8 @@ public class JcloudsLocationTest implements JcloudsLocationConfig {
                         .put(JcloudsLocationConfig.USER, "bob").put(JcloudsLocationConfig.LOGIN_USER_PASSWORD, "b0b")
                         .putAll(config).build());
 
-        UserCreation creation = jl.createUserStatements(null, jl.config().getBag());
-        return new StatementList(creation.statements).render(OsFamily.UNIX);
+        CreateUserStatements creation = CreateUserStatements.get(jl, null, jl.config().getBag());
+        return new StatementList(creation.statements()).render(OsFamily.UNIX);
     }
     
     @Test
--- a/utils/common/src/main/java/org/apache/brooklyn/util/text/Strings.java
+++ b/utils/common/src/main/java/org/apache/brooklyn/util/text/Strings.java
@@ -28,7 +28,6 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.StringTokenizer;
-
 import javax.annotation.Nullable;
 
 import org.apache.brooklyn.util.collections.MutableList;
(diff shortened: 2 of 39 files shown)


sjcorbett:
P.S. to preempt Jenkins: tests pass locally.

aledsage:
Thanks @sjcorbett - great to separate out this huge class, and improve this code!

I haven't had a proper chance to review yet. Does it change the names of any classes that might be referenced in persisted state? Do we need to add anything to `deserializingClassRenames.properties`?

sjcorbett:
@aledsage My intention was to avoid anything that was persisted. Of the classes touched: 
* `RebindToMachinePredicate` was a private static class that is created in `getRebindToMachinePredicate` rather than injected through config. 
* `UserCreation` shouldn't have been persisted unless someone has done something strange, but still exists and is usable as before.
* Nobody should have persisted `SUPPORTED_TEMPLATE_OPTIONS_PROPERTIES` and `SUPPORTED_TEMPLATE_BUILDER_PROPERTIES`. All related classes were anonymous inner classes.

locations/jclouds/src/test/java/org/apache/brooklyn/location/jclouds/JcloudsLocationTest.java
```diff
@@ -580,8 +578,8 @@ protected String getCreateUserStatementsFor(Map<ConfigKey<?>,?> config) {
                         .put(JcloudsLocationConfig.USER, "bob").put(JcloudsLocationConfig.LOGIN_USER_PASSWORD, "b0b")
                         .putAll(config).build());
 
-        UserCreation creation = jl.createUserStatements(null, jl.config().getBag());
-        return new StatementList(creation.statements).render(OsFamily.UNIX);
+        CreateUserStatements creation = CreateUserStatements.get(jl, null, jl.config().getBag());
```
aledsage: Feels like we could (should?) extract these tests into a `CreateUserStatementsTest` class. But doesn't have to be done in this PR obviously.

utils/common/src/main/java/org/apache/brooklyn/util/text/Strings.java
```diff
@@ -827,6 +828,33 @@ public static String ies(Iterator<?> x) {
...
 
+    /**
+     * Tries to convert v to a list of strings.
+     * <p>
+     * If v is <code>null</code> then the method returns an empty immutable list.
+     * If v is {@link Iterable} or an <code>Object[]</code> then toString is called on its elements.
+     * If v is a {@link String} then a singleton list containing v is returned.
+     * Otherwise the method throws an {@link IllegalArgumentException}.
+     */
+    public static List<String> toStringList(Object v) {
```
aledsage: Would be good to add unit tests for this (but we can do that in a separate PR, given that this PR has just moved it from one place to another).

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/templates/customize/CustomizeTemplateOptions2.java
```diff
@@ -0,0 +1,103 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.brooklyn.location.jclouds.templates.customize;
+
+// Has "2" in its name so JcloudsLocation can use it without having to use a fully qualified reference.
+public class CustomizeTemplateOptions2 {
```
aledsage: Unfortunate that `CustomizeTemplateOptions` is already taken. I don't like the "2" in the name, but not sure what to suggest that would be better (unless we can delete the existing interface `CustomizeTemplateOptions`).

I don't think the interface `JcloudsLocation.CustomizeTemplateOptions` will be mentioned in any persisted state, so deleting it is just about source-level compatibility with anything that uses it directly. I lean towards treating that as a part of our "internal" API, so deleting it. It should be easy to fix any downstream projects that then fail to compile.

We'd then use the nicer name of `CustomizeTemplateOptions` for this class.

@sjcorbett Thoughts?

aledsage: Or we could go for different naming, such as `TemplateOptionsCustomizer` (and `TemplateBuilderCustomizer`). That is tempting, in that it uses a noun rather verb for the classname (i.e. follows standard Java naming conventions).

sjcorbett: I like your second suggestion. Will rename them when I resolve the conflict.

aledsage:
Thanks @sjcorbett - looks great. Unfortunately there are merge conflicts again (probably due to https://github.com/apache/brooklyn-server/pull/496).

Once those are resolved, and you've taken a look through the comments to see what you think is worth addressing, then this is good to merge. Happy for you to do that as soon as you're ready, before any more conflicts come in!

geomacy:
Looks good @sjcorbett, a useful tidy-up to this big class; have had a look through PR but nothing to add to above comments.

sjcorbett:
Thanks @aledsage and @geomacy. I will merge this now.

PR #485, opened 2016-12-12
`BundleMaker` REST call allowing to add ZIP/JAR to catalog

ahgittin:
and testing things like inserting a manifest and loading files from the bundles

not directly useful in Brooklyn yet, but some preparatory tests and utils:

* confirming that we should be able to refactor BOMs to have scripts etc in separate files, installing them as JAR bundles
* making it easier to work with ZIPs and JARs that are supplied
  * inserting manifest entries if needed (to install a raw ZIP, so we are attractive to non-java users)
  * testing the above, without needing maven to build bundles

@aledsage @neykov @geomacy worth your taking a quick look at the above so you know they are available, also comments on below useful ... will prep something for ML but obv what i write will be better if you give me ideas sooner :)

i think we should change `BrooklynFeatureEnablement.FEATURE_LOAD_BUNDLE_CATALOG_BOM` to be `true`

also thinking we want to:

* extend REST API to allow `JAR` bundle posted to `/v1/catalog`, w `catalog.bom` at root
* same for `ZIP` without `MANIFEST.MF` if bundle name supplied as API parameter
* support in `br` for the above two, also allowing upload of a directory

that will support a dev workflow where a user can work on files locally and then `br deploy` the directory

we can then set up a separate `update-blueprint-of-entities` REST operation so people have an easy way to develop and update

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/catalog/internal/BasicBrooklynCatalog.java
+++ b/core/src/main/java/org/apache/brooklyn/core/catalog/internal/BasicBrooklynCatalog.java
@@ -57,11 +57,13 @@ import org.apache.brooklyn.util.guava.Maybe;
 import org.apache.brooklyn.util.javalang.AggregateClassLoader;
 import org.apache.brooklyn.util.javalang.JavaClassNames;
 import org.apache.brooklyn.util.javalang.LoadedClassLoader;
+import org.apache.brooklyn.util.osgi.VersionedName;
 import org.apache.brooklyn.util.text.Strings;
 import org.apache.brooklyn.util.time.Duration;
 import org.apache.brooklyn.util.time.Time;
 import org.apache.brooklyn.util.yaml.Yamls;
 import org.apache.brooklyn.util.yaml.Yamls.YamlExtract;
+import org.osgi.framework.Version;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.yaml.snakeyaml.Yaml;
@@ -407,7 +409,7 @@ public class BasicBrooklynCatalog implements BrooklynCatalog {
     }
     
     @SuppressWarnings({ "unchecked", "rawtypes" })
-    private Maybe<Map<?,?>> getFirstAsMap(Map<?,?> map, String firstKey, String ...otherKeys) {
+    private static Maybe<Map<?,?>> getFirstAsMap(Map<?,?> map, String firstKey, String ...otherKeys) {
         return (Maybe) getFirstAs(map, Map.class, firstKey, otherKeys);
     }
 
@@ -415,6 +417,53 @@ public class BasicBrooklynCatalog implements BrooklynCatalog {
         List<CatalogItemDtoAbstract<?, ?>> result = MutableList.of();
         collectCatalogItems(yaml, result, ImmutableMap.of());
         return result;
+    }
+
+    public static Map<?,?> getCatalogMetadata(String yaml) {
+        Map<?,?> itemDef = Yamls.getAs(Yamls.parseAll(yaml), Map.class);
+        return getFirstAsMap(itemDef, "brooklyn.catalog").orNull();        
+    }
+    
+    public static VersionedName getVersionedName(Map<?,?> catalogMetadata) {
+        String id = getFirstAs(catalogMetadata, String.class, "id").orNull();
+        String version = getFirstAs(catalogMetadata, String.class, "version").orNull();
+        String symbolicName = getFirstAs(catalogMetadata, String.class, "symbolicName").orNull();
+        symbolicName = findAssertingConsistentIfSet("symbolicName", symbolicName, getFirstAs(catalogMetadata, String.class, "symbolic-name").orNull());
+        // prefer symbolic name and version, if supplied
+        // else use id as symbolic name : version or just symbolic name
+        // (must match if both supplied)
+        if (Strings.isNonBlank(id)) {
+            if (CatalogUtils.looksLikeVersionedId(id)) {
+                symbolicName = findAssertingConsistentIfSet("symbolicName", symbolicName, CatalogUtils.getSymbolicNameFromVersionedId(id));
+                version = findAssertingConsistentIfSet("version", version, CatalogUtils.getVersionFromVersionedId(id));
+            } else {
+                symbolicName = findAssertingConsistentIfSet("symbolicName", symbolicName, id);
+            }
+        }
+        if (Strings.isBlank(symbolicName) && Strings.isBlank(version)) {
+            throw new IllegalStateException("Catalog BOM must define symbolicName and version");
+        }
+        if (Strings.isBlank(symbolicName)) {
+            throw new IllegalStateException("Catalog BOM must define symbolicName");
+        }
+        if (Strings.isBlank(version)) {
+            throw new IllegalStateException("Catalog BOM must define version");
+        }
+        return new VersionedName(symbolicName, Version.valueOf(version));
+    }
+    
+    private static String findAssertingConsistentIfSet(String context, String ...values) {
+        String v = null;
+        for (String vi: values) {
+            if (Strings.isNonBlank(vi)) {
+                if (Strings.isNonBlank(v)) {
+                    if (!v.equals(vi)) throw new IllegalStateException("Mismatch in "+context+": '"+v+"' or '"+vi+"' supplied");
+                } else {
+                    v = vi;
+                }
+            }
+        }
+        return v;
     }
 
     private void collectCatalogItems(String yaml, List<CatalogItemDtoAbstract<?, ?>> result, Map<?, ?> parentMeta) {
--- a/rest/rest-api/src/main/java/org/apache/brooklyn/rest/api/CatalogApi.java
+++ b/rest/rest-api/src/main/java/org/apache/brooklyn/rest/api/CatalogApi.java
@@ -76,23 +76,20 @@ public interface CatalogApi {
                     name = "item",
                     value = "Item to install, as JAR/ZIP or Catalog YAML (autodetected)",
                     required = true)
-            byte[] item,
-            @ApiParam(name="name", value="Symbolic name to use for bundle", required=false, defaultValue="") String bundleName, 
-            @ApiParam(name="version", value="Version to set for bundle", required=false, defaultValue="") String bundleVersion);
+            byte[] item);
     
     @POST
     @Beta
     @Consumes({"application/x-zip", "application/x-jar"})
-    @ApiOperation(value = "Add a catalog item (e.g. new type of entity, policy or location) by uploading OSGi bundle JAR, or ZIP if bundle name and optionally version are supplied. "
+    @ApiOperation(value = "Add a catalog item (e.g. new type of entity, policy or location) by uploading OSGi bundle JAR, or ZIP which will be turned into bundle JAR, "
+            + "containing catalog.bom containing bundle name and version. "
             + "Return value is 201 CREATED if bundle could be added.", response = String.class)
     public Response createFromArchive(
             @ApiParam(
                     name = "archive",
-                    value = "Bundle to install, in JAR format, or ZIP if bundle name and optionally version are supplied, optionally with catalog.bom contained within",
+                    value = "Bundle to install, in ZIP or JAR format, requiring catalog.bom containing bundle name and version",
                     required = true)
-            byte[] archive,
-            @ApiParam(name="name", value="Symbolic name to use for bundle", required=false, defaultValue="") String bundleName, 
-            @ApiParam(name="version", value="Version to set for bundle", required=false, defaultValue="") String bundleVersion);
+            byte[] archive);
     
     @POST
     @Consumes(MediaType.APPLICATION_XML)
--- a/rest/rest-api/src/main/java/org/apache/brooklyn/rest/api/CatalogApi.java
+++ b/rest/rest-api/src/main/java/org/apache/brooklyn/rest/api/CatalogApi.java
@@ -38,11 +38,13 @@ import org.apache.brooklyn.rest.domain.CatalogItemSummary;
 import org.apache.brooklyn.rest.domain.CatalogLocationSummary;
 import org.apache.brooklyn.rest.domain.CatalogPolicySummary;
 
+import com.google.common.annotations.Beta;
+
 import io.swagger.annotations.Api;
-import io.swagger.annotations.ApiResponse;
-import io.swagger.annotations.ApiResponses;
 import io.swagger.annotations.ApiOperation;
 import io.swagger.annotations.ApiParam;
+import io.swagger.annotations.ApiResponse;
+import io.swagger.annotations.ApiResponses;
 
 @Path("/catalog")
 @Api("Catalog")
@@ -50,14 +52,48 @@ import io.swagger.annotations.ApiParam;
 @Produces(MediaType.APPLICATION_JSON)
 public interface CatalogApi {
 
-    @Consumes
+    @Deprecated /** @deprecated since 0.11.0 use {@link #createFromYaml(String)} instead */
+    public Response create(String yaml);
+
+    @Consumes({MediaType.APPLICATION_JSON, "application/x-yaml",
+        // see http://stackoverflow.com/questions/332129/yaml-mime-type
+        "text/yaml", "text/x-yaml", "application/yaml"})
     @POST
-    @ApiOperation(value = "Add a catalog item (e.g. new type of entity, policy or location) by uploading YAML descriptor "
+    @ApiOperation(value = "Add a catalog item (e.g. new type of entity, policy or location) by uploading YAML descriptor. "
         + "Return value is map of ID to CatalogItemSummary, with code 201 CREATED.", response = String.class)
-    public Response create(
+    public Response createFromYaml(
             @ApiParam(name = "yaml", value = "YAML descriptor of catalog item", required = true)
             @Valid String yaml);
 
+    @POST
+    @Beta
+    @Consumes  // anything (if doesn't match other methods with specific content types
+    @ApiOperation(value = "Add items to the catalog, either YAML or JAR/ZIP, format autodetected. "
+            + "Specify a content-type header to skip auto-detection and invoke one of the more specific methods. "
+            + "Return value is 201 CREATED if bundle could be added.", response = String.class)
+    public Response createPoly(
+            @ApiParam(
+                    name = "item",
+                    value = "Item to install, as JAR/ZIP or Catalog YAML (autodetected)",
+                    required = true)
+            byte[] item,
+            @ApiParam(name="name", value="Symbolic name to use for bundle", required=false, defaultValue="") String bundleName, 
+            @ApiParam(name="version", value="Version to set for bundle", required=false, defaultValue="") String bundleVersion);
+    
+    @POST
+    @Beta
+    @Consumes({"application/x-zip", "application/x-jar"})
+    @ApiOperation(value = "Add a catalog item (e.g. new type of entity, policy or location) by uploading OSGi bundle JAR, or ZIP if bundle name and optionally version are supplied. "
+            + "Return value is 201 CREATED if bundle could be added.", response = String.class)
+    public Response createFromArchive(
+            @ApiParam(
+                    name = "archive",
+                    value = "Bundle to install, in JAR format, or ZIP if bundle name and optionally version are supplied, optionally with catalog.bom contained within",
+                    required = true)
+            byte[] archive,
+            @ApiParam(name="name", value="Symbolic name to use for bundle", required=false, defaultValue="") String bundleName, 
+            @ApiParam(name="version", value="Version to set for bundle", required=false, defaultValue="") String bundleVersion);
+    
     @POST
     @Consumes(MediaType.APPLICATION_XML)
     @Path("/reset")
--- a/rest/rest-resources/src/main/java/org/apache/brooklyn/rest/resources/CatalogResource.java
+++ b/rest/rest-resources/src/main/java/org/apache/brooklyn/rest/resources/CatalogResource.java
@@ -30,7 +30,6 @@ import java.util.Map;
 import java.util.NoSuchElementException;
 import java.util.Set;
 import java.util.jar.Attributes;
-import java.util.jar.JarFile;
 import java.util.jar.Manifest;
 
 import javax.annotation.Nullable;
@@ -78,9 +77,13 @@ import org.apache.brooklyn.util.core.osgi.BundleMaker;
 import org.apache.brooklyn.util.exceptions.Exceptions;
 import org.apache.brooklyn.util.guava.Maybe;
 import org.apache.brooklyn.util.os.Os;
+import org.apache.brooklyn.util.osgi.VersionedName;
+import org.apache.brooklyn.util.stream.Streams;
 import org.apache.brooklyn.util.text.StringPredicates;
 import org.apache.brooklyn.util.text.Strings;
 import org.apache.brooklyn.util.yaml.Yamls;
+import org.apache.commons.compress.archivers.zip.ZipArchiveEntry;
+import org.apache.commons.compress.archivers.zip.ZipFile;
 import org.osgi.framework.Bundle;
 import org.osgi.framework.Constants;
 import org.slf4j.Logger;
@@ -112,10 +115,7 @@ public class CatalogResource extends AbstractBrooklynRestResource implements Cat
     static Set<String> missingIcons = MutableSet.of();
 
     @Override
-    public Response createPoly(
-            byte[] item,
-            String bundleName, 
-            String bundleVersion) {
+    public Response createPoly(byte[] item) {
         Throwable yamlException = null;
         try {
             MutableList.copyOf( Yamls.parseAll(new InputStreamReader(new ByteArrayInputStream(item))) );
@@ -125,15 +125,12 @@ public class CatalogResource extends AbstractBrooklynRestResource implements Cat
         }
         
         if (yamlException==null) {
-            // treat as yaml
-            if (Strings.isNonBlank(bundleName) || Strings.isNonBlank(bundleVersion)) {
-                throw new IllegalArgumentException("Bundle name/version not permitted when installing catalog YAML");
-            }
+            // treat as yaml if it parsed
             return createFromYaml(new String(item));
         }
         
         try {
-            return createFromArchive(item, bundleName, bundleVersion);
+            return createFromArchive(item);
         } catch (Exception e) {
             throw Exceptions.propagate("Unable to handle input: not YAML or compatible ZIP. Specify Content-Type to clarify.", e);
         }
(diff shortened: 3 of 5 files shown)


aledsage:
@ahgittin (cc @neykov) I've only had a quick look, but a few initial comments:

1. Discuss on dev@brooklyn the idea of deploying blueprints via a directory/zip (rather than a user having to write mvn build stuff to produce a bundle). That's a major new feature, so should be discussed early rather than doing (some of) the work first - sorry if I missed an email thread! I strongly agree with adding the feature, but we should propose/discuss it first.

2. Discuss on dev@brooklyn about what structure we should support for the zip files (e.g. `CatalogBomScanner.CatalogPopulator` I think looks for `catalog.bom` and then any brooklyn.libraries it lists). I'm all for convention-over-configuration, but want us to discuss what those conventions should be.

3. You suggested changing `BrooklynFeatureEnablement.FEATURE_LOAD_BUNDLE_CATALOG_BOM` to be true - let's discuss that on the mailing list as well.

4. You said "not directly useful in Brooklyn yet", but you've added this to the REST api. What do you mean by "not directly useful"?

5. I'd move `BundleMaker` into a utility module, instead of brooklyn-core. It's nice to force us to keep things more modular, and not to introduce unnecessary dependencies on brooklyn things when writing library code (I'd get rid of the constructor that takes the ManagementContext, and leave it to the caller to extract the framework).

6. There's no `BundleMakerTest` (but `BundleMaker` is used in `CatalogMakeOsgiBundleTest` and `CatalogResourceTest`). It would be good to have that, exercising each method.

ahgittin:
Agree on all points - thanks.

Best
Alex

core/src/main/java/org/apache/brooklyn/util/core/osgi/BundleMaker.java
```diff
@@ -0,0 +1,340 @@
...
+            
+        } catch (Exception e) {
+            throw Exceptions.propagate("Error starting bundle from "+f, e);
+        }
+    }
+    
+    private boolean addUrlItemRecursively(ZipOutputStream zout, String root, String item, Predicate<String> filter) throws IOException {
+        InputStream itemFound = null;
+        try {
+            itemFound = resources.getResourceFromUrl(item);
```
neykov: Wow haven't realised it's possible to read the listing of a classpath path. Had to run the test myself to make sure it works :)
Is this standard, documented somewhere? Does it work in Karaf?

ahgittin:
@jfknoepfli gave interesting feedback:  would be useful to be able to add/update elements to catalog items/bundles .. eg `POST /catalog/ID/resources/img/logo.png`, supplying a new (or changed) logo/content, and then `classpath://...` references in the blueprint point to the new content -- something like an implicit bundle.

maybe more useful / safer in a snapshot context but something to consider for making users' lives easier, giving them this sort of incremental feature rather than require client-side to ZIP everything each time.  i can see it also where the ZIP is huge and it's just the YAML changing.  could be snapshot-only but with option to "publish snapshot as non-snapshot versioned release".

ahgittin:
I think discussion on ML has been 100% supportive of the direction this goes in, with a lot of ideas of where this could go, and no pushback on the piece part introduced here, including the REST API.  (I hadn't expected to get it in to the REST API so easily but when I did I added it to this PR so you are right @aledsage, it does add a usable piece.)

Re minor comments, I'm not in favour of a new module just for this, re a Brooklyn-free version I don't see the point as this isn't meant as a utility that would be any use outwith Brooklyn.  Similarly re unit tests for individual tests for methods in `BundleMaker` will be a lot of work to write and the assertions about the output not what we actually care about so not worth it, whereas testing them in larger stories gives assurance that it makes bundles usable from Brooklyn which is what we care about.

Yet to be tested in Karaf mode but if that presents issues can we treat that as a bug @neykov ?  (I don't know an easy way to test Brooklyn in Karaf mode.)

aledsage:
@ahgittin agree there is 100% support on the direction.

There was push-back on what exactly the REST API piece should look like:
* @geomacy suggested that we should only support a zip that had a valid MANIFEST.MF file, and that the `br` command could auto-generate that manifest to make the user's life simpler.
* @neykov asked why we should require the bundle symbolic name and version being passed at all, suggesting they could be inferred from the the catalog.bom in some way (most likely with additional metadata).

It also sounded from the mailing list thread like enabling `BrooklynFeatureEnablement.FEATURE_LOAD_BUNDLE_CATALOG_BOM` could be fiddly/hard (to make it work in HA mode, on rebind, etc).

It would be good to separate out the two (or three?) things that this PR does so that we can merge them incrementally, rather than them all being blocked until there's consensus on everything:

1. The underlying building blocks for manipulating bundles etc could get merged asap.
2. The changes to the REST api seem like they need more discussion on the mailing list.

We should also discuss on the mailing list more about how `FEATURE_LOAD_BUNDLE_CATALOG_BOM` should really work - e.g. how will we solve the HA/rebind issues. And we should write down (on the mailing list) what exactly will be supported for such bundles - presumably "just" reading a root `catalog.bom` file; this would reference other things in the bundle using `classpath://...` which is an ugly java'ism, but a good starting point.

aledsage:
@ahgittin (cc @neykov @geomacy) see https://github.com/apache/brooklyn-server/pull/576, for the part of this PR that adds the `BundleMaker` (without the REST api changes). What do you think about merging that now?

ahgittin:
After refactoring in #576 (thanks @aledsage) this PR contains _only_ the REST call.  As for the principal suggestions on the REST API, responding both to @geomacy and @neykov :

Requiring a `MANIFEST.MF` makes the input strongly java-centric; I'd like to appeal to people who write YAML blueprints with co-bundled scripts and images.  They'd wonder why they can't simply make and upload a ZIP.  They could probably be persuaded to supply a name and optional version on a CLI or UI, and understand why it is needed (hence supporting those args) but it would not be idiomatic to anyone but a java programmer to create a `META-INF/` dir with a `MANIFEST.MF` and its syntax.  Meanwhile it is very easy for us to convert the ZIP to a JAR on the server.  Feels like uncontroversial good UX.

OTOH for a java programmer a `MANIFEST.MF` is natural, and they'd want to drop the name/version args if they are in that file, and I see no reason to forbid that pattern.

Can this merge once above agreed?

aledsage:
@ahgittin (cc @neykov @geomacy) I've kicked off the REST api discussion again on the dev@brooklyn mailing list ("subject "Re: Uploading ZIPs for a better dev workflow").

geomacy:
@ahgittin I take your point [above](https://github.com/apache/brooklyn-server/pull/485#issuecomment-283280366) about a manifest being java centric.  That's persuaded me to not insist on it (I agree it would be nice to support it if the developer supplies one).

At the same time I think we should insist that they _somehow_ provide an explicit symbolic name and version.  I would have thought that the version should be mandatory?  How would we treat the bundle if it just had a name?  Do we treat it as a version 0.0.0-SNAPSHOT, so to speak? (i.e. lower than anything with an explicitly supplied version?)

ahgittin:
@geomacy yes, either insist on a version of default to `0.0.0.SNAPSHOT` ... no strong feelings here

geomacy:
Having thought some more about such a default I suggested the following on the mailing list:

I'd say let's not even be that flexible - my feeling is that the more flexible we are, the more confusing it is, and the harder to get right.  If we're going to add this capability let's just have one way to do it; if you are going to do a ZIP upload, you MUST have a catalog.bom, which MUST contain a manifest section, which MUST have both a symbolic name and a version.  That way everything's explicit and very clear, every time.  

It also means clients ('br' at least) can easily validate the upload even before execution by checking that all the manifest details are present.

ahgittin:
@geomacy Great point.  As discussed on ML, 100% agree.

> It also means clients ('br' at least) can easily validate the upload

This sounds like an anti-pattern to me.  There's very little benefit to duplicating in the client what the server will do anyway.  If the server is non-contactable any error is a moot point.  :)

For convenience, the current proposal from the ML is to require symbolic name and version at the first level in a catalog entry:

```
    brooklyn.catalog:
       symbolicName: com.example.Foo
       version: 1.0
       items:
       - id: foo
         version: 3.0   # optional version override
         itemType: entity
         item:
           type: blah
           ...
```

ahgittin:
(Marked issue as WIP while I implement the revised proposal.)

ahgittin:
New semantics applied, ready for re-review.

@geomacy How hard would it be to extend the CLI to support this new feature, specifically:

* `br catalog add some.zip` -- an explicit zip or jar
* `br catalog add /path/to/folder/` -- a folder which will be zipped or jarred up

That would make user flow very nice!

geomacy:
I don't think it should be too hard now that the API is in place, I will look into doing that as soon as I get the decks cleared a bit.

ahgittin:
awesome @geomacy

ahgittin:
**Caution**:  Bundles installed via this mechanism are not persisted so there will be errors until bundles are installed into any new environment.  This is deliberate, I'm trying to do smaller PRs!  It shouldn't be too hard to fix but @m4rkmckenna @hzbarcea is there any magic / OSGi standard mechanisms for sharing bundles across different containers and ensuring they're installed by default on new machines?

ahgittin:
bump @geomacy @aledsage @neykov

rest/rest-resources/src/test/java/org/apache/brooklyn/rest/resources/CatalogResourceTest.java
```diff
@@ -487,6 +495,79 @@ public void testAddMissingItem() {
...
+        List<String> actualInterfaces = ((Map<String, List<String>>) tag).get("traits");
+        List<Class<?>> expectedInterfaces = Reflections.getAllInterfaces(TestEntity.class);
+        assertEquals(actualInterfaces.size(), expectedInterfaces.size());
+        for (Class<?> expectedInterface : expectedInterfaces) {
+            assertTrue(actualInterfaces.contains(expectedInterface.getName()));
+        }
+
+        byte[] iconData = client().path("/catalog/icon/" + symbolicName + "/" + version).get(byte[].class);
+        assertEquals(iconData.length, 43);
+    }
```
tbouron: It would be worth to create more simple tests for:
- `version` mismatch
- `symbolicName` mismatch
- `symbolicName` and/or `version` missing
- uploading a zip, i.e. no manifest file

rest/rest-api/src/main/java/org/apache/brooklyn/rest/api/CatalogApi.java
```diff
@@ -38,27 +38,60 @@
...
+                    required = true)
+            byte[] item);
+    
+    @POST
+    @Beta
+    @Consumes({"application/x-zip", "application/x-jar"})
+    @ApiOperation(value = "Add a catalog item (e.g. new type of entity, policy or location) by uploading OSGi bundle JAR, or ZIP which will be turned into bundle JAR, "
+            + "containing catalog.bom containing bundle name and version. "
+            + "Return value is 201 CREATED if bundle could be added.", response = String.class)
+    public Response createFromArchive(
```
tbouron: This returns a 500 is case there is an issue with the archive. Should be 400 instead as this is because of a bad request

ahgittin:
Thanks @tbouron .  Seems this maybe picked up an httpcomponents version bump in the confusion, not helped by the failing brooklyn tests introduced in master (they are trying to fix in #599).  will have a look.

tbouron:
@ahgittin Another thing I just noticed is that we don't have the classic `POST /v1/catalog` endpoint that takes a YAML anymore in the Swagger UI:

screen shot 2017-03-21 at 09 59 24

geomacy:
@tbouron, those failing tests with CompositeEffectorYamlRebindTest and friends are resolved by https://github.com/apache/brooklyn-server/pull/599 and https://github.com/apache/brooklyn-server/pull/601, if you are wanting to test again you'll need to merge those.  

@ahgittin thanks for the reminder, I'll look again at this today

tbouron:
@geomacy @ahgittin Rebased on the latest master, the build error is gone (still need to [remove line 63](https://github.com/tbouron/brooklyn-library/blob/d7bf1595127dc96cd63bc845084fd44bad94df94/software/webapp/pom.xml#L63) though)

Retested on a clean distribution with no persisted state. Create a ZIP with the following bom:
```
brooklyn.catalog:
  symbolicName: test
  version: 0.1.0
  items:
    - id: entity-1
      itemType: entity
      name: Entity 1
      item:
        type: org.apache.brooklyn.entity.software.base.VanillaSoftwareProcess
    - id: entity-2
      itemType: entity
      name: Entity 2
      item:
        type: org.apache.brooklyn.entity.software.base.VanillaSoftwareProcess
```

This fails because of the symbolicName inheritance: both `entity-1` and `entity-2` have the same `symbolicName` therefore it needs to be overridden on each item. Not an issue per se, but the error is quite confusing at first:
```
Updating existing catalog entries is forbidden: test:0.1.0. Use forceUpdate argument to override.
```

I was also wondering why we are introducing the field `symbolicName`? Effectively, that's the same as the `id` (which internally is referenced as... `symbolicName`) From a user perspective, it would be easier to require `id` and most of the blueprints do declare `id`s for each items. WDYT?

geomacy:
I'd have thought there should not be a clash in the above example - from Alex's email [here](https://lists.apache.org/thread.html/ca13cfbc334725a6865bdb9e02a97d1a72af98fcdf04f5a58e32038c@%3Cdev.brooklyn.apache.org%3E) the idea is 

> That is, the fields symbolic-name and version are _required_ under the 
> `brooklyn.catalog` when POSTing a ZIP to the catalog endpoint, and are 
> used to define the bundle. The version (and icon, etc) are then 
> inherited by default by items defined therein. If the ZIP is a JAR 
> which contains a MANIFEST.MF, the bundle information (symbolic name and 
> version) must match exactly.

my understanding here was that the symbolicName wasn't involved in the definition of entity-1 and 2, just in the naming of the bundle?

tbouron:
@geomacy That was my understanding as well but my tests show a different behaviour. When I import my zip (with the above bom), `entity-1` is created under the id `test:0.1.0` which is wrong:

screen shot 2017-03-21 at 11 41 34

then the catalog process `entity-2` which has the same `symbolicName` and `version` and therefore fails.

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/resources/CatalogResource.java
```diff
@@ -106,6 +152,7 @@ public Response create(String yaml) {
         try {
             items = brooklyn().getCatalog().addItems(yaml);
         } catch (Exception e) {
+            e.printStackTrace();
```
aledsage: Should log this properly, rather than to stdout.

core/src/main/java/org/apache/brooklyn/core/catalog/internal/BasicBrooklynCatalog.java
```diff
@@ -417,6 +419,53 @@ else for (String key: otherKeys) {
...
+    public static Map<?,?> getCatalogMetadata(String yaml) {
+        Map<?,?> itemDef = Yamls.getAs(Yamls.parseAll(yaml), Map.class);
+        return getFirstAsMap(itemDef, "brooklyn.catalog").orNull();        
+    }
+    
+    public static VersionedName getVersionedName(Map<?,?> catalogMetadata) {
+        String id = getFirstAs(catalogMetadata, String.class, "id").orNull();
+        String version = getFirstAs(catalogMetadata, String.class, "version").orNull();
+        String symbolicName = getFirstAs(catalogMetadata, String.class, "symbolicName").orNull();
+        symbolicName = findAssertingConsistentIfSet("symbolicName", symbolicName, getFirstAs(catalogMetadata, String.class, "symbolic-name").orNull());
```
aledsage: Can we choose one of "symbolicName" or "symbolic-name", and just support one? Why support both?

core/src/main/java/org/apache/brooklyn/core/catalog/internal/BasicBrooklynCatalog.java
```diff
@@ -417,6 +419,53 @@ else for (String key: otherKeys) {
...
+        // prefer symbolic name and version, if supplied
+        // else use id as symbolic name : version or just symbolic name
+        // (must match if both supplied)
+        if (Strings.isNonBlank(id)) {
+            if (CatalogUtils.looksLikeVersionedId(id)) {
+                symbolicName = findAssertingConsistentIfSet("symbolicName", symbolicName, CatalogUtils.getSymbolicNameFromVersionedId(id));
```
aledsage: Not sure I agree with this (and I think it contradicts what was discussed on the mailing list). Do we want to support something like the (contrived) example below:
```
brooklyn.catalog:
  symbolicName: com.example.brooklyn.test-utils
  id: test-case
  version: 1.0
  item:
    type: org.apache.brooklyn.test.framework.TestCase
    ...
```
Here, the name-space for the entity "test-case" is entirely different from the namespace of the bundle ("com.example..."). Your code would force that the symbolicName and id are identical.

core/src/main/java/org/apache/brooklyn/core/catalog/internal/BasicBrooklynCatalog.java
```diff
@@ -417,6 +419,53 @@ else for (String key: otherKeys) {
...
+                version = findAssertingConsistentIfSet("version", version, CatalogUtils.getVersionFromVersionedId(id));
+            } else {
+                symbolicName = findAssertingConsistentIfSet("symbolicName", symbolicName, id);
+            }
+        }
+        if (Strings.isBlank(symbolicName) && Strings.isBlank(version)) {
```
aledsage: Should be `||` instead of `&&`? (or exception message is wrong).

aledsage: Ah, I should have read a few lines lower down. Deleting previous comment :-) Should have do as a review so you didn't see it until I was finished...

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/resources/CatalogResource.java
```diff
@@ -94,9 +113,36 @@ public CatalogItemSummary apply(@Nullable CatalogItem input) {
...
+        
+        if (yamlException==null) {
+            // treat as yaml if it parsed
+            return createFromYaml(new String(item));
+        }
+        
+        try {
+            return createFromArchive(item);
+        } catch (Exception e) {
+            throw Exceptions.propagate("Unable to handle input: not YAML or compatible ZIP. Specify Content-Type to clarify.", e);
```
aledsage: As per @tbouron's comment, this presumably results in the 500 response code, instead of a 4xx if it wasn't a valid zip?

Can we be more specific - e.g. if it wasn't a zip, then the `yamlException` may be a more useful response for the user. But that won't be logged.

rest/rest-resources/src/test/java/org/apache/brooklyn/rest/resources/CatalogResourceTest.java
```diff
@@ -487,6 +495,79 @@ public void testAddMissingItem() {
         //equivalent to HTTP response 404 text/html
         addAddCatalogItemWithInvalidBundleUrl("classpath://missing-jar-file.txt");
     }
+    
+    @Test
+    public void testOsgiBundleWithBom() throws Exception {
+        TestResourceUnavailableException.throwIfResourceUnavailable(getClass(), OsgiStandaloneTest.BROOKLYN_TEST_OSGI_ENTITIES_PATH);
```
aledsage: I worry that this test might only pass because this OSGi bundle's symbolic name starts with `org.apache.brooklyn...`. It is therefore white-listed. If we fail to find the resource `classpath:/org/apache/brooklyn/test/osgi/entities/icon.gif`, we'll look inside all the installed bundles whose symbolic names match the white-listed regexes.

If you changed this to use `BROOKLYN_TEST_OSGI_ENTITIES_COM_EXAMPLE_PATH`, does it still pass?

If so, how? How does it know that the item in the catalog came from this given bundle, when it has no `brooklyn.libraries` section?! Do we need to auto-add the `brooklyn.libraries` section when we are explicitly adding the new item to the catalog?

rest/rest-resources/src/test/java/org/apache/brooklyn/rest/resources/CatalogResourceTest.java
```diff
@@ -487,6 +495,79 @@ public void testAddMissingItem() {
...
+        String bom = Joiner.on("\n").join(
+                "brooklyn.catalog:",
+                "  id: " + symbolicName,
+                "  version: " + version,
+                "  itemType: entity",
+                "  name: My Catalog App",
+                "  description: My description",
+                "  icon_url: classpath:/org/apache/brooklyn/test/osgi/entities/icon.gif",
+                "  item:",
+                "    type: org.apache.brooklyn.core.test.entity.TestEntity");
```
aledsage: Can we change this to something inside the bundle (e.g. `BROOKLYN_TEST_OSGI_ENTITIES_COM_EXAMPLE_ENTITY`, if you switch to use the bundle I suggested above)?

rest/rest-resources/src/test/java/org/apache/brooklyn/rest/resources/CatalogResourceTest.java
aledsage: Agreed; I'm fine with us adding those tests incrementally though, if we commit to doing it immediately after this PR is merged.

core/src/main/java/org/apache/brooklyn/core/catalog/internal/BasicBrooklynCatalog.java
tbouron: @aledsage My understanding is that the `symbolic-name` comes from the `MANIFEST.MF` file. If not, I agree we should support only one.

core/src/main/java/org/apache/brooklyn/core/catalog/internal/BasicBrooklynCatalog.java
tbouron: I think that's exactly what happens with [my pervious example](https://github.com/apache/brooklyn-server/pull/485#issuecomment-288044805)

core/src/main/java/org/apache/brooklyn/core/catalog/internal/BasicBrooklynCatalog.java
tbouron: I revised my comment, the `catalogMetata` is the catalog metadata that comes only from the yaml. So I agree with @aledsage, we should support only `symbolicName`

core/src/main/java/org/apache/brooklyn/core/catalog/internal/BasicBrooklynCatalog.java
tbouron: After re-reading, I can see the point of enforcing that the `id` matches the symbolicName and/or version, if defined. But as @aledsage pointed out, this contradicts what we said on the ML, i.e. having one true way of getting the symbolic name + version.

So I would say we should delete this block

I can see

core/src/main/java/org/apache/brooklyn/core/catalog/internal/BasicBrooklynCatalog.java
```diff
@@ -417,6 +419,53 @@ else for (String key: otherKeys) {
...
 
+    public static Map<?,?> getCatalogMetadata(String yaml) {
+        Map<?,?> itemDef = Yamls.getAs(Yamls.parseAll(yaml), Map.class);
+        return getFirstAsMap(itemDef, "brooklyn.catalog").orNull();        
+    }
+    
+    public static VersionedName getVersionedName(Map<?,?> catalogMetadata) {
+        String id = getFirstAs(catalogMetadata, String.class, "id").orNull();
+        String version = getFirstAs(catalogMetadata, String.class, "version").orNull();
+        String symbolicName = getFirstAs(catalogMetadata, String.class, "symbolicName").orNull();
```
geomacy: Related to comment https://github.com/apache/brooklyn-server/pull/485#issuecomment-288044805 from @tbouron, will this not clash with the use of `symbolicName` made in `BasicBrooklynCatalog` to define catalog item ids?  The "symbolicName" we want here is related to the bundle and is distinct from the ids of the catalog items that will be created from the bom files in the zip.  Maybe we should call this `bundleSymbolicName`.

rest/rest-resources/src/test/java/org/apache/brooklyn/rest/resources/CatalogResourceTest.java
geomacy: The `CatalogBomScanner` automatically adds library information to the `catalog.bom` about the bundle that contains it; but see my other comments about `CatalogBomScanner` and deleting the temp files.

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/resources/CatalogResource.java
```diff
@@ -129,6 +176,88 @@ public Response create(String yaml) {
...
+                // if the above feature is not enabled, let's do it manually (as a contract of this method)
+                new CatalogBomScanner().new CatalogPopulator(
+                        ((LocalManagementContext) mgmt()).getOsgiManager().get().getFramework().getBundleContext(),
+                        mgmt()).addingBundle(bundle, null);
+            }
+            
+            return Response.status(Status.CREATED).build();
+            
+        } finally {
+            if (f!=null) f.delete();
```
geomacy: Is this safe, does the loaded bundle not require access to these files any more?  What will happen upon restarts, or a Karaf `clean`, especially if the bundle is a proper jar with classes/resources in it?

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/resources/CatalogResource.java
```diff
@@ -129,6 +176,88 @@ public Response create(String yaml) {
...
+                mf.getMainAttributes().putValue(Attributes.Name.MANIFEST_VERSION.toString(), "1.0");
+            }
+            
+            f2 = bm.copyAddingManifest(f, mf);
+            
+            Bundle bundle = bm.installBundle(f2, true);
+            
+            if (!BrooklynFeatureEnablement.isEnabled(BrooklynFeatureEnablement.FEATURE_LOAD_BUNDLE_CATALOG_BOM)) {
+                // if the above feature is not enabled, let's do it manually (as a contract of this method)
+                new CatalogBomScanner().new CatalogPopulator(
```
geomacy: The `CatalogBomScanner` is designed to be used as a singleton; the `CatalogPopulator` is an OSGI `BundleTracker` and is meant to be used with an open/close lifecycle.  The constructor does the `open()` but here nothing does the `close()`.  Upon `open()` It will fire the `addingBundle()` method on _**all**_ active bundles in the bundle context, and subsequently on any bundles that get added to that context; not just on this `bundle`.  

We will probably need to look at splitting `CatalogPopulator`  apart a bit more and factor out a tool that can be used on a single bundle in an ad-hoc way like this. We'll need to think about deactivating the bundle and removing catalog items, which is taken care of by the `BundleTracker` side of things in the implementation of `CatalogPopulator.removedBundle()`.

rest/rest-resources/src/test/java/org/apache/brooklyn/rest/resources/CatalogResourceTest.java
aledsage: I tested this - it does work:
```
    @Test
    public void testOsgiBundleWithBomNotInBrooklynNamespace() throws Exception {
        TestResourceUnavailableException.throwIfResourceUnavailable(getClass(), OsgiTestResources.BROOKLYN_TEST_OSGI_ENTITIES_COM_EXAMPLE_PATH);
        final String symbolicName = OsgiTestResources.BROOKLYN_TEST_OSGI_ENTITIES_COM_EXAMPLE_SYMBOLIC_NAME_FULL;
        final String version = OsgiTestResources.BROOKLYN_TEST_OSGI_ENTITIES_COM_EXAMPLE_VERSION;
        final String bundleUrl = OsgiTestResources.BROOKLYN_TEST_OSGI_ENTITIES_COM_EXAMPLE_URL;
        final String entityType = OsgiTestResources.BROOKLYN_TEST_OSGI_ENTITIES_COM_EXAMPLE_ENTITY;
        final String iconPath = OsgiTestResources.BROOKLYN_TEST_OSGI_ENTITIES_COM_EXAMPLE_ICON_PATH;
        BundleMaker bm = new BundleMaker(manager);
        File f = Os.newTempFile("osgi", "jar");
        Files.copyFile(ResourceUtils.create(this).getResourceFromUrl(bundleUrl), f);
        
        String bom = Joiner.on("\n").join(
                "brooklyn.catalog:",
                "  id: " + symbolicName,
                "  version: " + version,
                "  itemType: entity",
                "  name: My Catalog App",
                "  description: My description",
                "  icon_url: classpath:" + iconPath,
                "  item:",
                "    type: " + entityType);
        
        f = bm.copyAdding(f, MutableMap.of(new ZipEntry("catalog.bom"), (InputStream) new ByteArrayInputStream(bom.getBytes())));

        Response response = client().path("/catalog")
                .header(HttpHeaders.CONTENT_TYPE, "application/x-zip")
                .post(Streams.readFully(new FileInputStream(f)));

        
        assertEquals(response.getStatus(), Response.Status.CREATED.getStatusCode());

        CatalogEntitySummary entityItem = client().path("/catalog/entities/"+symbolicName + "/" + version)
                .get(CatalogEntitySummary.class);

        Assert.assertNotNull(entityItem.getPlanYaml());
        Assert.assertTrue(entityItem.getPlanYaml().contains(entityType));

        assertEquals(entityItem.getId(), CatalogUtils.getVersionedId(symbolicName, version));
        assertEquals(entityItem.getSymbolicName(), symbolicName);
        assertEquals(entityItem.getVersion(), version);

        // and internally let's check we have libraries
        RegisteredType item = getManagementContext().getTypeRegistry().get(symbolicName, version);
        Assert.assertNotNull(item);
        Collection<OsgiBundleWithUrl> libs = item.getLibraries();
        assertEquals(libs.size(), 1);
        OsgiBundleWithUrl lib = Iterables.getOnlyElement(libs);
        Assert.assertNull(lib.getUrl());

        assertEquals(lib.getSymbolicName(), symbolicName);
        assertEquals(lib.getVersion(), version);

        // now let's check other things on the item
        URI expectedIconUrl = URI.create(getEndpointAddress() + "/catalog/icon/" + symbolicName + "/" + entityItem.getVersion()).normalize();
        assertEquals(entityItem.getName(), "My Catalog App");
        assertEquals(entityItem.getDescription(), "My description");
        assertEquals(entityItem.getIconUrl(), expectedIconUrl.getPath());
        assertEquals(item.getIconUrl(), "classpath:" + iconPath);

        // an InterfacesTag should be created for every catalog item
        assertEquals(entityItem.getTags().size(), 1);
        Object tag = entityItem.getTags().iterator().next();
        List<String> actualInterfaces = ((Map<String, List<String>>) tag).get("traits");
        List<String> expectedInterfaces = ImmutableList.of(Entity.class.getName(), BrooklynObject.class.getName(), Identifiable.class.getName(), Configurable.class.getName());
        assertTrue(actualInterfaces.containsAll(expectedInterfaces), "actual="+actualInterfaces);

        byte[] iconData = client().path("/catalog/icon/" + symbolicName + "/" + version).get(byte[].class);
        assertEquals(iconData.length, 43);
        
        // Check that the catalog item is useable (i.e. can deploy the entity)
        String appYaml = Joiner.on("\n").join(
                "services:",
                "- type: " + symbolicName + ":" + version,
                "  name: myEntityName");
        
        Response appResponse = client().path("/applications")
                .header(HttpHeaders.CONTENT_TYPE, "application/x-yaml")
                .post(appYaml);

        assertEquals(appResponse.getStatus(), Response.Status.CREATED.getStatusCode());
        
        Entity entity = Iterables.tryFind(getManagementContext().getEntityManager().getEntities(), EntityPredicates.displayNameEqualTo("myEntityName")).get();
        assertEquals(entity.getEntityType().getName(), entityType);
    }
```

The special code is in `CatalogBomScanner`, where it calls `addLibraryDetails` (e.g. in the `scanForCatalog(bundle)` method).

core/src/main/java/org/apache/brooklyn/core/catalog/internal/BasicBrooklynCatalog.java
tbouron: @geomacy Yes it does clash with is the root cause of the issue I can see with my example blueprint. Why not simply use a field `bundle` for this?
```yaml
brooklyn.catalog
  bundle: org.apache.brooklyn.test
  version: 0.1.0
  id: test
  items:
  ...
```

aledsage:
Given that @tbouron has picked up these changes in https://github.com/apache/brooklyn-server/pull/608 (building on top of this PR), let's merge this now and then continue any discussion in @tbouron's PR.

Merging now.

neykov:
This broke the java rest client build, see https://builds.apache.org/job/brooklyn-master-build/954.

PR #488, opened 2016-12-13
Get ready for compiling with JDK 1.8 / Java 1.8

googlielmo:
These changes are needed to compile Brooklyn with JDK 1.8, if/when we switch to Java 1.8.
At the same time they work with 1.7, which remains the selected `java.version` for compiling.

 - minor source changes (generics)

### DIFF
--- a/camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/methods/DslComponent.java
+++ b/camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/methods/DslComponent.java
@@ -486,8 +486,8 @@ public class DslComponent extends BrooklynDslDeferredSupplier<Entity> {
             if (targetEntityMaybe.isAbsent()) return Maybe.absent("Target entity not available");
             EntityInternal targetEntity = (EntityInternal) targetEntityMaybe.get();
 
-            ConfigKey<?> key = targetEntity.getEntityType().getConfigKey(keyName);
-            Maybe<? extends Object> result = targetEntity.config().getNonBlocking(key != null ? key : ConfigKeys.newConfigKey(Object.class, keyName));
+            ConfigKey key = targetEntity.getEntityType().getConfigKey(keyName);
+            Maybe result = targetEntity.config().getNonBlocking(key != null ? key : ConfigKeys.newConfigKey(Object.class, keyName));
             return Maybe.<Object>cast(result);
         }
 
@@ -501,7 +501,7 @@ public class DslComponent extends BrooklynDslDeferredSupplier<Entity> {
                         @Override
                         public Object call() throws Exception {
                             Entity targetEntity = component.get();
-                            ConfigKey<?> key = targetEntity.getEntityType().getConfigKey(keyName);
+                            ConfigKey key = targetEntity.getEntityType().getConfigKey(keyName);
                             return targetEntity.getConfig(key != null ? key : ConfigKeys.newConfigKey(Object.class, keyName));
                         }})
                     .build();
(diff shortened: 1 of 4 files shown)


camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/methods/DslComponent.java
```diff
@@ -486,8 +486,8 @@ public DslConfigSupplier(DslComponent component, String keyName) {
             if (targetEntityMaybe.isAbsent()) return Maybe.absent("Target entity not available");
             EntityInternal targetEntity = (EntityInternal) targetEntityMaybe.get();
 
-            ConfigKey<?> key = targetEntity.getEntityType().getConfigKey(keyName);
-            Maybe<? extends Object> result = targetEntity.config().getNonBlocking(key != null ? key : ConfigKeys.newConfigKey(Object.class, keyName));
+            ConfigKey key = targetEntity.getEntityType().getConfigKey(keyName);
```
aledsage: Seems a shame to have to strip off the generics. Same below.

neykov:
I'm able to compile the project successfully without any changes with JDK 1.8.0_45 & 1.8.0_112.

googlielmo:
@neykov yes absolutely, the project compiles fine with JDK 8 as long as `java.version` stays at 1.7.
This PR will ensure that whenever we do the switch to Java 1.8, the project will continue to compile without errors.
Will update the description to clarify, thanks!

camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/methods/DslComponent.java
googlielmo: @aledsage I agree in principle.
In this particular case, as Java 8 has a deeper / more precise type inference algorithm, it refuses to compile, because it cannot guarantee that the type of the Maybe (the capture of ?) is actually compatible with both the `key` and the `newConfigKey` call. I wonder how it could be permitted in java 7, actually :-)
I will reinstate the generics with Object as that's the only type that's accepted unequivocally.

googlielmo:
@aledsage I've addressed your comment and reinstated generics.
In the case of `tryCoerceUntyped` I removed them altogether though. That's because we check types manually on purpose there, and I don't see the point of juggling with wildcard captures and casts just to make the compiler happy! ;-)

camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/methods/DslComponent.java
```diff
@@ -486,8 +486,8 @@ public DslConfigSupplier(DslComponent component, String keyName) {
             if (targetEntityMaybe.isAbsent()) return Maybe.absent("Target entity not available");
             EntityInternal targetEntity = (EntityInternal) targetEntityMaybe.get();
 
-            ConfigKey<?> key = targetEntity.getEntityType().getConfigKey(keyName);
-            Maybe<? extends Object> result = targetEntity.config().getNonBlocking(key != null ? key : ConfigKeys.newConfigKey(Object.class, keyName));
+            ConfigKey<Object> key = (ConfigKey<Object>) targetEntity.getEntityType().getConfigKey(keyName);
+            Maybe<Object> result = targetEntity.config().getNonBlocking(key != null ? key : ConfigKeys.newConfigKey(Object.class, keyName));
```
neykov: Could instead help the compiler by assigning `key != null ? key : ConfigKeys.newConfigKey(Object.class, keyName)` to an intermediate `ConfigKey<?>` value. Same below.
Given we are casting below anyway could remove generics altogether and stop pretending we are "safe".

utils/common/src/main/java/org/apache/brooklyn/util/javalang/coerce/EnumTypeCoercions.java
```diff
@@ -70,11 +70,11 @@ public E apply(String input) {
     }
     
     @SuppressWarnings({ "unchecked", "rawtypes" })
-    public static <T> Maybe<T> tryCoerceUntyped(String input, Class<T> targetType) {
+    public static Maybe tryCoerceUntyped(String input, Class targetType) {
         if (input==null) return null;
         if (targetType==null) return Maybe.absent("Null enum type");
         if (!targetType.isEnum()) return Maybe.absent("Type '"+targetType+"' is not an enum");
-        return tryCoerce(input, (Class<Enum>)targetType);
+        return tryCoerce(input, targetType);
```
neykov: I'd keep the generics as is and add a cast to `(Maybe<T>) ` of the return value (insert before `tryCoerce`).
Credit to Eclipse's quick fix for this one :).

PR #489, opened 2016-12-14
BROOKLYN-409: add test

aledsage:
Includes new osgi test bundle with prefix "com.example.brooklyn"

Without that new osgi bundle, the test might only pass because of the "white-listing" of org.apache.brooklyn prefixes. As the comment in the relevant pom.xml says, it's ok to use this mavenId that we don't own because the artifact is only used for test purposes. It is never part of a release.

---
For BROOKLYN-409, this test shows that it is already fixed in master! I believe it was fixed in this commit:
```
commit a72d1f20ed78504a9b690df1467465b781ac917a
Author: Aled Sage <aled.sage@gmail.com>
Date:   Mon Nov 7 17:11:45 2016 +0000

    BROOKLYN-379: entity type includes OSGi prefix
```


camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/catalog/CatalogOsgiVersionMoreEntityRebindTest.java
```diff
@@ -0,0 +1,91 @@
...
+        String policyType = OsgiTestResources.BROOKLYN_TEST_OSGI_ENTITIES_COM_EXAMPLE_POLICY;
+        
+        String yaml1 = Joiner.on("\n").join(
+                "brooklyn.catalog:",
+                "  id: wrapped-entity",
+                "  version: 1.0",
+                "  item:",
+                "    services:",
+                "    - type: " + TestEntity.class.getName());
+        addCatalogItems(yaml1);
```
neykov: `addCatalogItems` supports `String... catalogYaml` - no need to join them above.

utils/common/dependencies/osgi/com-example-entities/pom.xml
```diff
@@ -0,0 +1,89 @@
...
+            <version>${brooklyn.version}</version>
+        </dependency>
+    </dependencies>
+    
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-jar-plugin</artifactId>
+                <version>2.3.2</version>
```
neykov: The version is already defined (albeit a different one) in parent pom, remove from there.
Same for below.

aledsage:
Thanks @neykov - changes incorporated, and squashed into single commit. Merging now.

PR #490, opened 2016-12-14
Fix BROOKLYN-410: rebind locations from bundles

aledsage:
Builds on https://github.com/apache/brooklyn-server/pull/489 - please review and merge that first; then I'll rebase (which will make the first commit disappear).

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/mgmt/rebind/RebindIteration.java
+++ b/core/src/main/java/org/apache/brooklyn/core/mgmt/rebind/RebindIteration.java
@@ -988,7 +988,7 @@ public abstract class RebindIteration {
             } catch (Exception e) {
                 Exceptions.propagateIfFatal(e);
             }
-            return new ClassLoaderUtils(reflections.getClassLoader()).loadClass(jType);
+            return new ClassLoaderUtils(reflections.getClassLoader(), managementContext).loadClass(jType);
         }
 
         @SuppressWarnings("unchecked")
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/brooklyn/core/mgmt/rebind/RebindIteration.java
```diff
@@ -988,7 +988,7 @@ protected void setCatalogItemId(BrooklynObject item, String catalogItemId) {
             } catch (Exception e) {
                 Exceptions.propagateIfFatal(e);
             }
-            return new ClassLoaderUtils(reflections.getClassLoader()).loadClass(jType);
+            return new ClassLoaderUtils(reflections.getClassLoader(), managementContext).loadClass(jType);
```
neykov: I fail to see how this fixes the test (looking at the `ClassLoaderUtils` code). If `jType` has a prefix it should be able to load it anyway. If not then it will fail.

Looks good otherwise.

aledsage: Without this, on rebind when it calls `ClassLoaderUtils.getFramework()` it gets back null (i.e. the OSGi framework is not found). This is because `FrameworkUtil.getBundle(ClassLoaderUtils.class)` does not find a bundle. Looking at `ClassLoaderUtils.class.getClassLoader()`, it is `sun.misc.Launcher$AppClassLoader@77abfbdc`.

Presumably this would not happen if we were running Karaf, but happens when we just have the embedded felix container (i.e. in classic mode, and when running testng unit tests).

Passing in the management context means `getFramework()` is able to instead call `mgmt.getOsgiManager()`, and thus finds the OSGi `Framework`.

aledsage:
I've rebased (having merged #489). Once jenkins has run the tests, I'll merge - unless there are any other comments from anyone.

core/src/main/java/org/apache/brooklyn/core/mgmt/rebind/RebindIteration.java
neykov: Got it.

PR #491, opened 2016-12-14
(WIP) execAsync parameter for SshMachineLocation#execCommand

bostko:
This enables ssh machine location to always use that flag for execCommand and execScript
```
location:
  jclouds:vcloud-director:
    brooklyn.ssh.config.execAsync: true
services:
- type: catalog-item
```

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/util/core/internal/ssh/sshj/SshjTool.java
+++ b/core/src/main/java/org/apache/brooklyn/util/core/internal/ssh/sshj/SshjTool.java
@@ -39,6 +39,7 @@ import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicReference;
 
 import org.apache.brooklyn.core.BrooklynFeatureEnablement;
+import org.apache.brooklyn.util.collections.MutableMap;
 import org.apache.brooklyn.util.core.internal.ssh.BackoffLimitedRetryHandler;
 import org.apache.brooklyn.util.core.internal.ssh.ShellTool;
 import org.apache.brooklyn.util.core.internal.ssh.SshAbstractTool;
@@ -106,6 +107,9 @@ public class SshjTool extends SshAbstractTool implements SshTool {
     protected final long sshTriesTimeout;
     protected final BackoffLimitedRetryHandler backoffLimitedRetryHandler;
 
+    protected boolean execAsync;
+    protected Duration execAsyncPollingTimeout;
+
     /** Terminal type name for {@code allocatePTY} option. */
     final static String TERM = "vt100"; // "dumb"
 
@@ -139,6 +143,8 @@ public class SshjTool extends SshAbstractTool implements SshTool {
         protected int sshTries = 4;  //allow 4 tries by default, much safer
         protected long sshTriesTimeout = 2*60*1000;  //allow 2 minutes by default (so if too slow trying sshTries times, abort anyway)
         protected long sshRetryDelay = 50L;
+        protected boolean execAsync;
+        protected Duration execAsyncPollingTimeout;
 
         @Override
         public B from(Map<String,?> props) {
@@ -148,6 +154,8 @@ public class SshjTool extends SshAbstractTool implements SshTool {
             sshRetryDelay = getOptionalVal(props, PROP_SSH_RETRY_DELAY);
             connectTimeout = getOptionalVal(props, PROP_CONNECT_TIMEOUT);
             sessionTimeout = getOptionalVal(props, PROP_SESSION_TIMEOUT);
+            execAsync = getOptionalVal(props, PROP_EXEC_ASYNC);
+            execAsyncPollingTimeout = getOptionalVal(props, PROP_EXEC_ASYNC_POLLING_TIMEOUT);
             return self();
         }
         public B connectTimeout(int val) {
@@ -182,6 +190,8 @@ public class SshjTool extends SshAbstractTool implements SshTool {
         sshTries = builder.sshTries;
         sshTriesTimeout = builder.sshTriesTimeout;
         backoffLimitedRetryHandler = new BackoffLimitedRetryHandler(sshTries, builder.sshRetryDelay);
+        execAsync = builder.execAsync;
+        execAsyncPollingTimeout = builder.execAsyncPollingTimeout;
 
         sshClientConnection = SshjClientConnection.builder()
                 .hostAndPort(HostAndPort.fromParts(host, port))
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/apache/brooklyn/feed/ssh/SshFeed.java
```diff
@@ -270,6 +270,7 @@ private SshPollValue exec(String command, Map<String,String> env) throws IOExcep
         ConfigBag flags = ConfigBag.newInstanceExtending(config().getBag())
             .configure(SshTool.PROP_NO_EXTRA_OUTPUT, true)
             .configure(SshTool.PROP_OUT_STREAM, stdout)
+            .configure(SshMachineLocation.CLOSE_CONNECTION, true)
```
aledsage: This is really expensive. It means that we close all connections in our ssh-pool-cache in `SshMachineLocation`.
(As an aside, I wonder if that works properly! I hope that invalidating the cache won't terminate other ssh commands that are executing in other threads!).

aledsage: I say "really expensive" because establishing an ssh connection is cpu-intensive. By caching the connection and re-using it, we cut that cpu-usage dramatically.

aledsage:
@bostko there are merge conflicts.

Could you add a test to `SshMachineLocationIntegrationTest`, and maybe also `SshMachineLocationTest`, for use of async? See existing tests such as `SshjToolIntegrationTest.testAsyncExecReturnsExitCode()` - we don't need to duplicate all their functionality, we just want to confirm that async really does happen when configured in `SshMachineLocation`.

core/src/main/java/org/apache/brooklyn/util/core/internal/ssh/sshj/SshjTool.java
```diff
@@ -575,9 +585,16 @@ public int execCommands(Map<String,?> props, List<String> commands, Map<String,?
         }
         
         // If async is set, then do it as execScript
-        Boolean execAsync = getOptionalVal(props, PROP_EXEC_ASYNC);
+        Boolean execAsync = getOptionalVal(props, PROP_EXEC_ASYNC, this.execAsync);
         if (Boolean.TRUE.equals(execAsync) && BrooklynFeatureEnablement.isEnabled(BrooklynFeatureEnablement.FEATURE_SSH_ASYNC_EXEC)) {
-            return execScriptAsyncAndPoll(props, commands, env);
+            if (!props.containsKey(PROP_EXEC_ASYNC_POLLING_TIMEOUT.getName())) {
+                return execScriptAsyncAndPoll(
+                        MutableMap.<String, Object>builder().putAll(props).put(PROP_EXEC_ASYNC_POLLING_TIMEOUT.getName(), this.execAsyncPollingTimeout).build(),
```
neykov: I'd pass `props` as is here and change [this line](https://github.com/apache/brooklyn-server/blob/1a279401aaf68a6970f96950fe45e0e72581680b/core/src/main/java/org/apache/brooklyn/util/core/internal/ssh/sshj/SshjTool.java#L354) to use `SshjTool.this.execAsyncPollingTimeout` as the default. And fall back to `FIVE_MINUTES` if null.

core/src/main/java/org/apache/brooklyn/util/core/internal/ssh/sshj/SshjTool.java
```diff
@@ -575,9 +585,16 @@ public int execCommands(Map<String,?> props, List<String> commands, Map<String,?
...
+                        commands,
+                        env);
+            } else {
+                return execScriptAsyncAndPoll(props, commands, env);
+            }
```
neykov: Do the same change for `execScript` as well. Otherwise setting the config on the location will work only for some ssh calls (commands and not scripts).

core/src/main/java/org/apache/brooklyn/feed/ssh/SshFeed.java
neykov: Agree. Wouldn't the ssh feed benefit from `execAsync` as well? And not need this?

core/src/test/java/org/apache/brooklyn/location/ssh/SshjExecAsyncInSshMachineLocationIntegrationTest.java
```diff
@@ -0,0 +1,93 @@
...
+    @BeforeMethod(alwaysRun=true)
+    public void setUp() throws Exception {
+        super.setUp();
+    }
+
+    // TODO fix failing test
+    @Test(groups = "Integration")
+    @Override
+    public void testSshExecScript() throws Exception {
+        super.testSshExecScript();
```
bostko: I'd like to ask for help on that failing test.

aledsage:
I started looking into why `testSshExecScript()` was failing. First quick comment: you need to enable the feature:
```
    @BeforeMethod(alwaysRun=true)
    public void setUp() throws Exception {
        origAsyncExecEnablement = BrooklynFeatureEnablement.setEnablement(BrooklynFeatureEnablement.FEATURE_SSH_ASYNC_EXEC, true);
        super.setUp();
    }

    @AfterMethod(alwaysRun=true)
    public void tearDown() throws Exception {
        try {
            super.tearDown();
        } finally {
            if (origAsyncExecEnablement != null) {
                BrooklynFeatureEnablement.setEnablement(BrooklynFeatureEnablement.FEATURE_SSH_ASYNC_EXEC, origAsyncExecEnablement);
                origAsyncExecEnablement = null;
            }
            RecordingSshjTool.clear();
        }
    }
```

But it still fails. It doesn't manage to get the stdout/stderr from the long polling. If you change the test to also pass in `SshMachineLocation.NO_STDOUT_LOGGING.getName(), true` then it works.

The problems seems to be the interaction between the `PipedOutputStream` (added in `ExecWithLoggingHelpers.execWithLogging()`) and the `CountingOutputStream` (added in `SshjTool.execScriptAsyncAndPoll`'s `longPoll()`).

I don't have time to investigate further just now, unfortunately. I suggest we park this for a bit longer.

I'm not sure why we're wrapping it in a `PipedOutputStream` (which allows us to also receive the bytes and log them, using the `StreamGobbler` to respond to callbacks about the bytes being received). I'm not sure why we don't just wrap the `OutputStream` in a much simpler way, in a similar way to how guava wraps `CountingOutputStream` - where on each byte that is written it can check if that's the end of a line and log it accordingly.

But we don't want to be making changing like that just now, and certainly not as part of this PR.

bostko:
I will not use those changes. Closing for now.

PR #492, opened 2016-12-15
Allow a user to add or override deserialize configuration in karaf

tbouron:
This will make the code search for another file `etc/org.apache.brooklyn.persistence.class.rename.properties` during the deserialization and load the properties accordingly.

The file `etc/org.apache.brooklyn.persistence.class.rename.properties` does not exist by default but can be created manually. The aim is for any blueprint author to setup rewrite rules for the persistence state, if some classnames from its own blueprints change. Of course, the default rules provided by Brooklyn are always loaded

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/mgmt/persist/DeserializingClassRenamesProvider.java
+++ b/core/src/main/java/org/apache/brooklyn/core/mgmt/persist/DeserializingClassRenamesProvider.java
@@ -18,67 +18,74 @@
  */
 package org.apache.brooklyn.core.mgmt.persist;
 
-import java.io.IOException;
-import java.io.InputStream;
-import java.util.Enumeration;
+import java.util.List;
 import java.util.Map;
-import java.util.Properties;
 
-import org.apache.brooklyn.util.core.ResourceUtils;
-import org.apache.brooklyn.util.exceptions.Exceptions;
+import org.apache.brooklyn.util.collections.MutableMap;
 import org.apache.brooklyn.util.javalang.Reflections;
-import org.apache.brooklyn.util.stream.Streams;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import com.google.common.annotations.Beta;
-import com.google.common.collect.ImmutableMap;
-import com.google.common.collect.Maps;
+import com.google.common.collect.Lists;
 
+/*
+ * This provider keeps a cache of the class-renames, which is lazily populated (see {@link #cache}.
+ * Calling {@link #reset()} will set this cache to null, causing it to be reloaded next time
+ * it is requested.
+ *
+ * Loading the cache involves iterating over the {@link #loaders}, returning the union of
+ * the results from {@link Loader#load()}.
+ *
+ * Initially, the only loader is the basic {@link ClasspathConfigLoader}.
+ *
+ * However, when running in karaf the {@link OsgiConfigLoader} will be instantiated and added.
+ * See karaf/init/src/main/resources/OSGI-INF/blueprint/blueprint.xml
+ */
 @Beta
 public class DeserializingClassRenamesProvider {
+    private static final Logger LOG = LoggerFactory.getLogger(DeserializingClassRenamesProvider.class);
 
-    public static final String DESERIALIZING_CLASS_RENAMES_PROPERTIES_PATH = "classpath://org/apache/brooklyn/core/mgmt/persist/deserializingClassRenames.properties";
+    private static final List<ConfigLoader> loaders = Lists.newCopyOnWriteArrayList();
+    static {
+        loaders.add(new ClasspathConfigLoader());
+    }
     
     private static volatile Map<String, String> cache;
-    
-    @Beta
-    public static Map<String, String> loadDeserializingClassRenames() {
-        // Double-checked locking - got to use volatile or some such!
-        if (cache == null) {
-            synchronized (DeserializingClassRenamesProvider.class) {
-                if (cache == null) {
-                    cache = loadDeserializingClassRenamesCache();
-                }
-            }
-        }
-        return cache;
-    }
-    
-    private static Map<String, String> loadDeserializingClassRenamesCache() {
-        InputStream resource = new ResourceUtils(DeserializingClassRenamesProvider.class).getResourceFromUrl(DESERIALIZING_CLASS_RENAMES_PROPERTIES_PATH);
-        if (resource != null) {
-            try {
-                Properties props = new Properties();
-                props.load(resource);
-                
-                Map<String, String> result = Maps.newLinkedHashMap();
-                for (Enumeration<?> iter = props.propertyNames(); iter.hasMoreElements();) {
-                    String key = (String) iter.nextElement();
-                    String value = props.getProperty(key);
-                    result.put(key, value);
-                }
-                return result;
-            } catch (IOException e) {
-                throw Exceptions.propagate(e);
-            } finally {
-                Streams.closeQuietly(resource);
-            }
-        } else {
-            return ImmutableMap.<String, String>of();
-        }
+
+    public static List<ConfigLoader> getLoaders() {
+        return loaders;
     }
 
     @Beta
+    public static Map<String, String> loadDeserializingClassRenames() {
+        synchronized (DeserializingClassRenamesProvider.class) {
+            if (cache == null) {
+                MutableMap.Builder<String, String> builder = MutableMap.<String, String>builder();
+                for (ConfigLoader loader : loaders) {
+                    builder.putAll(loader.load());
+                }
+                cache = builder.build();
+                LOG.info("Class-renames cache loaded, size {}", cache.size());
+            }
+            return cache;
+        }
+    }
+
+    /**
+     * Handles inner classes, where the outer class has been renamed. For example:
+     * 
+     * {@code findMappedName("com.example.MyFoo$MySub")} will return {@code com.example.renamed.MyFoo$MySub}, if
+     * the renamed contains {@code com.example.MyFoo: com.example.renamed.MyFoo}.
+     */
+    @Beta
     public static String findMappedName(String name) {
-        return Reflections.findMappedNameAndLog(loadDeserializingClassRenames(), name);
+        return Reflections.findMappedNameAndLog(DeserializingClassRenamesProvider.loadDeserializingClassRenames(), name);
+    }
+
+    public static void reset() {
+        synchronized (DeserializingClassRenamesProvider.class) {
+            cache = null;
+        }
     }
 }
--- a/karaf/init/src/main/resources/OSGI-INF/blueprint/blueprint.xml
+++ b/karaf/init/src/main/resources/OSGI-INF/blueprint/blueprint.xml
@@ -47,10 +47,23 @@ limitations under the License.
         </cm:default-properties>
     </cm:property-placeholder>
 
+    <bean id="deserializingClassRenamesProvider"
+          class="org.apache.brooklyn.launcher.osgi.OsgiConfigLoader"
+          init-method="init"
+          destroy-method="destroy">
+
+        <property name="configAdmin" ref="configAdmin" />
+
+        <cm:managed-properties persistent-id="org.apache.brooklyn.classrename"
+                               update-method="updateProperties"
+                               update-strategy="component-managed" />
+    </bean>
+
     <bean id="launcher"
           class="org.apache.brooklyn.launcher.osgi.OsgiLauncher"
           init-method="init"
-          destroy-method="destroy">
+          destroy-method="destroy"
+          depends-on="deserializingClassRenamesProvider">
 
         <property name="brooklynVersion" ref="brooklynVersion" />
 
--- a/core/src/main/java/org/apache/brooklyn/core/mgmt/persist/ClasspathConfigLoader.java
+++ b/core/src/main/java/org/apache/brooklyn/core/mgmt/persist/ClasspathConfigLoader.java
@@ -1,0 +1,72 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.brooklyn.core.mgmt.persist;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Enumeration;
+import java.util.Map;
+import java.util.Properties;
+
+import org.apache.brooklyn.util.core.ResourceUtils;
+import org.apache.brooklyn.util.exceptions.Exceptions;
+import org.apache.brooklyn.util.stream.Streams;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Maps;
+
+/**
+ * Loads the class-renames from the configuration file on the classpath.
+ *
+ * @see {@link #DESERIALIZING_CLASS_RENAMES_PROPERTIES_PATH}
+ */
+public class ClasspathConfigLoader implements ConfigLoader {
+    private static final Logger LOG = LoggerFactory.getLogger(ClasspathConfigLoader.class);
+    private static final String DESERIALIZING_CLASS_RENAMES_PROPERTIES_PATH = "classpath://org/apache/brooklyn/core/mgmt/persist/deserializingClassRenames.properties";
+
+
+    @Override
+    public Map<String, String> load() {
+        try {
+            InputStream resource = new ResourceUtils(DeserializingClassRenamesProvider.class).getResourceFromUrl(DESERIALIZING_CLASS_RENAMES_PROPERTIES_PATH);
+
+            try {
+                Properties props = new Properties();
+                props.load(resource);
+
+                Map<String, String> result = Maps.newLinkedHashMap();
+                for (Enumeration<?> iter = props.propertyNames(); iter.hasMoreElements(); ) {
+                    String key = (String) iter.nextElement();
+                    String value = props.getProperty(key);
+                    result.put(key, value);
+                }
+                return result;
+            } catch (IOException e) {
+                throw Exceptions.propagate(e);
+            } finally {
+                Streams.closeQuietly(resource);
+            }
+        } catch (Exception e) {
+            LOG.warn("Failed to load class-renames from " + DESERIALIZING_CLASS_RENAMES_PROPERTIES_PATH + " (continuing)", e);
+            return ImmutableMap.<String, String>of();
+        }
+    }
+}
--- a/core/src/main/java/org/apache/brooklyn/core/mgmt/persist/ConfigLoader.java
+++ b/core/src/main/java/org/apache/brooklyn/core/mgmt/persist/ConfigLoader.java
@@ -1,0 +1,25 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.brooklyn.core.mgmt.persist;
+
+import java.util.Map;
+
+public interface ConfigLoader {
+    public Map<String, String> load();
+}
--- a/karaf/init/src/main/java/org/apache/brooklyn/launcher/osgi/OsgiConfigLoader.java
+++ b/karaf/init/src/main/java/org/apache/brooklyn/launcher/osgi/OsgiConfigLoader.java
@@ -1,0 +1,133 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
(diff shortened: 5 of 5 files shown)


core/src/main/java/org/apache/brooklyn/core/mgmt/persist/DeserializingClassRenamesProvider.java
```diff
@@ -37,7 +37,8 @@
 public class DeserializingClassRenamesProvider {
 
     public static final String DESERIALIZING_CLASS_RENAMES_PROPERTIES_PATH = "classpath://org/apache/brooklyn/core/mgmt/persist/deserializingClassRenames.properties";
-    
+    public static final String KARAF_DESERIALIZING_CLASS_RENAMES_PROPERTIES_PATH = "file:etc/deserializingClassRenames.properties";
+
```
m4rkmckenna: Filename should be namespaced `org.apache.brooklyn.****.properties` as karaf will add this to config admin

core/src/main/java/org/apache/brooklyn/core/mgmt/persist/DeserializingClassRenamesProvider.java
```diff
@@ -54,12 +55,20 @@
...
     
     private static Map<String, String> loadDeserializingClassRenamesCache() {
-        InputStream resource = new ResourceUtils(DeserializingClassRenamesProvider.class).getResourceFromUrl(DESERIALIZING_CLASS_RENAMES_PROPERTIES_PATH);
-        if (resource != null) {
+        return ImmutableMap.<String, String>builder()
+                .putAll(loadDeserializingClassRenamesCacheFromUrl(DESERIALIZING_CLASS_RENAMES_PROPERTIES_PATH))
+                .putAll(loadDeserializingClassRenamesCacheFromUrl(KARAF_DESERIALIZING_CLASS_RENAMES_PROPERTIES_PATH))
+                .build();
+    }
+
```
m4rkmckenna: For karaf distribution this is available from config admin

core/src/main/java/org/apache/brooklyn/core/mgmt/persist/DeserializingClassRenamesProvider.java
tbouron: I wasn't sure about the name, hence my note in the description.

@m4rkmckenna What about `org.apache.brooklyn.deserializeClassRename.properties` ?

core/src/main/java/org/apache/brooklyn/core/mgmt/persist/DeserializingClassRenamesProvider.java
tbouron: We cannot use config admin here as:

1. we are working with a static class
2. It won't work if Brooklyn is launched in classic mode

We can think of migrating this to a service once we will use Karaf only. Until then, we need to stick with this approach.

tbouron:
Just ran a build on my machine, all tests pass. I guess the Jenkins' issues are not related to this PR.

I also updated the name of the properties file for karaf from `etc/deserializingClassRenames.properties` to `etc/org.apache.brooklyn.persistence.class.rename.properties` based on @m4rkmckenna suggestions

neykov:
Some thoughts:
  * Classic doesn't have `etc` folder
  * The `file:etc/...` location is a hack which I don't like (but have used). It relies on the process working dir being KARAF_HOME.
  * I believe we can get hold of config admin even in static context - as long as the class is loaded inside the OSGi runtime
  * this files is actually **not** a properties file! It's a colon delimited key value list. Nothing stopping us from using properties notation (`=` delimiter)
  * Would make more sense if the author of java entities is able to package the renames in the new version of the bundle. Can we achieve this? May be listen for bundle loads and collect the renames once a bundle becomes loaded?
  * It certainly makes sense to have a run time way to add to the list of renames for some last minute on-site fixes.

m4rkmckenna:
@neykov i think you should be able to get `KARAF_HOME` from the env

tbouron:
> this files is actually not a properties file! It's a colon delimited key value list. Nothing stopping us from using properties notation (= delimiter)

True, I keep the same formatting and parsing for consistency but you're right, makes more sense to have a proper properties file

> Would make more sense if the author of java entities is able to package the renames in the new version of the bundle. Can we achieve this? May be listen for bundle loads and collect the renames once a bundle becomes loaded?

I'm pretty sure we can do this, it's probably a cleaner way actually

aledsage:
Given the class is static, I wondered about Brooklyn's startup injecting the karaf config via some static method call. But I prefer @neykov's suggestion that we get the karaf config admin via appropriate static method calls (if that's feasible).

@neykov @tbouron colons are acceptable separators in properties files: https://docs.oracle.com/cd/E23095_01/Platform.93/ATGProgGuide/html/s0204propertiesfileformat01.html

---
I wonder what we should do about https://github.com/apache/brooklyn-server/pull/144. It feels right that we support multiple files, rather than a user having to duplicate all of the brooklyn entries in their user-specific file (I think it's right that you've added the karaf file second, so it overrides the defaults).

I agree with Svet that we want a bundle author to be able to supply their own file. Let's tackle that by re-visiting #144, rather than in this PR.

core/src/main/java/org/apache/brooklyn/core/mgmt/persist/DeserializingClassRenamesProvider.java
```diff
@@ -72,7 +81,7 @@
             } finally {
                 Streams.closeQuietly(resource);
             }
-        } else {
+        } catch (Exception e) {
```
tbouron: This one is safe. It is for the `new ResourceUtils(DeserializingClassRenamesProvider.class).getResourceFromUrl(url);`

As per as the javadoc:
> Takes a string which is treated as a URL (with some extended "schemes" also expected), or as a path to something either on the classpath (absolute only) or the local filesystem (relative or absolute, depending on leading slash)
URLs can be of the form classpath://com/acme/Foo.properties as well as file:///home/... and http://acme.com/....
**Throws exception if not found**, using the context parameter passed into the constructor.
TODO may want OSGi, or typed object; should consider pax url
Returns:
a stream, or throws exception **(never returns null)**

tbouron:
Oh, good to know about #144. I was looking at how to get a hold of the global bundle context statically but no luck so far.

I agree that we should also allow each bundle to specify their own rewrite rules but that can be covered by #144

core/src/main/java/org/apache/brooklyn/core/mgmt/persist/DeserializingClassRenamesProvider.java
```diff
@@ -54,12 +55,20 @@
     }
     
     private static Map<String, String> loadDeserializingClassRenamesCache() {
-        InputStream resource = new ResourceUtils(DeserializingClassRenamesProvider.class).getResourceFromUrl(DESERIALIZING_CLASS_RENAMES_PROPERTIES_PATH);
-        if (resource != null) {
+        return ImmutableMap.<String, String>builder()
+                .putAll(loadDeserializingClassRenamesCacheFromUrl(DESERIALIZING_CLASS_RENAMES_PROPERTIES_PATH))
+                .putAll(loadDeserializingClassRenamesCacheFromUrl(KARAF_DESERIALIZING_CLASS_RENAMES_PROPERTIES_PATH))
```
bostko: `ImmutableMap.Builder#putAll` for duplicate entries will fail.
It is good to be able to override `DESERIALIZING_CLASS_RENAMES_PROPERTIES_PATH` with the ones provided in the file `KARAF_DESERIALIZING_CLASS_RENAMES_PROPERTIES_PATH`.

neykov:
As discussed offline with @tbouron it's possible to get OSGi services from static context. See See `org.apache.brooklyn.util.core.osgi.Compat` for an example. Replacing `ManagementContext` with `ConfigurationAdmin` should be all that's needed.
Didn't know that colon is a valid properties separator @aledsage. That's neat.
+1 for multiple files.

geomacy:
hi @svet, @tbouron, the `ConfigurationAdmin` is an OSGI compendium service, is that a dependency that we'd be happy to add to brooklyn-core?  I think I would prefer not to, but if so , then we could get config admin from a static context as @neykov says. On the other hand we could avoid the compendium dependency, and instead change the `DeserializingClassRenamesProvider` class to become a non-static provider - it could be a singleton object that gets set up in the launchers (with an appropriate implementation, one way for OSGI launcher, one way for classic), probably in the `initPersistence` stage.  After that, then 
`DeserializingClassRenamesProvider.loadDeserializingClassRenames()`
becomes 
`DeserializingClassRenamesProvider.INSTANCE.loadDeserializingClassRenames()`

Just a thought!

tbouron:
Thanks to all of you @aledsage @neykov and @geomacy.

This now use `ConfigurationAdmin` to load custom class deserialization renames (and supports updates too).
I created https://github.com/apache/brooklyn-dist/pull/72 that adds and `org.apache.brooklyn.classrename.cfg` in `etc` + integration test. Although, the test suite does not work so I tested this manually by:
- adding `a=b` in `etc/org.apache.brooklyn.classrename.cfg`
- Launching the karaf distribution
- checking that the config appears by doing `config:list`
- checking that the configuration is picked up on startup by scanning the logs for `Reading OSGi configuration from org.apache.brooklyn.classrename.cfg`
- updating the config file and checking the load for `DeserializingClassRenamesProvider.OsgiConfigLoader.updateProperties: clearing cache, so class-renames will be reloaded`

core/src/main/java/org/apache/brooklyn/core/mgmt/persist/DeserializingClassRenamesProvider.java
```diff
@@ -20,65 +20,217 @@
...
-        // Double-checked locking - got to use volatile or some such!
-        if (cache == null) {
-            synchronized (DeserializingClassRenamesProvider.class) {
-                if (cache == null) {
-                    cache = loadDeserializingClassRenamesCache();
+        synchronized (DeserializingClassRenamesProvider.class) {
+            if (cache == null) {
+                ImmutableMap.Builder<String, String> builder = ImmutableMap.<String, String>builder();
+                for (Loader loader : loaders) {
+                    builder.putAll(loader.load());
```
bostko: Hi @tbouron  I think you missed my comment, it is still valid.

> _bostko 7 days ago_
> `ImmutableMap.Builder#putAll` for duplicate entries will fail.
> It is good to be able to override `DESERIALIZING_CLASS_RENAMES_PROPERTIES_PATH` with the ones provided in the file `KARAF_DESERIALIZING_CLASS_RENAMES_PROPERTIES_PATH`.

bostko: I am not sure whether the case is the same, but still I do not see a guarantee for uniqueness of loaders loaded items.

tbouron: Good catch @bostko, I'll change that

tbouron: @bostko done

karaf/init/src/main/java/org/apache/brooklyn/launcher/osgi/OsgiLauncher.java
```diff
@@ -199,4 +199,7 @@ public void setLocalBrooklynProperties(String localBrooklynProperties) {
         this.localBrooklynProperties = localBrooklynProperties;
     }
 
+    public void setDeserializingClassRenamesProvider(DeserializingClassRenamesProvider.OsgiConfigLoader val) {
```
geomacy: use instead the "depends-on" attribute in the blueprint:

```
    <bean id="launcher"
          class="org.apache.brooklyn.launcher.osgi.OsgiLauncher"
          init-method="init"
          destroy-method="destroy"
          depends-on="deserializingClassRenamesProvider">

      <!-- set properties but avoid the deserializingClassRenamesProvider one -->
      
    </bean>
```

tbouron: Cool! That's feels cleaner

tbouron: @geomacy Do I still need the `<property name="deserializingClassRenamesProvider" ref="deserializingClassRenamesProvider" />` ?

geomacy: No, I'd remove the property, since it's a no-op in the Java code it's not needed. (ie. remove it from both the blueprint.xml and the Java code)

tbouron: Cool, thanks @geomacy, it's done

core/src/main/java/org/apache/brooklyn/core/mgmt/persist/DeserializingClassRenamesProvider.java
```diff
@@ -20,65 +20,217 @@
...
+        }
+
+        @Override
+        public Map<String, String> load() {
+            if (configAdmin == null) {
+                LOG.warn("No OSGi configuration-admin available - cannot load {}.cfg", KARAF_DESERIALIZING_CLASS_RENAMES_PROPERTIES);
+                return ImmutableMap.of();
+            }
+            
+            String filter = '(' + Constants.SERVICE_PID + '=' + KARAF_DESERIALIZING_CLASS_RENAMES_PROPERTIES + ')';
```
geomacy: Add a TODO here, you could avoid using ConfigurationAdmin in the code and doing all that filtering/ listConfigurations yourself. It should be possible to get cm to do that for you with a managed bean;  but  it’s not a big deal, it just means you have written a little more code  than you needed to.

The way the configuration admin service is typically intended to be used is that you let it inject properties into the beans that you declare - i.e. your code doesn't need to get hold of the ConfigurationService class at all.  See https://www.eclipse.org/gemini/blueprint/documentation/reference/2.0.0.RELEASE/html/compendium.html section 11.1.2.1 for background.

I was thinking along the lines of changing the DeserializingClassRenamesProvider to be a regular bean that could be injected as a service, but the pattern of callers would make that a quite extensive change, affecting a lot of classes.  But it shouldn't be hard to make it a singleton bean, and have it be initialised by the launchers. Then in Karaf (OsgiLauncher) you would have an "DeserializingClassRenamesProviderInitializer" managed bean in the blueprint.xml,

```xml
<bean id="deserializingRenamesInit" class="org.apache.brooklyn.launcher.osgi.DeserializingClassRenamesProviderInitializer">
   <osgix:managed-properties persistent-id="org.apache.brooklyn.core.deserializing.renames" update-method="updateCallback"/>
</bean>
```

and then the DeserializingClassRenamesProviderInitializer would have a method to initialize the provider, along the following lines and ignoring all niceties of synchronization, exceptions etc.

```java
public class DeserializingClassRenamesProviderInitializer {

    public void updateCallback(Map properties) {

        DeserializingClassRenamesProvider.initializeSingleton(properties);
    }
}
```

and changing the provider class to

```java
public class DeserializingClassRenamesProvider {

    private volatile Map<String, String> cache;

    public static DeserializingClassRenamesProvider INSTANCE;

    private DeserializingClassRenamesProvider(Map properties) {
        // reset cache from properties
    }

    public static void initializeSingleton(Map properties) {
        INSTANCE = new DeserializingClassRenamesProvider(properties);
    }

    // not static!
    public Map<String, String> loadDeserializingClassRenames() {
       // return the cache
    }
}
```

geomacy:
👍 LGTM ; added a comment about possible simplification but I don't think it's needed absolutely. If this works as-is then it's already fit for purpose.

core/src/main/java/org/apache/brooklyn/core/mgmt/persist/DeserializingClassRenamesProvider.java
geomacy: ps the above is just a sketch, and would need adapted to what you have already done

core/src/main/java/org/apache/brooklyn/util/core/osgi/Compat.java
```diff
@@ -39,15 +40,19 @@ public static Compat getInstance() {
     }
 
     private final ServiceTracker managementContextTracker;
+    private final ServiceTracker configurationAdminTracker;
```
aledsage: We don't need this any more - I think you can get rid of it?

karaf/init/src/main/resources/OSGI-INF/blueprint/blueprint.xml
```diff
@@ -47,10 +47,23 @@ limitations under the License.
...
+                               update-method="updateProperties"
+                               update-strategy="component-managed" />
+    </bean>
+
     <bean id="launcher"
           class="org.apache.brooklyn.launcher.osgi.OsgiLauncher"
           init-method="init"
-          destroy-method="destroy">
+          destroy-method="destroy"
+          depends-on="deserializingClassRenamesProvider">
```
aledsage: Nice! I didn't know about `depends-on`.

aledsage:
LGTM - would be nice to delete the unused `Compat.getConfigurationAdmin()`, which is being added in this PR.

tbouron:
@aledsage Unused code removed

aledsage:
Thanks @tbouron Failing test `DslYamlTest.testDslAttributeWhenReady` looks unrelated (we should look into that in a separate PR). You can kick off jenkins again by running `git commit --amend; git push --force`.

@m4rkmckenna's right that `org.osgi.service.cm` version should be parameterised. You could do that here, or we can do it in a separate PR - let me know what you'd prefer.

If you don't get a chance to kick off jenkins again, then I'll try to check out your branch and run the tests - and then merge (but don't have time to do that just now).

For any other reviewers, note that https://github.com/apache/brooklyn-dist/pull/72 should be reviewed + tested + merged after this PR is merged.

tbouron:
@aledsage I parameterised the `org.osgi.service.cm`'s version

core/src/main/java/org/apache/brooklyn/core/mgmt/persist/DeserializingClassRenamesProvider.java
neykov: @geomacy I believe `updateCallback` is not called on the initial creation of the bean, only after configuration changes subsequently. This will cause the properties to be loaded only after someone goes and edits the file post-boot.

core/src/main/java/org/apache/brooklyn/core/mgmt/persist/DeserializingClassRenamesProvider.java
```diff
@@ -20,65 +20,217 @@
...
+        public void updateProperties(Map properties) {
+            LOG.debug("DeserializingClassRenamesProvider.OsgiConfigLoader.updateProperties: clearing cache, so class-renames will be reloaded");
+            DeserializingClassRenamesProvider.reset();
+        }
+
+        @Override
+        public Map<String, String> load() {
+            if (configAdmin == null) {
+                LOG.warn("No OSGi configuration-admin available - cannot load {}.cfg", KARAF_DESERIALIZING_CLASS_RENAMES_PROPERTIES);
+                return ImmutableMap.of();
```
neykov: What would cause `configAdmin` to be missing at this point? This class is instantiated only in OSGi context, so this would be an error condition.

karaf/init/pom.xml
```diff
@@ -75,7 +75,18 @@
...
+            <!--
+                See http://stackoverflow.com/questions/18844987/creating-a-blueprint-bean-from-an-inner-class:
+                we unfortunately need to exclude DeserializingClassRenamesProvider from the imports. 
+                Its inner class is referenced in a bean of blueprint.xml. This confuses the mvn 
+                plugin, which would otherwise add it to the Import-Package (thus causing the bundle
+                to fail to install).
+            -->
+            <Import-Package>
+                !org.apache.brooklyn.core.internal.BrooklynProperties.Factory, 
+                !org.apache.brooklyn.core.mgmt.persist.DeserializingClassRenamesProvider,
```
neykov: Could make the nested class standalone instead?

launcher/src/main/java/org/apache/brooklyn/launcher/BrooklynLauncher.java
```diff
@@ -180,13 +180,13 @@ public BrooklynLauncher publicAddress(InetAddress publicAddress) {
...
      * Specifies additional flags to be passed to {@link BrooklynWebServer}.
-     */ 
+     */
     public BrooklynLauncher webServerFlags(Map<String,?> webServerFlags) {
         this.webconsoleFlags  = webServerFlags;
         return this;
     }
 
-    /** 
+    /**
```
neykov: No changes in this file.

core/pom.xml
```diff
@@ -237,6 +237,11 @@
             <classifier>tests</classifier>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.osgi</groupId>
+            <artifactId>org.osgi.service.cm</artifactId>
+            <version>${org.osgi.service.cm}</version>
+        </dependency>
```
neykov: We are using 
```
        <dependency>
            <groupId>org.osgi</groupId>
            <artifactId>org.osgi.compendium</artifactId>
            <version>${org.osgi.compendium.version}</version>
            <scope>provided</scope>
        </dependency>
```
elsewhere for the same dependency.

core/src/main/java/org/apache/brooklyn/core/mgmt/persist/DeserializingClassRenamesProvider.java
neykov: On the other hand agree that it would be nice if we move the OSGi specific functionality (`OsgiConfigLoader`) to `karaf-init` so we don't introduce osgi dependencies in core.

geomacy: ah, I thought it was used to do the initial set as well...

core/src/main/java/org/apache/brooklyn/core/mgmt/persist/DeserializingClassRenamesProvider.java
tbouron: @neykov I'm not entirely sure but I remember having the case. It feels safer to have a check just in case, instead of having a potential NPE.

That being said, it was probably before the refactoring

launcher/src/main/java/org/apache/brooklyn/launcher/BrooklynLauncher.java
tbouron: Argh, that's my IDE performing syntax cleaning, I'll revert

karaf/init/pom.xml
m4rkmckenna: Agree OSGi / internal classes are usually a bad idea

tbouron:
@neykov @m4rkmckenna @aledsage Comment addressed

core/src/main/java/org/apache/brooklyn/core/mgmt/persist/OsgiConfigLoader.java
```diff
@@ -0,0 +1,131 @@
...
+ * future releases.
+ *
+ * See http://stackoverflow.com/questions/18844987/creating-a-blueprint-bean-from-an-inner-class:
+ * we unfortunately need to include {@code !org.apache.brooklyn.core.mgmt.persist.DeserializingClassRenamesProvider}
+ * in the Import-Package, as the mvn plugin gets confused due to the use of this inner class
+ * within the blueprint.xml.
+ *
+ * @see {@link #KARAF_DESERIALIZING_CLASS_RENAMES_PROPERTIES}
+ */
+public class OsgiConfigLoader implements ConfigLoader {
```
neykov: Can this be moved to `karaf-init`, along with the `org.osgi.compendium` dependency?

tbouron: Can we see that in a following PR @neykov ? I already changed a lot of things to address your comments but they keep coming. I'm afraid it will never get merged if we follow this trend.

neykov: Agree it wasn't an easy change @tbouron. I'd prefer if it gets changed in this PR, then good to merge after squashing the WIP commits.

tbouron: @neykov Moved the `OsgiConfigLoader` class in `karaf-init` and squashed commits

neykov:
Thanks @tbouron, merging.

PR #493, opened 2016-12-15
Make the Repeater class use a threads to execute the checks

grkvlt:
The Future returned by the Executor can be cancelled if it does not return a value after the required timeout.

### DIFF
--- a/utils/common/src/main/java/org/apache/brooklyn/util/repeat/Repeater.java
+++ b/utils/common/src/main/java/org/apache/brooklyn/util/repeat/Repeater.java
@@ -43,6 +43,7 @@ import com.google.common.base.Predicate;
 import com.google.common.base.Predicates;
 import com.google.common.base.Stopwatch;
 import com.google.common.util.concurrent.Callables;
+import com.google.common.util.concurrent.MoreExecutors;
 
 /**
  * Simple mechanism to repeat an operation periodically until a condition is satisfied.
@@ -88,6 +89,7 @@ public class Repeater implements Callable<Boolean> {
     private boolean rethrowException = false;
     private Predicate<? super Throwable> rethrowImmediatelyCondition = Exceptions.isFatalPredicate();
     private boolean warnOnUnRethrownException = true;
+    private ExecutorService executor = MoreExecutors.sameThreadExecutor();
 
     public Repeater() {
         this(null);
@@ -145,6 +147,26 @@ public class Repeater implements Callable<Boolean> {
     }
 
     /**
+     * Use a new thread for every iteration of the loop.
+     *
+     * @return {@literal this} to aid coding in a fluent style.
+     */
+    public Repeater threaded() {
+        this.executor = Executors.newSingleThreadExecutor();
+        return this;
+    }
+
+    /**
+     * @see #threaded()
+     * @param executor an {@link ExecutorService} to use when creating threads.
+     * @return {@literal this} to aid coding in a fluent style.
+     */
+    public Repeater threaded(ExecutorService executor) {
+        this.executor = executor;
+        return this;
+    }
+
+    /**
      * Set how long to wait between loop iterations.
      *
      * @param period how long to wait between loop iterations.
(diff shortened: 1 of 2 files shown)


geomacy:
hi @grkvlt looks good at an initial glance, but are there tests that could be added to demonstrate the improvement?

neykov:
I like the idea, very useful sometimes.

Brooklyn already spins up a huge number of threads and this change will add even more even if not needed (tasks needs cancelling). Do you think each task will benefit from the change? If not, what if we provide a wrapper to `Repeater` to be used selectively? 
Could be a very thin wrapper which passes a wrapped body which is executed in a separate thread. The body executed by the repeater just does `future.get(timeout)`.

aledsage:
@grkvlt I agree with @neykov's concern about thread usage. Maybe we could make it configurable, so the default is the existing behaviour (which would use a guava `MoreExecutors.directExecutor()`), and only a new single-threaded executor if configured to do so.

grkvlt:
@neykov @aledsage @geomacy thanks for the feedback, I'll make this configurable and add some tests

grkvlt:
@neykov @aledsage I used `MoreExecutors.sameThreadExecutor()` as the default, but added `threaded()` and `threaded(ExecutorService)` methods for the builder to change to either `Executors.newSingleThreadExecutor()` or whatever the user wants.

utils/common/src/main/java/org/apache/brooklyn/util/repeat/Repeater.java
```diff
@@ -301,89 +332,96 @@ public Repeater limitTimeTo(Duration duration) {
...
         result.checkNoError();
-        if (!result.get()) 
+        if (!result.get()) {
             throw new IllegalStateException(description+" unsatisfied after "+Duration.of(timer));
+        }
     }
-    
+
     public ReferenceWithError<Boolean> runKeepingError() {
-        Preconditions.checkState(body != null, "repeat() method has not been called to set the body");
```
neykov: Any reason to remove the body != null check?

utils/common/src/main/java/org/apache/brooklyn/util/repeat/Repeater.java
```diff
@@ -301,89 +332,96 @@ public Repeater limitTimeTo(Duration duration) {
...
+                    }
+                    return ReferenceWithError.newInstanceMaskingError(false, lastError);
+                }
+    
+                Time.sleep(delayThisIteration);
             }
-
-            Time.sleep(delayThisIteration);
+        } finally {
+           executor.shutdownNow();
```
neykov: Don't shut down if passed from caller.

utils/common/src/main/java/org/apache/brooklyn/util/repeat/Repeater.java
```diff
@@ -301,89 +332,96 @@ public Repeater limitTimeTo(Duration duration) {
...
-                    if (log.isDebugEnabled()) log.debug(description, e);
-                    hasLoggedTransientException = true;
+        try {
+            while (true) {
+                Duration delayThisIteration = delayOnIteration.apply(iterations);
+                iterations++;
+
+                Future<?> call = executor.submit(body);
+                try {
+                    call.get(delayThisIteration.toMilliseconds(), TimeUnit.MILLISECONDS);
```
neykov: I think a separate time out is needed here, especially when used with `backoff`, but happy to go with this.

utils/common/src/main/java/org/apache/brooklyn/util/repeat/Repeater.java
grkvlt: Oops!

utils/common/src/main/java/org/apache/brooklyn/util/repeat/Repeater.java
grkvlt: Should have been changed to a `checkNotNull()` call

utils/common/src/main/java/org/apache/brooklyn/util/repeat/Repeater.java
```diff
@@ -141,6 +148,27 @@ public Repeater repeat(Callable<?> body) {
...
+    }
+
+    /**
+     * @see #threaded()
+     * @param executor an externally managed {@link ExecutorService} to use when creating threads.
+     * @return {@literal this} to aid coding in a fluent style.
+     */
+    public Repeater threaded(ExecutorService executor) {
+        this.executor = executor;
+        return this;
```
neykov: A bit on the pedantic side - can you add `shutdown =false` here. Can imagine first calling `threaded()` and passing the builder to another method only to add a custom executor using `threaded(myExecutor)`. Good to merge after that.

grkvlt:
@neykov have a look now, I updated the comments to be more explicit too

software/base/src/main/java/org/apache/brooklyn/entity/software/base/lifecycle/LifecycleEffectorTasks.java
```diff
@@ -0,0 +1,1072 @@
+/*
```
neykov: Was this file committed by mistake? LGTM otherwise.

neykov:
Thanks for updates @grkvlt, merging.

PR #494, opened 2016-12-15
Simplify BitUtils methods

grkvlt:
Couldn't help it. All tests still pass.

### DIFF
--- a/utils/common/src/main/java/org/apache/brooklyn/util/math/BitUtils.java
+++ b/utils/common/src/main/java/org/apache/brooklyn/util/math/BitUtils.java
@@ -18,25 +18,22 @@
  */
 package org.apache.brooklyn.util.math;
 
+import com.google.common.primitives.Bytes;
+import com.google.common.primitives.Ints;
+
 public class BitUtils {
 
     /** reverses the bits in a byte, i.e.  128 = 0b1000000 = bit list {0,0,0,0,0,0,0,1},
      * reversed yields 1 = 0b00000001 = bit list {1,0,0,0,0,0,0,0} */
     public static byte reverseBitSignificance(byte b) {
-        int result = 0;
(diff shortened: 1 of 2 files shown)


utils/common/src/main/java/org/apache/brooklyn/util/math/BitUtils.java
```diff
@@ -18,25 +18,22 @@
...
      * reversed yields 1 = 0b00000001 = bit list {1,0,0,0,0,0,0,0} */
     public static byte reverseBitSignificance(byte b) {
-        int result = 0;
-        for (int i=0; i<8; i++) {
-            result <<= 1;
-            if ((b&1)==1) result++;
-            b >>= 1;
-        }
-        return (byte)result;
+        return (byte) (Integer.reverse(b) >> 24);
```
neykov: Wow [`Integer.reverse`](https://github.com/openjdk-mirror/jdk7u-jdk/blob/master/src/share/classes/java/lang/Integer.java#L1204) does some interesting voodoo magic :).
Should this be unsigned shift?

aledsage:
@grkvlt can you respond to @neykov's question "can you confirm signed vs unsigned shift", and then good to merge.

utils/common/src/main/java/org/apache/brooklyn/util/math/BitUtils.java
grkvlt: No. The following tests should convince you:
```
        Assert.assertEquals(BitUtils.reverseBitSignificanceInByte(0b00000001), (byte)0b10000000);
        Assert.assertEquals(BitUtils.reverseBitSignificanceInByte(0x0f), (byte)0xf0);
```

neykov:
thanks @grkvlt, merging.

PR #495, opened 2016-12-16
Fix JMX Karaf import

drigodwin:
JMX in a Tomcat deployment was failing in Karaf with the exception `java.lang.RuntimeException: java.io.IOException: java.lang.ClassNotFoundException: javax.management.openmbean.CompositeDataSupport`. This was because of a missing dependency from the wrap of `org.glassfish.external/opendmk_jmxremote_optional_jar`. 

Unfortunately adding an `$Import-Package=javax.management.openmbean` here would exclude the other files included from the jar so `$DynamicImport-Package` is used.

### DIFF
--- a/karaf/features/src/main/feature/feature.xml
+++ b/karaf/features/src/main/feature/feature.xml
@@ -229,7 +229,7 @@
 
     <feature name="brooklyn-software-base"  version="${project.version}"  description="Brooklyn Software Base">
         <bundle>mvn:org.apache.brooklyn/brooklyn-software-base/${project.version}</bundle>
-        <bundle dependency="true">wrap:mvn:org.glassfish.external/opendmk_jmxremote_optional_jar/${opendmk_jmxremote_optional_jar.version}$DynamicImport-Package=javax.management.openmbean</bundle>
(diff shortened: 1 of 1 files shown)


aledsage:
@drigodwin see https://ops4j1.jira.com/wiki/display/paxurl/Wrap+Protocol, section 20.4.2 "Indicating Optional Dependencies": I think we just need a `,*` at the end of `Import-Package=...` for it to not exclude everything else?

See the example lower down in the same file:
```
<bundle dependency="true">wrap:mvn:com.squareup.okhttp/okhttp/2.2.0$Import-Package=org.apache.http.*;resolution:=optional,android.util.*;resolution:=optional,okio;version=%27[1.2.0,1.3.0)%27,*</bundle>
```

neykov:
@drigodwin have you tried @aledsage's suggestion of appending ",*" to the `Import-Package`? Note that some characters could need to be escaped.

If that doesn't work then `DynamicImport-Package` is a fair fallback and LGTM.

drigodwin:
I've tested the `.*` and that also works, thanks @aledsage. I've swapped back to using `$Import-Package` here.

neykov:
thanks, merging.

PR #497, opened 2016-12-20
Add findOpenSocketsOnNode to ReachableSocketFinder

sjcorbett:
This changes the semantics of ReachableSocketFinder. Previously it would return as soon as any host+port was reached. Now it determines the status of each socket in order. The checks are concurrent but, for example, if the third of five sockets is reachable the status of the first and second will also be determined.

These changes are motivated by some other work on JcloudsLocation that I will open a pull request for shortly.

### DIFF
--- a/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsUtil.java
+++ b/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsUtil.java
@@ -34,7 +34,6 @@ import java.util.Properties;
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
-
 import javax.annotation.Nullable;
 
 import org.apache.brooklyn.core.config.Sanitizer;
@@ -349,31 +348,56 @@ public class JcloudsUtil {
         return getFirstReachableAddress(node, Duration.FIVE_MINUTES);
     }
     
+    /**
+     * Uses {@link Networking#isReachablePredicate()} to determine reachability.
+     * @see #getReachableAddresses(NodeMetadata, Duration, Predicate)
+     */
     public static String getFirstReachableAddress(NodeMetadata node, Duration timeout) {
-        return getFirstReachableAddress(node, timeout, new Networking.IsReachablePredicate());
+        return getFirstReachableAddress(node, timeout, Networking.isReachablePredicate());
     }
 
+    /** @see #getReachableAddresses(NodeMetadata, Duration, Predicate) */
     public static String getFirstReachableAddress(NodeMetadata node, Duration timeout, Predicate<? super HostAndPort> socketTester) {
+        Iterable<HostAndPort> addresses = getReachableAddresses(node, timeout, socketTester);
+        HostAndPort address = Iterables.getFirst(addresses, null);
+        if (address != null) {
+            return address.getHostText();
+        } else {
+            throw new IllegalStateException("No reachable IPs for " + node + "; check whether the node is " +
+                    "reachable and whether it meets the requirements of the HostAndPort tester: " + socketTester);
+        }
+    }
+
+    /**
+     * @return The public and private addresses of node that are reachable.
+     * @see #getReachableAddresses(Iterable, Duration, Predicate)
+     */
+    public static Iterable<HostAndPort> getReachableAddresses(NodeMetadata node, Duration timeout, Predicate<? super HostAndPort> socketTester) {
         final int port = node.getLoginPort();
-        List<HostAndPort> sockets = FluentIterable
-                .from(Iterables.concat(node.getPublicAddresses(), node.getPrivateAddresses()))
+        return getReachableAddresses(Iterables.concat(node.getPublicAddresses(), node.getPrivateAddresses()), port, timeout, socketTester);
+    }
+
+    /** @see #getReachableAddresses(Iterable, Duration, Predicate) */
+    public static Iterable<HostAndPort> getReachableAddresses(Iterable<String> hosts, final int port, Duration timeout, Predicate<? super HostAndPort> socketTester) {
+        FluentIterable<HostAndPort> sockets = FluentIterable
+                .from(hosts)
                 .transform(new Function<String, HostAndPort>() {
                         @Override public HostAndPort apply(String input) {
                             return HostAndPort.fromParts(input, port);
-                        }})
-                .toList();
-        
-        ListeningExecutorService executor = MoreExecutors.listeningDecorator(Executors.newCachedThreadPool());
-        try {
-            ReachableSocketFinder finder = new ReachableSocketFinder(socketTester, executor);
-            HostAndPort result = finder.findOpenSocketOnNode(sockets, timeout);
-            return result.getHostText();
-        } catch (Exception e) {
-            Exceptions.propagateIfFatal(e);
-            throw new IllegalStateException("Unable to choose IP for "+node+"; check whether the node is reachable or whether it meets requirements of the HostAndPort tester: " + socketTester , e);
-        } finally {
-            executor.shutdownNow();
-        }
+                        }});
+        return getReachableAddresses(sockets, timeout, socketTester);
+    }
+
+    /**
+     * Uses {@link ReachableSocketFinder} to determine which sockets are reachable. Iterators
+     * are unmodifiable and are lazily evaluated.
+     * @param sockets The host-and-ports to test
+     * @param timeout Max time to try to connect to the ip:port
+     * @param socketTester A predicate determining reachability.
+     */
+    public static Iterable<HostAndPort> getReachableAddresses(Iterable<HostAndPort> sockets, Duration timeout, Predicate<? super HostAndPort> socketTester) {
+        ReachableSocketFinder finder = new ReachableSocketFinder(socketTester);
+        return finder.findOpenSocketsOnNode(sockets, timeout);
     }
 
     // Suggest at least 15 minutes for timeout
--- a/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsUtil.java
+++ b/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsUtil.java
@@ -95,11 +95,13 @@ import com.google.common.util.concurrent.ListeningExecutorService;
 import com.google.common.util.concurrent.MoreExecutors;
 import com.google.inject.Module;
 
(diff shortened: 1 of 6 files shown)


aledsage:
LGTM. It's slightly concerning that `JcloudsUtil.getFirstReachableAddress` might now block until the reachability check on the VM's private (unreachable) address times out. But I presume that is what you're alluding to for "some other work on JcloudsLocation".

To elaborate on my concern, imagine we are (somehow) given the IPs in order: `[10.0.0.1, 10.0.0.2]` but where only 10.0.0.2 is reachable. We might very quickly establish that `10.0.0.2` is reachable (in its future), while the check for `10.0.0.1` might take a really long time (e.g. because it's behind a firewall that just doesn't respond at all). When iterating over the the results of `tryReachable()` in-order, we'll pause for a long time on the future.get() of `10.0.0.1`, before we move onto `10.0.0.2` (which will have been evaluated really quickly).

I don't know whether it's worth trying to avoid that. One way we could do it is for `getReachableAddresses()` to take a callback argument, so it can be told immediately about each success/fail/timeout, and also told when everything has been attempted. That would be a more complicated (internal) API to use. Is it worth it?

sjcorbett:
I appreciate your concern. I feel it depends on how you interpret the word first in `getFirstReachableAddress`. Previously it related to reachability, now it relates to the list of addresses input. My desire is to use to the latter meaning. The mitigation is that the maximum length of time that callers will wait is governed by the timeout argument, which in the only places this is really used is whatever value the user has given for `POLL_FOR_FIRST_REACHABLE_ADDRESS`, or five minutes if they didn't. The reachability checks are still concurrent so it won't get significantly worse when there are more addresses.

The changes forthcoming that I mentioned make how we determine reachability and the addresses used for management of entities more flexible. In particular users will be able to indicate whether they want a node's public or private addresses to be preferred. It is in this scenario that I think we need to wait until we've exhausted the preferences before choosing an address from the other set.

sjcorbett:
After some thought my inclination is to modify this pull request to make both strategies possible.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsUtil.java
```diff
@@ -354,30 +355,60 @@ public static String getFirstReachableAddress(ComputeServiceContext context, Nod
...
+                    "reachable and whether it meets the requirements of the HostAndPort tester: " + socketTester);
+        }
+    }
+
+    /**
+     * @return The public and private addresses of node that are reachable.
+     * @see #getReachableAddresses(Iterable, Duration, Predicate)
+     */
+    public static Iterable<HostAndPort> getReachableAddresses(NodeMetadata node, Duration timeout, Predicate<? super HostAndPort> socketTester) {
         final int port = node.getLoginPort();
```
neykov: I acknowledge that this is existing code, but shouldn't we get the port from the port forwarder? External ports could be different from the actual listening port on the machine.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsUtil.java
```diff
@@ -354,30 +355,60 @@ public static String getFirstReachableAddress(ComputeServiceContext context, Nod
...
+     * @param timeout Max time to try to connect to the ip:port
+     * @param socketTester A predicate determining reachability.
+     */
+    public static Iterable<HostAndPort> getReachableAddresses(Iterable<HostAndPort> sockets, Duration timeout, Predicate<? super HostAndPort> socketTester) {
         ListeningExecutorService executor = MoreExecutors.listeningDecorator(Executors.newCachedThreadPool());
         try {
             ReachableSocketFinder finder = new ReachableSocketFinder(socketTester, executor);
-            HostAndPort result = finder.findOpenSocketOnNode(sockets, timeout);
-            return result.getHostText();
-        } catch (Exception e) {
```
neykov: The `catch` here is useful for adding context to the exception - the node. Could be useful when troubleshooting and cross-referncing logs.

neykov:
Agree with @aledsage's concerns. I think we should never wait the full `timeout` period if we already have reachable addresses. There frequently will be addresses which we can't reach from the Brooklyn instance and which don't have to slow down the provisioning.

The check is mostly us waiting for the machine and the sshd daemon to come up. Once it's working and we can connect with at least one address, then any other reachable addresses will be connectable within seconds.

What could trip the above logic is if a private IP has a duplicate on the local network which responds on the ssh port.

sjcorbett:
@aledsage @neykov I have updated this pull request to hopefully get the best of both worlds. The point of contention was waiting the full timeout for a response when another check had succeeded. I've changed the tasks in `ReachableSocketFinder` to wait a five second grace period before abandoning their check when another result is available. For example, if the input is `[10.0.0.1, 10.0.0.2]` and `.2` is available first it will wait a little longer for `.1` to be available before concluding that `.2` is the first reachable address. Is this satisfactory to you?

@neykov regarding connecting to private IPs I have changes forthcoming that test addresses in tandem with credentials to avoid this.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsUtil.java
sjcorbett: No exception is thrown here any more. The deleted section applied to `getFirstReachableAddress` which still throws with a helpful message if no address if found. `getReachableAddresses` returns an empty iterable.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsUtil.java
sjcorbett: I think I can improve this in the PR to come.

utils/common/src/main/java/org/apache/brooklyn/util/net/ReachableSocketFinder.java
```diff
@@ -69,86 +65,113 @@ public ReachableSocketFinder(Predicate<? super HostAndPort> socketTester, Listen
...
+                public Optional<HostAndPort> call() {
+                    // Whether the socket was reachable (vs. the result of call, which is whether the repeater is done).
+                    final AtomicBoolean theResultWeCareAbout = new AtomicBoolean();
+                    Repeater.create("socket-reachable")
+                            .limitTimeTo(timeout)
+                            .backoffTo(Duration.FIVE_SECONDS)
+                            .until(new Callable<Boolean>() {
+                                @Override
+                                public Boolean call() throws TimeoutException {
+                                    boolean reachable = socketTester.apply(socket);
```
neykov: Socket connect timeout is OS dependent and could be well over a minute. This will trip the grace period timeout below.

utils/common/src/main/java/org/apache/brooklyn/util/net/ReachableSocketFinder.java
```diff
@@ -69,86 +65,113 @@ public ReachableSocketFinder(Predicate<? super HostAndPort> socketTester, Listen
...
-                            if (socketTester.apply(socket)) {
-                                reachableSocket.compareAndSet(null, socket);
-                                latch.countDown();
+                    protected Optional<HostAndPort> computeNext() {
+                        if (count < futures.size()) {
+                            final Future<Optional<HostAndPort>> future = futures.get(count++);
+                            try {
+                                return future.get(timeout.toUnit(TimeUnit.MILLISECONDS), TimeUnit.MILLISECONDS);
+                            } catch (Exception e) {
+                                Exceptions.propagateIfInterrupt(e);
```
neykov: Suggest using `propagateIfFatal` instead. Could swallow important info like `OutOfMemoryError`.

utils/common/src/main/java/org/apache/brooklyn/util/net/ReachableSocketFinder.java
```diff
@@ -69,86 +65,113 @@ public ReachableSocketFinder(Predicate<? super HostAndPort> socketTester, Listen
...
-                    public void run() {
-                        try {
-                            if (socketTester.apply(socket)) {
-                                reachableSocket.compareAndSet(null, socket);
-                                latch.countDown();
+                    protected Optional<HostAndPort> computeNext() {
+                        if (count < futures.size()) {
+                            final Future<Optional<HostAndPort>> future = futures.get(count++);
+                            try {
+                                return future.get(timeout.toUnit(TimeUnit.MILLISECONDS), TimeUnit.MILLISECONDS);
```
neykov: Can we guarantee that all threads return after `timeout` expires? Could we get in a state where each (or at least the first few calls of) `computeNext` blocks?

utils/common/src/main/java/org/apache/brooklyn/util/net/ReachableSocketFinder.java
sjcorbett: `e` is an `Exception` rather than a `Throwable` so I don't think this is a problem.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsUtil.java
neykov: +1 my bad

utils/common/src/main/java/org/apache/brooklyn/util/net/ReachableSocketFinder.java
sjcorbett: The state you posit is possible if you initialise the class with an executor with low throughput. A `singleThreadExecutor` would be a bad idea. In practice the only place this is used ([JcloudsUtil#getReachableAddresses](https://github.com/apache/brooklyn-server/blob/5d0d505cab112fe79b7c2c43e3a08769a0099781/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsUtil.java#L399)) uses `Executors.newCachedThreadPool`. After that I think it depends on the level of concurrency possible on the underlying system. Both of these conditions affected the previous implementation.

utils/common/src/main/java/org/apache/brooklyn/util/net/ReachableSocketFinder.java
sjcorbett: Can we do anything about this?

sjcorbett:
@neykov I've updated again. Took a little reworking due to the uncertainty introduced about when the executor was allowed to shut down. Resorted to having each call to `tryReachable` use its own executor. This is only less efficient than before if you use the same `ReachableSocketFinder` multiple times. I think we can consider it an optimisation for the future.

neykov:
Thanks @sjcorbett. LGTM, merging.

PR #498, opened 2016-12-20
Use SIGTERM in stop script inside AbstractSoftwareProcessSshDriver

bostko:
- fixes BROOKLYN-418

### DIFF
--- a/software/base/src/main/java/org/apache/brooklyn/entity/software/base/AbstractSoftwareProcessSshDriver.java
+++ b/software/base/src/main/java/org/apache/brooklyn/entity/software/base/AbstractSoftwareProcessSshDriver.java
@@ -27,6 +27,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
+import com.google.common.base.Joiner;
 import org.apache.brooklyn.api.entity.EntityLocal;
 import org.apache.brooklyn.api.entity.drivers.downloads.DownloadResolver;
 import org.apache.brooklyn.config.ConfigKey;
@@ -613,22 +614,29 @@ public abstract class AbstractSoftwareProcessSshDriver extends AbstractSoftwareP
                 // no pid, not running; 1 is not running
                 s.requireResultCode(Predicates.or(Predicates.equalTo(0), Predicates.equalTo(1)));
             } else if (STOPPING.equals(phase)) {
+                String stopCommand = Joiner.on('\n').join("PID=$(cat "+pidFile + ")",
+                        "test -n \"$PID\" || exit 0",
+                        "SIGTERM_USED=\"\"",
+                        "for i in $(seq 1 16); do",
+                        "  if ps -p $PID > /dev/null ; then",
+                        "     kill $PID",
+                        "     echo Attempted to stop PID $PID by sending SIGTERM.",
+                        "  else",
+                        "     echo Process $PID stopped successfully.",
+                        "     SIGTERM_USED=\"true\"",
+                        "     break",
+                        "  fi",
+                        "  sleep 1",
+                        "done",
+                        "if test -z $SIGTERM_USED; then",
+                        "  kill -9 $PID",
+                        "  echo Sent SIGKILL to $PID",
+                        "fi",
+                        "rm -f " + pidFile);
                 if (processOwner != null) {
-                    s.body.append(
-                            "export PID=$(" + BashCommands.sudoAsUser(processOwner, "cat "+pidFile) + ")",
-                            "test -n \"$PID\" || exit 0",
-                            BashCommands.sudoAsUser(processOwner, "kill $PID"),
-                            BashCommands.sudoAsUser(processOwner, "kill -9 $PID"),
-                            BashCommands.sudoAsUser(processOwner, "rm -f "+pidFile)
-                    );
+                    s.body.append(BashCommands.sudoAsUser(processOwner, stopCommand));
                 } else {
-                    s.body.append(
-                            "export PID=$(cat "+pidFile+")",
-                            "test -n \"$PID\" || exit 0",
-                            "kill $PID",
-                            "kill -9 $PID",
-                            "rm -f "+pidFile
-                    );
+                    s.body.append(stopCommand);
                 }
             } else if (KILLING.equals(phase)) {
                 if (processOwner != null) {
--- a/software/base/src/test/java/org/apache/brooklyn/entity/software/base/lifecycle/ScriptHelperIntegrationTest.java
+++ b/software/base/src/test/java/org/apache/brooklyn/entity/software/base/lifecycle/ScriptHelperIntegrationTest.java
@@ -1,0 +1,182 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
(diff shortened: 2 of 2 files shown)


software/base/src/main/java/org/apache/brooklyn/entity/software/base/AbstractSoftwareProcessSshDriver.java
```diff
@@ -613,22 +614,27 @@ protected ScriptHelper newScript(Map<String, ?> flags, String phase) {
...
+                        "  kill -9 $PID",
+                        "  echo Process $PID stopped with SIGKILL",
+                        "fi");
                 if (processOwner != null) {
-                    s.body.append(
-                            "export PID=$(" + BashCommands.sudoAsUser(processOwner, "cat "+pidFile) + ")",
-                            "test -n \"$PID\" || exit 0",
-                            BashCommands.sudoAsUser(processOwner, "kill $PID"),
-                            BashCommands.sudoAsUser(processOwner, "kill -9 $PID"),
-                            BashCommands.sudoAsUser(processOwner, "rm -f "+pidFile)
```
aledsage: Did you mean to remove the `rm -rf` of the pidFile? If so, why?

software/base/src/test/java/org/apache/brooklyn/entity/software/base/lifecycle/ScriptHelperIntegrationTest.java
```diff
@@ -0,0 +1,176 @@
...
+                        "trap \"trap_handler_command\" SIGTERM",
+                        "while true; do",
+                        "  sleep 1",
+                        "done");
+        driver.copyResource(ImmutableMap.of(), new ReaderInputStream(new StringReader(launchContents), "UTF-8"), "launch.sh", true);
+        driver.executeLaunchCommand("nohup bash launch.sh > /dev/null &");
+        ScriptHelper stopCommandScriptHelper = driver.stopCommandScriptHelper();
+        stopCommandScriptHelper.execute();
+        Assert.assertEquals(StringUtils.countMatches(stopCommandScriptHelper.getResultStdout(), "is still running"), 1, "SIGTERM should be tried one time");
+        Assert.assertFalse(stopCommandScriptHelper.getResultStdout().contains("stopped with SIGKILL"), "SIGKILL should not be sent after SIGTERM fails.");
```
aledsage: Are you saying that if `kill <pid>` returns non-zero, then it should not subsequently retry or do a `kill -9 <pid>`? Why not? It feels to me that we should escalate and do the `kill -9`.

Is it also worth asserting that the `stopCommandScriptHelper` exit code is correct?

software/base/src/test/java/org/apache/brooklyn/entity/software/base/lifecycle/ScriptHelperIntegrationTest.java
```diff
@@ -0,0 +1,176 @@
...
+                        "  sleep 1",
+                        "done");
+        driver.copyResource(ImmutableMap.of(), new ReaderInputStream(new StringReader(launchContents), "UTF-8"), "launch.sh", true);
+        driver.executeLaunchCommand("nohup bash launch.sh > /dev/null &");
+        ByteArrayOutputStream stdOut = new ByteArrayOutputStream(10);
+        SshMachineLocation machineLocation = (SshMachineLocation) Iterables.getFirst(entity.getLocations(), null);
+        machineLocation.execCommands(
+                ImmutableMap.<String, Object>of(ExecWithLoggingHelpers.STDOUT.getName(), stdOut),
+                "check process is stopped",
+                ImmutableList.of("cat "+driver.getRunDir() + '/' + VanillaSoftwareProcessSshDriver.PID_FILENAME),
```
aledsage: Why are you executing this? Is it really "check process is stopped"? You don't do anything with the response. Do we expect it be running, given that launch has just been called?

software/base/src/test/java/org/apache/brooklyn/entity/software/base/lifecycle/ScriptHelperIntegrationTest.java
bostko: exit code is 2 and not 13 as in the test bash.

software/base/src/test/java/org/apache/brooklyn/entity/software/base/lifecycle/ScriptHelperIntegrationTest.java
bostko: I use the response from that command to get the pid of the command and checking that that pid is not running anymore after the stop command.

software/base/src/main/java/org/apache/brooklyn/entity/software/base/AbstractSoftwareProcessSshDriver.java
```diff
@@ -613,22 +614,29 @@ protected ScriptHelper newScript(Map<String, ?> flags, String phase) {
                 // no pid, not running; 1 is not running
                 s.requireResultCode(Predicates.or(Predicates.equalTo(0), Predicates.equalTo(1)));
             } else if (STOPPING.equals(phase)) {
+                String stopCommand = Joiner.on('\n').join("PID=$(cat "+pidFile + ")",
+                        "test -n \"$PID\" || exit 0",
+                        "SIGTERM_USED=\"\"",
+                        "for i in $(seq 1 16); do",
+                        "  if ps -p $PID > /dev/null ; then",
+                        "     kill $PID",
+                        "     echo $PID is still running.",
```
neykov: Would be more correct if you change it to "Sent SIGKILL to $PID".

software/base/src/main/java/org/apache/brooklyn/entity/software/base/AbstractSoftwareProcessSshDriver.java
```diff
@@ -613,22 +614,29 @@ protected ScriptHelper newScript(Map<String, ?> flags, String phase) {
...
+                        "  else",
+                        "     echo Process $PID stopped successfully.",
+                        "     SIGTERM_USED=\"true\"",
+                        "     break",
+                        "  fi",
+                        "  sleep 1",
+                        "done",
+                        "if test -z $SIGTERM_USED; then",
+                        "  kill -9 $PID",
+                        "  echo Process $PID stopped with SIGKILL",
```
neykov: Same here - `Sent SIGKILL to $PID`. The process could be locked in kernel space so even `kill -9` won't kill it.

software/base/src/main/java/org/apache/brooklyn/entity/software/base/AbstractSoftwareProcessSshDriver.java
```diff
@@ -613,22 +614,29 @@ protected ScriptHelper newScript(Map<String, ?> flags, String phase) {
...
+                        "fi",
+                        "rm " + pidFile);
```
bostko: It script fails stop will exit with non-zero code. 
I think scripts started by Apache Brooklyn shouldn't be modified read-only externally and if so stop is better to fail.

neykov: My point is that `rm` will try to get user input at this point which is what `-f` tries to avoid. It will still fail on read-only fail.

neykov:
LGTM, merging.

PR #499, opened 2016-12-21
BROOKLYN-417: default to no-auth

aledsage:


sjcorbett:
@aledsage Your changes look good to me. Can you confirm that the two integration tests you have added both pass? I've checked that `testStartsWebServerWithoutAuthentication` fails on master.

aledsage:
Thanks @sjcorbett, I can confirm they both pass for me. Merging now, and will cherry-pick into the 0.10.0 branch.

PR #500, opened 2016-12-21
Upgrade free marker to 2.3.25 (from 2.3.22)

aledsage:
Freemarker 2.3.25 adds support for iterating over map entries, which will be useful for some entity implementations.

@ygy could you review this please?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -137,7 +137,7 @@
         <jetty-schemas.version>3.1.M0</jetty-schemas.version>
         <airline.version>0.6</airline.version>
         <mockwebserver.version>20121111</mockwebserver.version>
-        <freemarker.version>2.3.22</freemarker.version>
+        <freemarker.version>2.3.25-incubating</freemarker.version>
         <commons-io.version>2.4</commons-io.version>
         <jsonPath.version>2.0.0</jsonPath.version>
         <commons-compress.version>1.4</commons-compress.version>


pom.xml
```diff
@@ -137,7 +137,7 @@
         <jetty-schemas.version>3.1.M0</jetty-schemas.version>
         <airline.version>0.6</airline.version>
         <mockwebserver.version>20121111</mockwebserver.version>
-        <freemarker.version>2.3.22</freemarker.version>
+        <freemarker.version>2.3.25-incubating</freemarker.version>
```
bostko: I'd rather not upgrade to incubating release.

bostko:
There are numerous ways to iterate a map. 
@ygy what is the use case which new freemarker supports?

pom.xml
aledsage: @bostko the freemarker project was recently moved to Apache - all such projects start as "incubating", and must therefore have "-incubating" in their release names, until they graduate as a top-level project. It was the same for Apache Brooklyn.

The "-incubating" says nothing about the quality or maturity of the product/release.

aledsage:
@bostko see http://freemarker.org/docs/versions_2_3_25.html, in the section "changes on the FTL side":

```
Extended the list directive to support listing hashes (such as Map-s), like <#list map as k, v>${k}: ${v}</#list>, where k and v are the key and value in the key-value pairs of the hash.
```

Also see http://freemarker.org/docs/versions_2_3_24.html, but I don't think those changes relate to iterating over maps.

neykov:
LGTM, merging.

PR #501, opened 2016-12-22
JCLOUDS-1108 workaround: GCE hardwareId short-form

aledsage:


locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -1484,6 +1493,48 @@ public Template buildTemplate(ComputeService computeService, ConfigBag config, C
...
+     */
+    private String transformHardwareId(String hardwareId, ConfigBag config) {
+        String provider = getProvider();
+        String region = getRegion();
+        if (Strings.isBlank(region)) region = config.get(CLOUD_REGION_ID);
+        
+        if (!"google-compute-engine".equals(provider)) {
+            return hardwareId;
+        }
+        if (hardwareId == null || hardwareId.toLowerCase().startsWith("http") || hardwareId.contains("/")) {
```
duncangrant: Should null check be separate?  It presumably doesn't represent a long-form id but rather a misuse of the function.

aledsage: Agreed - given it's a private method, we can be stricter about what is passed in (e.g. throw `NullPointerException` on null) particularly as the one caller guards the call with `Strings.isNonBlank(config.get(HARDWARE_ID))`.

aledsage:
I've changed the null check, and squashed it into one commit. I've re-run all the (Live) tests in `JcloudsGceHardwareProfilesStubbedLiveTest` - they passed for me locally.

Once jenkins has run the unit tests again, I'll merge this.

PR #503, opened 2017-01-05
Drop availability zone suffixes when configuring AWS regions

sjcorbett:
This fixes the deployment of blueprints that give a specific availability zone as the region. For example:
```yaml
location:
  jclouds:aws-ec2:
    region: us-east-1b
services:
- type: org.apache.brooklyn.entity.webapp.tomcat.Tomcat8Server
```
The problem was introduced in the fix for https://issues.apache.org/jira/browse/BROOKLYN-399 (see #458). Setting `LocationConstants.PROPERTY_REGIONS` to a region like `us-east-1b` causes jclouds to throw an `IllegalStateException` in `TemplateBuilder#locationId`.


aledsage:
LGTM; merging

PR #504, opened 2017-01-05
Make the brooklyn server config discoverable by OSGi and downstream projects

tbouron:
This is a follow up PR of https://github.com/apache/brooklyn-dist/pull/72

It moves the brooklyn server config back into `brooklyn-server` and create a feature that will be used in `brooklyn-dist` (https://github.com/apache/brooklyn-dist/pull/75)

### DIFF
--- a/karaf/config/pom.xml
+++ b/karaf/config/pom.xml
@@ -1,0 +1,74 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+    Licensed to the Apache Software Foundation (ASF) under one
+    or more contributor license agreements.  See the NOTICE file
+    distributed with this work for additional information
+    regarding copyright ownership.  The ASF licenses this file
+    to you under the Apache License, Version 2.0 (the
+    "License"); you may not use this file except in compliance
+    with the License.  You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing,
+    software distributed under the License is distributed on an
+    "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+    KIND, either express or implied.  See the License for the
+    specific language governing permissions and limitations
+    under the License.
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>brooklyn-karaf</artifactId>
+        <groupId>org.apache.brooklyn</groupId>
+        <version>0.11.0-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>brooklyn-server-config</artifactId>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.codehaus.mojo</groupId>
+                <artifactId>build-helper-maven-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>attach-artifact</id>
+                        <phase>package</phase>
+                        <goals>
+                            <goal>attach-artifact</goal>
+                        </goals>
+                        <configuration>
+                            <artifacts>
+                                <artifact>
+                                    <file>
+                                        ${project.basedir}/src/main/resources/org.apache.brooklyn.classrename.cfg
+                                    </file>
+                                    <type>cfg</type>
+                                    <classifier>classrename</classifier>
+                                </artifact>
+                                <artifact>
+                                    <file>
+                                        ${project.basedir}/src/main/resources/org.apache.brooklyn.core.catalog.bomscanner.cfg
+                                    </file>
+                                    <type>cfg</type>
+                                    <classifier>core.catalog.bomscanner</classifier>
+                                </artifact>
+                                <artifact>
+                                    <file>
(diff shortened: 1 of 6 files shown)


tbouron:
Well, jenkins won't be happy for https://github.com/apache/brooklyn-dist/pull/75 until this PR is merged (it is uses the new feature created here)

drigodwin:
I don't think we can merge this and [brooklyn-dist/75](https://github.com/apache/brooklyn-dist/pull/75) in it's current state as it will break the classic launch brooklyn.

tbouron:
Good catch @drigodwin, the classic mode would not work without the `default.catalog.bom` in `brooklyn-server`.

I re-added this file in https://github.com/apache/brooklyn-dist/pull/75 and removed it here, should be fine now.

karaf/config/src/main/resources/org.apache.brooklyn.core.catalog.bomscanner.cfg
```diff
@@ -0,0 +1,27 @@
...
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS,
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+#
+################################################################################
+
+# CSV Whitelist of regular expressions to match bundle symbolic id if bundle is to be permitted to add
```
aledsage: Unrelated to this PR (because you're just moving the files from brooklyn-dist), but we should probably include a comment that says about `brooklyn.experimental.feature.osgi.catalog_bom=true`. Without enabling that functionality, this file will presumably have no effect.

See https://github.com/apache/brooklyn-server/blob/master/core/src/main/java/org/apache/brooklyn/core/BrooklynFeatureEnablement.java#L122-L127

aledsage:
I take it we're happy to leave the itests (e.g. https://github.com/apache/brooklyn-dist/blob/master/karaf/itest/src/test/java/org/apache/brooklyn/core/catalog/internal/CatalogBomScannerTest.java) in brooklyn-dist, even though it's brooklyn-server that declares the config files? I think it needs the full `brooklyn-dist` bulid in order to run those itests.

karaf/config/pom.xml
```diff
@@ -0,0 +1,74 @@
...
+                                    </file>
+                                    <type>cfg</type>
+                                    <classifier>classrename</classifier>
+                                </artifact>
+                                <artifact>
+                                    <file>
+                                        ${project.basedir}/src/main/resources/org.apache.brooklyn.core.catalog.bomscanner.cfg
+                                    </file>
+                                    <type>cfg</type>
+                                    <classifier>core.catalog.bomscanner</classifier>
```
tbouron: @aledsage This PR just make the configuration files available through karaf features. If you want to changes names I think it should be addressed in another PR, especially because it won't be backward compatible.

karaf/config/src/main/resources/org.apache.brooklyn.core.catalog.bomscanner.cfg
tbouron: I think it should be addressed in another PR

aledsage:
@m4rkmckenna and @neykov - would appreciate your comments on this and https://github.com/apache/brooklyn-dist/pull/75. I think this is motivated by your comments "config files should be put in place at compile"?

Can you confirm that brooklyn-server is the right place for these config files, rather than in brooklyn-dist?

What about the other files in https://github.com/apache/brooklyn-dist/tree/master/karaf/apache-brooklyn/src/main/resources/etc (e.g. org.ops4j.pax.web.cfg etc)? Is it right that they stay where they are in brooklyn-dist?

tbouron:
@aledsage My comments below

> I take it we're happy to leave the itests (e.g. https://github.com/apache/brooklyn-dist/blob/master/karaf/itest/src/test/java/org/apache/brooklyn/core/catalog/internal/CatalogBomScannerTest.java) in brooklyn-dist, even though it's brooklyn-server that declares the config files? I think it needs the full brooklyn-dist bulid in order to run those itests.

Precisely

 > What about the other files in https://github.com/apache/brooklyn-dist/tree/master/karaf/apache-brooklyn/src/main/resources/etc (e.g. org.ops4j.pax.web.cfg etc)? Is it right that they stay where they are in brooklyn-dist?

Those files are not related to `brooklyn-server` but karaf itself hence why I left them over there

neykov:
Re itests - that's correct @aledsage, the only reason to have them here is that we need the karaf dist built before running them.
Slight preference to keeping all the config files together with the (possibly any downstreams) assembly dist  cherry-picking files.

m4rkmckenna:
@neykov My personal pref is to keep the config files in the same repository as the code that uses it ... I hate PRs that span multiple repositories but i guess its not that big a deal

tbouron:
@neykov Same as @m4rkmckenna, I think it's cleaner to have the config files in the same repo as the code that use them. Now I can always revert back, close this PR and do all this in `brooklyn-dist`.

What do you guys think?

neykov:
My reasoning is:
  * This is runtime configuration and as such is defined by the requirements of the distribution the files end up in.
  * That's analogous to `default.catalog.bom` - each distribution has its own needs of what goes in it - it's not the Brooklyn "framework" that decides what's in the catalog - it just consumes it.
  * When in brooklyn-dist, could pack all the config files together to be re-used.

That said, both approaches are technically correct and this is turning into bike-shedding discussion so I'm happy with any solution.

tbouron:
Ok, looks like this is controversial so let's change nothing, it will be simpler and quicker.

I'll close this PR, re-add the files in `brooklyn-dist` on https://github.com/apache/brooklyn-dist/pull/75 and do the feature changes over there.

PR #505, opened 2017-01-06
BROOKLYN-404: fix ConcurrentModificationException in BrooklynProperties

aledsage:
As discussed in https://issues.apache.org/jira/browse/BROOKLYN-404, the fix is to change `BrooklynProperties` so it no longer exposes the map interface. This means we can much more easily control its mutators and its accessors to avoid such `ConcurrentModificationException`s.

This PR does *not* do anything about @neykov's additional suggestion in jira: "Taking a step back I think we should either let users update the properties in a non-hackish way (that is include the mutators in the interface which getConfig() returns) or introduce an alternative "scratch" space to serve the same purpose."

Currently, if someone wants to modify the `BrooklynProperties`, it needs to be cast to `BrooklynProperties` (whereas `managementContext.getConfig()` returns `StringConfigMap`, which is a super-interface of `BrooklynProperties`).

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/internal/BrooklynProperties.java
+++ b/core/src/main/java/org/apache/brooklyn/core/internal/BrooklynProperties.java
@@ -50,7 +50,7 @@ import com.google.common.base.Supplier;
  * thereafter.
  */
 @SuppressWarnings("rawtypes")
-public interface BrooklynProperties extends Map, StringConfigMap {
+public interface BrooklynProperties extends StringConfigMap {
 
     public static class Factory {
         private static final Logger LOG = LoggerFactory.getLogger(BrooklynProperties.Factory.class);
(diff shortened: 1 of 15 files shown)


sjcorbett:
@aledsage could you rebase on master?

core/src/main/java/org/apache/brooklyn/core/internal/BrooklynProperties.java
```diff
@@ -181,6 +181,7 @@ public static Builder fromProperties(BrooklynProperties brooklynProperties) {
             }
 
             @Override
+            @SuppressWarnings("deprecation")
```
sjcorbett: Minor - would be better to change `toString` to use `MoreObjects` rather than the deprecated `Objects` rather than silencing the warning.

sjcorbett:
Looks good; will merge.

PR #506, opened 2017-01-06
AbstractYamlTest: make setUpPlatform overridable

aledsage:
This is useful for writing more advanced tests that want to use YAML - in particular, I want it for refactoring performance/load tests such as https://github.com/apache/brooklyn-library/blob/master/qa/src/test/java/org/apache/brooklyn/qa/load/LoadTest.java (which need a major overhaul - I made a start on that; PR to follow soon hopefully).


sjcorbett:
Looks good.

PR #508, opened 2017-01-06
Alternate names for JcloudsProviders

drigodwin:
This adds the ability to have alternate names for JcloudsProviders stored a property file `org/apache/brooklyn/core/mgmt/persist/jcloudsProviderRenames.properties` (for classic) or a config file `org.apache.brooklyn.jcloudsrename` (for karaf). To do this I've refactored the existing structure for Class Renames.


aledsage:
@drigodwin can you kick off the jenkins build again please (e.g. by making a change to the commit id by running `git commit --ammend; git push -f`).

The jenkins test failure looks unrelated, so we should look at that in a separate PR:
```
2017-01-06 14:40:00,515 INFO  TESTNG INVOKING: "Surefire test" - org.apache.brooklyn.core.entity.ApplicationLifecycleStateTest.testStartsThenChildFailsButWithQuorumCausesAppToSucceed()
2017-01-06 14:40:00,586 INFO  No Camp-YAML parser registered for parsing catalog item DSL; skipping DSL-parsing
2017-01-06 14:40:00,604 WARN  Setting TestEntityImpl{id=pr3k9o55xr} on-fire due to problems when expected running, up=false, not-up-indicators: {myIndicator=Simulate not-up of child}
2017-01-06 14:40:00,605 WARN  Setting Application[jncpaf3w] on-fire due to problems when expected running, up=true, problems: {service-lifecycle-indicators-from-children-and-members=Required entity not healthy: TestEntityImpl{id=pr3k9o55xr}}
2017-01-06 14:40:00,616 INFO  TESTNG FAILED: "Surefire test" - org.apache.brooklyn.core.entity.ApplicationLifecycleStateTest.testStartsThenChildFailsButWithQuorumCausesAppToSucceed() finished in 99 ms
java.lang.AssertionError: entity=Application[jncpaf3w]; attribute=Sensor: service.state (org.apache.brooklyn.core.entity.lifecycle.Lifecycle) expected [running] but found [on-fire]
	at org.apache.brooklyn.test.Asserts.fail(Asserts.java:754)
	at org.apache.brooklyn.test.Asserts.failNotEquals(Asserts.java:144)
	at org.apache.brooklyn.test.Asserts.assertEquals(Asserts.java:466)
	at org.apache.brooklyn.core.entity.EntityAsserts.assertAttributeEquals(EntityAsserts.java:58)
	at org.apache.brooklyn.core.entity.EntityAsserts$4.run(EntityAsserts.java:126)
	at org.apache.brooklyn.test.Asserts$RunnableAdapter.call(Asserts.java:1351)
	at org.apache.brooklyn.test.Asserts.succeedsContinually(Asserts.java:1034)
	at org.apache.brooklyn.test.Asserts.succeedsContinually(Asserts.java:1013)
	at org.apache.brooklyn.core.entity.EntityAsserts.assertAttributeEqualsContinually(EntityAsserts.java:124)
	at org.apache.brooklyn.core.entity.ApplicationLifecycleStateTest.assertHealthContinually(ApplicationLifecycleStateTest.java:196)
	at org.apache.brooklyn.core.entity.ApplicationLifecycleStateTest.testStartsThenChildFailsButWithQuorumCausesAppToSucceed(ApplicationLifecycleStateTest.java:170)
```

core/src/main/java/org/apache/brooklyn/core/mgmt/persist/DeserializingClassRenamesProvider.java
```diff
@@ -43,49 +37,30 @@
...
-        loaders.add(new ClasspathConfigLoader());
-    }
-    
-    private static volatile Map<String, String> cache;
+    private static DeserializingClassRenamesProvider instance;
 
-    public static List<ConfigLoader> getLoaders() {
-        return loaders;
+    public static DeserializingClassRenamesProvider getInstance(){
+        if (instance == null) instance = new DeserializingClassRenamesProvider();
```
aledsage: Need to be careful about synchronization for singletons. See Bloch's Effective Java second edition, item 3.

I don't think we need to instantiate it lazily - we can do it as soon as this class is loaded. I'd change it to:
```
private static final DeserializingClassRenamesProvider INSTANCE = new DeserializingClassRenamesProvider();
```

core/src/main/java/org/apache/brooklyn/core/mgmt/persist/DeserializingJcloudsRenamesProvider.java
```diff
@@ -0,0 +1,43 @@
...
+
+    private static DeserializingJcloudsRenamesProvider instance;
+
+    public static DeserializingJcloudsRenamesProvider getInstance(){
+        if (instance == null) instance = new DeserializingJcloudsRenamesProvider();
+        return instance;
+    }
+
+    private DeserializingJcloudsRenamesProvider(){
+        super(Arrays.asList(new ConfigLoader[]{
```
aledsage: Nit-picking... I'd the more concise: `super(ImmutableList.of(new JcloudsProviderRenameConfigLoader()));`

And I'd change the `DeserializingProvider` constructor to take `(List<? extends ConfigLoader> loaders)`.

core/src/main/java/org/apache/brooklyn/core/mgmt/persist/DeserializingClassRenamesProvider.java
```diff
@@ -43,49 +37,30 @@
...
-     * 
+     *
      * {@code findMappedName("com.example.MyFoo$MySub")} will return {@code com.example.renamed.MyFoo$MySub}, if
      * the renamed contains {@code com.example.MyFoo: com.example.renamed.MyFoo}.
      */
     @Beta
-    public static String findMappedName(String name) {
-        return Reflections.findMappedNameAndLog(DeserializingClassRenamesProvider.loadDeserializingClassRenames(), name);
+    public String findMappedName(String name) {
+        return Reflections.findMappedNameAndLog(DeserializingClassRenamesProvider.getInstance().loadDeserializingMapping(), name);
```
aledsage: This can just call `loadDeserializingMapping(...)` - it should be the same instance as is returned by `DeserializingClassRenamesProvider.getInstance()`.

core/src/main/java/org/apache/brooklyn/core/mgmt/persist/DeserializingJcloudsRenamesProvider.java
aledsage: Same for `DeserializingClassRenamesProvider`

core/src/main/java/org/apache/brooklyn/core/mgmt/persist/ClasspathConfigLoader.java
```diff
@@ -18,55 +18,17 @@
...
-
 /**
  * Loads the class-renames from the configuration file on the classpath.
  *
  * @see {@link #DESERIALIZING_CLASS_RENAMES_PROPERTIES_PATH}
  */
-public class ClasspathConfigLoader implements ConfigLoader {
-    private static final Logger LOG = LoggerFactory.getLogger(ClasspathConfigLoader.class);
-    private static final String DESERIALIZING_CLASS_RENAMES_PROPERTIES_PATH = "classpath://org/apache/brooklyn/core/mgmt/persist/deserializingClassRenames.properties";
+public class ClasspathConfigLoader extends PropertiesConfigLoader {
```
aledsage: I don't think this class pulls its weight anymore. I'd delete it, and just have the thing that constructed it call `new PropertiesConfigLoader(DESERIALIZING_CLASS_RENAMES_PROPERTIES_PATH)`.

I might then rename `PropertiesConfigLoader` to indicate that it's loading the properties from the classpath.

core/src/main/java/org/apache/brooklyn/core/mgmt/persist/JcloudsProviderRenameConfigLoader.java
```diff
@@ -0,0 +1,30 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.brooklyn.core.mgmt.persist;
+
+public class JcloudsProviderRenameConfigLoader extends PropertiesConfigLoader {
```
aledsage: As for ClasspathConfigLoader, I don't think this class is pulling its weight - I'd delete it.

core/src/main/java/org/apache/brooklyn/core/mgmt/persist/DeserializingProvider.java
```diff
@@ -0,0 +1,79 @@
...
+
+    private volatile Map<String, String> cache;
+
+    public List<ConfigLoader> getLoaders() {
+        return loaders;
+    }
+
+    @Beta
+    public Map<String, String> loadDeserializingMapping() {
+        synchronized (DeserializingProvider.class) {
```
aledsage: Does it need to synchronize on the class? Now that this is non-static, can we not just do `synchronized (this) {` (and same in `reset()` obviously).

core/src/main/java/org/apache/brooklyn/core/mgmt/persist/PropertiesConfigLoader.java
```diff
@@ -0,0 +1,72 @@
...
+    private String propertiesPath;
+
+    protected PropertiesConfigLoader(String propertiesPath){
+        this.propertiesPath = propertiesPath;
+    }
+
+    @Override
+    public Map<String, String> load() {
+        try {
+            InputStream resource = new ResourceUtils(DeserializingClassRenamesProvider.class).getResourceFromUrl(propertiesPath);
```
aledsage: I'd use `new ResourceUtils(PropertiesConfigLoader.class, ...)`. Otherwise it's just confusing what the relationship is between this and `DeserializingClassRenamesProvider`.

karaf/init/src/main/java/org/apache/brooklyn/launcher/osgi/ClasspathOsgiConfigLoader.java
```diff
@@ -0,0 +1,70 @@
...
+ * future releases.
+ *
+ * See http://stackoverflow.com/questions/18844987/creating-a-blueprint-bean-from-an-inner-class:
+ * we unfortunately need to include {@code !org.apache.brooklyn.core.mgmt.persist.DeserializingClassRenamesProvider}
+ * in the Import-Package, as the mvn plugin gets confused due to the use of this inner class
+ * within the blueprint.xml.
+ *
+ * @see {@link #KARAF_DESERIALIZING_CLASS_RENAMES_PROPERTIES}
+ */
+public class ClasspathOsgiConfigLoader extends OsgiConfigLoader{
```
aledsage: In what way is this "Classpath" OsgiConfigLoader?

karaf/init/src/main/java/org/apache/brooklyn/launcher/osgi/JcloudsProviderRenameOsgiConfigLoader.java
```diff
@@ -0,0 +1,57 @@
...
+ */
+package org.apache.brooklyn.launcher.osgi;
+
+import org.apache.brooklyn.core.mgmt.persist.DeserializingJcloudsRenamesProvider;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Map;
+
+public class JcloudsProviderRenameOsgiConfigLoader extends OsgiConfigLoader{
```
aledsage: Nitpick: personal preference for a space before the opening `{`

karaf/init/src/main/java/org/apache/brooklyn/launcher/osgi/ClasspathOsgiConfigLoader.java
aledsage: I'd probably go for "ClassRenameOsgiConfigLoader"

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsProviderAndApiLoader.java
```diff
@@ -84,13 +85,15 @@ public static boolean isApi(String id) {
     }
     
     public static Optional<ProviderMetadata> getProvider(String id) {
+        id = DeserializingJcloudsRenamesProvider.getInstance().applyJcloudsRenames(id);
```
aledsage: I wonder if we should overwrite the provider config key with the new name. Otherwise, we'll need to leave these values in the properties file forever. But I'm fine with us not doing that in this PR.

locations/jclouds/src/test/java/org/apache/brooklyn/location/jclouds/JcloudsProviderAndApiLoaderTest.java
```diff
@@ -68,6 +69,35 @@ public void testRegisteredProvider() throws Exception {
             ProviderRegistry.unregisterProvider(provider);
         }
     }
+
+    @Test
+    public void testRenamedRegisteredProvider() throws Exception {
+        String id = "my-example-provider";
+        String renamedId = "my-example-provider-renamed";
```
aledsage: Personal preference for using "oldId" and "newId". When I first read the variable names, It thought that "renamedId" was going to be what the provider had been renamed *to*, rather than what it was renamed *from*.

core/src/main/java/org/apache/brooklyn/core/mgmt/persist/DeserializingJcloudsRenamesProvider.java
drigodwin: If I did this, as described [here](http://stackoverflow.com/questions/2776975/how-can-i-add-to-list-extends-number-data-structures), I wouldn't be able to add new elements to the list as is required by the `OsgiConfigLoader`.

aledsage: Ah, good point. Given that the super-constructor is protected, then that's ok. Could instead use: `super(MutableList.of(new JcloudsProviderRenameConfigLoader()));`

As a side-note, if the constructor were public, then it would be better to take a copy of the list (rather than relying on the caller passing in a mutable list).

If we do continue to rely on the constructor being called with a mutable list (rather than it taking a copy), then probably worth adding a javadoc comment about that parameter.

drigodwin:
retest this please

aledsage:
Thanks @drigodwin - LGTM; merging.

Not sure why jenkins is still showing "validating", 20 hours later. The link is to https://builds.apache.org/job/brooklyn-server-pull-requests/1547, which was kicked off at 09-Jan-2017 15:17:14 and the console output shows "BUILD SUCCESS". Therefore I'm going to assume the build was fine and that it's a jenkins thing!

sjcorbett:
I get an ugly exception Brooklyn with this change:
```
2017-01-10 12:06:39,640 WARN  Failed to load properties file from classpath://org/apache/brooklyn/core/mgmt/persist/jcloudsProviderRenames.properties (continuing)
org.apache.brooklyn.util.exceptions.PropagatedRuntimeException: Error getting resource 'classpath://org/apache/brooklyn/core/mgmt/persist/jcloudsProviderRenames.properties' for class org.apache.brooklyn.core.mgmt.persist.PropertiesConfigLoader
        at org.apache.brooklyn.util.exceptions.Exceptions.propagate(Exceptions.java:165) ~[org.apache.brooklyn-brooklyn-utils-common-0.11.0-SNAPSHOT.jar:0.11.0-SNAPSHOT]
Caused by: java.io.IOException: Error accessing classpath://org/apache/brooklyn/core/mgmt/persist/jcloudsProviderRenames.properties: java.io.IOException: org/apache/brooklyn/core/mgmt/persist/jcloudsProviderRenames.properties not found on classpath
        at org.apache.brooklyn.util.core.ResourceUtils.getResourceFromUrl(ResourceUtils.java:232) ~[org.apache.brooklyn-brooklyn-core-0.11.0-SNAPSHOT.jar:0.11.0-SNAPSHOT]
Caused by: java.io.IOException: org/apache/brooklyn/core/mgmt/persist/jcloudsProviderRenames.properties not found on classpath
        at org.apache.brooklyn.util.core.ResourceUtils.getResourceViaClasspath(ResourceUtils.java:371) ~[org.apache.brooklyn-brooklyn-core-0.11.0-SNAPSHOT.jar:0.11.0-SNAPSHOT]
```
Should another file have been included in the pull request?

drigodwin:
@sjcorbett it is missing that file, I haven't seen that exception however. I've opened #511 to add it.

PR #509, opened 2017-01-09
Add Azure Location

drigodwin:
This adds the Azure classic and Azure ARM locations from jclouds labs to the karaf features.

### DIFF
--- a/karaf/features/src/main/feature/feature.xml
+++ b/karaf/features/src/main/feature/feature.xml
@@ -265,6 +265,10 @@
 
         <bundle dependency="true">mvn:org.apache.jclouds.provider/google-compute-engine/${jclouds.version}</bundle>
 
+        <bundle dependency="true">mvn:org.apache.jclouds.labs/azurecompute/${jclouds.version}</bundle>
+        <bundle dependency="true">mvn:org.apache.jclouds.labs/azurecompute-arm/${jclouds.version}</bundle>
+        <bundle dependency="true">mvn:org.apache.jclouds.api/oauth/${jclouds.version}</bundle>
+
         <feature>jclouds-services</feature>
         <bundle dependency="true">wrap:mvn:com.jcraft/jsch.agentproxy.sshj/0.0.9$Bundle-Name=wrap%20%3A%3A%20jsch.agentproxy.sshj</bundle>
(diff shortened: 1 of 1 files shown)


aledsage:
@drigodwin looks good. In what way is this still "work in progress"?

Looking at how features are defined in https://github.com/jclouds/jclouds-karaf/blob/master/feature/src/main/resources/feature.xml, I wonder if we want to (overengineer?!) structure it like:
```
    <!--
        Features for providers defined in jclouds-labs, similar to the way core providers are defined in
        https://github.com/jclouds/jclouds-karaf/blob/rel/jclouds-karaf-2.0.0/feature/src/main/resources/feature.xml
    -->
    <feature name="jclouds-labs-azurecompute" description='Components to access Azure Compute Classic' version="${jclouds.version}">
        <feature version='${jclouds.version}'>jclouds-compute</feature>
        <bundle>mvn:org.apache.jclouds.labs/azurecompute/${jclouds.version}</bundle>
    </feature>

    <feature name="jclouds-labs-azurecompute-arm" description='Components to access Azure Compute (ARM)' version="${jclouds.version}">
        <feature version='${jclouds.version}'>jclouds-compute</feature>
        <bundle dependency="true">mvn:org.apache.jclouds.api/oauth/${jclouds.version}</bundle>
        <bundle>mvn:org.apache.jclouds.labs/azurecompute-arm/${jclouds.version}</bundle>
    </feature>
```

And then include those features in the feature `brooklyn-locations-jclouds`.

---
As an aside (for different PR), we can now use the jclouds feature `jclouds-google-compute-engine` rather than us having to explicitly depend on the bundle.

drigodwin:
It's still a wip as there were some oddities that emerged during testing @aledsage, they looked like they could be unrelated but thought it best if merging didn't happen until they were sorted out.

m4rkmckenna:
@drigodwin @aledsage I tweaked the startup order in 6b90a4a this now works as expected

drigodwin:
I've tested this by deploying Tomcat to Azure classic and it worked fine, thanks @m4rkmckenna  👍

geomacy:
👍 LGTM and others have tested; merging.

PR #510, opened 2017-01-10
Fix logging of creation string in JcloudsLocation

sjcorbett:
Use `getCreationString` rather than `setup.getDescription` throughout `JcloudsLocation`. 

Fixes the annoying nulls in log messages when locations are provisioned:
```
2017-01-10 10:54:17,954 INFO  Creating VM null in JcloudsLocation[...]
2017-01-10 11:00:34,114 ERROR Failed to start VM for null;
```
etc.

See  8405d8774a0380a99b9dd1fbd2daf7fc15265cb8 for the introduction of `getCreationString`.

### DIFF
--- a/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
+++ b/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
@@ -651,7 +651,7 @@ public class JcloudsLocation extends AbstractCloudMachineProvisioningLocation im
         Stopwatch provisioningStopwatch = Stopwatch.createStarted();
 
         try {
-            LOG.info("Creating VM "+setup.getDescription()+" in "+this);
+            LOG.info("Creating VM "+getCreationString(setup)+" in "+this);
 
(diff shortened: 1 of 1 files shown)


locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -651,7 +651,7 @@ protected MachineLocation obtainOnce(ConfigBag setup) throws NoMachinesAvailable
         Stopwatch provisioningStopwatch = Stopwatch.createStarted();
 
         try {
-            LOG.info("Creating VM "+setup.getDescription()+" in "+this);
+            LOG.info("Creating VM "+getCreationString(setup)+" in "+this);
```
m4rkmckenna: [minor / ignore] I would change this to  `LOG.info("Creating VM {} in {}", getCreationString(setup), this);`

probably applies elsewhere

aledsage:
LGTM - thanks @sjcorbett; merging.

It looks like we'll no longer log the `setup.getDescription()` when it's non-null, but that's fine. It's pretty much always null in the logs I've seen, I think, and the getCreationString looks like it'll be nicer.

PR #511, opened 2017-01-10
Add jcloudsProviderRenames.properties

drigodwin:
This adds the `jcloudsProviderRenames.properties` file missing from #508


sjcorbett:
@drigodwin thanks.

PR #513, opened 2017-01-10
Use stripped version of ScriptBytecodeAdapter.isCase

bostko:
- Optimize performance and avoid locks used in groovy dynamic invocation
  Should fix BROOKLYN-424

### DIFF
--- a/utils/groovy/src/main/java/org/apache/brooklyn/util/groovy/GroovyJavaMethods.java
+++ b/utils/groovy/src/main/java/org/apache/brooklyn/util/groovy/GroovyJavaMethods.java
@@ -22,6 +22,7 @@ import java.util.concurrent.Callable;
 
 import org.apache.brooklyn.util.concurrent.CallableFromRunnable;
 import org.apache.brooklyn.util.exceptions.Exceptions;
+import org.codehaus.groovy.runtime.DefaultGroovyMethods;
 import org.codehaus.groovy.runtime.ScriptBytecodeAdapter;
 import org.codehaus.groovy.runtime.callsite.CallSite;
 import org.codehaus.groovy.runtime.callsite.CallSiteArray;
@@ -65,7 +66,7 @@ public class GroovyJavaMethods {
     @SuppressWarnings("unchecked")
     public static <T> Callable<T> callableFromRunnable(final Runnable job) {
         try {
-            if (ScriptBytecodeAdapter.isCase(job, Callable.class)) {
+            if (safeGroovyIsCase(job, Callable.class)) {
                 return (Callable<T>)ScriptBytecodeAdapter.asType(job, Callable.class);
             } else {
                 return CallableFromRunnable.newInstance(job, null);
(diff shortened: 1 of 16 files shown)


utils/groovy/src/main/java/org/apache/brooklyn/util/groovy/GroovyJavaMethods.java
```diff
@@ -75,6 +76,20 @@
         }
     }
 
+    /**
+     * Alternative implementation of {@link ScriptBytecodeAdapter#isCase(Object, Object)}<br>
+     * Stripped down to work only for caseExpression of type <code>java.lang.Class</code>.<br>
```
geomacy: If we're OK to do that then do we even need this method? Couldn't we just use instanceof/isAssignableFrom as appropriate?
```java
if ( job instanceof Callable) {
```
etc?

utils/groovy/src/main/java/org/apache/brooklyn/util/groovy/GroovyJavaMethods.java
```diff
@@ -75,6 +76,20 @@
...
+     * It behaves the same way only for cases when caseExpression <code>java.lang.Class</code> does <b>not</b> implement <code>isCase</code> method.<br>
+     * It goes directly to {@link DefaultGroovyMethods#isCase(Object, Object)} method instead of using Groovy dynamic invocation.<br>
+     * This safes extra operations and locks used in Groovy dynamic invocation. See <a href="https://issues.apache.org/jira/browse/BROOKLYN-424">BROOKLYN-424</a>.
```
geomacy: Typo "safes".   Maybe something like "This saves extra operations and avoids the locks used ...."

utils/groovy/src/main/java/org/apache/brooklyn/util/groovy/GroovyJavaMethods.java
```diff
@@ -75,6 +76,20 @@
...
+     */
+    public static boolean scriptBytecodeAdapter_isCase(Object switchValue, Class caseExpression) {
```
geomacy: If we do keep this method, I'd suggest we change the name, since it now has nothing to do with ScriptByteCodeAdapter, maybe call it something like `safeGroovyIsCase`.

utils/groovy/src/main/java/org/apache/brooklyn/util/groovy/GroovyJavaMethods.java
bostko: I think it is the same.
I prefer reusing the actual method being used so you can easily compare implementations of  `ScriptBytecodeAdapter#isCase(Object, Object)` and `safeGroovyIsCase` just by checking that at the end both refer to `DefaultGroovyMethods.isCase`

geomacy: My point is that _if_ we don't need the functionality that `ScriptBytecodeAdapter` was giving us, then we should keep it as simple as possible; otherwise people will be looking at this code thinking "why is the call to this Groovy stuff needed?" when in fact it's not. It will be confusing. There's no need for it to do "isCase" any more, is there?

aledsage:
@bostko I'm reviewing this now, and will submit an additional commit on top of this PR shortly. I think a lot of things can be improved by instead calling `JavaGroovyEquivalents.groovyTruth(...)` and `JavaGroovyEquivalents.elvis(...)` so that we entirely avoid groovy code! I'm making those changes now and will let you know when that's done/tested.

utils/groovy/src/main/java/org/apache/brooklyn/util/groovy/GroovyJavaMethods.java
aledsage: If we keep it, let's make it package-private (so tests can still call it).

utils/groovy/src/main/java/org/apache/brooklyn/util/groovy/GroovyJavaMethods.java
aledsage: I wondered about that as well - that was my initial reaction. However, I'm fine with doing it like this for two reasons.

First, we can change most callers to instead use `JavaGroovyEquivalents` to avoid calling groovy code at all. Second, groovy allows some weird stuff with meta-programming, so I think it's safest to call the real Groovy code.

aledsage:
@bostko see https://github.com/bostko/brooklyn-server/pull/3

aledsage:
LGTM; merging - thanks @bostko

PR #514, opened 2017-01-11
Allow oauth.endpoint in location

drigodwin:
`oauth.endpoint` is required by Azure ARM but isn't passed through to the jcloudsProperties.


andreaturli:
lgtm!

PR #516, opened 2017-01-13
Add unit tests for Karaf features

googlielmo:
Add pax exam unit tests to brooklyn server / karaf.

Currently, the unit tests check that the Karaf features load correctly.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -60,7 +60,6 @@
         <dependency>
             <groupId>com.hierynomus</groupId>
             <artifactId>sshj</artifactId>
-            <version>${sshj.version}</version>
         </dependency>
         <dependency>
             <groupId>com.thoughtworks.xstream</groupId>
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -35,6 +35,20 @@
         <relativePath>../parent/pom.xml</relativePath>
     </parent>
 
+    <properties>
+        <osgi.import>
+            !android.util*,
+            !org.apache.brooklyn.rt.felix,
+            !org.apache.felix.framework,            
+            *
+        </osgi.import>
+        <osgi.export>
+            brooklyn.*,org.apache.brooklyn.*;-noimport:=true
+        </osgi.export>
+        <osgi.private>
+        </osgi.private>
+    </properties>
+    
     <dependencies>
         <dependency>
             <groupId>org.apache.brooklyn</groupId>
@@ -284,6 +298,7 @@
                     </dependency>
                 </dependencies>
             </plugin>
+            <!--
             <plugin>
                 <groupId>org.apache.felix</groupId>
                 <artifactId>maven-bundle-plugin</artifactId>
@@ -307,6 +322,35 @@
                     </instructions>
                 </configuration>
             </plugin>
+            -->
+            <plugin>
+                <groupId>org.apache.felix</groupId>
+                <artifactId>maven-bundle-plugin</artifactId>
+                <extensions>false</extensions>
+                <executions>
+                    <execution>
+                        <id>bundle</id>
+                        <phase>package</phase>
+                        <goals>
+                            <goal>bundle</goal>
+                        </goals>
+                        <configuration>
+                            <instructions>
+                                <Bundle-Name>${project.name}</Bundle-Name>
+                                <Bundle-SymbolicName>${project.groupId}.${project.artifactId}</Bundle-SymbolicName>
+                                <Export-Package>${osgi.export}</Export-Package>
+                                <Import-Package>${osgi.import}</Import-Package>
+                                <DynamicImport-Package>${osgi.dynamic.import}</DynamicImport-Package>
+                                <Private-Package>${osgi.private}</Private-Package>
+                                <Require-Bundle>${osgi.bundles}</Require-Bundle>
+                                <Bundle-Activator>${osgi.activator}</Bundle-Activator>
+                                <Export-Service>${osgi.export.service}</Export-Service>
+                            </instructions>
+                            <classifier>bundle</classifier>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>            
         </plugins>
         <pluginManagement>
             <plugins>
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -35,20 +35,6 @@
         <relativePath>../parent/pom.xml</relativePath>
     </parent>
 
-    <properties>
-        <osgi.import>
-            !android.util*,
-            !org.apache.brooklyn.rt.felix,
-            !org.apache.felix.framework,            
-            *
-        </osgi.import>
-        <osgi.export>
-            brooklyn.*,org.apache.brooklyn.*;-noimport:=true
-        </osgi.export>
-        <osgi.private>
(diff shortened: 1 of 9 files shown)


geomacy:
hi @googlielmo do you know there are pax exam tests for karaf in brooklyn-dist?  If I recall correctly, the tests are there because they are testing the whole of brooklyn (not just brooklyn-server).

googlielmo:
hi @geomacy, yes I'm aware of that; as these are actually unit tests (not integration tests), the idea is to test the `brooklyn-server` features at every build, to spot regressions early on, especially in view of a simplification of dependencies (which is another effort going on in parallel). These being local tests, I think they can co-exist with the broader ones in dist. Wdyt?

geomacy:
@googlielmo I guess that seems sensible, just wondered if you knew about them. thanks

googlielmo:
I should have clarified the (somewhat limited) scope of this PR in the first place, thank you for giving me the occasion to do it @geomacy 🙂

googlielmo:
Rebased on master (not squashed)

karaf/itests/pom.xml
```diff
@@ -0,0 +1,180 @@
...
+
+    <artifactId>brooklyn-features-itests</artifactId>
+    <name>Brooklyn Karaf Tests</name>
+
+    <properties>
+        <osgi.version>6.0.0</osgi.version>
+        <pax.exam.version>4.9.1</pax.exam.version>
+        <pax-url.version>2.5.2</pax-url.version>
+        <servlet.api.version>3.1.0</servlet.api.version>
+        <validation.api.version>1.1.0.Final</validation.api.version>
```
m4rkmckenna: Can you move these to the parent pom

karaf/itests/pom.xml
```diff
@@ -0,0 +1,180 @@
...
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.servicemix.tooling</groupId>
+                <artifactId>depends-maven-plugin</artifactId>
+                <version>1.3</version>
```
m4rkmckenna: Should be parameterised

karaf/features/src/main/feature/feature.xml
```diff
@@ -260,7 +260,7 @@
             https://github.com/square/okhttp/pull/2246
         -->
         <!-- TODO use ServiceMix bundle for okhttp when dependency on okio is fixed there -->
-        <bundle dependency="true">wrap:mvn:com.squareup.okhttp/okhttp/2.2.0$Import-Package=org.apache.http.*;resolution:=optional,android.util.*;resolution:=optional,okio;version=%27[1.2.0,1.3.0)%27,*</bundle>
+        <bundle dependency="true">mvn:org.apache.servicemix.bundles/org.apache.servicemix.bundles.okhttp/2.2.0_2</bundle>
```
m4rkmckenna: Can you explain your reasoning for this ... I remember there being an issue with the servicemix wrapped version

Also version should be parameterised

googlielmo: Hi @m4rkmckenna servicemix accepted my pull request https://github.com/apache/servicemix-bundles/pull/91 so it's safe to use their version now!

googlielmo:
Thanks for the comments @m4rkmckenna. Will address them with a new commit.

karaf/itests/src/test/java/org/apache/brooklyn/karaf/itests/FeatureInstallationTest.java
```diff
@@ -0,0 +1,83 @@
...
+import org.ops4j.pax.exam.karaf.options.LogLevelOption;
+import org.ops4j.pax.exam.options.MavenArtifactUrlReference;
+import org.ops4j.pax.exam.spi.reactors.ExamReactorStrategy;
+import org.ops4j.pax.exam.spi.reactors.PerClass;
+
+
+@RunWith(PaxExam.class)
+@ExamReactorStrategy(PerClass.class)
+public class FeatureInstallationTest extends TestBase {
+
```
googlielmo: Had a look at this with @andreaturli who wrote the first unit tests in this PR.
It's fair to say that at the moment this is quite similar to the `verify` phase of the karaf-maven-plugin, especially as we're using the `PerClass` Exam reactor strategy.
At the same time, yes we can add more tests / assertions to check the health of karaf / brooklyn, good point! Will add some.

aledsage:
@googlielmo (cc @m4rkmckenna) given Marks' comment about seeing value in this, then I'm fine with us merging it and then iterating on it to add more assertions (as we find time to do that). That feels better than having the PR sitting around for longer. Does that sound sensible to you @m4rkmckenna?

@googlielmo can you squash your commits please, and ensure there is a nice description on the commit(s)? (We want the commits descriptions to be understandable when glancing through the history of Brooklyn commits, where we can't easily tell which PR they were in; commit messages like "wip" don't tell us what it's for).

core/pom.xml
```diff
@@ -60,6 +60,7 @@
         <dependency>
             <groupId>com.hierynomus</groupId>
             <artifactId>sshj</artifactId>
+            <version>${sshj.version}</version>
```
aledsage: Do we really need this line? It's defined int he `dependencyManagement` section of `parent/pom.xml`, so we shouldn't have to repeat it here.

karaf/pom.xml
```diff
@@ -34,21 +34,29 @@
...
-    <pax.url.version>2.4.3</pax.url.version>
+    <pax.exam.version>4.9.1</pax.exam.version>
+    <pax.url.version>2.5.2</pax.url.version>
     <tinybundles.version>1.0.0</tinybundles.version>
 
     <!-- feature repositories -->
     <servicemix.version>6.0.0</servicemix.version>
     <reflections.bundle.version>0.9.9_1</reflections.bundle.version> <!-- see reflections.version -->
 
+    <maven.buildhelper.plugin.version>1.10</maven.buildhelper.plugin.version>
```
aledsage: Where is this used?

karaf/pom.xml
```diff
@@ -34,21 +34,29 @@
...
+    <maven.resources.plugin.version>3.0.1</maven.resources.plugin.version>
+    <maven.depends.plugin.version>1.3</maven.depends.plugin.version>
+
     <maven.compiler.source>${java.version}</maven.compiler.source>
     <maven.compiler.target>${java.version}</maven.compiler.target>
+
+    <okhttp.bundle.version>2.2.0_2</okhttp.bundle.version>
```
aledsage: I wonder where this should be defined - should it be in the root `pom.xml`, alongside all the other dependency versions? Or should it be here (closer to where it's used in feature.xml)? I lean towards the former. But no strong feelings.

googlielmo:
@aledsage I squashed and edited the commit messages to something more useful.
If it's okay to you and @m4rkmckenna I second the idea of merging it and then iterating on it to add more assertions.

drigodwin:
I agree that we should merge this and iterate, since there have been no more comments I'm going to do that now. Thanks @googlielmo @m4rkmckenna @aledsage

PR #517, opened 2017-01-13
AbstractMemberTrackingPolicy does not call MEMBER_ADDED on group members when rebinding

sjcorbett:
This means you don't get more `added` events for things that were already in the group. I think this is the expected behaviour.


geomacy:
👍  makes sense, test passes, merging.

PR #518, opened 2017-01-13
Specify the Azure ARM region

drigodwin:
The Azure ARM jclouds provider loads hardware details from all regions unless the one on which we are going to launch is specified.


aledsage:
LGTM. Longer term, it would be nice to generalise this (e.g. should we apply it for all clouds, or would that break some if they use "zone" instead of "region" for example).

Merging now.

PR #519, opened 2017-01-13
Add experimental CORS server support

bostko:
Include org.apache.cxf.rs.security.cors.CrossOriginResourceSharing implementation

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/BrooklynFeatureEnablement.java
+++ b/core/src/main/java/org/apache/brooklyn/core/BrooklynFeatureEnablement.java
@@ -61,6 +61,8 @@ public class BrooklynFeatureEnablement {
     /** whether feeds are automatically registered when set on entities, so that they are persisted */
     public static final String FEATURE_FEED_REGISTRATION_PROPERTY = FEATURE_PROPERTY_PREFIX+".feedRegistration";
 
+    public static final String FEATURE_CORS_CXF_PROPERTY = FEATURE_PROPERTY_PREFIX + ".corsCxfFeature";
+
     public static final String FEATURE_CATALOG_PERSISTENCE_PROPERTY = FEATURE_PROPERTY_PREFIX+".catalogPersistence";
     
     /** whether the default standby mode is {@link HighAvailabilityMode#HOT_STANDBY} or falling back to the traditional
--- a/rest/rest-resources/src/main/java/org/apache/brooklyn/rest/filter/CorsImplSupplierFilter.java
+++ b/rest/rest-resources/src/main/java/org/apache/brooklyn/rest/filter/CorsImplSupplierFilter.java
@@ -1,0 +1,166 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.brooklyn.rest.filter;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Preconditions;
+import com.google.common.reflect.TypeToken;
+import org.apache.brooklyn.api.mgmt.ManagementContext;
+import org.apache.brooklyn.config.ConfigKey;
+import org.apache.brooklyn.config.StringConfigMap;
+import org.apache.brooklyn.core.BrooklynFeatureEnablement;
+import org.apache.brooklyn.core.config.ConfigKeys;
+import org.apache.brooklyn.core.config.ConfigPredicates;
+import org.apache.brooklyn.core.config.ConfigUtils;
+import org.apache.cxf.rs.security.cors.CrossOriginResourceSharingFilter;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.annotation.Nullable;
+import javax.ws.rs.container.ContainerRequestContext;
+import javax.ws.rs.container.ContainerResponseContext;
+import javax.ws.rs.ext.Provider;
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * <p>
+ * Enables support for Cross Origin Resource Sharing (CORS) filtering on requests in BrooklynWebServer.
+ * If enabled, the allowed origins for the CORS headers should be configured
+ * using the <code>brooklyn.experimental.feature.corsCxfFeature.allowedOrigins=[]</code> property.
+ * </p>
+ * <p>
+ * If <code>brooklyn.experimental.feature.corsCxfFeature.allowedOrigins</code> is not is not supplied then allowedOrigins will be a wildcard on all domains.<br>
+ * Not specifying <code>allowedOrigins</code> is strongly discouraged.
+ * </p>
+ * <p>
+ * Currently there is no support for varying these headers on a per-API-resource basis, that is, the same configured headers are applied to all requests.
+ * </p>
+ * <p>
+ * Apache Brooklyn API requests should be exposed to third party web apps with great attention.
+ * </p>
+ * Apache Brooklyn API calls do not use CORS annotations so findResourceMethod is set to false.
+ */
+@Provider
+public class CorsImplSupplierFilter extends CrossOriginResourceSharingFilter {
+    /**
+     * @see CrossOriginResourceSharingFilter#setAllowOrigins(List<String>)
+     */
+    public static final ConfigKey<List<String>> ALLOW_ORIGINS = ConfigKeys.newConfigKey(new TypeToken<List<String>>() {},
+            "allowedOrigins",
+            "List of allowed origins. Access-Control-Allow-Origin header will be returned to client if Origin header in request is matching exactly a value among the list allowed origins. " +
+                    "If AllowedOrigins is empty or not specified then all origins are allowed. " +
+                    "No wildcard allowed origins are supported.",
+            Collections.<String>emptyList());
+    
+    /**
+     * @see CrossOriginResourceSharingFilter#setAllowHeaders(List<String>)
+     */
+    public static final ConfigKey<List<String>> ALLOW_HEADERS = ConfigKeys.newConfigKey(new TypeToken<List<String>>() {},
+            "allowedHeaders",
+            "List of allowed headers for preflight checks.",
+            Collections.<String>emptyList());
+
+    /**
+     * @see CrossOriginResourceSharingFilter#setAllowCredentials(boolean)
+     */
+    public static final ConfigKey<Boolean> ALLOW_CREDENTIALS = ConfigKeys.newBooleanConfigKey(
+            "allowCredentials",
+            "The value for the Access-Control-Allow-Credentials header. If false, no header is added. If true, the\n" +
+                    "     * header is added with the value 'true'. False by default.",
+            false);
+
+    /**
+     * @see CrossOriginResourceSharingFilter#setExposeHeaders(List<String>)
+     */
+    public static final ConfigKey<List<String>> EXPOSE_HEADERS = ConfigKeys.newConfigKey(new TypeToken<List<String>>() {},
+            "exposedHeaders",
+            "A list of non-simple headers to be exposed via Access-Control-Expose-Headers.",
+            Collections.<String>emptyList());
+
+    /**
+     * @see CrossOriginResourceSharingFilter#setMaxAge(Integer)
+     */
+    public static final ConfigKey<Integer> MAX_AGE = ConfigKeys.newIntegerConfigKey(
+            "maxAge",
+            "The value for Access-Control-Max-Age.",
+            null);
+
+    /**
+     * @see CrossOriginResourceSharingFilter#setPreflightErrorStatus(Integer)
+     */
+    public static final ConfigKey<Integer> PREFLIGHT_FAIL_STATUS = ConfigKeys.newIntegerConfigKey(
+            "preflightFailStatus",
+            "Preflight error response status, default is 200.",
+            200);
+
+    public static final ConfigKey<Boolean> BLOCK_CORS_IF_UNAUTHORIZED = ConfigKeys.newBooleanConfigKey(
+            "blockCorsIfUnauthorized",
+            "Do not apply CORS if response is going to be with UNAUTHORIZED status.",
+            false);
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(CorsImplSupplierFilter.class);
+
+    private boolean corsEnabled = false;
+    
+    // For karaf loading
+    public CorsImplSupplierFilter() {}
+
+    @VisibleForTesting
+    public CorsImplSupplierFilter(@Nullable ManagementContext mgmt) {
+        Preconditions.checkNotNull(mgmt,"ManagementContext should be suppplied to CORS filter.");
+        setCorsEnabled(true);
+        setFindResourceMethod(false);
+        setCorsEnabled(BrooklynFeatureEnablement.isEnabled(BrooklynFeatureEnablement.FEATURE_CORS_CXF_PROPERTY));
+        StringConfigMap corsProperties = ConfigUtils.filterForPrefixAndStrip(
+                mgmt.getConfig().submap(ConfigPredicates.nameStartsWith(BrooklynFeatureEnablement.FEATURE_CORS_CXF_PROPERTY + ".")).asMapWithStringKeys(),
+                BrooklynFeatureEnablement.FEATURE_CORS_CXF_PROPERTY + ".");
+        setAllowOrigins(corsProperties.getConfig(ALLOW_ORIGINS));
+        setAllowHeaders(corsProperties.getConfig(ALLOW_HEADERS));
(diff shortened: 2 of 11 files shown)


m4rkmckenna:
@bostko how is the filter configured?

core/src/main/java/org/apache/brooklyn/core/BrooklynFeatureEnablement.java
```diff
@@ -61,6 +61,37 @@
     /** whether feeds are automatically registered when set on entities, so that they are persisted */
     public static final String FEATURE_FEED_REGISTRATION_PROPERTY = FEATURE_PROPERTY_PREFIX+".feedRegistration";
 
+    /**
```
geomacy: I didn't say that CORS was _always_ insecure; used judiciously, it can be a perfectly acceptable mechanism for many use-cases.  Used badly or inappropriately, it can open you up to increased security risks, but the same can be said about lots of things.  I would remove the warnings from this comment, and simply explain what the feature is that is being enabled.  It is up to users to determine whether CORS is applicable in their case and to enable it if so.

core/src/main/java/org/apache/brooklyn/core/BrooklynFeatureEnablement.java
```diff
@@ -61,6 +61,37 @@
...
+     *
+     * It is best when web app communicates just with its own server, not with multiple servers.
+     * It's the Apache Brooklyn server that should be the single point of contact to moderate and control access to the information from the AMP API, which should never be independently exposed to a web UI.
+     * This sort of architecture can give you additional headaches behind proxies, with firewalls, etc.
+     * For another thing, CORS can be used and can be secure enough up to a point if implemented right (and that's not a trivial 'if'),
+     * but it is still an additional attack vector that can be exploited by mitm attacks etc.
+     * In short, the proposed architecture and use of CORS is more complex, less secure,
+     * and more difficult to manage than the alternative of web client ---> Apache Brooklyn Server ----> fan out to backend servers + Apache Brooklyn API + etc.
+     *
+     * Notes by Geoff Macartney.
```
geomacy: Definitely don't blame me! :-)

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/filter/CorsImplSupplierFilter.java
```diff
@@ -0,0 +1,99 @@
...
+import org.slf4j.LoggerFactory;
+
+import javax.annotation.Nullable;
+import javax.ws.rs.container.ContainerRequestContext;
+import javax.ws.rs.container.ContainerResponseContext;
+import javax.ws.rs.ext.Provider;
+import java.util.Collections;
+import java.util.List;
+
+/**
```
geomacy: Why repeat this?  In any case, as above, just get rid of these warnings.

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/filter/CorsImplSupplierFilter.java
```diff
@@ -0,0 +1,99 @@
...
+ */
+@Provider
+public class CorsImplSupplierFilter extends CrossOriginResourceSharingFilter {
+    public static final ConfigKey<List<String>> ALLOWED_ORIGINS = ConfigKeys.newConfigKey(new TypeToken<List<String>>() {}, BrooklynFeatureEnablement.FEATURE_CORS_CXF_PROPERTY + ".allowedOrigins");
+    private static final Logger LOGGER = LoggerFactory.getLogger(CorsImplSupplierFilter.class);
+
+    private static final boolean brooklynFeatureEnabled = BrooklynFeatureEnablement.isEnabled(BrooklynFeatureEnablement.FEATURE_CORS_CXF_PROPERTY);
+    static {
+        if (brooklynFeatureEnabled) {
+            LOGGER.warn("CORS brooklyn feature enabled.");
```
geomacy: info not warn

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/filter/CorsImplSupplierFilter.java
```diff
@@ -0,0 +1,99 @@
...
+        }
+    }
+
+    public CorsImplSupplierFilter(@Nullable ManagementContext managementContext) {
+        setFindResourceMethod(false);
+        if (managementContext != null) {
+            setAllowOrigins(JavaGroovyEquivalents.<List<String>>elvis(managementContext.getConfig().getConfig(ALLOWED_ORIGINS), Collections.<String>emptyList()));
```
geomacy: why use `elvis` here?  Also prefer `ImmutableList.of()` to emptyList.

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/filter/CorsImplSupplierFilter.java
```diff
@@ -0,0 +1,99 @@
...
+ * For another thing, CORS can be used and can be secure enough up to a point if implemented right (and that's not a trivial 'if'),
+ * but it is still an additional attack vector that can be exploited by mitm attacks etc.
+ * In short, the proposed architecture and use of CORS is more complex, less secure,
+ * and more difficult to manage than the alternative of web client ---> Apache Brooklyn Server ----> fan out to backend servers + Apache Brooklyn API + etc.
+ *
+ * Notes by Geoff Macartney.
+ */
+@Provider
+public class CorsImplSupplierFilter extends CrossOriginResourceSharingFilter {
+    public static final ConfigKey<List<String>> ALLOWED_ORIGINS = ConfigKeys.newConfigKey(new TypeToken<List<String>>() {}, BrooklynFeatureEnablement.FEATURE_CORS_CXF_PROPERTY + ".allowedOrigins");
```
geomacy: We should use Karaf's Configuration Admin service to do the configuration in the Karaf case. 

See for example how this is done in [the karaf blueprint](https://github.com/apache/brooklyn-server/blob/master/karaf/init/src/main/resources/OSGI-INF/blueprint/blueprint.xml#L36-L49) and used to [inject configuration into the launcher](https://github.com/apache/brooklyn-server/blob/master/karaf/init/src/main/resources/OSGI-INF/blueprint/blueprint.xml#L84-L94).

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/filter/CorsImplSupplierFilter.java
bostko: It is possible for `ALLOWED_ORIGINS` to be null or empty, if so then I find elvis use full to setting it to an empty list.
`Collections.emptyList()` is used in `CrossOriginResourceSharingFilter` default constructor so I use the same class which is also immutable.

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/filter/CorsImplSupplierFilter.java
neykov: @geomacy I see `osgilauncher.cfg` as configuration done by the distribution creator and `brooklyn.cfg` configured by the user. That's why I think it makes sense to have this as configurable in `brooklyn.cfg` since the origin is something that the user will decide on.

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/filter/CorsImplSupplierFilter.java
```diff
@@ -0,0 +1,79 @@
...
+    private static final boolean brooklynFeatureEnabled = BrooklynFeatureEnablement.isEnabled(BrooklynFeatureEnablement.FEATURE_CORS_CXF_PROPERTY);
+    static {
+        if (brooklynFeatureEnabled) {
+            LOGGER.info("CORS brooklyn feature enabled.");
+        }
+    }
+
+    public CorsImplSupplierFilter(@Nullable ManagementContext managementContext) {
+        setFindResourceMethod(false);
+        if (managementContext != null) {
```
neykov: Why would this be null in normal operation. Better throw instead of silently ignoring the problem.

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/filter/CorsImplSupplierFilter.java
m4rkmckenna: @neykov Disagree the cors configuration should be in a separate PID

As the CORS configuration has the potential to be hundreds of lines long

rest/rest-server/src/test/java/org/apache/brooklyn/rest/CorsFilterLauncherTest.java
```diff
@@ -0,0 +1,168 @@
...
+        useServerForTest(apiLauncher.start());
+    
+        HttpClient client = client();
+        HttpToolResponse response = HttpTool.execAndConsume(client, httpOptionsRequest("server/status", "GET", shouldAllowOrigin));
+        assertNull(response.getHeaderLists().get(HEADER_AC_ALLOW_ORIGIN), "Access Control Header should not be available.");
+        assertOkayResponse(response, "");
+    
+        response = HttpTool.execAndConsume(client, httpOptionsRequest("script/groovy", shouldAllowOrigin, "POST"));
+        assertNull(response.getHeaderLists().get(HEADER_AC_ALLOW_ORIGIN), "Access Control Header should not be available.");
+        assertOkayResponse(response, "");
```
tbouron: You don't actually test that the request is rejected if it comes from another domain

rest/rest-server/src/test/java/org/apache/brooklyn/rest/CorsFilterLauncherTest.java
```diff
@@ -0,0 +1,168 @@
...
+import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertNull;
+
+/**
+ * If brooklyn.experimental.feature.corsCxfFeature.allowedOrigins is not supplied then allowedOrigins will be all domains.
+ */
+public class CorsFilterLauncherTest extends BrooklynRestApiLauncherTestFixture {
+    @Test
+    public void testEnabledCorsSendsBasicAllowResponse() throws IOException {
+        BrooklynFeatureEnablement.enable(BrooklynFeatureEnablement.FEATURE_CORS_CXF_PROPERTY);
```
tbouron: I would break down all tests into single unit test. Will be simpler to see and check what has been tested or not

rest/rest-server/src/test/java/org/apache/brooklyn/rest/CorsFilterLauncherTest.java
```diff
@@ -0,0 +1,168 @@
...
+                        HttpHeaders.CONTENT_TYPE, "application/text"),
+                "return 0;".getBytes());
+        assertAcNotAllowOrigin(response);
+        assertOkayResponse(response, "{\"result\":\"0\"}");
+    }
+    
+    @Test
+    public void testCorsIsDisabled() throws IOException {
+        BrooklynFeatureEnablement.disable(BrooklynFeatureEnablement.FEATURE_CORS_CXF_PROPERTY);
+        final String shouldAllowOrigin = "http://foo.bar.com";
```
tbouron: Same as above

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/filter/CorsImplSupplierFilter.java
neykov: Personal preference to stick to the standard configuration approach to Brooklyn and not split it in smaller files.
Hundreds of lines long sounds like something has gone horribly wrong. I'd worry more about sending it on the network.

rest/rest-server/src/test/java/org/apache/brooklyn/rest/CorsFilterLauncherTest.java
```diff
@@ -0,0 +1,168 @@
...
+import java.net.URI;
+import java.util.List;
+
+import static org.apache.brooklyn.rest.CsrfTokenFilterLauncherTest.assertOkayResponse;
+import static org.apache.cxf.rs.security.cors.CorsHeaderConstants.*;
+import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertNull;
+
+/**
+ * If brooklyn.experimental.feature.corsCxfFeature.allowedOrigins is not supplied then allowedOrigins will be all domains.
```
geomacy: Would it be worth adding a test for that?

core/src/main/java/org/apache/brooklyn/core/BrooklynFeatureEnablement.java
```diff
@@ -61,6 +61,36 @@
     /** whether feeds are automatically registered when set on entities, so that they are persisted */
     public static final String FEATURE_FEED_REGISTRATION_PROPERTY = FEATURE_PROPERTY_PREFIX+".feedRegistration";
 
+    /**
+     * <p>
+     * Enabling CORS is strongly discouraged.
```
bostko: I think it is valuable to have more information.
What others think?

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/filter/CorsImplSupplierFilter.java
geomacy: I don't think the config will be hundreds of lines as it is only the domain of the requesting site that needs to be added to the header, not separate entries for each page that may do the request.  There will only be a handful of domains in most use-cases, I expect.

@neykov for the configuration, I agree up to a point; we probably need to think a bit about recommendations for what sort of configs are most appropriate in brooklyn.cfg, and what are better done in the conventional Karaf way of having a PID for a feature, e.g. for CORS you would have a cfg file such as `org.apache.brooklyn.rest.filter.cors.cfg` with the appropriate config in it.  However, as there is only one property to be configured, I think that is probably overkill for this scenario.

core/src/main/java/org/apache/brooklyn/core/BrooklynFeatureEnablement.java
m4rkmckenna: Remove

geomacy: The point is that you can't tell at this level whether CORS is appropriate or not; that's more an architectural question. Sometimes CORS is a proper solution to a given need to integrate multiple server backends onto a page in a web application (e.g. with a plugin based web application).  Sometimes it is a hack to glue in some REST call that you should really be handling server side.  We can't simply say 'use of CORS is discouraged'.

geomacy:
I think that's all the comments I want to make except that it would be good to have an actual web page test case that demonstrates it working; maybe @m4rkmckenna could help out with that?  

Apart from the particular comments above I'm happy to say this looks good to me.

core/src/main/java/org/apache/brooklyn/core/BrooklynFeatureEnablement.java
neykov: CORS per se is a good thing. When combined with some other features like basic auth could be a terrible idea because the credentials need to be shared with all clients. Much easier for malicious users to extract the credentials from the javascript env and use them any way they like.
Basic auth would be the reason why I'd stick to proxy until we get something smarter like web tokens. 

I like @geomacy's suggestion above, but perhaps with the proxy recommendation added to it.

rest/rest-server/src/test/java/org/apache/brooklyn/rest/CorsFilterLauncherTest.java
geomacy: I don't think it's meant to be rejected, if I understand it right, it is the browser that will discard the response as it won't have the header it needs to bypass the same origin policy?

bostko: @geomacy that's right.

rest/rest-server/src/test/java/org/apache/brooklyn/rest/CorsFilterLauncherTest.java
bostko: You will save only `BrooklynRestApiLauncher apiLauncher` instantiation.
I do not think making a big test method will be simpler to read.
However I welcome suggestions for shortening assertions.

tbouron: I meant, the current test function contains 4 separate tests. I would break those 4 down into their own test function to be more readable.

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/filter/CorsImplSupplierFilter.java
```diff
@@ -0,0 +1,81 @@
...
+            LOGGER.info("CORS brooklyn feature enabled.");
+        }
+    }
+    
+    @VisibleForTesting
+    public CorsImplSupplierFilter(@Nullable ManagementContext mgmt) {
+        setFindResourceMethod(false);
+        Preconditions.checkNotNull(mgmt,"ManagementContext should be suppplied to CORS filter.");
+        setAllowOrigins(JavaGroovyEquivalents.<List<String>>elvis(
+                mgmt.getConfig().getConfig(ALLOWED_ORIGINS), Collections.<String>emptyList()));
```
m4rkmckenna: @bostko Please expose all configuration the filter provides as it is all needed

See https://cxf.apache.org/javadoc/latest/org/apache/cxf/rs/security/cors/CrossOriginResourceSharingFilter.html

neykov: If we start supporting all config options of the filter, then I could see this being its own PID.

bostko: @neykov could you give guidance how this can be done?

bostko:
PR is updated.
To test it in the karaf distribution you need to have `etc/org.apache.brooklyn.rest.filter.cors.cfg`

bostko:
@m4rkmckenna are there comments which I missed to address.

m4rkmckenna:
@bostko One from @geomacy RE javadoc (DRY) ... I think the rest have been addressed

geomacy:
Looks good, and other reviewers approved.  Will merge now.

PR #520, opened 2017-01-16
Limit parallelism of start/stop steps on SoftwareProcess

neykov:
Setting a `ReleaseableLatch` as the latch of a `SoftwareProcess` allows for controlling the behaviour of the start/stop effector.
When `ReleaseableLatch.Factory.maxConcurrency(int)` is used as the value of the latch (either directly or by referncing a `MaxConcurrencySensor`) only the specified maximum number of threads will be able to execute the immediate step the latch is guarding.

It's very similar to `DynamicCluster.MAX_CONCURRENT_CHILD_COMMANDS` but allows for finer-grained control (though is specific to `SoftwareProcess`).

Merge together with https://github.com/apache/brooklyn-library/pull/87.

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/sensor/MaxConcurrencySensor.java
+++ b/core/src/main/java/org/apache/brooklyn/core/sensor/MaxConcurrencySensor.java
@@ -1,0 +1,91 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.brooklyn.core.sensor;
+
+import org.apache.brooklyn.api.mgmt.Task;
+import org.apache.brooklyn.config.ConfigKey;
+import org.apache.brooklyn.core.config.ConfigKeys;
+import org.apache.brooklyn.core.effector.AddSensor;
+import org.apache.brooklyn.core.entity.Entities;
+import org.apache.brooklyn.util.core.config.ConfigBag;
+import org.apache.brooklyn.util.core.task.Tasks;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Can be used as:
+ * <pre>
+ * {@code
+ * brooklyn.initializers:
+ * - type: org.apache.brooklyn.core.sensor.MaxConcurrencySensor
+ *   brooklyn.config:
+ *     name: start-latch-value
+ *     latch.concurrency.max: 10
+ * }
+ *
+ * and is the short hand for:
+ *
+ * <pre>
+ * {@code
+ * brooklyn.initializers:
+ * - type: org.apache.brooklyn.core.sensor.StaticSensor
+ *   brooklyn.config:
+ *     name: start-latch-value
+ *     static.value:
+ *       $brooklyn.object:
+ *         type: org.apache.brooklyn.core.sensor.ReleaseableLatch$Factory
+ *         factoryMethod.name: newMaxConcurrencyLatch
+ *         factoryMethod.args: [10]
+ * }
+ * </pre>
+ */
+public class MaxConcurrencySensor extends AddSensor<ReleaseableLatch> {
+    private static final Logger log = LoggerFactory.getLogger(MaxConcurrencySensor.class);
+
+    public static final ConfigKey<String> SENSOR_TYPE = ConfigKeys.newConfigKeyWithDefault(AddSensor.SENSOR_TYPE, ReleaseableLatch.class.getName());
+    public static final ConfigKey<Integer> MAX_CONCURRENCY = ConfigKeys.newIntegerConfigKey(
+            "latch.concurrency.max",
+            "The maximum number of threads that can execute the step for the latch this sensors is used at, in parallel.",
+            Integer.MAX_VALUE);
+    private Object maxConcurrency;
+
+    public MaxConcurrencySensor(ConfigBag params) {
+        super(params);
+        maxConcurrency = params.getStringKey(MAX_CONCURRENCY.getName());
+    }
+
+    @Override
+    public void apply(@SuppressWarnings("deprecation") final org.apache.brooklyn.api.entity.EntityLocal entity) {
+        super.apply(entity);
+        final Task<ReleaseableLatch> resolveValueTask = DependentConfiguration.maxConcurrency(maxConcurrency);
+
+        class SetValue implements Runnable {
+            @Override
+            public void run() {
+                ReleaseableLatch releaseableLatch = resolveValueTask.getUnchecked();
+                log.debug(this+" setting sensor "+sensor+" to "+releaseableLatch+" on "+entity);
+                entity.sensors().set(sensor, releaseableLatch);
+            }
+        }
+        Task<ReleaseableLatch> setValueTask = Tasks.<ReleaseableLatch>builder().displayName("Setting " + sensor + " on " + entity).body(new SetValue()).build();
+
+        Entities.submit(entity, Tasks.sequential("Resolving and setting " + sensor + " on " + entity, resolveValueTask, setValueTask));
+    }
+
+}
--- a/core/src/main/java/org/apache/brooklyn/core/sensor/MaxConcurrencySensor.java
+++ b/core/src/main/java/org/apache/brooklyn/core/sensor/MaxConcurrencySensor.java
@@ -18,11 +18,14 @@
  */
 package org.apache.brooklyn.core.sensor;
 
+import org.apache.brooklyn.api.entity.EntityInitializer;
 import org.apache.brooklyn.api.mgmt.Task;
+import org.apache.brooklyn.api.sensor.AttributeSensor;
 import org.apache.brooklyn.config.ConfigKey;
 import org.apache.brooklyn.core.config.ConfigKeys;
 import org.apache.brooklyn.core.effector.AddSensor;
 import org.apache.brooklyn.core.entity.Entities;
+import org.apache.brooklyn.core.entity.EntityInternal;
 import org.apache.brooklyn.util.core.config.ConfigBag;
 import org.apache.brooklyn.util.core.task.Tasks;
 import org.slf4j.Logger;
@@ -55,24 +58,29 @@ import org.slf4j.LoggerFactory;
  * }
  * </pre>
  */
-public class MaxConcurrencySensor extends AddSensor<ReleaseableLatch> {
+public class MaxConcurrencySensor implements EntityInitializer {
     private static final Logger log = LoggerFactory.getLogger(MaxConcurrencySensor.class);
 
+    public static final ConfigKey<String> SENSOR_NAME = ConfigKeys.newStringConfigKey("name", "The name of the sensor to create");
     public static final ConfigKey<String> SENSOR_TYPE = ConfigKeys.newConfigKeyWithDefault(AddSensor.SENSOR_TYPE, ReleaseableLatch.class.getName());
     public static final ConfigKey<Integer> MAX_CONCURRENCY = ConfigKeys.newIntegerConfigKey(
             "latch.concurrency.max",
             "The maximum number of threads that can execute the step for the latch this sensors is used at, in parallel.",
             Integer.MAX_VALUE);
+
     private Object maxConcurrency;
+    private String sensorName;
 
     public MaxConcurrencySensor(ConfigBag params) {
-        super(params);
-        maxConcurrency = params.getStringKey(MAX_CONCURRENCY.getName());
+        this.sensorName = params.get(SENSOR_NAME);
+        this.maxConcurrency = params.getStringKey(MAX_CONCURRENCY.getName());
     }
 
     @Override
     public void apply(@SuppressWarnings("deprecation") final org.apache.brooklyn.api.entity.EntityLocal entity) {
-        super.apply(entity);
+        final AttributeSensor<ReleaseableLatch> sensor = Sensors.newSensor(ReleaseableLatch.class, sensorName);
+        ((EntityInternal) entity).getMutableEntityType().addSensor(sensor);
+
         final Task<ReleaseableLatch> resolveValueTask = DependentConfiguration.maxConcurrency(maxConcurrency);
 
         class SetValue implements Runnable {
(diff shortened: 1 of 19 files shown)


core/src/main/java/org/apache/brooklyn/util/core/task/ValueResolverIterator.java
```diff
@@ -0,0 +1,173 @@
+package org.apache.brooklyn.util.core.task;
```
sjcorbett: This file is missing the ASF license header.

core/src/main/java/org/apache/brooklyn/core/sensor/MaxConcurrencySensor.java
```diff
@@ -0,0 +1,91 @@
...
+import org.apache.brooklyn.util.core.task.Tasks;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Can be used as:
+ * <pre>
+ * {@code
+ * brooklyn.initializers:
+ * - type: org.apache.brooklyn.core.sensor.StaticSensor
```
sjcorbett: `StaticSensor` should be `MaxConcurrencySensor`

software/base/src/main/java/org/apache/brooklyn/entity/software/base/SoftwareProcessDriverLifecycleEffectorTasks.java
```diff
@@ -169,7 +173,7 @@ protected String startProcessesAtMachine(final Supplier<MachineLocation> machine
     }
 
     @Override
-    protected void postStartCustom() {
+    protected void postStartCustom(AtomicReference<ReleaseableLatch> startLatchRef) {
```
sjcorbett: There are subclasses of `SoftwareProcessDriverLifecycleEffectorTasks` and `MachineLifecycleEffectorTasks` in brooklyn-library (e.g. https://github.com/apache/brooklyn-library/blob/master/software/cm/ansible/src/main/java/org/apache/brooklyn/entity/cm/ansible/AnsibleLifecycleEffectorTasks.java#L138, which should really have an `@Override` annotation) and elsewhere in the wild so we should at least deprecate the no-arg methods.

sjcorbett:
Compilation fails for me with the following error:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.3:testCompile (default-testCompile) on project brooklyn-software-base: Compilation failure: Compilation failure:
[ERROR] /Users/sam/code/brooklyn.io/brooklyn/brooklyn-server/software/base/src/test/java/org/apache/brooklyn/entity/software/base/SoftwareProcessEntityLatchTest.java:[135,17] reference to configure is ambiguous
[ERROR] both method <V>configure(org.apache.brooklyn.config.ConfigKey<V>,V) in org.apache.brooklyn.api.internal.AbstractBrooklynObjectSpec and method <V>configure(org.apache.brooklyn.config.ConfigKey<V>,org.apache.brooklyn.api.mgmt.Task<? extends V>) in org.apache.brooklyn.api.internal.AbstractBrooklynObjectSpec match
[ERROR] /Users/sam/code/brooklyn.io/brooklyn/brooklyn-server/software/base/src/test/java/org/apache/brooklyn/entity/software/base/SoftwareProcessEntityLatchTest.java:[134,58] incompatible types: org.apache.brooklyn.api.entity.Entity cannot be converted to org.apache.brooklyn.entity.software.base.SoftwareProcessEntityTest.MyService
[ERROR] /Users/sam/code/brooklyn.io/brooklyn/brooklyn-server/software/base/src/test/java/org/apache/brooklyn/entity/software/base/SoftwareProcessEntityLatchTest.java:[169,25] reference to configure is ambiguous
[ERROR] both method <V>configure(org.apache.brooklyn.config.ConfigKey<V>,V) in org.apache.brooklyn.api.internal.AbstractBrooklynObjectSpec and method <V>configure(org.apache.brooklyn.config.ConfigKey<V>,org.apache.brooklyn.api.mgmt.Task<? extends V>) in org.apache.brooklyn.api.internal.AbstractBrooklynObjectSpec match
```
Maven is building with Java 8:
```
$ mvn -v
Apache Maven 3.3.9 (bb52d8502b132ec0a5a3f4c09453c07478323dc5; 2015-11-10T16:41:47+00:00)
Maven home: /opt/apache-maven
Java version: 1.8.0_05, vendor: Oracle Corporation
Java home: /Library/Java/JavaVirtualMachines/jdk1.8.0_05.jdk/Contents/Home/jre
Default locale: en_US, platform encoding: UTF-8
OS name: "mac os x", version: "10.11.6", arch: "x86_64", family: "mac"
```
For what it's worth IntelliJ doesn't like line 135 either.

sjcorbett:
The bulk of these changes look good. It would help me if you gave some context for 112269ae484dcc5283d9cabc8de2e987f013e79f.

neykov:
@sjcorbett It's generalizing `DslDeferredFunctionCall.resolve` to be reused in the newly added code.

When using DSL the actual value we want to resolve could be wrapped in several layers of Suppliers. `ValueResolver` will resolve recursively until it hits a non-supplier. That's not desirable in some situations - we are interested in the intermediate values as well. What the above method did was to resolve iteratively until it hit a certain condition (the value is of a specific type or is non-resolvable).

I needed the same behaviour for the latches changes so I abstracted it in the `ValueResolverIterator`. It will resolve the value one step at a time, returning each step of the iteration or `absent` if there's a resolve error.

I'll add some of this as a javadoc.

software/base/src/main/java/org/apache/brooklyn/entity/software/base/SoftwareProcessDriverLifecycleEffectorTasks.java
neykov: I was wondering how to do that in a backwards way. Couldn't think of a good solution.
If we keep the old signature it won't get called so it's actually worse than getting a compile error.
Could do something with thread locals, calling the deprecated method and passing the argument. Do you think that's something worth doing?

neykov:
Addressed comments, let's discuss API backwards compatibility.

sjcorbett:
Jenkins is failing on `SoftwareProcessEntityLatchTest.testConcurrency` and `testReleaseableLatchBlocks` 

By the way, can we get rid of `maxConcurrentChildCommands` when this is merged? I don't see any reason to keep it. It's annotated `@Beta`.

software/base/src/main/java/org/apache/brooklyn/entity/software/base/SoftwareProcessDriverLifecycleEffectorTasks.java
sjcorbett: We could force subclasses to error at compile time by leaving the removed methods in place and making them `final`. I'd prefer that to silently ignoring the method. Your suggestion of continuing to call the deprecated method seems like the least bad approach.

neykov:
After seeing how many classes override the methods decided to keep the original signatures and deprecate them, wrapping the call with a thread local.
Test failures fixed.

software/base/src/main/java/org/apache/brooklyn/entity/software/base/lifecycle/MachineLifecycleEffectorTasks.java
```diff
@@ -444,15 +460,27 @@ public MachineLocation call() throws NoMachinesAvailableException {
         }
     }
 
-    /** Wraps a call to {@link #preStartCustom(MachineLocation)}, after setting the hostname and address. */
+    /**
+     * Wraps a call to {@link #preStartCustom(MachineLocation)}, after setting the hostname and address.
+     * @deprecated since 0.11.0. Use {@link #preStartAtMachineAsync(Supplier)} instead.
```
sjcorbett: Should this point at `preStartAtMachineAsync(Supplier, AtomicReference)`?

neykov: good spot, updated.

neykov:
Rebased and resolved merge conflicts.

software/base/src/main/java/org/apache/brooklyn/entity/software/base/lifecycle/MachineLifecycleEffectorTasks.java
```diff
@@ -944,14 +1002,19 @@ protected static boolean canStop(StopMode stopMode, boolean isStopped) {
...
     /** 
      * Override to check whether stop can be executed.
      * Throw if stop should be aborted.
      */
-    protected void preStopConfirmCustom() {
+    protected void preStopConfirmCustom(AtomicReference<ReleaseableLatch> stopLatchRef) {
         // Opportunity to block stop() until other dependent components are ready for it
-        Object val = entity().getConfig(SoftwareProcess.STOP_LATCH);
-        if (val != null) log.debug("{} finished waiting for stop-latch {}; continuing...", entity(), val);
+        stopLatchRef.set(waitForLatch(entity(), SoftwareProcess.STOP_LATCH));
```
geomacy: Is it advisable to do this outside the protected method, to avoid possibility of subclasses forgetting to call super(). - i.e. wrap the`protected` method in something `private` that sets the latch reference then calls the protected method.  Perhaps this line should go into `doStopLatching` just before the call to `doStop()`.  Same for the other latches, START etc.

software/base/src/main/java/org/apache/brooklyn/entity/software/base/lifecycle/MachineLifecycleEffectorTasks.java
```diff
@@ -731,15 +774,30 @@ protected void restartChildren(ConfigBag parameters) {
...
     /**
      * As {@link #stop} but calling {@link #suspendAnyProvisionedMachines} rather than
      * {@link #stopAnyProvisionedMachines}.
      */
     public void suspend(ConfigBag parameters) {
-        doStop(parameters, new SuspendAnyProvisionedMachinesTask());
+        doStopLatching(parameters, new SuspendAnyProvisionedMachinesTask());
+    }
+
+    protected void doStopLatching(ConfigBag parameters, Callable<StopMachineDetails<Integer>> stopTask) {
```
geomacy: Despite the name, this doesn't actually aquire the latch - that's left to the `doStop` - see comment on `preStopConfirm` below.

software/base/src/main/java/org/apache/brooklyn/entity/software/base/lifecycle/MachineLifecycleEffectorTasks.java
neykov: Good cleanup, will update.

software/base/src/main/java/org/apache/brooklyn/entity/software/base/lifecycle/MachineLifecycleEffectorTasks.java
neykov: It does now after latest changes :)

core/src/main/java/org/apache/brooklyn/core/sensor/ReleaseableLatch.java
```diff
@@ -0,0 +1,95 @@
...
+            private int permits;
+            private transient final Semaphore sem;
+
+            public MaxConcurrencyLatch(int permits) {
+                this.permits = permits;
+                this.sem = new Semaphore(permits);
+            }
+
+            @Override
+            public void acquire(Entity caller) {
```
neykov: I've designed the API such that it doesn't assume a semaphore below. Could be used in different ways - for example to replicate the `maxConcurrentChildCommands` behaviour. So while it's not used in this particular implementation, could be useful for alternative implementations.

software/base/src/main/java/org/apache/brooklyn/entity/software/base/lifecycle/MachineLifecycleEffectorTasks.java
```diff
@@ -1083,4 +1146,39 @@ protected void clearEntityLocationAttributes(Location machine) {
         entity().sensors().set(Attributes.SUBNET_ADDRESS, null);
     }
 
+    public static ReleaseableLatch waitForLatch(EntityInternal entity, ConfigKey<Boolean> configKey) {
+        Maybe<?> rawValue = entity.config().getRaw(configKey);
+        if (rawValue.isAbsent()) {
+            return ReleaseableLatch.NOP;
+        } else {
+            ValueResolverIterator<Boolean> iter = resolveLatchIterator(entity, rawValue.get(), configKey);
+            Maybe<ReleaseableLatch> releasableLatchMaybe = iter.next(ReleaseableLatch.class);
```
neykov: `next(Class<?> type)` will return the first element instance of `type` - it's doing the looping under the hood.

core/src/main/java/org/apache/brooklyn/core/sensor/ReleaseableLatch.java
```diff
@@ -0,0 +1,95 @@
...
+
+    ReleaseableLatch NOP = new Factory.NopLatch();
+
+    static class Factory {
+        private static class NopLatch extends AbstractReleaseableLatch {
+            @Override public void acquire(Entity caller) {}
+            @Override public void release(Entity caller) {}
+        }
+
+        private static class MaxConcurrencyLatch extends AbstractReleaseableLatch {
```
neykov: Two reasons:
  * Personal preference to inline small classes like this.
  * More importantly to hide the implementation and make it private.

core/src/main/java/org/apache/brooklyn/core/sensor/ReleaseableLatch.java
geomacy: makes sense; I think it probably would be worth adding the comment that the entity MAY be ignored by implementations, and callers shouldn't assume any specific behaviour based on entity, i.e. don't assume you have to match `acquire()`/`release()` on the same entity.

software/base/src/main/java/org/apache/brooklyn/entity/software/base/lifecycle/MachineLifecycleEffectorTasks.java
geomacy: ah cool

core/src/main/java/org/apache/brooklyn/core/sensor/ReleaseableLatch.java
neykov: Updated doc. I'd expect the entity would match on acquire/release - added info for the argument as well.

neykov:
@geomacy, @sam addressed comments.

sjcorbett:
@neykov looks good. I'm going to merge this. Could I ask you to add a couple of lines describing the new feature to the section on latches in the docs (https://brooklyn.apache.org/v/latest/yaml/advanced-example.html#latches)?

One final thought for future work is how to improve behaviour when the user does something they shouldn't do. For example, if someone deployed this (totally not contrived!) example:
```yaml
services:

  brooklyn.initializers:
  - type: org.apache.brooklyn.core.sensor.MaxConcurrencySensor
    brooklyn.config:
      name: my-latch
      latch.concurrency.max: 1
      targetType: org.apache.brooklyn.core.sensor.ReleaseableLatch

  brooklyn.config: 
    launch.command: sleep 2
    checkRunning.command: true
    install.latch: $brooklyn:self().attributeWhenReady("my-latch")
    start.latch: $brooklyn:self().attributeWhenReady("my-latch")
    stop.latch: $brooklyn:self().attributeWhenReady("my-latch")
```
They'll get an application that is stuck starting forever and that can't be stopped either. We could consider having a (long) timeout on the `acquire` or perhaps more interesting would be to have reentrant latches. Some strong wording in the documentation that this is a bad idea would be sufficient for the moment.

neykov:
@sam Addressed above comment in
https://github.com/apache/brooklyn-server/pull/539
https://github.com/apache/brooklyn-docs/pull/146

PR #521, opened 2017-01-16
fix unexpected AddChildrenEffector behaviour and DSL toString/reparse behaviour

ahgittin:
so far just adds tests indicating the ugly behaviour;
next will:
* improve impl to avoid the ugly behaviour
* facilitate evaluation of DSL in configuration parameters


ahgittin:
in the first commit, noticed that some of the DSL `toString()` methods returned non-re-parseable output.  i've tidied most of the `DslComponent.toString()` impls to be consistent, and cleaner.

also added nice newer DSL tests, and support for `entity("ID")` to look at internal ID's

ahgittin:
*ready for review*

camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/DslDeferredFunctionCall.java
```diff
@@ -99,48 +99,81 @@ public Object call() throws Exception {
...
+                Method m = method.get();
+    
+                checkCallAllowed(m);
+    
+                try {
+                    // Value is most likely another BrooklynDslDeferredSupplier - let the caller handle it,
+                    return Maybe.of(Reflections.invokeMethodFromArgs(instance, m, instanceArgs));
+                } catch (IllegalArgumentException | IllegalAccessException | InvocationTargetException e) {
+                    // If the method is there but not executable for whatever reason fail with a fatal error, don't return an absent.
+                    throw Exceptions.propagate(new InvocationTargetException(e, "Error invoking '"+fnName+instanceArgs+"' on '"+instance+"'"));
```
aledsage: Minor: I don't think this should be an `InvocationTargetException` - `e` was not "thrown by an invoked method or constructor". If `e` is an IllegalArgumentException or IllegalAccessException, then it was caused by trying to invoke it; if `e` is an `InvocationTargetException` then it is already wrapped. We probably just want `Exceptions.propagateAnnotated("...", e)`

tbouron:
@aledsage Did you try the blueprint  I posted above? I couldn't make it work, I'm guessing I missed something obvious

camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/DslDeferredFunctionCall.java
```diff
@@ -62,7 +62,7 @@ public DslDeferredFunctionCall(Object o, String fn, List<Object> args) {
     @Override
     public Task<Object> newTask() {
         return Tasks.builder()
-                .displayName("Deferred function call " + object + "." + fnName + "(" + toString(args) + ")")
+                .displayName("Deferred function call " + object + "." + fnName + args + ")")
```
neykov: Is `args` guaranteed to be non-null here? You'd get `fnNamenull` if not.

ahgittin:
@tbouron Yes ... for once it gives a good error message:  `requires either blueprint_type or blueprint_yaml`.  You called it `brooklyn_yaml`!

@aledsage Good comment re `InvocationTargetException`, adjusted that

also fixed the parentheses (with local static methods).

PR #522, opened 2017-01-17
Class renames for advanced-networking vcloud-director

aledsage:
See https://github.com/brooklyncentral/advanced-networking/pull/116


iyovcheva:
Looks good

aledsage:
Thinking about this more, given that brooklyncentral/advanced-networking is not part of Apache Brooklyn, maybe these renames should not be included in core brooklyn. It could be the responsibility of downstream projects to do this (if they are bundling advanced-networking).

Longer term, if we support multiple properties files that are aggregated, then advanced-networking can supply its own renames file.

PR #523, opened 2017-01-17
Fix Azure ARM VM names

drigodwin:
Azure ARM only allows alphanumeric characters in it's naming. The default for Brooklyn is to use `-` characters to split up blocks of alphanumeric chars which breaks this. This adds an exception for `azurecompute-arm` in the same style as VM name length.

Tested with Tomcat 7 on Azure ARM, AWS and OpenStack

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/location/cloud/CloudLocationConfig.java
+++ b/core/src/main/java/org/apache/brooklyn/core/location/cloud/CloudLocationConfig.java
@@ -74,8 +74,8 @@ public interface CloudLocationConfig {
     public static final ConfigKey<Integer> VM_NAME_MAX_LENGTH = ConfigKeys.newIntegerConfigKey(
         "vmNameMaxLength", "Maximum length of VM name", 60);
 
-    public static final ConfigKey<String> VM_NAME_DISALLOWED_PATTERN = ConfigKeys.newStringConfigKey(
-            "vmNameAllowedChars", "The character pattern to remove from a VM name", "[^a-zA-Z0-9\\-_]");
+    public static final ConfigKey<String> VM_NAME_ALLOWED_CHARACTERS = ConfigKeys.newStringConfigKey(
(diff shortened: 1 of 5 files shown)


locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsMachineNamer.java
```diff
@@ -41,4 +42,16 @@ public Integer getCustomMaxNameLength(ConfigBag setup) {
         return null;
     }
 
+    protected String generateNewIdOfLength(ConfigBag setup, int len) {
```
neykov: Missing `@Override`.

core/src/main/java/org/apache/brooklyn/core/location/cloud/CloudLocationConfig.java
```diff
@@ -74,6 +74,9 @@
     public static final ConfigKey<Integer> VM_NAME_MAX_LENGTH = ConfigKeys.newIntegerConfigKey(
         "vmNameMaxLength", "Maximum length of VM name", 60);
 
+    public static final ConfigKey<String> VM_NAME_DISALLOWED_PATTERN = ConfigKeys.newStringConfigKey(
+            "vmNameAllowedChars", "The character pattern to remove from a VM name", "[^a-zA-Z0-9\\-_]");
```
neykov: Name mismatch - `VM_NAME_DISALLOWED_PATTERN`, `vmNameAllowedChars`. Also not obvious it's a regex, figured it out after looking at the default value.

Prefer to have a white-list rather than a black list. I see that's a bit tricky to do with regexes. Could iterate through the characters of the string and match each one against the regex. wdyt?

drigodwin:
Addressed comments and again tested with Tomcat 7 on Azure ARM, AWS and OpenStack

neykov:
Thanks @drigodwin , merging.

PR #524, opened 2017-01-17
Misc improvements

sjcorbett:
A bunch of things I've found useful.

### DIFF
--- a/rest/rest-server/src/main/java/org/apache/brooklyn/rest/filter/BrooklynPropertiesSecurityFilter.java
+++ b/rest/rest-server/src/main/java/org/apache/brooklyn/rest/filter/BrooklynPropertiesSecurityFilter.java
@@ -148,8 +148,13 @@ public class BrooklynPropertiesSecurityFilter implements Filter {
         String authorization = request.getHeader("Authorization");
         if (authorization != null) {
             String userpass = new String(Base64.decodeBase64(authorization.substring(6)));
-            user = userpass.substring(0, userpass.indexOf(":"));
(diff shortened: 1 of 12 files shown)


rest/rest-server/src/main/java/org/apache/brooklyn/rest/filter/BrooklynPropertiesSecurityFilter.java
```diff
@@ -148,8 +148,11 @@ protected boolean authenticate(HttpServletRequest request) {
         String authorization = request.getHeader("Authorization");
         if (authorization != null) {
             String userpass = new String(Base64.decodeBase64(authorization.substring(6)));
-            user = userpass.substring(0, userpass.indexOf(":"));
-            pass = userpass.substring(userpass.indexOf(":") + 1);
+            int idxColon = userpass.indexOf(":");
+            if (idxColon >= 0) {
+                user = userpass.substring(0, idxColon);
+                pass = userpass.substring(idxColon + 1);
+            }
```
geomacy: previously you'd have got an index-out-of-bounds exception if the request had an invalid auth string without a colon (not likely perhaps but not impossible);  like this it will just leave user and pass empty and try to authenticate with them.  Would it be worth throwing a more explanatory exception here?

sjcorbett: Good point. It should just return false (i.e. unauthorised).

geomacy:
👍  lgtm, merging.

PR #525, opened 2017-01-17
openstack-swift support

aledsage:


locations/jclouds/src/main/java/org/apache/brooklyn/core/mgmt/persist/jclouds/JcloudsBlobStoreBasedObjectStore.java
```diff
@@ -36,6 +36,7 @@
 import org.jclouds.blobstore.options.ListContainerOptions;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.apache.brooklyn.location.jclouds.BlobStoreContextFactoryImpl;
```
tbouron: You have some unused imports in the class since you delete fews lines

aledsage:
Thanks @tbouron - unused imports removed, rebased against master, and waiting for jenkins to re-run the tests. Then will merge.

tbouron:
@aledsage Jenkins is happy now 👍

PR #526, opened 2017-01-18
Add removeSecurityGroup by id

geomacy:
A small enhancement to support use cases where you have the id of a group you want to delete; this addition allows you to delete the group with just the id, rather than have an extra round-trip to fetch the group so you can then delete it.


sjcorbett:
Looks sensible. It would be good to see some tests for SecurityGroupEditor in a subsequent pull request.

PR #527, opened 2017-01-19
Adds test for catalog-effector from OSGi bundle

aledsage:


sjcorbett:
Looks good. Would be nice to be able to make an assertion on the effect of the effector, but perhaps that's one for another time.

PR #528, opened 2017-01-19
BROOKLYN-426: avoid propagating NoClassDefFoundError

aledsage:
See https://issues.apache.org/jira/browse/BROOKLYN-426 for details.

Tested this with a badly behaving bundle, and confirmed that we were able to list the catalog contents when this fix is used.


core/src/main/java/org/apache/brooklyn/util/core/LoaderDispatcher.java
```diff
@@ -44,6 +44,10 @@
                 return Maybe.<Class<?>>of(SystemFrameworkLoader.get().loadClassFromBundle(className, bundle));
             } catch (ClassNotFoundException e) {
                 return Maybe.absent("Failed to load class " + className + " from bundle " + bundle, e);
+            } catch (NoClassDefFoundError e) {
```
m4rkmckenna: **MINOR**
```
} catch (ClassNotFoundException | NoClassDefFoundError  e) {
    return Maybe.absent("Failed to load class " + className + " from bundle " + bundle, e);
}
```

PR #529, opened 2017-01-19
ConnectivityResolver

sjcorbett:
`ConnectivityResolver` is a special customiser used by `JcloudsLocation` to determine the address and credentials with which a location should be contacted. By default it should behave as the previous logic in `JcloudsLocation`, but it exposes options to have Brooklyn prefer to contact VMs on private addresses and can be injected on a per-entity basis. For example:

```yaml
services:
- type: server
  location: the-same-private-network-as-brooklyn
  brooklyn.initializers:
  - type: org.apache.brooklyn.location.jclouds.DefaultConnectivityResolver
    brooklyn.config:
      mode: ONLY_PRIVATE
- type: server
  location: another-cloud
  # implicit use of the default network info customizer.
```

Would result in the first entity being managed on the instance's private address (and deployment failing if this was not possible) and the second being managed on its public address. Graceful fallback is possible by replacing `ONLY_PRIVATE` with `PREFER_PRIVATE`. There are `PUBLIC` variants of each of these.

These changes were prompted by a thread on the Brooklyn mailing list last December: http://markmail.org/thread/xpr6nsyimv7goewy. Some of the items discussed in that thread have been left for future work. It depends on PRs #497 and #524.

~~I'd like opinions on:~~
* ~~naming: "BasicLocationNetworkInfoCustomizer" is cumbersome.~~
* ~~whether the class should extend `BasicJcloudsLocationCustomizer`. I can see it being useful to anyone that wants to subclass BLNIC but at the moment it's totally unused.~~
* ~~whether we should include the `publishNetworks` option right now. It controls the publication of `host.address.{public,private}.n` sensors. The idea is good but the feature feels a bit half-baked right now.~~
* ~~whether `ManagementAddressResolveOptions` is the right way to convey information between the location and the customiser. Again, it's a bit ungainly.~~

### DIFF
--- a/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
+++ b/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
@@ -2693,15 +2693,21 @@ public class JcloudsLocation extends AbstractCloudMachineProvisioningLocation im
         for (LoginCredentials creds : credentialsToTry) {
             machinesToTry.put(createTemporarySshMachineLocation(hostAndPort, creds, sshProps), creds);
         }
+        final Duration repeaterTimeout = timeout;
         try {
             Callable<Boolean> checker = new Callable<Boolean>() {
                 @Override
                 public Boolean call() {
                     for (Map.Entry<SshMachineLocation, LoginCredentials> entry : machinesToTry.entrySet()) {
                         SshMachineLocation machine = entry.getKey();
+                        Duration statusTimeout = Duration.THIRTY_SECONDS.isShorterThan(repeaterTimeout)
+                                ? Duration.THIRTY_SECONDS
+                                : repeaterTimeout;
                         int exitstatus = machine.execScript(
                                 ImmutableMap.of(
-                                        SshTool.PROP_SSH_TRIES_TIMEOUT.getName(), Duration.THIRTY_SECONDS.toMilliseconds(),
+                                        SshTool.PROP_CONNECT_TIMEOUT.getName(), statusTimeout.toMilliseconds(),
+                                        SshTool.PROP_SESSION_TIMEOUT.getName(), statusTimeout.toMilliseconds(),
+                                        SshTool.PROP_SSH_TRIES_TIMEOUT.getName(), statusTimeout.toMilliseconds(),
                                         SshTool.PROP_SSH_TRIES.getName(), 1),
                                 "check-connectivity",
                                 ImmutableList.of("true"));
--- a/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
+++ b/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
@@ -430,6 +430,11 @@ public class JcloudsLocation extends AbstractCloudMachineProvisioningLocation im
         return result;
     }
 
+    public ConnectivityResolver getLocationNetworkInfoCustomizer(ConfigBag setup) {
+        ConnectivityResolver configured = setup.get(CONNECTIVITY_RESOLVER);
+        return (configured != null) ? configured : new DefaultConnectivityResolver();
+    }
+
     protected Collection<MachineLocationCustomizer> getMachineCustomizers(ConfigBag setup) {
         Collection<MachineLocationCustomizer> customizers = setup.get(MACHINE_LOCATION_CUSTOMIZERS);
         return (customizers == null ? ImmutableList.<MachineLocationCustomizer>of() : customizers);
@@ -626,18 +631,46 @@ public class JcloudsLocation extends AbstractCloudMachineProvisioningLocation im
         }
     }
 
+    protected ConnectivityResolverOptions.Builder getConnectivityOptionsBuilder(ConfigBag setup, boolean isWindows) {
+        boolean waitForSshable = !"false".equalsIgnoreCase(setup.get(JcloudsLocationConfig.WAIT_FOR_SSHABLE));
+        boolean waitForWinRmable = !"false".equalsIgnoreCase(setup.get(JcloudsLocationConfig.WAIT_FOR_WINRM_AVAILABLE));
+        boolean waitForConnectable = isWindows ? waitForWinRmable : waitForSshable;
+
+        boolean usePortForwarding = setup.get(JcloudsLocationConfig.USE_PORT_FORWARDING);
+        boolean skipJcloudsSshing = usePortForwarding ||
+                Boolean.FALSE.equals(setup.get(JcloudsLocationConfig.USE_JCLOUDS_SSH_INIT));
+
+        ConnectivityResolverOptions.Builder builder = ConnectivityResolverOptions.builder()
+                .waitForConnectable(waitForConnectable)
+                .usePortForwarding(usePortForwarding)
+                .skipJcloudsSshing(skipJcloudsSshing);
+
+        String pollForFirstReachable = setup.get(JcloudsLocationConfig.POLL_FOR_FIRST_REACHABLE_ADDRESS);
+        boolean pollEnabled = !"false".equalsIgnoreCase(pollForFirstReachable);
+
+        if (pollEnabled) {
+            Predicate<? super HostAndPort> reachableAddressesPredicate = getReachableAddressesPredicate(setup);
+            Duration pollTimeout = "true".equals(pollForFirstReachable)
+                                   ? Duration.FIVE_MINUTES
+                                   : Duration.of(pollForFirstReachable);
+            builder.pollForReachableAddresses(reachableAddressesPredicate, pollTimeout, true);
+        }
+        return builder;
+    }
+
     protected MachineLocation obtainOnce(ConfigBag setup) throws NoMachinesAvailableException {
         AccessController.Response access = getManagementContext().getAccessController().canProvisionLocation(this);
         if (!access.isAllowed()) {
             throw new IllegalStateException("Access controller forbids provisioning in "+this+": "+access.getMsg());
         }
 
-        boolean waitForSshable = !"false".equalsIgnoreCase(setup.get(WAIT_FOR_SSHABLE));
-        boolean waitForWinRmable = !"false".equalsIgnoreCase(setup.get(WAIT_FOR_WINRM_AVAILABLE));
-        boolean usePortForwarding = setup.get(USE_PORT_FORWARDING);
-        boolean skipJcloudsSshing = Boolean.FALSE.equals(setup.get(USE_JCLOUDS_SSH_INIT)) || usePortForwarding;
+        Predicate<? super HostAndPort> reachablePredicate = getReachableAddressesPredicate(setup);
+        ConnectivityResolverOptions options = getConnectivityOptionsBuilder(setup, false).build();
+
+        // FIXME How do we influence the node.getLoginPort, so it is set correctly for Windows?
+        // Setup port-forwarding, if required
         JcloudsPortForwarderExtension portForwarder = setup.get(PORT_FORWARDER);
-        if (usePortForwarding) checkNotNull(portForwarder, "portForwarder, when use-port-forwarding enabled");
+        if (options.usePortForwarding()) checkNotNull(portForwarder, "portForwarder, when use-port-forwarding enabled");
 
         final ComputeService computeService = getComputeService(setup);
         CloudMachineNamer cloudMachineNamer = getCloudMachineNamer(setup);
@@ -676,12 +709,14 @@ public class JcloudsLocation extends AbstractCloudMachineProvisioningLocation im
                 // Setup the template
                 template = buildTemplate(computeService, setup, customizers);
                 boolean expectWindows = isWindows(template, setup);
-                if (!skipJcloudsSshing) {
+                if (!options.skipJcloudsSshing()) {
                     if (expectWindows) {
                         // TODO Was this too early to look at template.getImage? e.g. customizeTemplate could subsequently modify it.
                         LOG.warn("Ignoring invalid configuration for Windows provisioning of "+template.getImage()+": "+USE_JCLOUDS_SSH_INIT.getName()+" should be false");
-                        skipJcloudsSshing = true;
-                    } else if (waitForSshable) {
+                        options = options.toBuilder()
+                                .skipJcloudsSshing(true)
+                                .build();
+                    } else if (options.waitForConnectable()) {
                         userCredentials = initTemplateForCreateUser(template, setup);
                     }
                 }
@@ -723,7 +758,6 @@ public class JcloudsLocation extends AbstractCloudMachineProvisioningLocation im
                 throw new IllegalStateException("No nodes returned by jclouds create-nodes in " + getCreationString(setup));
 
             boolean windows = isWindows(node, setup);
-            boolean waitForConnectable = (windows) ? waitForWinRmable : waitForSshable;
 
             if (windows) {
                 int newLoginPort = node.getLoginPort() == 22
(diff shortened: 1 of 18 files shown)


geomacy:
Will have a look through this now, however, on the points for opinion above, my 2¢ worth is:

- naming: "Customizer" isn't really needed (we don't insist on e.g. VanillaSoftwareProcessEntity), and this class isn't really "Basic" (which to me refers to the "default no-op" behaviour of `BasicJcloudsLocationCustomizer`). Also "Location" doesn't add much to the description, when you already have "Network".  The role of the class is about choosing which network you want, not really about "Info" as such so how about `org.apache.brooklyn.location.jclouds.NetworkPreference`.
- what to extend: I suppose another way to look at it is, "if someone changes behaviour in `BasicJcloudsLocationCustomizer` is that something we would want this to inherit"?   No strong feeling but I tend to think that we would want that, so it should extend it.
- `publishNetworks` - I'd say prefer leaving things as separate as possible, so would suggest leaving that to a separate PR.
- resolve options - will have a look.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/BasicLocationNetworkInfoCustomizer.java
```diff
@@ -0,0 +1,473 @@
...
+     *     <li>If the machine is connectable and pollForFirstReachableAddress is set in options then use all
+     *     {@link #getReachableAddresses reachable} addresses.</li>
+     *     <li>Use the first address that is resolvable with {@link #isAddressResolvable}.</li>
+     *     <li>Use the first address in the node's public then private addresses.</li>
+     * </ol>
+     */
+    protected Iterable<HostAndPort> getManagementCandidates(
+            JcloudsLocation location, NodeMetadata node, ConfigBag config, ManagementAddressResolveOptions options) {
+        final Optional<HostAndPort> hostAndPortOverride = options.getHostAndPortOverride();
+        boolean lookupAwsHostname = Boolean.TRUE.equals(config.get(JcloudsLocation.LOOKUP_AWS_HOSTNAME));
```
geomacy: It would be nice to factor out the AWS specifics into a separate tool; I know this has come from the existing JcloudsLocation code but as this is a new class maybe there is an opportunity to avoid repeating this pattern?

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -1738,97 +1740,6 @@ protected LoginCredentials initTemplateForCreateUser(Template template, ConfigBa
...
-                return HostAndPort.fromParts(address, defaultPort);
-            }
-        }
-        LOG.warn("No resolvable address in {} ({}/{}); using first; may cause future failures",
-                new Object[]{addresses, getCreationString(config), node});
-        String host = Iterables.get(addresses, 0);
-        return HostAndPort.fromParts(host, defaultPort);
-    }
-
     private boolean isAddressResolvable(String addr) {
```
geomacy: This can be deleted now.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/BasicLocationNetworkInfoCustomizer.java
geomacy: p.s. I don't mean this PR _requires_ this change as such; but it might be worth adding a "TODO" to highlight the point.

geomacy:
I tested this successfully:  put a brooklyn on ibm-bluebox-lon-pub, and used it to deploy the blueprint 
```
location: ibm-bluebox-lon-pub
services:
- type: server
  name: custom
  brooklyn.initializers:
  - type: org.apache.brooklyn.location.jclouds.BasicLocationNetworkInfoCustomizer
    brooklyn.config:
      mode: ONLY_PRIVATE
- type: server
  name: default
  # implicit use of the default network info customizer.
```

i.e. both servers are deployed to the same location as brooklyn, which is configured to create a public address (`jclouds.openstack-nova.auto-create-floating-ips: true`).

The server 'custom' above was assigned a host address on the private network, while the 'default' one was assigned it on the public network:
```
$ for ename in custom default ; do printf "%s " ${ename} ;  br app app1 ent ${ename} sensor host.address ; done
custom 10.101.1.232
default 169.50.81.104
```

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -625,6 +630,30 @@ public MachineLocation obtain(Map<?,?> flags) throws NoMachinesAvailableExceptio
         }
     }
 
+    protected ManagementAddressResolveOptions getManagementAddressResolveOptions(
+            NodeMetadata node, ConfigBag setup, Optional<HostAndPort> sshHostAndPortOverride, Optional<LoginCredentials> userCredentials) {
+        boolean waitForSshable = !"false".equalsIgnoreCase(setup.get(WAIT_FOR_SSHABLE));
```
sjcorbett: The config keys it needs are only available as arguments to certain methods (`obtain`, primarily) so we can't pull it up to the constructor.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -2607,24 +2522,32 @@ public boolean apply(@Nullable WinRmMachineLocation machine) {
...
+        // Handle where jclouds gives us the wrong login user (!) and both a password + ssh key.
+        // Try all the permutations to find the one that works.
+        Iterable<LoginCredentials> credentialsToTry = generateCredentials(node.getCredentials(), setup.get(LOGIN_USER));
+        return waitForSshable(computeService, node, managementHostAndPort, credentialsToTry, setup);
+    }
+
+    /** @deprecated Since 0.11.0. Use {@link #waitForSshableGuessCredentials} instead. */
+    @Deprecated
+    protected LoginCredentials waitForSshable(ComputeService computeService, NodeMetadata node, HostAndPort managementHostAndPort, ConfigBag setup) {
+        return waitForSshable(computeService, node, managementHostAndPort, setup);
```
sjcorbett: Good spot.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/BasicLocationNetworkInfoCustomizer.java
```diff
@@ -0,0 +1,472 @@
...
+                       : defaultPort;
+            final HostAndPort override = HostAndPort.fromParts(hostAndPortOverride.get().getHostText(), port);
+            LOG.debug("Using host and port override for management candidates of {}: {}", location, override);
+            return ImmutableList.of(override);
+        }
+
+        // Treat AWS as a special case because the DNS fully qualified hostname in AWS is
+        // (normally?!) a good way to refer to the VM from both inside and outside of the region.
+        // TODO This is a bit weird: if the statement below is true then getHostnameAws will find the first
+        // reachable address, which repeats if case after this one.
```
neykov: So it logins to the machine to find which IP to login to :).
Note that this is different from the next case. It similarly fetches the first reachable IP, but then connects to it and executes a `curl` command, returning the `hostname` value.
Would make more sense to move this after the code which already figured out the correct IPs and use them to fetch the hostname. I'd expect this to be available from EC2 API so in future could improve it to fetch it directly through jclouds.
For amazon it would return the `hostname`. Wonder how it interacts with `PREFER_PRIVATE`. Seems the return value gets coerced to `InetAddress` (which will resolve the hostname from the POV of the Brooklyn instance) and set on the `MachineLocation`.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/BasicLocationNetworkInfoCustomizer.java
```diff
@@ -0,0 +1,472 @@
...
+
+        Iterable<HostAndPort> managementCandidates = getManagementCandidates(location, node, config, options);
+        Iterable<LoginCredentials> credentialCandidates = getCredentialCandidates(location, node, options, config);
+
+        // Try each pair of address and credential until one succeeds.
+        if (options.expectReachable() && options.pollForFirstReachableAddress() && shouldCheckCredentials()) {
+            for (HostAndPort hap : managementCandidates) {
+                for (LoginCredentials cred : credentialCandidates) {
+                    LOG.trace("Testing host={} with credential={}", hap, cred);
+                    if (checkCredential(location, hap, cred, config, options.isWindows())) {
```
neykov: The check is still vulnerable to the local network IP poisoning. The reachability check will succeed immediately for the local IP, not giving enough time for the public IP to become available. And then the credential check will fail.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -2610,24 +2514,32 @@ public boolean apply(@Nullable WinRmMachineLocation machine) {
...
         if (Strings.isNonBlank(loginUserOverride)) {
             users.add(loginUserOverride);
         }
-
-        // See https://issues.apache.org/jira/browse/BROOKLYN-186
-        // Handle where jclouds gives us the wrong login user (!) and both a password + ssh key.
-        // Try all the permutations to find the one that works.
-        List<LoginCredentials> credentialsToTry = Lists.newArrayList();
+        List<LoginCredentials> credentialsToTry = new ArrayList<>();
+        Lists.newArrayList();
```
neykov: Remove.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/BasicLocationNetworkInfoCustomizer.java
```diff
@@ -0,0 +1,472 @@
...
+        if (options.skipJcloudsSshing() && options.expectReachable()) {
+            if (options.isWindows()) {
+                return ImmutableList.of(options.getInitialCredentials());
+            } else {
+                return location.generateCredentials(node.getCredentials(), setup.get(JcloudsLocationConfig.LOGIN_USER));
+            }
+        }
+
+        // 1. Were they configured by the user?
+        LoginCredentials customCredentials = setup.get(JcloudsLocation.CUSTOM_CREDENTIALS);
```
neykov: `CUSTOM_CREDENTIALS` is actually in `JcloudsLocationConfig` and `JcloudsLocation` extends it - better to reference statics with the class that declares them. Here and elsewhere in this class.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/BasicLocationNetworkInfoCustomizer.java
```diff
@@ -0,0 +1,472 @@
...
+            if (hit.hasNext()) HostAndPort.fromHost(hit.next());
+        }
+        if (hapChoice == null) {
+            throw new IllegalStateException("Exhausted all options when determining address for " + location);
+        }
+
+        if (credChoice == null) {
+            credChoice = Iterables.getFirst(credentialCandidates, null);
+            if (credChoice == null) {
+                throw new IllegalStateException("Exhausted all options when determining credential for " + location);
```
neykov: A more appropriate message would be `No credentials configured for location`. Since `credentialCandidates` is empty.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/BasicLocationNetworkInfoCustomizer.java
```diff
@@ -0,0 +1,472 @@
...
+            LOG.trace("Choosing first management candidate given node={} and mode={}", node, getMode());
+            hapChoice = Iterables.getFirst(managementCandidates, null);
+        }
+        if (hapChoice == null) {
+            LOG.trace("Choosing first address of node={} in mode={}", node, getMode());
+            final Iterator<String> hit = getNodeAddressesWithMode(node).iterator();
+            if (hit.hasNext()) HostAndPort.fromHost(hit.next());
+        }
+        if (hapChoice == null) {
+            throw new IllegalStateException("Exhausted all options when determining address for " + location);
```
neykov: Would happen only when jclouds returns no IP addresses at all so could amend with "jclouds did not return any IP addresses matching " + getMode()

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/LocationNetworkInfoCustomizer.java
```diff
@@ -0,0 +1,46 @@
...
+import com.google.common.annotations.Beta;
+
+/**
+ * LocationNetworkInfoCustomizers are used by {@link JcloudsLocation} to determine the host and port
+ * on which connections to locations should be made and the credentials that should be used.
+ *
+ * @see JcloudsLocationConfig#LOCATION_NETWORK_INFO_CUSTOMIZER
+ */
+@Beta
+public interface LocationNetworkInfoCustomizer extends JcloudsLocationCustomizer {
```
neykov: Don't think this should be a `JcloudsLocationCustomizer`. Yes it customises the behaviour of the location. My thinking is that `JcloudsLocationCustomizer` is not properly named though - it's too generic and anything altering the behaviour of the location matches it. Something better describing it would be `LocationLifecycleCallback`.
To back the above - you are not implementing any of the methods of the customiser.

The primary task of this interface is to return the addresses we can ssh into. A name more fitting in this description would be `LocationConnectivityResolver`, `LocationAddressResolver`.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -429,6 +429,11 @@ protected CloudMachineNamer getCloudMachineNamer(ConfigBag config) {
         return result;
     }
 
+    public LocationNetworkInfoCustomizer getLocationNetworkInfoCustomizer(ConfigBag setup) {
+        LocationNetworkInfoCustomizer configured = setup.get(LOCATION_NETWORK_INFO_CUSTOMIZER);
```
neykov: I think a config key is suitable for this. Just FYI there is the [`extension`](https://github.com/apache/brooklyn-server/blob/master/core/src/main/java/org/apache/brooklyn/core/location/AbstractLocation.java#L729-L749) API as well. We could improve it in future and use it to customise the behaviour of the location with pluggable components, leaving the keys for (declarative) configuration.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -752,58 +783,32 @@ protected MachineLocation obtainOnce(ConfigBag setup) throws NoMachinesAvailable
...
-                userCredentials = LoginCredentials.fromCredentials(node.getCredentials());
+                userCredentials = creds;
             }
+
             // store the credentials, in case they have changed
             putIfPresentButDifferent(setup, JcloudsLocationConfig.PASSWORD, userCredentials.getOptionalPassword().orNull());
             putIfPresentButDifferent(setup, JcloudsLocationConfig.PRIVATE_KEY_DATA, userCredentials.getOptionalPrivateKey().orNull());
 
             // Wait for the VM to be reachable over SSH
+            // TODO: this has already been tested by locationnetworkinfocustomizer
```
neykov: Guard it with with a `!CHECK_CREDENTIALS` then?

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -2476,37 +2382,10 @@ protected String getFirstReachableAddress(NodeMetadata node, ConfigBag setup) {
         String result;
```
neykov: Don't think the method `getFirstReachableAddress` is still referenced. Mark it deprecated.

neykov:
* naming
  * I think a more appropriate name for the interface would be `ConnectivityResolver` and for the implementation `DefaultConnectivityResolver` (could even drop the location part).
* whether the class should extend BasicJcloudsLocationCustomizer:
  * No
* whether we should include the publishNetworks option right now:
  * it's a nice to have feature, though not sure the current format is optimal, see comments above. If in doubt could include it disabled by default.
* whether ManagementAddressResolveOptions is the right way to convey information between the location and the customiser: 
  * My first thought was that `JcloudsLocation`  + `config` should be enough, potentially including `credentialsOverride`. Why include both `config` and `options` if it can be reconstructed on the other side.  I favour minimal interfaces which don't duplicate data. But on the other hand I'm happy to leave as is, especially if the bean is reused at other places. Could clean it up in future.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
sjcorbett: Think I misunderstood you here. Were you referring to the constructor for `ManagementAddressResolveOptions`?

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/BasicLocationNetworkInfoCustomizer.java
sjcorbett: Isn't this an intentional trade-off per the discussion about grace periods in `ReachableSocketFinder` in https://github.com/apache/brooklyn-server/pull/497?

sjcorbett:
@geomacy @neykov thanks for your thorough comments. Please see 728fb46 for changes addressing your suggestions and c9fcf15 for a simplification of `ReachableSocketFinder`.

geomacy:
looks good, will merge

PR #530, opened 2017-01-19
JcloudsLocation is not releasing its customizers (BROOKLYN-427)

geomacy:
preRelease and postRelease of JcloudsLocationCustomizers are not being 
called as expected during the release of a Jclouds machine location.
The problem is that the "setup" used in JcloudsLocation.obtainOnce includes a copy of the flags passed from MachineLifecycleEffectorTasks#stopAnyProvisionedMachines,
which includes the customizers. However, the "setup" used in the "release"
method of JcloudsLocation is taken from the JcloudsLocation itself, which
is the provisioning (i.e. parent) location, which doesn't have those 
customizers in it.
It would actually make more sense to get the location customizers from
the machine location, as a given parent could have mutiple machines
with different customizers.

The changes here include an addition to the
`JcloudsCustomizerInstantiationYamlDslTest#testCustomizers`
to demonstrate the problem, and a fix that adds the config to 
the machine location.

### DIFF
--- a/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
+++ b/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
@@ -72,6 +72,7 @@ import org.apache.brooklyn.core.location.cloud.AbstractCloudMachineProvisioningL
 import org.apache.brooklyn.core.location.cloud.AvailabilityZoneExtension;
 import org.apache.brooklyn.core.location.cloud.names.AbstractCloudMachineNamer;
 import org.apache.brooklyn.core.location.cloud.names.CloudMachineNamer;
+import org.apache.brooklyn.core.location.internal.LocationInternal;
 import org.apache.brooklyn.core.mgmt.internal.LocalLocationManager;
 import org.apache.brooklyn.core.mgmt.persist.LocationWithObjectStore;
 import org.apache.brooklyn.core.mgmt.persist.PersistenceObjectStore;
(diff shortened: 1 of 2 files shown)


locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -2219,7 +2227,7 @@ public void release(MachineLocation rawMachine) {
 
         Exception tothrow = null;
 
-        ConfigBag setup = config().getBag();
+        ConfigBag setup = ((LocationInternal)machine).config().getBag();
```
neykov: Is the cast needed?

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -1065,6 +1066,13 @@ protected MachineLocation obtainOnce(ConfigBag setup) throws NoMachinesAvailable
                 LOG.warn("Problem generating log message summarising completion of jclouds machine provisioning "+machineLocation+" by "+this, e);
             }
 
+            if (customizers.size() > 0) {
+                machineLocation.config().set(JCLOUDS_LOCATION_CUSTOMIZERS, customizers);
+            }
+            if (machineCustomizers.size() > 0) {
+                machineLocation.config().set(MACHINE_LOCATION_CUSTOMIZERS, machineCustomizers);
+            }
```
neykov: Can you instead add the config for `JCLOUDS_LOCATION_CUSTOMIZERS` at [construction time](https://github.com/apache/brooklyn-server/blob/master/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java#L2056-L2099) so migrated settings are all at one place.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
geomacy: IIRC it is needed to be able to do the `getBag()`

geomacy:
Addressed review [comment](https://github.com/apache/brooklyn-server/pull/530#discussion_r97329773) "Can you instead add the config for JCLOUDS_LOCATION_CUSTOMIZERS..."

geomacy:
test failure unrelated?

geomacy:
let's try again

neykov:
Thanks for updates, merging.

PR #531, opened 2017-01-20
initial work to support HttpEntity

andreaturli:
- add HttpCommnadEffector
- add CompositeEffector
- add EntityInitializers util class to resolve DSL injected as params
  into the HttpCommandEffector

---

It is based on https://github.com/apache/brooklyn-server/pull/152 which was not merged as it didn't handle rebind properly, happy to sort it out in this PR but I've not touched rebinding yet.

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/effector/CompositeEffector.java
+++ b/core/src/main/java/org/apache/brooklyn/core/effector/CompositeEffector.java
@@ -18,10 +18,11 @@
  */
 package org.apache.brooklyn.core.effector;
 
+import static org.apache.brooklyn.core.entity.trait.Startable.START;
+import static org.apache.brooklyn.core.entity.trait.Startable.STOP;
+
 import java.util.List;
 import java.util.Map;
-
-import javax.annotation.Nullable;
 
 import org.apache.brooklyn.api.effector.Effector;
 import org.apache.brooklyn.api.entity.EntityLocal;
@@ -36,11 +37,8 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.google.common.annotations.Beta;
-import com.google.common.base.Optional;
 import com.google.common.base.Preconditions;
-import com.google.common.base.Predicate;
 import com.google.common.collect.ImmutableList;
-import com.google.common.collect.Iterables;
 import com.google.common.collect.Lists;
 import com.google.common.reflect.TypeToken;
 
@@ -48,11 +46,14 @@ import com.google.common.reflect.TypeToken;
 public class CompositeEffector extends AddEffector {
 
     private static final Logger LOG = LoggerFactory.getLogger(CompositeEffector.class);
+    private static final String ORIGINAL_PREFIX = "original-";
 
-    public static final ConfigKey<List<String>> EFFECTORS = ConfigKeys.newConfigKey(new TypeToken<List<String>>() {}, "effectors",
+    public static final ConfigKey<List<String>> EFFECTORS = ConfigKeys.newConfigKey(new TypeToken<List<String>>() {
+                                                                                    }, "effectors",
             "Effector names to be chained together in the composite effector", ImmutableList.<String>of());
     public static final ConfigKey<Boolean> OVERRIDE = ConfigKeys.newBooleanConfigKey("override",
             "Wheter additional defined effectors should override pre-existing effector with same name or not (default: false)", Boolean.FALSE);
+
     public CompositeEffector(ConfigBag params) {
         super(newEffectorBuilder(params).build());
     }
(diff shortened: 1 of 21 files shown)


core/src/test/java/org/apache/brooklyn/core/effector/http/HttpCommandEffectorIntegrationTest.java
```diff
@@ -0,0 +1,125 @@
...
+    @AfterMethod(alwaysRun=true)
+    public void tearDown() throws Exception {
+        if (app != null) Entities.destroyAll(app.getManagementContext());
+    }
+
+    @Test(groups="Integration")
+    public void testHttpEffector() throws Exception {
+        new HttpCommandEffector(ConfigBag.newInstance()
+                .configure(HttpCommandEffector.EFFECTOR_NAME, "GithubApacheAccount")
+                .configure(HttpCommandEffector.EFFECTOR_URI, "https://api.github.com/users/apache")
```
aledsage: I'd recommend http://httpbin.org/ for integration testing - it gives very predictable responses based on the inputs, allowing you to test a wide variety of things such as request/response headers, get/put/post/delete etc.

core/src/test/java/org/apache/brooklyn/core/effector/http/HttpCommandEffectorIntegrationTest.java
```diff
@@ -0,0 +1,125 @@
...
+
+public class HttpCommandEffectorIntegrationTest {
+
+    final static Effector<String> EFFECTOR_GITHUB_APACHE_ACCOUNT = Effectors.effector(String.class, "GithubApacheAccount").buildAbstract();
+
+    private TestApplication app;
+    private EntityLocal entity;
+    
+    @BeforeMethod(alwaysRun=true)
+    public void setUp() throws Exception {
```
aledsage: You can write a unit test (rather than just an integration test) if you use something similar to https://github.com/apache/brooklyn-server/blob/master/core/src/test/java/org/apache/brooklyn/feed/http/HttpFeedTest.java

aledsage:
It would be good to also have some yaml-based tests (e.g. see https://github.com/apache/brooklyn-server/blob/master/camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/SshCommandEffectorYamlTest.java)

aledsage:
And rebind tests would be very good (e.g. see https://github.com/apache/brooklyn-server/blob/master/camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/AbstractYamlRebindTest.java and its sub-classes - the important thing is to call `rebind()` in the test method after creating your entities, and then doing some basic testing that the entity is still functional).

core/src/main/java/org/apache/brooklyn/core/effector/CompositeEffector.java
```diff
@@ -0,0 +1,133 @@
...
+import com.google.common.collect.Iterables;
+import com.google.common.collect.Lists;
+import com.google.common.reflect.TypeToken;
+
+@Beta
+public class CompositeEffector extends AddEffector {
+
+    private static final Logger LOG = LoggerFactory.getLogger(CompositeEffector.class);
+
+    public static final ConfigKey<List<String>> EFFECTORS = ConfigKeys.newConfigKey(new TypeToken<List<String>>() {}, "effectors",
```
aledsage: I see we pass in the same parameters to every effector. I wonder whether we should also accept a list of maps, so that we could define the parameters to the individual effectors. But that would make the code more complicated.

It would be interesting to see what the yaml would look like for that.

core/src/test/java/org/apache/brooklyn/core/effector/http/HttpCommandEffectorIntegrationTest.java
andreaturli: thanks @aledsage, much better, I'll use it

andreaturli:
I've added more tests for the effectors:

- add unit tests: HttpCommandEffectorTest and CompositeEffectorTest
- add YamlTests: HttpCommandEffectorYamlTest and CompositeEffectorYamlTest.java
- add YamlRebindTests: HttpCommandEffectorYamlRebindTest and CompositeEffectorYamlRebindTest

---

please @aledsage can you have another look?

andreaturli:
rebuild please

andreaturli:
@aledsage, @grkvlt can you have another look at this PR, please?

ahgittin:
very nice @andreaturli -- and excellent tests.

is there a corresponding PR against the docs to explain how to use this?

looks good so merging

andreaturli:
I will open a PR for docs ASAP, thanks @ahgittin for merging it

PR #533, opened 2017-01-20
persistence: include bundle-prefix for whitelisted classes

aledsage:


aledsage:
I tested this manually in karaf, in three ways:
1. started brooklyn-latest karaf, deployed an app, stopped and restarted (confirming that rebind worked)
2. started an older brooklyn karaf (approx 0.10.0), deployed an app, stooped and restarted brooklyn-latest
3. started an older brooklyn non-karaf (approx 0.10.0), deployed an app, stooped and  restarted brooklyn-latest karaf

The rebind worked fine in each case.

For (3), I also tested this with customer-specific jars (with catalog items and apps referencing those classes). On restarting the latest brooklyn, I added these jars as bundles in the catalog "libraries" section. This required a few more steps to get the upgrade working, but it confirmed that the part covered by this PR worked as expected.

neykov:
LGTM. The white-list has precendece over the bundle prefixes. Serializing them in all cases is a no-op for white listed bundles. But very useful for rebind when the white-list changes.

There's a failing test:
  OsgiClassPrefixerTest.testGetPrefixWithWhitelistedBundle:73->assertAbsent:104 val=Optional.of(org.apache.brooklyn.my-bundle:) expected [false] but found [true]

Good to merge after fixing the test.

aledsage:
Thanks @neykov - merging now.

PR #536, opened 2017-01-26
Avoid NPE on getSubnetIp()

alasdairhodge:
Avoids direct access to `node` field, which may be `null`.


alasdairhodge:
Jenkins failure is unrelated! Details to follow in mailing list.

aledsage:
LGTM - this now matches the code in master. Merging into `0.9.x` branch.

PR #537, opened 2017-01-26
Avoid config access during rebind

alasdairhodge:
Config access during rebind causes unfortunate edge-case for customer using an externalised config supplier that delegates to another entity (so not really "external") which is not guaranteed to have rebound.

Lazily instantiate `runScriptFactory` to prevent config access during rebind.


alasdairhodge:
Jenkins failure is unrelated! Details to follow in mailing list.

aledsage:
Thanks @alasdairhodge - LGTM

FYI the code in master was changed in https://github.com/apache/brooklyn-server/commit/c2669da89c79ff62cfb0082b040274494e59170f#diff-9889943c3686c124b0b7e8ee65d35316. There we do an extra check on lazy-creation for if `jcloudsParent` to give a nicer error than a simple NPE.

However, nobody should be calling `submitRunScript()`. It was deprecated in 0.9.0. It can fail with an unpleasant exception after rebind if the underlying VM was terminated.

Merging this as-is.

PR #538, opened 2017-01-27
Silence the info log

neykov:
* SSHJ logging goes to debug log only
* Users don't need to see config cache size


aledsage:
Thanks @neykov - LGTM; merging.

PR #539, opened 2017-01-27
Log a warning in case of max-concurrency latch double-acquire

neykov:


core/src/main/java/org/apache/brooklyn/core/sensor/ReleaseableLatch.java
```diff
@@ -66,6 +73,13 @@ public MaxConcurrencyLatch(int permits) {
 
             @Override
             public void acquire(Entity caller) {
+                if (!ownerEntities.add(caller)) {
+                    LOG.warn("Entity {} trying to double-acquire permit with ~{} permits available and ~{} threads waiting in queue",
```
aledsage: Perhaps change this log message slightly:
```
LOG.warn("Entity {} acquiring permit multiple times with ~{} permits available ...
```
It could otherwise be mis-understood as saying that the entity was trying to do it, but that it's being skipped.

Worth adding javadoc to say what the semantics are for reentrant calls (and multiple calls by the same entity from different threads).

core/src/main/java/org/apache/brooklyn/core/sensor/ReleaseableLatch.java
```diff
@@ -56,8 +60,11 @@
         }
 
         private static class MaxConcurrencyLatch implements ReleaseableLatch {
+            private static final Logger LOG = LoggerFactory.getLogger(MaxConcurrencyLatch.class);
+
             private int permits;
             private transient final Semaphore sem;
+            private transient final Set<Entity> ownerEntities = new ConcurrentHashSet<Entity>();
```
aledsage: The `transient` makes me think you're trying to handle persistence of the `MaxConcurrencyLatch` - is that right?

If so, then this will fail after rebind because the field `sem` will be null. We'd need a readResolve method to set it, or a getter that lazily instantiates the semaphore.

core/src/main/java/org/apache/brooklyn/core/sensor/ReleaseableLatch.java
```diff
@@ -75,7 +89,11 @@ public void acquire(Entity caller) {
 
             @Override
             public void release(Entity caller) {
-                sem.release();
+                try {
+                    sem.release();
+                } finally {
+                    ownerEntities.remove(caller);
```
aledsage: I think this is fine. If we wanted to get more complicated, we could use a MultiMap or some such so that we know how many times the entity has called acquire(). But that doesn't feel worth the complexity, so I like the code as-is.

core/src/main/java/org/apache/brooklyn/core/sensor/ReleaseableLatch.java
neykov: Updated, thanks for the input.

core/src/main/java/org/apache/brooklyn/core/sensor/ReleaseableLatch.java
neykov: Good spot, haven't considered this case. It's handled and covered by tests though - there's a `readResolve` further down. Added a comment.

core/src/main/java/org/apache/brooklyn/core/sensor/ReleaseableLatch.java
neykov: Agree, this was my thinking as well. Can always update it when we need to address a more complex use-case.

aledsage:
Thanks @neykov - LGTM; merging

PR #543, opened 2017-01-31
Remove uses of EntityLocal

aledsage:
Deletes usage of the deprecated `EntityLocal` (folk should just use `Entity` wherever possible, as it's a lot simpler to explain/understand!).

This is careful to not change the signature of any interfaces or methods that folk might be overriding. It does change some static utility methods though, to take `Entity` instead of `EntityLocal` as a param, because that is safe.


neykov:
LGTM. Tested locally. Merging.

PR #544, opened 2017-01-31
Prefer centos 7

duncangrant:
As discusson on the Apache Brooklyn discussion mailing list this PR updates BrooklynImageChooser to prefer centos over ubuntu

### DIFF
--- a/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/BrooklynImageChooser.java
+++ b/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/BrooklynImageChooser.java
@@ -73,7 +73,7 @@ public class BrooklynImageChooser implements Cloneable {
         return img.getName().toLowerCase().matches("(.*[^a-z])?"+pattern.toLowerCase()+"([^a-z].*)?");
     }
     
-    public double punishmentForOldOsVersions(Image img, OsFamily family, double minVersion) {
+    protected double punishmentForOldOsVersions(Image img, OsFamily family, double minVersion) {
         OperatingSystem os = img.getOperatingSystem();
         if (os!=null && family.equals(os.getFamily())) {
             String v = os.getVersion();
@@ -128,15 +128,18 @@ public class BrooklynImageChooser implements Cloneable {
 
     
         // prefer these guys, in stock brooklyn provisioning
-        score += punishmentForOldOsVersions(img, OsFamily.UBUNTU, 11);
-        score += punishmentForOldOsVersions(img, OsFamily.CENTOS, 6);
 
         OperatingSystem os = img.getOperatingSystem();
         if (os!=null) {
             if (os.getFamily()!=null) {
                 // preference for these open, popular OS (but only wrt versions above) 
-                if (os.getFamily().equals(OsFamily.CENTOS)) score += 2;
+                if (os.getFamily().equals(OsFamily.CENTOS)) {
+                    score += punishmentForOldOsVersions(img, OsFamily.CENTOS, 7);
+                    score += 3;
+
+                }
                 else if (os.getFamily().equals(OsFamily.UBUNTU)) {
+                    score += punishmentForOldOsVersions(img, OsFamily.UBUNTU, 12);
                     score += 2;
 
                     // prefer these LTS releases slightly above others (including above CentOS)
--- a/locations/jclouds/src/test/java/org/apache/brooklyn/location/jclouds/BrooklynImageChooserTest.java
+++ b/locations/jclouds/src/test/java/org/apache/brooklyn/location/jclouds/BrooklynImageChooserTest.java
@@ -1,0 +1,167 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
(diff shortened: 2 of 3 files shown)


geomacy:
Do you think this should be made configurable as [suggested by Mark](https://mail-archives.apache.org/mod_mbox/brooklyn-dev/201701.mbox/%3CCABKEZSbm-%2B7LuWwydxWG4nbfzNLJbiOV5d2%3Dy9ehfYz%3DpWfBrg%40mail.gmail.com%3E) during that conversation?  Or is that something that we should do, but separately from this change?

duncangrant:
@geomacy I read that as more of an open question for which there was only one response (from @googlielmo ) in favour.  I'm personally don't like the idea of making this configurable as I think it is better to handle that in locations and blueprints.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/BrooklynImageChooser.java
```diff
@@ -135,10 +135,14 @@ public double score(Image img) {
...
                 // preference for these open, popular OS (but only wrt versions above) 
-                if (os.getFamily().equals(OsFamily.CENTOS)) score += 2;
-                else if (os.getFamily().equals(OsFamily.UBUNTU)) {
+                if (os.getFamily().equals(OsFamily.CENTOS)) {
                     score += 2;
 
+                    if(os.getVersion()!= null && os.getVersion().startsWith("7.")) score += 0.1;
+                }
+                else if (os.getFamily().equals(OsFamily.UBUNTU)) {
+                    score += 1;
```
aledsage: Will the version always start with "7.", or might it just be "CentOS 7"? I'd be inclined to say `startsWith("7")`.

Do we want to favour more recent CentOS over slightly older ones (e.g. 7.3 is better than 7.1)? Not sure of the best way to do that in a future-proof way. We could do something similar to `punishmentForOldOsVersions`, but to reward newer versions. However, we'd need to be careful that if someone says `7.20170104` that we don't give it a ridiculous high score!)

Changing Ubuntu to `+= 1` puts it at the same level as RHEL, AMZN_LINUX and DEBIAN. I wonder if we should make it `+= 1.5`? Or we could make CentOS `+= 3`?

Minor personal preference: I'd refactor to move the `punishmentForOldOsVersions(img, OsFamily.CENTOS, 6)` call into the `if (...CENTOS)` block here. Similar for ubuntu.

aledsage:
See `JcloudsImageChoiceStubbedLiveTest` and `JcloudsAwsImageChoiceStubbedLiveTest`. I suspect those will need fixed, and perhaps we could extend their coverage (or would that be brittle?)

e.g. `JcloudsAwsImageChoiceStubbedLiveTest.testDefault` should now fail, because it should give a centos machine.

m4rkmckenna:
@geomacy @duncangrant It was a question and should not be included in this pr 

But I can see value in exposing all of the hardcoded weights that are within this class. IMO downstream projects may want to change / tweak this core behaviour

geomacy:
That's grand @m4rkmckenna, @duncangrant, just wanted to check.

duncangrant:
@aledsage I've tried to take into account your comments and I've written some unit tests to assert the logic etc.  I think that I've probably made it harder to make it configurable so maybe still needs a rethink.

ygy:
CentOS 7.3 has a bug which affects ssh daemon reloading - https://bugs.centos.org/view.php?id=12757
`CreateUserPolicy` is directly impaired, because it uses jclouds to reload the ssh daemon config

I have created the bug report yesterday. It is confirmed using AWS images and clean VMs installed from the official CentOS iso files. CentOS 7.2 is not affected.

duncangrant:
I've removed the code that selects the latest centos so that it behaves as before - with no preference for any version of centos

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/BrooklynImageChooser.java
```diff
@@ -73,7 +73,7 @@ protected static boolean imageNameContainsWordCaseInsensitive(Image img, String
         return img.getName().toLowerCase().matches("(.*[^a-z])?"+pattern.toLowerCase()+"([^a-z].*)?");
     }
     
-    public double punishmentForOldOsVersions(Image img, OsFamily family, double minVersion) {
+    private double punishmentForOldOsVersions(Image img, OsFamily family, double minVersion) {
```
neykov: Can you make this `protected`. If someone needs to fine-tune the selection the easiest thing to do would be to extend this class and configure an override in the location.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/BrooklynImageChooser.java
```diff
@@ -128,15 +128,18 @@ public double score(Image img) {
...
             if (os.getFamily()!=null) {
                 // preference for these open, popular OS (but only wrt versions above) 
-                if (os.getFamily().equals(OsFamily.CENTOS)) score += 2;
+                if (os.getFamily().equals(OsFamily.CENTOS)) {
+                    score += punishmentForOldOsVersions(img, OsFamily.CENTOS, 7);
+                    score += 3;
+
+                }
                 else if (os.getFamily().equals(OsFamily.UBUNTU)) {
+                    score += punishmentForOldOsVersions(img, OsFamily.UBUNTU, 11);
```
neykov: Suggest bumping this to at least 12, ideally 14.

aledsage: @neykov It later does:
```
                    if ("14.04".equals(os.getVersion())) score += 0.2;
                    else if ("12.04".equals(os.getVersion())) score += 0.1;
```
so is rewarding 12+, even if it's not punishing 11.

I agree that bumping this to 12 would make sense though. If we changed it to 14, we'd end up preferring 12.10 or 13.10 over  12.04.5 (even though the last one is supported until April 2017, while the first two are not supported).

However, I doubt we ever test our blueprints on 12.04!

PR #545, opened 2017-01-31
Delete EntityFactory

aledsage:
One should instead always use `EntitySpec`, e.g. with `DynamicCluster` memberSpec config.

This PR also:
* refactors tests so that they use memberSpec instead of entityFactory.
* adds BlockingEntity.STARTUP_DELAY/SHUTDOWN_DELAY

Note that some of the code I'm deleting we didn't properly deprecate, but I'm ok with that because it was sort-of deprecated, if indirectly! For example, `DynamicFabric.FACTORY` config key was not deprecated but the type of the config key (`EntityFactory`) has been deprecated since 0.7.0. Similarly, the classes `AbstractConfigurableEntityFactory` and `ClosureEntityFactory` were not deprecated but the interface `EntityFactory` was deprecated so it's clear enough that these classes should not have been used (even if direct use of them would not have given a nice deprecation warning).


neykov:
LGTM. Tests passing locally, also tried a successful deployment. Merging.

PR #546, opened 2017-01-31
Sequencer entity

grkvlt:
An entity and a group that manage sequential sensor values

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/entity/stock/SequenceEntity.java
+++ b/core/src/main/java/org/apache/brooklyn/entity/stock/SequenceEntity.java
@@ -28,6 +28,8 @@ import org.apache.brooklyn.core.entity.trait.Startable;
 import org.apache.brooklyn.entity.group.SequenceGroup;
 import org.apache.brooklyn.util.core.flags.SetFromFlag;
 
+import com.google.common.base.Supplier;
+
 /**
  * An entity that supplies a sequence of values through an effector.
  * <p>
@@ -38,15 +40,12 @@ import org.apache.brooklyn.util.core.flags.SetFromFlag;
  *   brooklyn.config:
  *     sequence.start: 0
  *     sequence.increment: 1
- *     sequence.format: "global-%03d"
  * }</pre>
  */
 @ImplementedBy(SequenceEntityImpl.class)
-public interface SequenceEntity extends Entity, Startable {
+public interface SequenceEntity extends Entity, Startable, Supplier<Integer> {
 
     AttributeSensor<Integer> SEQUENCE_VALUE = SequenceGroup.SEQUENCE_VALUE;
-
-    AttributeSensor<String> SEQUENCE_STRING = SequenceGroup.SEQUENCE_STRING;
 
     @SetFromFlag("sequenceStart")
     ConfigKey<Integer> SEQUENCE_START = SequenceGroup.SEQUENCE_START;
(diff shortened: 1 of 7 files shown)


core/src/test/java/org/apache/brooklyn/entity/group/SequenceGroupTest.java
```diff
@@ -0,0 +1,165 @@
...
+    private TestApplication app;
+    private SequenceGroup group;
+    private TestEntity e1, e2, e3;
+
+    @BeforeMethod(alwaysRun=true)
+    @Override
+    public void setUp() throws Exception {
+        super.setUp();
+
+        app = TestApplication.Factory.newManagedInstanceForTests();
```
aledsage: super-class will have instantiated app - don't do it here, and delete the private field `app`. Also delete `tearDown` as that is done by super.

core/src/test/java/org/apache/brooklyn/entity/group/SequenceGroupTest.java
```diff
@@ -0,0 +1,165 @@
...
+        group = app.createAndManageChild(EntitySpec.create(SequenceGroup.class)
+                .configure(SequenceGroup.SEQUENCE_START, 12345)
+                .configure(SequenceGroup.SEQUENCE_INCREMENT, 678)
+                .configure(SequenceGroup.SEQUENCE_VALUE_SENSOR, value)
+                .configure(SequenceGroup.SEQUENCE_STRING_SENSOR, string)
+                .configure(SequenceGroup.SEQUENCE_FORMAT, "0x%04X"));
+        group.setEntityFilter(EntityPredicates.hasInterfaceMatching(".*TestEntity"));
+
+        assertEqualsIgnoringOrder(group.getMembers(), ImmutableSet.of(e1, e2, e3));
+        assertAttributeEquals(e1, value, 12345);
```
aledsage: Can we be sure that `e1`, `e2`, e3` will be the order of the members? Is that because it is called sequentially with the order that the entities were created?

core/src/main/java/org/apache/brooklyn/entity/stock/SequenceEntity.java
```diff
@@ -0,0 +1,84 @@
...
+    Void increment();
+
+    @Effector(description = "Return the current numeric value of the sequence")
+    Integer currentValue();
+
+    @Effector(description = "Return the current string representation of the sequence")
+    String currentString();
+
+    @Effector(description = "Update and return the next numeric value of the sequence")
+    Integer nextValue();
```
aledsage: I'd prefer we follow the naming conventions of `AtomicInteger` - so rename `increment` to `incrementAndGet`, and delete `nextValue` + `nextString`.

I'm inclined to let the caller do the formatting of the result, rather than having separate `nextValue` and `nextString`. The difference between those two methods, and their interaction, feels as though it might confuse.

grkvlt: Good idea, that would make things more obvious and simpler.

aledsage:
@grkvlt (cc @drigodwin) given we have `DynamicCluster.CLUSTER_MEMBER_ID`, where do we need this sequencer entity as opposed to just using that?

grkvlt:
@aledsage the use case in mind was to sequentially number across multiple clusters (i.e a fabric spread around several locations)

grkvlt:
@drigodwin we need to document this, I will add some descriptive text explaining this for Brooklyn docs

ahgittin:
I feel like this will soon be superseded, by being able to point at some shared-scoped atomic integer and request a `getAndIncrement` directly in YAML as part of `start`.  Haven't looked in detail but if that impl could be done as an extension of `DynamicGroup` could the code here be simplified?

That said not that opposed to adding for now if useful with an expectation it might fall away.

grkvlt:
@mikezaccardo is this still useful, e.g. for HLF blueprints?

mikezaccardo:
@grkvlt this would indeed be useful for the Hyperledger Fabric multi-cluster in place of this entity: https://github.com/cloudsoft/brooklyn-hyperledger/blob/master/catalog/hyperledger/multi-cluster.bom#L205-L251

PR #548, opened 2017-02-01
Clean up now-orphaned locations after unmanaging an entity

alasdairhodge:
Invoking "unmanage" operation on an entity leads to orphaned locations. More seriously, orphaned `JcloudsSshMachineLocation` instances still contain defunct references to the unmanaged entity, which subsequently causes dangling-reference errors on rebind.

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/mgmt/internal/LocalEntityManager.java
+++ b/core/src/main/java/org/apache/brooklyn/core/mgmt/internal/LocalEntityManager.java
@@ -19,7 +19,6 @@
 package org.apache.brooklyn.core.mgmt.internal;
 
 import static com.google.common.base.Preconditions.checkNotNull;
-import groovy.util.ObservableList;
 
 import java.lang.reflect.Proxy;
 import java.util.Collection;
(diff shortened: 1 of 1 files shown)


alasdairhodge:
Feedback welcome, @ahgittin and @aledsage

alasdairhodge:
Build failure due to missing 0.9.x artifacts in Nexus...

alasdairhodge:
Investigating test failure now...

alasdairhodge:
So the unit tests assert that `unmanage` leaves orphaned locations. Obviously, I think this is broken behaviour, but let's see what the consensus is.

core/src/main/java/org/apache/brooklyn/core/mgmt/internal/LocalEntityManager.java
```diff
@@ -475,6 +483,19 @@ private void unmanage(final Entity e, ManagementTransitionMode mode, boolean has
...
+        for (Entity e : getEntities()) {
+            if (e == unmanagedEntity)
+                continue;
+            orphanLocations.removeAll(e.getLocations());
+        }
+        LocationManager lm = managementContext.getLocationManager();
+        for (Location l : orphanLocations) {
+            lm.unmanage(l);
+        }
+    }
```
alasdairhodge: Perhaps this method should be moved onto `ManagementContextInternal` and broadened to GC orphans of all types? As it stands, we're coupling `LocalEntityManger` to `(Local)LocationManager`.

core/src/main/java/org/apache/brooklyn/core/mgmt/internal/LocalEntityManager.java
```diff
@@ -419,6 +422,7 @@ private void unmanage(final Entity e, ManagementTransitionMode mode, boolean has
                 // migrating away or in-place active partial rebind:
                 ((EntityInternal)e).getManagementSupport().onManagementStopping(info);
                 stopTasks(e);
+                pruneOrphanedLocations(e);
```
alasdairhodge: Not sure whether this is the right thing to do in this situation, as we don't actually remove the entity from the internal collections.

alasdairhodge:
Linked issue: https://issues.apache.org/jira/browse/BROOKLYN-432

alasdairhodge:
Closing this PR, as my "fix" has some undesired consequences: unmanaging a jclouds machine lcoation causes any underlying VM to be released!

PR #549, opened 2017-02-03
SecurityGroupEditor- Allow for different jclouds prefixes on different clouds

geomacy:
jclouds on AWS uses 'jclouds#' while on Openstack it is 'jclouds-'.
Also fix some incorrect log messages and add a bit of detail to them.

### DIFF
--- a/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/SecurityGroupEditor.java
+++ b/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/SecurityGroupEditor.java
@@ -34,6 +34,7 @@ import org.slf4j.LoggerFactory;
 
 import java.util.Set;
 import java.util.concurrent.Callable;
+import java.util.regex.Pattern;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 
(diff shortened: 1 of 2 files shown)


locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/SecurityGroupEditor.java
```diff
@@ -156,11 +162,14 @@ public AmbiguousGroupName(String s) {
      * @throws AmbiguousGroupName in the unexpected case that the cloud returns more than one matching group.
      */
     public Optional<SecurityGroup> findSecurityGroupByName(final String name) {
-        final String query = name.startsWith(JCLOUDS_PREFIX) ? name : JCLOUDS_PREFIX + name;
         final Iterable<SecurityGroup> groupsMatching = findSecurityGroupsMatching(new Predicate<SecurityGroup>() {
+            final String rawName = name.matches(IS_JCLOUDS_PREFIXED_REGEX)
+                ? name.substring(JCLOUDS_PREFIX_REGEX.length())
+                : name;
```
neykov: Now that you are using a regex could simplify to name.replaceAll("^jclouds[#-]", "")

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/SecurityGroupEditor.java
```diff
@@ -156,11 +162,14 @@ public AmbiguousGroupName(String s) {
...
+
             @Override
             public boolean apply(final SecurityGroup input) {
-                return input.getName().equals(query);
+                return input.getName().matches(JCLOUDS_PREFIX_REGEX + rawName);
```
neykov: Need to escape the name with "Pattern.quote(rawName)". Or better yet use the same method as above to strip the prefix and compare the non-prefixed names.

geomacy:
changes made as suggested thanks @neykov, and squashed

PR #552, opened 2017-02-08
Adds DynamicCluster.testClustersHaveIndependentCounters

aledsage:
@drigodwin mentioned that DynamicCluster's memberIds seemed to keep incrementing up when a second cluster was provisioned (i.e. the second cluster didn't start counting at 0). This adds a test to assert we start counting from 0 from each new cluster.


drigodwin:
For future reference, the problem I saw was specifically with ZookeeperEnsemble, the default value for the numbering is shared between clusters ([see here](https://github.com/apache/brooklyn-library/blob/4bcc6ef1360b5a2e34dcac089f9ba60784f59172/software/messaging/src/main/java/org/apache/brooklyn/entity/zookeeper/ZooKeeperEnsemble.java#L53)).

PR #553, opened 2017-02-08
MultimapSerializer works with non-String keys

sjcorbett:
MultimapSerializer works with non-String keys

### DIFF
--- a/core/src/test/java/org/apache/brooklyn/util/core/json/MultimapSerializerTest.java
+++ b/core/src/test/java/org/apache/brooklyn/util/core/json/MultimapSerializerTest.java
@@ -1,0 +1,72 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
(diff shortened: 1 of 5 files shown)


sjcorbett:
Thanks @geomacy

core/src/test/java/org/apache/brooklyn/util/core/json/MultimapSerializerTest.java
```diff
@@ -0,0 +1,72 @@
...
+
+    @Test
+    public void testSerializeEntityKey() throws Exception {
+        Entity entity = app.createAndManageChild(EntitySpec.create(BasicEntity.class));
+        Multimap<Entity, Integer> map = MultimapBuilder.hashKeys().arrayListValues().build();
+        map.put(entity, 1);
+        map.put(entity, 2);
+        map.put(entity, 3);
+        String json = mapper.writer().writeValueAsString(map);
+        assertEquals(json, "{\"BasicEntityImpl{id=" + entity.getId() + "}" + "\":[1,2,3]}");
```
neykov: What would the output be if you tried to serialise a complex object, say `map.put(new SimpleBean(), 1)`?

sjcorbett: It should be the `toString` of that object. A SimpleBean would be something like `"SimpleBean@4e5cdbfd"`. A Multimap with an `ImmutableMap` key would look like: `{"{a=b}":[3]}`

sjcorbett: Assuming you mean SimpleBean in SoftwareProcessShellEnvironmentTest it'd actually be `"org.apache.brooklyn.entity.software.base.SoftwareProcessShellEnvironmentTest$SimpleBean@xxxxxxxx"`.

neykov: Thanks. I meant a generic `SimpleBean` :).

PR #555, opened 2017-02-09
Add DurationPredicates

mikezaccardo:
Adds `DurationPredicates` which contains `Predicate`s for common `Duration` comparison operations.

### DIFF
--- a/utils/common/src/main/java/org/apache/brooklyn/util/time/DurationPredicates.java
+++ b/utils/common/src/main/java/org/apache/brooklyn/util/time/DurationPredicates.java
@@ -1,0 +1,162 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.brooklyn.util.time;
+
+import java.util.concurrent.TimeUnit;
+
+import com.google.common.base.Preconditions;
+import com.google.common.base.Predicate;
+import com.google.common.base.Stopwatch;
+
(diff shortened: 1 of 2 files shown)


utils/common/src/main/java/org/apache/brooklyn/util/time/DurationPredicates.java
```diff
@@ -0,0 +1,119 @@
...
+     *            the returned {@link Predicate}.
+     * @return A {@link Predicate} that checks if the {@link Duration} that was
+     *         supplied to this method is longer than any {@link Stopwatch}
+     *         supplied to the returned {@link Predicate#apply(Object)}.
+     */
+    public static Predicate<Stopwatch> isLongerThanStopwatch(final Duration duration) {
+        return new Predicate<Stopwatch>() {
+            @Override
+            public boolean apply(Stopwatch other) {
+                return duration != null && other != null && duration.isLongerThan(other);
```
grkvlt: This is confusing. A `Predicate<Stopwatch>` is making a boolean statement about `Stopwatch`-es, so in this case maybe reverse the check and the name?

utils/common/src/main/java/org/apache/brooklyn/util/time/DurationPredicates.java
```diff
@@ -0,0 +1,119 @@
...
+     *            the returned {@link Predicate}.
+     * @return A {@link Predicate} that checks if the {@link Duration} that was
+     *         supplied to this method is shorter than any {@link Duration}
+     *         supplied to to the returned {@link Predicate#apply(Object)}.
+     */
+    public static Predicate<Duration> isShorterThanDuration(final Duration duration) {
+        return new Predicate<Duration>() {
+            @Override
+            public boolean apply(Duration other) {
+                return duration != null && other != null && duration.isShorterThan(other);
```
grkvlt: Again, the `Duration` the predicate speaks about here is the one called `other`, the `duration` is constant over the life of the predicate.

utils/common/src/main/java/org/apache/brooklyn/util/time/DurationPredicates.java
```diff
@@ -0,0 +1,119 @@
...
+     *            the returned {@link Predicate}.
+     * @return A {@link Predicate} that checks if the {@link Duration} that was
+     *         supplied to this method is shorter than any {@link Stopwatch}
+     *         supplied to to the returned {@link Predicate#apply(Object)}.
+     */
+    public static Predicate<Stopwatch> isShorterThanStopwatch(final Duration duration) {
+        return new Predicate<Stopwatch>() {
+            @Override
+            public boolean apply(Stopwatch other) {
+                return duration != null && other != null && duration.isShorterThan(other);
```
grkvlt: Same.

grkvlt:
@mikezaccardo remember we need tests for these too!

ahgittin:
LGTM

PR #556, opened 2017-02-10
Speed up tests

aledsage:
Starting an embedded OSGi container is slow - over half a second per `setUp()` call (and thus per test method) that uses it. It will likely be considerably slower on the jenkins build machine etc. This PR separates out those classes that include an embedded OSGi container into those tests that actually need it and those that don't. It primarily moves code around (e.g. separating out the osgi-based tests from `CatalogYamlLocationTest` into a new `CatalogOsgiYamlLocationTest`).

It also cherry-pick @ahgittin's commit that was included in https://github.com/apache/brooklyn-server/pull/480 (which marks some more slow tests as "integration").

If you're wondering why just moving the code has resulted in a bunch of extra lines (`+1,956 −1,038` according to the "Files changed" summary), there are two primary reasons: first there are 6 new classes which have the apache header and a bunch of imports; second the `CatalogOsgiYamlEntityTest` repeats some tests that are done in `CatalogOsgiEntityTest`. The reason is that there are subtleties in how bundles are used to load catalog items (especially when composing multiple items, and sub-typing). It's therefore worth having slow versions of some of these tests that use actual OSGi bundles.


neykov:
LGTM, merging.

PR #557, opened 2017-02-13
transfer icons from catalog items to entities

ahgittin:
and set icons on the common `server` and `cluster` items

previously while we do a lot to support icons and send detail via the API,
we don't _set_ them on entities unless explicitly in the definition.
since we often have them on catalog items, transferring them seems a sensible idea.


drigodwin:
This looks fine, what are the sources and licenses of the icons @ahgittin?

drigodwin:
This looks fine, what are the sources and licenses of the icons @ahgittin?

ahgittin:
good catch -- they were from a free icon site, but i must find those properly!

ahgittin:
Have changed icons to ones for which the licensing is clear and added license info in commit history; will open new PR for corresponding license data in `brooklyn-dist`

ahgittin:
Have also pushed license changes; you can see them here:

* https://github.com/apache/brooklyn-dist/commit/7087633e71d2131a68e52266ea808dd50ed0fc33

Good catch @drigodwin !

PR #558, opened 2017-02-13
BROOKLYN-433: regex/obj config key constraint in yaml

aledsage:
Docs PR will follow, to update the constraints section in https://github.com/apache/brooklyn-docs/blob/master/guide/yaml/yaml-reference.md

### DIFF
--- a/camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/ConfigParametersYamlTest.java
+++ b/camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/ConfigParametersYamlTest.java
@@ -22,6 +22,8 @@ import static org.testng.Assert.assertEquals;
 import static org.testng.Assert.assertNotNull;
 import static org.testng.Assert.assertTrue;
 
+import static com.google.common.base.Preconditions.checkNotNull;
+
 import java.util.Date;
 import java.util.List;
 import java.util.Map;
@@ -31,6 +33,7 @@ import org.apache.brooklyn.api.entity.EntitySpec;
 import org.apache.brooklyn.api.location.PortRange;
 import org.apache.brooklyn.config.ConfigKey;
 import org.apache.brooklyn.core.config.ConfigKeys;
+import org.apache.brooklyn.core.config.ConstraintViolationException;
 import org.apache.brooklyn.core.entity.BrooklynConfigKeys;
 import org.apache.brooklyn.core.location.PortRanges;
 import org.apache.brooklyn.core.sensor.Sensors;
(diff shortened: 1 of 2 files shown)


camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/ConfigParametersYamlTest.java
```diff
@@ -750,6 +754,175 @@ public void testPortSetAsAttributeOnSoftwareProcess() throws Exception {
...
+                "services:",
+                "- type: entity-with-keys",
+                "  brooklyn.config:",
+                "    testRequired: myprefix-myVal");
+
+        try {
+            createStartWaitAndLogApplication(yamlNoVal);
+            Asserts.shouldHaveFailedPreviously();
+        } catch (ConstraintViolationException e) {
+            // success
```
bostko: What about putting a message assertion?

core/src/main/java/org/apache/brooklyn/core/objs/BasicSpecParameter.java
```diff
@@ -310,6 +317,41 @@ private static Predicate parseConstraints(Object obj, BrooklynClassLoadingContex
...
+                    Object key = Iterables.getOnlyElement(constraint.keySet());
+                    Object val = constraint.get(key);
+                    if (BUILT_IN_CONSTRAINT_FACTORIES.containsKey(key)) {
+                        Function<Object, Predicate<?>> factory = BUILT_IN_CONSTRAINT_FACTORIES.get(key);
+                        return factory.apply(val);
+                    } else {
+                        throw new IllegalArgumentException("The constraint '" + constraint + "' for a catalog input is not recognized as a built-in (" + BUILT_IN_CONSTRAINTS.keySet() + ")");
+                    }
+                } else {
+                    throw new IllegalArgumentException("The constraint '" + constraint + "' for a catalog input is not recognized - cannot handle multiple key:vales");
```
bostko: Change to `The config key constraint '" + constraint + "' is not supported - it can handle only single key:value constraint.`

neykov:
cc @tbouron - should we add support in UI as well?

tbouron:
@neykov Depends how that is currently implemented: are we doing this validation client or server side?
Although in this particular case, doing this validation client side would require the REST API to return the pattern but I think it currently returns only the type of constraint, i.e. `required` or `regex`.

But I'm definitely 👍 to add it to the UI: better to fail fast

aledsage:
@tbouron It's validated server-side. Longer term, we should improve the validation errors + warnings that the server-side sends back, along with the info needed for content-assist. That's a bigger topic!

@bostko thanks for comments - addressed, rebased and pushed. Once jenkins build completes again, I'll merge it.

tbouron:
@aledsage In that case, LGTM 👍 

Would be great to follow up after this PR is merged to make the REST API return more information about the constraint (and modify the UI to react to it)

aledsage:
Docs PR now merged - https://github.com/apache/brooklyn-docs/pull/151

PR #560, opened 2017-02-16
Attempted non-determinate test fixes

ahgittin:
Prompted by the following (from https://builds.apache.org/job/brooklyn-server-pull-requests/1730/ ):

```
Test Result (3 failures / +3)

org.apache.brooklyn.launcher.CleanOrphanedLocationsIntegrationTest.testCleanedCopiedPersistedState
org.apache.brooklyn.launcher.CleanOrphanedLocationsIntegrationTest.testSelectionWithOrphanedLocationsInData
org.apache.brooklyn.launcher.BrooklynLauncherTest.testErrorsCaughtByApiAndRestApiWorks
```

The first two have been observed in several places, and might be fixed by the first commit here.  The second one is less frequent but the second commit here might help and should improve logging.

### DIFF
--- a/launcher/src/main/java/org/apache/brooklyn/launcher/BrooklynWebServer.java
+++ b/launcher/src/main/java/org/apache/brooklyn/launcher/BrooklynWebServer.java
@@ -20,6 +20,7 @@ package org.apache.brooklyn.launcher;
 
 import java.io.File;
 import java.io.InputStream;
+import java.net.BindException;
 import java.net.InetAddress;
 import java.net.URI;
 import java.security.KeyPair;
@@ -82,6 +83,8 @@ import org.apache.brooklyn.util.os.Os;
 import org.apache.brooklyn.util.stream.Streams;
 import org.apache.brooklyn.util.text.Identifiers;
 import org.apache.brooklyn.util.text.Strings;
+import org.apache.brooklyn.util.time.Duration;
+import org.apache.brooklyn.util.time.Time;
 import org.apache.brooklyn.util.web.ContextHandlerCollectionHotSwappable;
 import org.eclipse.jetty.http.HttpVersion;
 import org.eclipse.jetty.jaas.JAASLoginService;
@@ -447,7 +450,15 @@ public class BrooklynWebServer {
         rootContext.setTempDirectory(Os.mkdirs(new File(webappTempDir, "war-root")));
 
         server.setHandler(handlers);
-        server.start();
+        try {
+            server.start();
+        } catch (BindException e) {
+            // port discovery routines may take some time to clear, e.g. 250ms for SO_TIMEOUT
+            // tests fail because of this; see if adding a delay improves things
+            log.warn("Initial server start-up failed binding (retrying after a delay): "+e);
+            Time.sleep(Duration.millis(500));
+            server.start();
+        }
         //reinit required because some webapps (eg grails) might wipe our language extension bindings
         BrooklynInitialization.reinitAll();
 
--- a/launcher/src/main/java/org/apache/brooklyn/launcher/BrooklynWebServer.java
+++ b/launcher/src/main/java/org/apache/brooklyn/launcher/BrooklynWebServer.java
@@ -453,8 +453,7 @@ public class BrooklynWebServer {
         try {
             server.start();
         } catch (BindException e) {
-            // port discovery routines may take some time to clear, e.g. 250ms for SO_TIMEOUT
-            // tests fail because of this; see if adding a delay improves things
+            // retry once just in case it was some fluke or race
             log.warn("Initial server start-up failed binding (retrying after a delay): "+e);
             Time.sleep(Duration.millis(500));
             server.start();
--- a/launcher/src/test/java/org/apache/brooklyn/launcher/CleanOrphanedLocationsIntegrationTest.java
+++ b/launcher/src/test/java/org/apache/brooklyn/launcher/CleanOrphanedLocationsIntegrationTest.java
@@ -41,6 +41,7 @@ import org.apache.brooklyn.core.server.BrooklynServerPaths;
 import org.apache.brooklyn.core.test.entity.LocalManagementContextForTests;
 import org.apache.brooklyn.test.Asserts;
 import org.apache.brooklyn.util.collections.MutableSet;
+import org.apache.brooklyn.util.javalang.JavaClassNames;
 import org.apache.brooklyn.util.os.Os;
 import org.apache.brooklyn.util.time.Duration;
 import org.slf4j.Logger;
@@ -147,6 +148,7 @@ public class CleanOrphanedLocationsIntegrationTest extends AbstractCleanOrphaned
 
     @Test
     public void testCleanedCopiedPersistedState() throws Exception {
+        LOG.info(JavaClassNames.niceClassAndMethod()+" taking persistence from "+persistenceDirWithOrphanedLocations);
         BrooklynLauncher launcher = BrooklynLauncher.newInstance()
                 .webconsole(false)
                 .brooklynProperties(OsgiManager.USE_OSGI, false)
--- a/launcher/src/test/java/org/apache/brooklyn/launcher/CleanOrphanedLocationsIntegrationTest.java
+++ b/launcher/src/test/java/org/apache/brooklyn/launcher/CleanOrphanedLocationsIntegrationTest.java
@@ -121,6 +121,7 @@ public class CleanOrphanedLocationsIntegrationTest extends AbstractCleanOrphaned
         PersistenceExceptionHandler persistenceExceptionHandler = PersistenceExceptionHandlerImpl.builder().build();
         ((RebindManagerImpl) rebindManager).setPeriodicPersistPeriod(Duration.ONE_SECOND);
         rebindManager.setPersister(persister, persistenceExceptionHandler);
+        ((RebindManagerImpl) rebindManager).forcePersistNow();
     }
 
     @Test
--- a/launcher/src/test/java/org/apache/brooklyn/launcher/CleanOrphanedLocationsIntegrationTest.java
+++ b/launcher/src/test/java/org/apache/brooklyn/launcher/CleanOrphanedLocationsIntegrationTest.java
@@ -20,6 +20,7 @@ package org.apache.brooklyn.launcher;
 
 import java.io.File;
 import java.util.Set;
+import java.util.zip.ZipFile;
 
 import org.apache.brooklyn.api.mgmt.ManagementContext;
 import org.apache.brooklyn.api.mgmt.ha.HighAvailabilityMode;
@@ -41,6 +42,10 @@ import org.apache.brooklyn.core.server.BrooklynServerPaths;
 import org.apache.brooklyn.core.test.entity.LocalManagementContextForTests;
 import org.apache.brooklyn.test.Asserts;
 import org.apache.brooklyn.util.collections.MutableSet;
+import org.apache.brooklyn.util.core.ResourceUtils;
+import org.apache.brooklyn.util.core.file.ArchiveUtils;
+import org.apache.brooklyn.util.core.osgi.BundleMaker;
+import org.apache.brooklyn.util.exceptions.Exceptions;
 import org.apache.brooklyn.util.javalang.JavaClassNames;
 import org.apache.brooklyn.util.os.Os;
 import org.apache.brooklyn.util.time.Duration;
@@ -67,13 +72,26 @@ public class CleanOrphanedLocationsIntegrationTest extends AbstractCleanOrphaned
     private Set<ManagementContext> mgmts;
     private ManagementContext mgmt;
 
+    private String copyResourcePathToTempPath(String resourcePath) {
+        BundleMaker bm = new BundleMaker(ResourceUtils.create(this));
+        bm.setDefaultClassForLoading(CleanOrphanedLocationsIntegrationTest.class);
+        File jar = bm.createJarFromClasspathDir(resourcePath);
+        File output = Os.newTempDir("brooklyn-test-resouce-from-"+resourcePath);
+        try {
+            ArchiveUtils.extractZip(new ZipFile(jar), output.getAbsolutePath());
+            return output.getAbsolutePath();
+        } catch (Exception e) {
+            throw Exceptions.propagate(e);
+        }
+    }
+    
     @BeforeMethod(alwaysRun = true)
     @Override
     public void setUp() throws Exception {
         super.setUp();
-        persistenceDirWithOrphanedLocations = getClass().getResource(PERSISTED_STATE_PATH_WITH_ORPHANED_LOCATIONS).getFile();
-        persistenceDirWithoutOrphanedLocations = getClass().getResource(PERSISTED_STATE_PATH_WITHOUT_ORPHANED_LOCATIONS).getFile();
-        persistenceDirWithMultipleLocationsOccurrence = getClass().getResource(PERSISTED_STATE_PATH_WITH_MULTIPLE_LOCATIONS_OCCURRENCE).getFile();
+        persistenceDirWithOrphanedLocations = copyResourcePathToTempPath(PERSISTED_STATE_PATH_WITH_ORPHANED_LOCATIONS);
+        persistenceDirWithoutOrphanedLocations = copyResourcePathToTempPath(PERSISTED_STATE_PATH_WITHOUT_ORPHANED_LOCATIONS);
+        persistenceDirWithMultipleLocationsOccurrence = copyResourcePathToTempPath(PERSISTED_STATE_PATH_WITH_MULTIPLE_LOCATIONS_OCCURRENCE);
 
         destinationDir = Os.newTempDir(getClass());
         
--- a/core/src/main/java/org/apache/brooklyn/core/catalog/internal/BasicBrooklynCatalog.java
+++ b/core/src/main/java/org/apache/brooklyn/core/catalog/internal/BasicBrooklynCatalog.java
@@ -1053,7 +1053,7 @@ public class BasicBrooklynCatalog implements BrooklynCatalog {
         log.debug("Adding manual catalog item to "+mgmt+": "+type);
         checkNotNull(type, "type");
         if (manualAdditionsCatalog==null) loadManualAdditionsCatalog();
-        manualAdditionsClasses.addClass(type);
+        manualAdditionsClasses.registerClass(type);
         CatalogItem<?, ?> result = manualAdditionsCatalog.classpath.addCatalogEntry(type);
         
         // Clear spec-cache (in-case overwriting)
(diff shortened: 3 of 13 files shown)


launcher/src/main/java/org/apache/brooklyn/launcher/BrooklynWebServer.java
```diff
@@ -447,7 +450,15 @@ public synchronized void start() throws Exception {
...
         server.setHandler(handlers);
-        server.start();
+        try {
+            server.start();
+        } catch (BindException e) {
+            // port discovery routines may take some time to clear, e.g. 250ms for SO_TIMEOUT
+            // tests fail because of this; see if adding a delay improves things
+            log.warn("Initial server start-up failed binding (retrying after a delay): "+e);
+            Time.sleep(Duration.millis(500));
+            server.start();
```
neykov: Just binding to a socket won't result in a `TIME_WAIT` state (the `isPortAvailable` check). `TIME_WAIT` is the result of a previous closed connection (previous test) or a client trying to connect in the small window the probe socket is open.
One thing that will improve false positive tests is removing [`setReuseAddress(true)`](https://github.com/apache/brooklyn-server/blob/master/utils/common/src/main/java/org/apache/brooklyn/util/net/Networking.java#L105). The `SO_TIMEOUT=250` in the same method isn't used at all since it applies to reads which we don't do here. So it can go away as well. `TIME_WAIT` timeout could be in the minutes range.

Instead of papering over the problem suggest we go for the root cause directly and the Apache servers are an excellent test bed for it. I can do the networking changes if you prefer?

LGTM other than this.

neykov: Another possible cause would be parallel tests taking the port just after the check. For tests we could change the logic (here and elsewhere jmx?) to let sockets pick their own listening port.

ahgittin: @neykov Why would `setReuseAddress(true)` improve false positives?  As long as we make the same call when we start the server it should be identical?  (Or are you thinking that the root cause is that the server-start call isn't using that, and that is what you'd fix?)

I agree finding and fixing the root cause (and recognising and logging better when we encounter it?) is even better, but I don't know what it is!

I don't think we do parallel tests so I suspect that isn't the issue.  Letting sockets pick their own ports is a big change so I'd rather not go down that line.

ahgittin:
Based on one run this improves things.  I've tried closing and reopening but that no longer triggers a rebuild.  Damn server-side smartness!

ahgittin:
Aha nvm Jenkins is now running, it just took a little while.

launcher/src/main/java/org/apache/brooklyn/launcher/BrooklynWebServer.java
neykov: > Why would setReuseAddress(true) improve false positives? As long as we make the same call when we start the server it should be identical? (Or are you thinking that the root cause is that the server-start call isn't using that, and that is what you'd fix?)

That's right - `setReuseAddress(true)` is used in `isPortAvailable`, but not in the actual server bind call (for web or jmx). `SO_REUSEADDR` allows you to bind to a port in `TIME_WAIT` state. So the availability check will succeed, but the server bind will fail.

I wouldn't add `setReuseAddress(true)` to the server, but will instead remove it from `isPortAvailable`.

We have parallel tests not from the same maven run, but from parallel jenkins jobs (another process).

neykov: Another use case which would cause bind to fail is described in the comments of [isPortAvailable](https://github.com/apache/brooklyn-server/blob/master/utils/common/src/main/java/org/apache/brooklyn/util/net/Networking.java#L96-L98). Again the solution is to remove `SO_REUSEADDR`.
Will prepare a PR which does the change - let's see if it will go well with the PR build.

neykov: https://github.com/apache/brooklyn-server/pull/563 - will let it run a few times to see how it behaves on the salves. Obviously not failing doesn't mean anything.

While experimenting I found that `SO_REUSEADDR` is activated **by default** on `ServerSocket`.

neykov: One thing was clear though - `TIME_WAIT` doesn't fail binding the `ServerSocket`. It's something else running that's bugging the test. Could be another Brooklyn job that manages to steal the port or something unrelated doing funny things with the interfaces.

ahgittin:
Curious. I'll amend comments here and merge if that's okay. Happy to remove
the setReuseAddr or not depending who gets to it first.

neykov:
+1 go ahead.

ahgittin:
Applied @neykov 's comments.  If all tests pass then good to merge.

ahgittin:
nope, last failure shows this didn't fix the orphaned state failures either :(

ahgittin:
the debug log was still around, so we get a bit more info:

```
2017-02-18 02:00:01,331 DEBUG o.a.b.c.m.p.FileBasedObjectStore [main]: File-based objectStore will use directory /home/jenkins/jenkins-slave/workspace/brooklyn-server-pull-requests%402/launcher/target/test-classes/orphaned-locations-test-data/data-with-orphaned-locations
2017-02-18 02:00:01,332 DEBUG o.a.b.c.m.p.FileBasedObjectStore [main]: Created dir /home/jenkins/jenkins-slave/workspace/brooklyn-server-pull-requests%402/launcher/target/test-classes/orphaned-locations-test-data/data-with-orphaned-locations for FileBasedObjectStore{basedir=/home/jenkins/jenkins-slave/workspace/brooklyn-server-pull-requests%402/launcher/target/test-classes/orphaned-locations-test-data/data-with-orphaned-locations}
2017-02-18 02:00:01,345 DEBUG o.a.b.c.m.p.FileBasedObjectStore [main]: Persistence mode AUTO, directory /home/jenkins/jenkins-slave/workspace/brooklyn-server-pull-requests%402/launcher/target/test-classes/orphaned-locations-test-data/data-with-orphaned-locations, no previous state
2017-02-18 02:00:01,349 DEBUG o.a.b.c.m.p.FileBasedObjectStore [main]: Created dir /home/jenkins/jenkins-slave/workspace/brooklyn-server-pull-requests%402/launcher/target/test-classes/orphaned-locations-test-data/data-with-orphaned-locations/entities for FileBasedObjectStore{basedir=/home/jenkins/jenkins-slave/workspace/brooklyn-server-pull-requests%402/launcher/target/test-classes/orphaned-locations-test-data/data-with-orphaned-locations}
2017-02-18 02:00:01,367 DEBUG o.a.b.c.m.p.FileBasedObjectStore [main]: Created dir /home/jenkins/jenkins-slave/workspace/brooklyn-server-pull-requests%402/launcher/target/test-classes/orphaned-locations-test-data/data-with-orphaned-locations/locations for FileBasedObjectStore{basedir=/home/jenkins/jenkins-slave/workspace/brooklyn-server-pull-requests%402/launcher/target/test-classes/orphaned-locations-test-data/data-with-orphaned-locations}
2017-02-18 02:00:01,367 DEBUG o.a.b.c.m.p.FileBasedObjectStore [main]: Created dir /home/jenkins/jenkins-slave/workspace/brooklyn-server-pull-requests%402/launcher/target/test-classes/orphaned-locations-test-data/data-with-orphaned-locations/policies for FileBasedObjectStore{basedir=/home/jenkins/jenkins-slave/workspace/brooklyn-server-pull-requests%402/launcher/target/test-classes/orphaned-locations-test-data/data-with-orphaned-locations}
2017-02-18 02:00:01,367 DEBUG o.a.b.c.m.p.FileBasedObjectStore [main]: Created dir /home/jenkins/jenkins-slave/workspace/brooklyn-server-pull-requests%402/launcher/target/test-classes/orphaned-locations-test-data/data-with-orphaned-locations/enrichers for FileBasedObjectStore{basedir=/home/jenkins/jenkins-slave/workspace/brooklyn-server-pull-requests%402/launcher/target/test-classes/orphaned-locations-test-data/data-with-orphaned-locations}
2017-02-18 02:00:01,368 DEBUG o.a.b.c.m.p.FileBasedObjectStore [main]: Created dir /home/jenkins/jenkins-slave/workspace/brooklyn-server-pull-requests%402/launcher/target/test-classes/orphaned-locations-test-data/data-with-orphaned-locations/feeds for FileBasedObjectStore{basedir=/home/jenkins/jenkins-slave/workspace/brooklyn-server-pull-requests%402/launcher/target/test-classes/orphaned-locations-test-data/data-with-orphaned-locations}
2017-02-18 02:00:01,368 DEBUG o.a.b.c.m.p.FileBasedObjectStore [main]: Created dir /home/jenkins/jenkins-slave/workspace/brooklyn-server-pull-requests%402/launcher/target/test-classes/orphaned-locations-test-data/data-with-orphaned-locations/catalog for FileBasedObjectStore{basedir=/home/jenkins/jenkins-slave/workspace/brooklyn-server-pull-requests%402/launcher/target/test-classes/orphaned-locations-test-data/data-with-orphaned-locations}
2017-02-18 02:00:01,368 DEBUG o.a.b.c.m.p.FileBasedObjectStore [main]: Created dir /home/jenkins/jenkins-slave/workspace/brooklyn-server-pull-requests%402/launcher/target/test-classes/orphaned-locations-test-data/data-with-orphaned-locations/plane for FileBasedObjectStore{basedir=/home/jenkins/jenkins-slave/workspace/brooklyn-server-pull-requests%402/launcher/target/test-classes/orphaned-locations-test-data/data-with-orphaned-locations}
2017-02-18 02:00:01,373 DEBUG o.a.b.c.m.p.FileBasedObjectStore [main]: Created dir /home/jenkins/jenkins-slave/workspace/brooklyn-server-pull-requests%402/launcher/target/test-classes/orphaned-locations-test-data/data-with-orphaned-locations/nodes for FileBasedObjectStore{basedir=/home/jenkins/jenkins-slave/workspace/brooklyn-server-pull-requests%402/launcher/target/test-classes/orphaned-locations-test-data/data-with-orphaned-locations}
2017-02-18 02:00:01,373 DEBUG o.a.b.c.m.h.ManagementPlaneSyncRecordPersisterToObjectStore [main]: ManagementPlaneMemento-persister will use store FileBasedObjectStore{basedir=/home/jenkins/jenkins-slave/workspace/brooklyn-server-pull-requests%402/launcher/target/test-classes/orphaned-locations-test-data/data-with-orphaned-locations}
2017-02-18 02:00:01,374 DEBUG o.a.b.c.m.p.BrooklynMementoPersisterToObjectStore [main]: Loaded rebind lists; took 0ms: 0 entities, 0 locations, 0 policies, 0 enrichers, 0 feeds, 0 catalog items; from /home/jenkins/jenkins-slave/workspace/brooklyn-server-pull-requests%402/launcher/target/test-classes/orphaned-locations-test-data/data-with-orphaned-locations
2017-02-18 02:00:01,374 DEBUG o.a.b.c.m.p.BrooklynMementoPersisterToObjectStore [main]: Loaded rebind raw data; took 0ms; 0 entities, 0 locations, 0 policies, 0 enrichers, 0 feeds, 0 catalog items, from /home/jenkins/jenkins-slave/workspace/brooklyn-server-pull-requests%402/launcher/target/test-classes/orphaned-locations-test-data/data-with-orphaned-locations
```

seems the directory `/home/jenkins/jenkins-slave/workspace/brooklyn-server-pull-requests%402/launcher/target/test-classes/orphaned-locations-test-data/data-with-orphaned-locations` did not exist

ahgittin:
at the moment the contents look healthy in https://builds.apache.org/job/brooklyn-server-pull-requests/ws/launcher/target/test-classes/orphaned-locations-test-data/data-with-orphaned-locations/locations/ ... i guess either there is something wrong with the path (need to confirm `%402` but assuming `%40` is octal for a space that seems reasonable) or something else is clobbering the files on disk (but I'd expect a parallel build in the same space to wipe out the `target/` directory and thus be very damaging).

probably good practise to read the files from the classpath and save them somewhere else, rather than assuming an unpacked directory.  that would fix the problem.  but i'd like to to understand it in case it's causes other issues.

ahgittin:
(elsewhere in the log it refers to path `/home/jenkins/jenkins-slave/workspace/brooklyn-server-pull-requests@2/launcher/target/surefire/surefirebooter8075241968007199755.jar!/META-INF/MANIFEST.MF` so `%40` is hex for `@`)

ahgittin:
a healthy run also shows the `Creating directory` message, but the key difference seems to be:

```
2017-02-14 16:16:13,575 DEBUG o.a.b.c.m.h.ManagementPlaneSyncRecordPersisterToObjectStore [main]: ManagementPlaneMemento-persister will use store FileBasedObjectStore{basedir=/Users/alex/Data/cloudsoft/dev/gits/brooklyn/brooklyn-server/launcher/target/test-classes/orphaned-locations-test-data/data-with-orphaned-locations}
2017-02-14 16:16:13,575 DEBUG o.a.b.c.m.p.BrooklynMementoPersisterToObjectStore [main]: Loaded rebind lists; took 0ms: 2 entities, 5 locations, 0 policies, 0 enrichers, 0 feeds, 0 catalog items; from /Users/alex/Data/cloudsoft/dev/gits/brooklyn/brooklyn-server/launcher/target/test-classes/orphaned-locations-test-data/data-with-orphaned-locations
```

ahgittin:
that comment comes immediately after the `File[] subPathDirFiles = subPathDir.listFiles(fileFilter);` so does seems like the dir is empty when the failing test runs (as opposed to some other cause)

ahgittin:
also, this line

```
2017-02-18 02:00:01,345 DEBUG o.a.b.c.m.p.FileBasedObjectStore [main]: Persistence mode AUTO, directory /home/jenkins/jenkins-slave/workspace/brooklyn-server-pull-requests%402/launcher/target/test-classes/orphaned-locations-test-data/data-with-orphaned-locations, no previous state
```

appears only in the failed run, and code confirms it should be logged iff the dir is empty / non-existant.  so how is the dir being wiped?!

ahgittin:
i wonder if the problem is that the `@` is being escaped when it shouldn't.  would explain why it fails non-deterministically, probably sometimes it runs in the non-`@2` directory.  and can't tell why it is getting escaped at all ... we do `Class.getResource(..).getFile()` and that's what we use.  on my box it doesn't seem to be escaped if i mess it up to run in a dir containing an `@`.

ahgittin:
working on the basis that the `getResource(..).getFile()` call might do URL-escaping when we say `getResource()` and then not properly unescape it when we do `getFile()`.  that technique isn't guaranteed to work (definitely not if tests are run from JARs), this is just another failure case we hadn't encountered.

will rewrite to extract that entire folder.

ahgittin:
This now extracts classpath folders to a temp folder on the filespace.  Builds on #485 (using an intermediate JAR) for simplicity.  Assuming the `@` escaping was the cause of orphans this should fix it -- in any case it improves the hacky `getFile()` done on a classpath resource URL.

Note #485 needs review first now.  Personally I think that PR should be simpler than it has been as it's a pretty clear improvement and further PRs can improve it further -- but if it's contentious I can refactor this to avoid the intermediate archive and thus the dependency (though personally not convinced it's worth it).

aledsage:
Wow, I wasn't expecting to see this set of commits of file-edits based on the name of the PR and the description!

It would be great to review + merge your improvements for non-deterministic test failures independent of trying to review + merge the `BundleMaker` changes. We can then get it merged quickly, and get the benefits immediately.

Can you separate those out into their own PR, rather than building on the `BundleMaker` PR (and thus blocking the merging of those changes until that other PR is dealt with)?

ahgittin:
@aledsage I didn't pull in #485 for fun.  The simplest fix for some of the failures involves building an intermediate bundle thus

> Builds on #485 (using an intermediate JAR) for simplicity.

I'll split out the REST portion of #485 and look at your comments after which this will be simpler to review.

ahgittin:
@aledsage I like how you split out #485; this has been rebased and conflicts merged.  Should be easy to review now!

launcher/src/test/java/org/apache/brooklyn/launcher/CleanOrphanedLocationsIntegrationTest.java
```diff
@@ -66,13 +72,26 @@
...
     @BeforeMethod(alwaysRun = true)
     @Override
     public void setUp() throws Exception {
         super.setUp();
-        persistenceDirWithOrphanedLocations = getClass().getResource(PERSISTED_STATE_PATH_WITH_ORPHANED_LOCATIONS).getFile();
-        persistenceDirWithoutOrphanedLocations = getClass().getResource(PERSISTED_STATE_PATH_WITHOUT_ORPHANED_LOCATIONS).getFile();
-        persistenceDirWithMultipleLocationsOccurrence = getClass().getResource(PERSISTED_STATE_PATH_WITH_MULTIPLE_LOCATIONS_OCCURRENCE).getFile();
+        persistenceDirWithOrphanedLocations = copyResourcePathToTempPath(PERSISTED_STATE_PATH_WITH_ORPHANED_LOCATIONS);
+        persistenceDirWithoutOrphanedLocations = copyResourcePathToTempPath(PERSISTED_STATE_PATH_WITHOUT_ORPHANED_LOCATIONS);
+        persistenceDirWithMultipleLocationsOccurrence = copyResourcePathToTempPath(PERSISTED_STATE_PATH_WITH_MULTIPLE_LOCATIONS_OCCURRENCE);
```
neykov: The proper way to escape the URL is `new File(url.toURI())`.

launcher/src/main/java/org/apache/brooklyn/launcher/BrooklynWebServer.java
```diff
@@ -447,7 +450,14 @@ public synchronized void start() throws Exception {
...
         server.setHandler(handlers);
-        server.start();
+        try {
+            server.start();
+        } catch (BindException e) {
+            // retry once just in case it was some fluke or race
+            log.warn("Initial server start-up failed binding (retrying after a delay): "+e);
+            Time.sleep(Duration.millis(500));
+            server.start();
+        }
```
neykov: Wdyt about catching the second `BindException` in the test that used to fail (in case the backoff is not good enough) and dump some [troubleshooting details](https://github.com/apache/incubator-brooklyn/pull/657/files#diff-2219dcab34c5379452f6ac98854d8dd0R88)?

Seems the problem is rare and it's hard to get to the root cause just by re-running jobs on the slaves.

launcher/src/test/java/org/apache/brooklyn/launcher/CleanOrphanedLocationsIntegrationTest.java
ahgittin: really?  javadoc says `File(String pathname)`, no indication that it should accept a URI.  but i haven't tried.

in any case i think the extract is a better pattern than assuming resources are in `classes/` dir on filesystem -- wdyt?

launcher/src/main/java/org/apache/brooklyn/launcher/BrooklynWebServer.java
ahgittin: good idea for the future.  i maybe wouldn't do it here (in `/src/main` code), but instead attach it as a failure listener for all/selected tests.

ahgittin:
interesting comments @neykov -- but think this can be merged without those (once you've finished)

launcher/src/test/java/org/apache/brooklyn/launcher/CleanOrphanedLocationsIntegrationTest.java
neykov: There's the [File(URI)](https://docs.oracle.com/javase/7/docs/api/java/io/File.html#File(java.net.URI)) constructor. That's tests so it's pretty safe to assume they execute from the file system.

Since what you wrote works correctly I'm happy to leave it as is.

neykov:
Good idea about the listener. Merging.

ahgittin:
still seems not merged @neykov

aledsage:
LGTM as well; merging now :-)

neykov:
Thanks @ahgittin, got distracted and left it at `git push`.
Merged before seeing @aledsage's comment.

PR #559, opened 2017-02-15
Activity api include background

ahgittin:
As requested by @tbouron

PR #561, opened 2017-02-16
better presentation for sensor publishing tasks

ahgittin:
previously they didn't even have a name; now they have a nice name, description, and a tag.
there is some attempt to optimize the use of toString so it isn't hugely computationally expensive,
although this will increase expense a bit; i tend to think it's worth it for increased visibility.
(if we are publishing vast numbers of sensor events maybe we are doing something wrong, and if this
is identified as the bottleneck we can parameterise it, and in any case when we come to be multi-host
we'll need to revisit how we do this as currently we're thinking a good datastore is good enough for
the intended volumes, as opposed to a dedicated message bus.)

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/mgmt/internal/LocalSubscriptionManager.java
+++ b/core/src/main/java/org/apache/brooklyn/core/mgmt/internal/LocalSubscriptionManager.java
@@ -42,6 +42,7 @@ import org.apache.brooklyn.api.sensor.Sensor;
 import org.apache.brooklyn.api.sensor.SensorEvent;
 import org.apache.brooklyn.api.sensor.SensorEventListener;
 import org.apache.brooklyn.core.entity.Entities;
+import org.apache.brooklyn.core.mgmt.BrooklynTaskTags;
 import org.apache.brooklyn.core.sensor.BasicSensorEvent;
 import org.apache.brooklyn.util.collections.MutableList;
 import org.apache.brooklyn.util.collections.MutableMap;
@@ -86,10 +87,13 @@ public class LocalSubscriptionManager extends AbstractSubscriptionManager {
         return allSubscriptions.size();
     }
 
+    /** The total number of sensor change events generated (irrespective of number subscribers, see {@link #getTotalEventsDelivered()}) */
     public long getTotalEventsPublished() {
         return totalEventsPublishedCount.get();
     }
     
+    /** The total number of sensor change events submitted for delivery, counting multiple deliveries for multipe subscribers (see {@link #getTotalEventsPublished()}),
+     * but excluding initial notifications, and incremented when submitted ie prior to delivery */
     public long getTotalEventsDelivered() {
         return totalEventsDeliveredCount.get();
     }
@@ -111,7 +115,7 @@ public class LocalSubscriptionManager extends AbstractSubscriptionManager {
             s.subscriberExecutionManagerTagSupplied = true;
         } else {
             s.subscriberExecutionManagerTag = 
-                s.subscriber instanceof Entity ? "subscription-delivery-entity-"+((Entity)s.subscriber).getId()+"["+s.subscriber+"]" : 
+                s.subscriber instanceof Entity ? "subscription-delivery-entity-"+((Entity)s.subscriber).getId() : 
                 s.subscriber instanceof String ? "subscription-delivery-string["+s.subscriber+"]" : 
                 "subscription-delivery-object["+s.subscriber+"]";
             s.subscriberExecutionManagerTagSupplied = false;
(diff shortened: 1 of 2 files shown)


ahgittin:
noticed when working on #559 .  related to but not implementing ML discussion at this time around whether `entities/xxx/activities` API should default/optionally `includeBackground`.  when listing bg tasks this gets rid of the annoying blank ones in the list; now you see:

image

and clicking through you can see:

image


Note the addition of the `SENSOR` tag.  This could be used in UI's to filter sensor publication tasks /cc @tbouron .

Also note that if entity `child` publishes a sensor, eg in the course of `start`, and someone else, say `parent` is subscribed to it, the event does _not_ currently show up when you ask for `child`'s background tasks (because it runs with the `parent` as the context entity) nor the `start` task's subtasks (as we don't normally want to execute it as an in-queue subtask).  Thus that event will _not_ show up even if you `includeBackground` on the `start` task; it will only show up if you query `parent` entity's tasks (with `includeBackground` if we change the default for that endpoint as discussed on ML).  However it will still report the `child.start` as his "submitted by".  If that is a problem we should add an index on `submittedBy` for tasks so we can properly and efficiently get `includeBackground` tasks submitted by a task even across entity boundaries.  However ATM I think it is a minor but liveable quirk.  (In any case it isn't being introduced here, it is just becoming clearer because other things are being clarified.)

core/src/main/java/org/apache/brooklyn/core/mgmt/internal/LocalSubscriptionManager.java
```diff
@@ -111,7 +115,7 @@ public long getTotalEventsDelivered() {
             s.subscriberExecutionManagerTagSupplied = true;
         } else {
             s.subscriberExecutionManagerTag = 
-                s.subscriber instanceof Entity ? "subscription-delivery-entity-"+((Entity)s.subscriber).getId()+"["+s.subscriber+"]" : 
+                s.subscriber instanceof Entity ? "subscription-delivery-entity-"+((Entity)s.subscriber).getId() :
```
tbouron: Do we want to remove the subscriber name here?

core/src/main/java/org/apache/brooklyn/core/mgmt/internal/LocalSubscriptionManager.java
```diff
@@ -237,42 +212,85 @@ public synchronized boolean unsubscribe(SubscriptionHandle sh) {
...
+            .add(BrooklynTaskTags.SENSOR_TAG)
+            .build()
+            .asUnmodifiable();
+        
+        StringBuilder name = new StringBuilder("sensor ");
+        StringBuilder description = new StringBuilder("Sensor ");
+        String sensorName = s.sensor==null ? null : s.sensor.getName();
+        String sourceName = event.getSource()==null ? null : event.getSource().getId();
+        name.append(sourceName);
+        name.append(":");
```
tbouron: Should the `:` be appended **only** if the sourceName is different than `null`?

core/src/main/java/org/apache/brooklyn/core/mgmt/internal/LocalSubscriptionManager.java
```diff
@@ -237,42 +212,85 @@ public synchronized boolean unsubscribe(SubscriptionHandle sh) {
...
+        name.append(sensorName);
+        
+        description.append(sensorName);
+        description.append(" on ");
```
tbouron: Should the `on` be appended only if the `sensorName` is different than null?

core/src/main/java/org/apache/brooklyn/core/mgmt/internal/LocalSubscriptionManager.java
```diff
@@ -237,42 +212,85 @@ public synchronized boolean unsubscribe(SubscriptionHandle sh) {
...
+        description.append(sourceName);
+        description.append(" publishing to ");
+        description.append(s.subscriber instanceof Entity ? ((Entity)s.subscriber).getId() : s.subscriber);
+        
+        if (includeDescriptionForSensorTask(event)) {
+            name.append(" ");
+            name.append(event.getValue());
```
tbouron: Not sure we need the value in the name as we already have it in the description

tbouron:
Also great to have a `SENSOR` tag, this will allows us to do filtering more easily.

core/src/main/java/org/apache/brooklyn/core/mgmt/internal/LocalSubscriptionManager.java
ahgittin: yes, the tostring repeats the ID, adds the java type which might be nice, but really a bit clunky imo

core/src/main/java/org/apache/brooklyn/core/mgmt/internal/LocalSubscriptionManager.java
ahgittin: yes, good spot

PR #562, opened 2017-02-16
Provide a scratchpad area per management context

neykov:
Replaces analogous but incorrect usage of `ManagementContext.getConfig()`.

### DIFF
--- a/camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/BrooklynCampPlatform.java
+++ b/camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/BrooklynCampPlatform.java
@@ -65,7 +65,7 @@ public class BrooklynCampPlatform extends AggregatingCampPlatform implements Has
 
     /** finds and returns the {@link CampPlatform} registered for the management context, or null if none */
     @Nullable public static CampPlatform findPlatform(ManagementContext mgmt) {
-        return mgmt.getConfig().getConfig(BrooklynCampConstants.CAMP_PLATFORM);
+        return mgmt.getScratchpad().get(BrooklynCampConstants.CAMP_PLATFORM);
     }
     
     // --- brooklyn setup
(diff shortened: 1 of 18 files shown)


camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/test/lite/CampPlatformWithJustBrooklynMgmt.java
```diff
@@ -30,7 +30,9 @@
 
     public CampPlatformWithJustBrooklynMgmt(ManagementContext mgmt) {
         this.mgmt = mgmt;
+        // Deprecated in 0.11.0. Add to release notes and remove in next release.
```
geomacy: maybe tweak the comment to clarify that it is the `use of getConfig().put() on the following line` that is deprecated in favour of the use of the scratchpad. (and same goes for similar comments elsewhere)

neykov:
Good idea about a limited interface @geomacy. If you look at the change sites you'll find out that a `context` is hard to identify. On the other hand using a `ConfigKey` for the key gives you a pretty good idea what the value is. Perhaps I should enforce the use of `ConfigKeys` in the interface.

@tbouron I share your concerns so I added https://github.com/apache/brooklyn-docs/pull/150 as well. There are (still) valid uses of `put` on `BrooklynProperties` left in the code base so would be careful with adding logging there. Also it will start logging for the backwards compatibility changes which is also not a great UX.

tbouron:
@neykov Agree, having this warning within the docs seems to be the way to go.

neykov:
@geomacy Addressed comments, can you take another look.

camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/BrooklynCampPlatform.java
```diff
@@ -65,7 +65,7 @@ public BrooklynCampPlatform(PlatformRootSummary root, ManagementContext manageme
 
     /** finds and returns the {@link CampPlatform} registered for the management context, or null if none */
     @Nullable public static CampPlatform findPlatform(ManagementContext mgmt) {
-        return mgmt.getConfig().getConfig(BrooklynCampConstants.CAMP_PLATFORM);
+        return (CampPlatform) mgmt.getScratchpad().get(BrooklynCampConstants.CAMP_PLATFORM);
```
geomacy: Don't need the cast, here and elsewhere.

neykov:
Good point about casts, updated.

geomacy:
👍  lgtm, merging

PR #563, opened 2017-02-16
[WIP] Don't SO_REUSEADDR when probing for available ports

neykov:
Do not merge, the changes here are meant to test the behaviour on Apache's jenkins slaves.

The changes have the side effect of causing Brooklyn to bump it's default port on each restart if there were existing connections.

### DIFF
--- a/launcher/src/test/java/org/apache/brooklyn/launcher/WebAppRunnerTest.java
+++ b/launcher/src/test/java/org/apache/brooklyn/launcher/WebAppRunnerTest.java
@@ -78,14 +78,13 @@ public class WebAppRunnerTest {
     
     @Test
     public void testStartWar1() throws Exception {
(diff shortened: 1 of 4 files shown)


neykov:
Using non-`SO_REUSEADDR` sockets seems to work fine on the apache build slaves. On the other hand it prevents us from re-binding to the same port after restarting Brooklyn, so not ideal.
If the problems persist I suggest we add troubleshooting info in the tests. Things like the output of `netstat`, `ip addr show` & similar, which could guide us to the root cause of the failures. See https://github.com/apache/incubator-brooklyn/pull/657 for other ideas.

https://github.com/apache/incubator-brooklyn/pull/662 addresses a similar problem, worth having in mind if someone tries to dig in deeper.

PR #571, opened 2017-02-27
Better error reporting

ahgittin:
populate error field of `ApiError`, and better logging on REST errors

### DIFF
--- a/rest/rest-resources/src/main/java/org/apache/brooklyn/rest/resources/ApplicationResource.java
+++ b/rest/rest-resources/src/main/java/org/apache/brooklyn/rest/resources/ApplicationResource.java
@@ -275,6 +275,7 @@ public class ApplicationResource extends AbstractBrooklynRestResource implements
             spec = createEntitySpecForApplication(yaml);
         } catch (Exception e) {
             Exceptions.propagateIfFatal(e);
+            log.warn("Failed REST deployment, could not create spec: "+e);
             UserFacingException userFacing = Exceptions.getFirstThrowableOfType(e, UserFacingException.class);
             if (userFacing!=null) {
                 log.debug("Throwing "+userFacing+", wrapped in "+e);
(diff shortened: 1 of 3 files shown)


ahgittin:
Test failure due to Apache build system configuration:  `insufficient memory for the Java Runtime Environment to continue`

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/resources/ApplicationResource.java
```diff
@@ -374,7 +378,13 @@ public Response createPoly(byte[] inputToAutodetectType) {
 
 
         if (spec != null) {
-            return launch(potentialYaml, spec);
+            try {
+                return launch(potentialYaml, spec);
+            } catch (Exception e) {
+                Exceptions.propagateIfFatal(e);
+                log.warn("Failed REST deployment launching "+spec+": "+e);
+                throw WebResourceUtils.badRequest(e, "Error launching blueprint (autodetecting)");
```
aledsage: How can we tell this is a `badRequest`, rather than some other internal error?

aledsage:
@ahgittin bump: see previous comment - how can we tell it's a badRequest that caused the exception?

To force a rebuild by jenkins, you can do `git commit --amend; git push -f`.

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/resources/ApplicationResource.java
ahgittin: This seemed like the best status code to use and it's what we use on line 294/297.  Do you think there's a better one?  If we do nothing it will give Internal Server Error which seems worse?

ahgittin:
Tests pass -- merging.  If `badRequest` is too bad we can improve that subsequently.  As noted we use `badRequest` already if `launch` fails when we know it's YAML; if it fails in auto-detect mode we should probably throw the same.  Not giving an error equates to 500 server error which is probably wrong; chances are launch fails either because we couldn't parse or the synchronous launch steps are invalid, which unless the provided yaml references something that is broken, is going to be because the supplied yaml itself is broken IE **bad request**.

PR #572, opened 2017-02-27
Increase timeout in ReachableSocketFinderTest

sjcorbett:
Use the default grace period. I think leaving it as `ZERO` was an oversight on my behalf.


sjcorbett:
@geomacy @neykov could you review this?

geomacy:
+1 Looks good to me, merging.

PR #573, opened 2017-02-28
Do not runtime-inherit catalog item in the spec to fix config key list problem

ahgittin:
fixes https://issues.apache.org/jira/browse/BROOKLYN-445 and failing tests included here

means catalog BOMs should be written such that any entity/policy/etc which wants to access resources in the bundle is defined as its own item

### DIFF
--- a/api/src/main/java/org/apache/brooklyn/api/internal/AbstractBrooklynObjectSpec.java
+++ b/api/src/main/java/org/apache/brooklyn/api/internal/AbstractBrooklynObjectSpec.java
@@ -75,7 +75,6 @@ public abstract class AbstractBrooklynObjectSpec<T,SpecT extends AbstractBrookly
     protected AbstractBrooklynObjectSpec(Class<? extends T> type) {
         checkValidType(type);
         this.type = type;
-        this.catalogItemId = ApiObjectsFactory.get().getCatalogItemIdFromContext();
     }
     
     @SuppressWarnings("unchecked")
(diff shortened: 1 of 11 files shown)


neykov:
Changes look good.
Do I correctly understand that the following is no longer possible?
```
brooklyn.catalog:
  brooklyn.libraries:
    ...
  items:
  - id: test
    version: 0.0.1-SNAPSHOT
    item:
      type: Application
      brooklyn.children:
      - type: VanillaSoftwareProcess
         brooklyn.config:
           files.install:
             classpath://something-in-libraries: uploaded-script.sh
```

Or is it some other case that changes behaviour?

ahgittin:
@neykov correct.  I will add this as a failing test and suggest that this PR:

    https://github.com/apache/brooklyn-server/pull/338

seek to confirm that the test passes.

(marking this WIP until i fix)

ahgittin:
@neykov Prescient suggestions.  Completely forget there were _two_ other paths where `catalogItemId` was being set to an inheritance/context value.

It doesn't cause problems with config keys but it is messy and should be fixed.  Natural place to do it is in #338 -- have added several more tests and TODO items here which should be looked at as part of that.

ahgittin:
@neykov @geomacy Think I've tidied the last bits.  Could this be merged?  Then we can focus on #338 getting in.

neykov:
The changes include unrelated change in the class loading fallbacks - increasing the priority of the bundle prefix in relation to the other fallbacks. Contrary to the comment in https://github.com/apache/brooklyn-server/pull/573/commits/40330602af958d1e3486feea81933168d44e936d, the changes include behaviour changes, not only logging improvements. Would've been useful to have it in a separate commit + PR for better visibility if we ever need to come back to the changes.
The `RedisClusterIntegrationTest` would've been better added to the `brooklyn-camp` project which is the place to add CAMP tests which don't fit elsewhere - we really need to split them in an camp-integration-test project.
Comments above are not worth blocking the merge. Merging now.

sjcorbett:
Regarding `RedisClusterIntegrationTest`, it specifically only failed when run through the http server. The same blueprint passed when tested with `AbstractYamlTest`.

ahgittin:
Quite right re https://github.com/apache/brooklyn-server/pull/573/commits/40330602af958d1e3486feea81933168d44e936d -- hit a weird bug in testing related to this but missed it when making the PRs.

Agree re `Redis` except for the point Sam mentions.  Good to have a few tests that go through the server.

PR #574, opened 2017-02-28
don't NPE fail is application is GC'd at same time as location

ahgittin:
i think we want to make sure a record is generated, even if it is incomplete


aledsage:
LGTM; merging.

PR #575, opened 2017-02-28
Clean up location metadata and OSGi packaging

ahgittin:
Discovered when trying to use the REST API that we didn't get much of the location detail we might want; in particular we didn't have info on provider/endpoint/region, and the parent spec was not accurately identified.  Tracked the latter to confusion around `ORIGINAL_SPEC` which I've clarified and deprecated the confusing other field, and the former to a place where flags weren't pulled in, just config.

Also had trouble accessing some of the data because it required a cast to `JcloudsLocation`, which pulled in an OSGi dependency on the main brooklyn-jclouds package, and classes in that package pulled in jclouds bundles through their signatures, giving a huge set of bundle constraints in my downstream project.  Have refactored here so that we can access many common things through an `api` package which doesn't pull in jclouds dependencies.

Finally, noticed some other places where more machine detail would probably be wanted, but wasn't provided.

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/location/BasicLocationRegistry.java
+++ b/core/src/main/java/org/apache/brooklyn/core/location/BasicLocationRegistry.java
@@ -347,7 +347,12 @@ public class BasicLocationRegistry implements LocationRegistry {
             .putAll(locationFlags)
             .putIfAbsentAndNotNull(LocationInternal.NAMED_SPEC_NAME, ld.getName())
             .putIfAbsentAndNotNull(LocationInternal.ORIGINAL_SPEC, ld.getName());
-        return getLocationSpec(ld.getSpec(), newLocationFlags.getAllConfigRaw());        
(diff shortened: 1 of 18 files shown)


core/src/main/java/org/apache/brooklyn/core/location/BasicLocationRegistry.java
```diff
@@ -372,7 +377,10 @@ public final Location getLocationManaged(String spec, Map locationFlags) {
 
             if (resolver != null) {
                 try {
-                    return (Maybe) Maybe.of(resolver.newLocationSpecFromString(spec, locationFlags, this));
+                    LocationSpec<? extends Location> specO = resolver.newLocationSpecFromString(spec, locationFlags, this);
```
duncangrant: I feel like the naming in this function has gotten a bit lost.  We have a spec and a specO and it took me a bit too much time to understand what I was seeing.  Can't think of a better naming convention now.

PR #576, opened 2017-02-28
BundleMaker (part1)

aledsage:
This is extracted from https://github.com/apache/brooklyn-server/pull/485 - it excludes the changes to the REST api; it improves test coverage and fixes a couple of things in `BundleMaker` (generics, and NPE in `createJarFromClasspathDir` when there is no manifest).

The intent is that we can merge this soon, without having to wait for consensus on the REST api changes.

The first commit is a squashing-together of @ahgittin's  two commits in https://github.com/apache/brooklyn-server/pull/485, but with the rest-api changes removed. I made no other changes to that commit (hence the author being Alex).

I've also marked `BundleMaker` as `@Beta` - it's really an internal tool; we don't want end-users relying on it as part of our public api. Would be good to have a nicer way of indicating what is a "public api"!


ahgittin:
looks good

PR #577, opened 2017-03-01
Bugfix. Move a test from src/main to src/test

bostko:


ahgittin:
Oops!  Thanks @bostko -- done.

PR #579, opened 2017-03-02
Log User info on effector invocation via REST API

googlielmo:
- Add user information


geomacy:
Mark's happy; merging.

PR #580, opened 2017-03-02
Set the username for Windows locations correctly

neykov:
Not able to provision windows machines with current master. The username when deploying to a winrm target is always empty. Fix it by following the same approach the SSH branch takes to populate the username.

Tested by running `WinRmMachineLocationLiveTest` against Amazon.


sjcorbett:
Looks good. `WinRmMachineLocationLiveTest` passed for me.

PR #583, opened 2017-03-08
Generalised Entities.waitFor()

alasdairhodge:
Provides general-purpose blocking wait on any entity predicate. Example:
````
    Entities.waitFor(
        machineEntity,
        EntityPredicates.attributeEqualTo(MACHINE_STATE, MachineState.REBOOTING),
        Duration.FIVE_SECONDS);
````

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/entity/Entities.java
+++ b/core/src/main/java/org/apache/brooklyn/core/entity/Entities.java
@@ -18,6 +18,9 @@
  */
 package org.apache.brooklyn.core.entity;
 
+import static org.apache.brooklyn.core.entity.EntityPredicates.attributeEqualTo;
+import static org.apache.brooklyn.util.guava.Functionals.isSatisfied;
+
 import java.io.Closeable;
 import java.io.IOException;
 import java.io.PrintWriter;
@@ -32,7 +35,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.Stack;
-import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Semaphore;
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by bostko** [comment]
LGTM

File: core/src/main/java/org/apache/brooklyn/core/entity/Entities.java [hunk 1]
```diff
@@ -1139,6 +1141,29 @@ public static void warnOnIgnoringConfig(Entity entity, ConfigKey<?> key) {
...
+            String description = "Waiting for " + condition + " on " + entity;
+            Tasks.setBlockingDetails(description);
+
+            Repeater repeater = Repeater.create(description)
+                .until(isSatisfied(entity, condition))
+                .limitTimeTo(timeout)
+                .backoffTo(Duration.ONE_SECOND)
+                .rethrowException();
+
+            if (!repeater.run()) {
```
geomacy: How about making this capture any exception with `runKeepingError` so that the `throw` below could set it as a cause? Also then you can change the message below to refer to whatever the actual error was, rather than assuming it's a timeout.

core/src/test/java/org/apache/brooklyn/core/entity/EntitiesTest.java
```diff
@@ -130,5 +129,21 @@ public void testCreateGetContainsAndRemoveTags() throws Exception {
...
+    @Test
+    public void testWaitFor() throws Exception {
+        entity = app.createAndManageChild(EntitySpec.create(TestEntity.class));
+        Duration timeout = Duration.ONE_MILLISECOND;
+
+        Entities.waitFor(entity, applicationIdEqualTo(app.getApplicationId()), timeout);
+
+        try {
+            Entities.waitFor(entity, applicationIdEqualTo(app.getApplicationId() + "-wrong"), timeout);
+            Assert.fail();
```
geomacy: Prefer `org.apache.brooklyn.test.Asserts#shouldHaveFailedPreviously()`

core/src/test/java/org/apache/brooklyn/core/entity/EntitiesTest.java
```diff
@@ -130,5 +129,21 @@ public void testCreateGetContainsAndRemoveTags() throws Exception {
...
+
+        } catch (Exception e) {
+            // expected timeout
```
geomacy: Do an assertion on the exception so you're sure it's really what you expected, using one of the `Asserts.expectedFailureContains...` variants.

core/src/main/java/org/apache/brooklyn/core/entity/Entities.java
alasdairhodge: Timeout is indicated by the `run` method returning `false`, rather than via an exception. Since we set `rethrowException`, anything thrown by the passed `Callable` will be rethrown on timeout.

alasdairhodge: On the subject of exceptions, I toyed with the idea of throwing `TimeoutException` instead of `IllegalStateException`, but am opting for consistency/backwards compatibility with the existing `waitForServiceUp` methods. (The standard `TimeoutException` is checked.)

alasdairhodge:
Closing in favour of #584 against `master`.

PR #586, opened 2017-03-09
Conveniences for yaml

aledsage:
A few conveniences to make it easier to write my yaml blueprints! See individual commits for details.

If anything in here looks controversial to you then please let me know, and I'll split things out into separate PRs.

### DIFF
--- a/camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/methods/BrooklynDslCommon.java
+++ b/camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/methods/BrooklynDslCommon.java
@@ -63,6 +63,7 @@ import org.apache.brooklyn.util.core.task.Tasks;
 import org.apache.brooklyn.util.exceptions.Exceptions;
 import org.apache.brooklyn.util.guava.Maybe;
 import org.apache.brooklyn.util.javalang.Reflections;
+import org.apache.brooklyn.util.javalang.coerce.TypeCoercer;
 import org.apache.brooklyn.util.net.Urls;
 import org.apache.commons.beanutils.BeanUtils;
 import org.slf4j.Logger;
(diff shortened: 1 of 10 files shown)


tbouron:
Tested, LGTM.

I'm just not understanding the use-case for the `deferred` in `$brooklyn:object`, could you give a bit more details on this @aledsage ?

aledsage:
@tbouron Thanks. My use-case for `deferred` is a bit unusual! The "general" use-case is where you want to delay the creation/evaluation of the object until it's actually needed. And to have it being called every time it is evaluated, rather than evaluated only once and its value stored. One might want this for a number of reasons:

1. If evaluated statically, then that result is persisted (which may include confidential info). If `deferred`, then the persisted value will be the DSL.

2. There is a difference in behaviour for if a `$brooklyn:object` has an arg that is deferred (e.g. an arg that is `$brooklyn:config(...)`) versus if all the args can be statically evaluated: with the former, the value is the "deferred DSL"; with the latter, it is evaluated immediately. This allows one to force the two ways to behave the same.

My use-case is a bit sneaky (and one could argue misusing config, because the value changes each time you ask for it)! I am writing a yaml-based test for a service, and am using `org.apache.brooklyn.test.framework.TestHttpCall`. When constructing the URL, I need to include a timestamp (for "now") in the url query parameters. I therefore want some way to get the current timestamp at the time the url is being used (rather the time when the entity was constructed). For example:

```
      type: org.apache.brooklyn.test.framework.TestHttpCall
      brooklyn.config:
        url:
          $brooklyn:formatString:
          - "%s/usage/?startTime=2016-01-01T00:00:00Z&endTime=%s"
          - $brooklyn:component($brooklyn:config("targetId")).attributeWhenReady("main.uri")
          - $brooklyn:formatString:
            - "%sZ"
            - $brooklyn:object:
                type: org.apache.brooklyn.util.time.Time
                factoryMethod.name: makeDateString
                factoryMethod.args:
                - $brooklyn:object:
                    deferred: true
                    type: java.lang.System
                    factoryMethod.name: currentTimeMillis
                - yyyy-MM-dd'T'HH:mm:ss
                - UTC
```

Without the `deferred: true`, the `endTime` is the time at which the entity was constructed. With the `deferred`, it is the time at which the entity's `start` effector is called (i.e. when it reads the `url` config value).

tbouron:
@aledsage That makes sense, the example helped a lot to understand.

LGTM then

aledsage:
Thanks @tbouron - merging now.

PR #590, opened 2017-03-13
Use boolean Options correctly in ItemLister

sjcorbett:
Replaces list-objects' `--catalog-only` and `--ignore-impls` arguments with `--all-classes` and `--include-impls` respectively.

ItemLister defined a few options as booleans whose default was true. Airline treats boolean options as having zero arity and whose existence implies truth. See Airline's [MetadataLoader](https://github.com/airlift/airline/blob/0.7/src/main/java/io/airlift/airline/model/MetadataLoader.java#L156) class.

I've removed the default from all other boolean options to make it clearer that the value is set by Airline.

CloudExplorerLiveTest passes.

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/enricher/AbstractEnricher.java
+++ b/core/src/main/java/org/apache/brooklyn/core/enricher/AbstractEnricher.java
@@ -45,8 +45,9 @@ import com.google.common.collect.Maps;
 */
 public abstract class AbstractEnricher extends AbstractEntityAdjunct implements Enricher {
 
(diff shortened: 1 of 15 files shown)


neykov:
LGTM, merging. @tbouron are there any docs that should be updated for the community catalog?

tbouron:
Yes, there is one: the `reference` field in https://github.com/brooklyncentral/blueprint-repository

Good catch @neykov !

neykov:
Thanks @tbouron. Looks like it's good, doesn't use the options changed here.

PR #591, opened 2017-03-13
Fix ssh login when provider's image requires TTY

ygy:


bostko:
I would move tty disable flag after the createUser statement.

bostko:
Is it possible that `PROP_ALLOCATE_PTY` breaks for Centos 6 images?
Is it tested on other clouds. I thing it should be optional.
I have to suggestions for implementing it:
- Move DONT_REQUIRE_TTY_FOR_SUDO check above. Unfortunatelly this may require rearanging  of wait for reachable as well.
- Make `SshMachineLocation.SSH_CONFIG_GIVEN_TO_PROPS` configurable or make that parameter configurable.

aledsage:
LGTM.

In jclouds, it always sets allocatePTY when using sshj: https://github.com/jclouds/jclouds/blob/rel/jclouds-2.0.0/drivers/sshj/src/main/java/org/jclouds/sshj/SshjSshClient.java#L438
In contrast, brooklyn only sets it if configured to do so: https://github.com/apache/brooklyn-server/blob/rel/apache-brooklyn-0.10.0/core/src/main/java/org/apache/brooklyn/util/core/internal/ssh/sshj/SshjTool.java#L837-L839
Brooklyn defaults this to not allocatePTY: https://github.com/apache/brooklyn-server/blob/rel/apache-brooklyn-0.10.0/core/src/main/java/org/apache/brooklyn/util/core/internal/ssh/SshTool.java#L75

Brooklyn will then by default execute a command to disable requiretty: https://github.com/apache/brooklyn-server/blob/rel/apache-brooklyn-0.10.0/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java#L904-L915
That behaviour is configurable via https://github.com/apache/brooklyn-server/blob/rel/apache-brooklyn-0.10.0/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocationConfig.java#L179-L188

Making this change just means we do the same as jclouds already does. This change will only affect us if we've set `useJcloudsSshInit: false`, and then only for the initial createUser command execution.

We have much more thoroughly tested the jclouds behaviour across many clouds, because `useJcloudsSshInit` defaults to true.

aledsage:
@bostko I've tested this on aws and softlayer, on centos 6 & 7, and ubuntu 16, 14, 12 and 10. I'll add some more live tests for those in a separate PR later.

I don't think we should rely on `DONT_REQUIRE_TTY_FOR_SUDO`: that controls whether we execute an extra command so that we disable requiretty in the `/etc/sudoers` file! It might be set to false. Note that the extra command also sets `SshTool.PROP_ALLOCATE_PTY` to true (as @ygy is doing here).

`SshMachineLocation.SSH_CONFIG_GIVEN_TO_PROPS` feels overkill here. This is an unusual use-case: it only affects the arg passed to the single createUser command, and then only if one has set `useJcloudsSshInit: false`. I think that a user would sensibly expect a config option like `SSH_CONFIG_GIVEN_TO_PROPS` to be used more widely. It's unclear to me where else we'd want to use it, given one can pass through other config for the SshMachineLocation to use already.

aledsage:
@bostko is there a specific reason you think it would break CentOS 6 deployments? Or just that you wanted to ensure it had been tested before this was merged?

bostko:
I wanted to check whether it is tested against CentOS 6.
Thank you for inspecting that!

PR #593, opened 2017-03-14
BROOKLYN-451: Adds yaml-dsl rebind tests (for historic state)

aledsage:
Tests + fixes https://issues.apache.org/jira/browse/BROOKLYN-451

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/util/core/xstream/ObjectWithDefaultStringImplConverter.java
+++ b/core/src/main/java/org/apache/brooklyn/util/core/xstream/ObjectWithDefaultStringImplConverter.java
@@ -1,0 +1,89 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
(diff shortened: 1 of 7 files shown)


core/src/main/java/org/apache/brooklyn/util/core/xstream/XmlSerializer.java
```diff
@@ -80,6 +80,11 @@ protected MapperWrapper wrapMapper(MapperWrapper next) {
 
         xstream.registerConverter(new EnumCaseForgivingConverter());
         xstream.registerConverter(new Inet4AddressConverter());
+        
+        // See ObjectWithDefaultStringImplConverter (and its usage) for why we want to auto-detect 
+        // annotations (usages of this is in the camp project, so we can't just list it statically
+        // here unfortunately).
+        xstream.autodetectAnnotations(true);
```
aledsage: What are the performance implications of this?!

m4rkmckenna: From http://x-stream.github.io/annotations-tutorial.html#AutoDetect

> In auto-detection mode XStream will have to examine any unknown class type for annotations. This will slow down the marshalling process until all processed types have been examined once.

core/src/main/java/org/apache/brooklyn/util/core/xstream/ObjectWithDefaultStringImplConverter.java
```diff
@@ -0,0 +1,86 @@
...
+    private final Class<?> defaultImpl = String.class;
+
+    public ObjectWithDefaultStringImplConverter(ConverterLookup lookup, ClassLoader loader) {
+        this.lookup = lookup;
+        this.loader = loader;
+    }
+
+    @Override
+    public boolean canConvert(@SuppressWarnings("rawtypes") Class type) {
+        return true;
```
m4rkmckenna: Worth adding a comment here as to why this is hardcoded

neykov:
LGTM

core/src/main/java/org/apache/brooklyn/util/core/xstream/XmlSerializer.java
aledsage: Seems to be about 5-10% overhead (when testing with a simple EntityMemento that has lots of string keys and brooklyn.parameters). That's certainly acceptable, to fix the issue!

aledsage:
Thanks @neykov @m4rkmckenna. I've added the comment you suggested. Merging now.

PR #594, opened 2017-03-14
Revert "add original failing test from @sjcorbett"

Graeme-Miller:
This reverts commit d7cbaeed46d7fedf8e49df015baa812352060f45.

That commit added a dependency on brooklyn-software-nosql.

However, this causes a circular dependency as brooklyn-software-nosql depends on brooklyn-server.


ahgittin:
I don't see where the circular dependency is.  I believe you, but I'm confused!  `brooklyn-software-nosql` doesn't depend on `rest-server` -- looks like the only thing that does is `launcher`.

This is a useful test as REST seems to configure things slightly differently.  Feels okay for `rest-server` to have test deps elsewhere ... but if there is a better place e.g. `qa` that's fine too.  Would prefer not to revert but if the circular dep is a blocker fine to revert and then we can add elsewhere.  But @Graeme-Miller can you give more details on this circular dep?

Graeme-Miller:
Ok, so the circular dependency is:
* dependency has been added between brooklyn-rest-server and brooklyn-software-nosql
* brooklyn-rest-server is a part of brooklyn-server, as is brooklyn-core.
* brooklyn-software-nosql depends on brooklyn-core.

Brooklyn builds brooklyn-server first, then later it builds brooklyn-software-nosql. With this added dependency, that fails as brooklyn-server relies on brooklyn-software-nosql. 

We can't reverse the order as brooklyn-software-nosql relies on brooklyn-server (because of its brooklyn-core dependency).

In short, if you update the version numbers and try and rebuild, it will fail. As brooklyn-software-nosql has not been built by the time brooklyn-server is being built.

ahgittin:
Gotcha. It's a sequential dependency as software-nosql is in library built
after software. D'oh. Thanks Graeme.

@sjcorbett where should we move it? The qa project?

Best
Alex

sjcorbett:
@ahgittin In the test's current form it sounds like the qa project is sensible. Alternatively you could rephrase it to drop the use of Redis. It was coincidental that I found the problem while experimenting with that entity.

ahgittin:
I've added the test instead to the `qa` project, in https://github.com/apache/brooklyn-library/pull/94 .

Reverted the changes here, adding a manual commit to restore access to the `server` field.  Thanks @Graeme-Miller .

PR #595, opened 2017-03-15
Workaround for WinRmCommandSensor to see CmdFeed in OSGi

bostko:
Move CmdFeed to the same package as WinRmCommandSensor


aledsage:
FYI the problem is that the package `org.apache.brooklyn.feed.windows` is used in two different OSGi bundles: `brooklyn-software-winrm` contains `CmdFeed` and `WindowsPerformanceCounterFeed`; and `brooklyn-core` contains `WindowsPerformanceCounterPollConfig`. When another bundle imports the package, it can only get it from one other bundle (that exports the package). Therefore we get one or other of these.

@bostko why do we have `WindowsPerformanceCounterPollConfig` in core, rather than in `brooklyn-software-winrm`? I'm inclined to move that from core, and delete the package from core.

Thoughts?

aledsage:
Looking at the git history, the `WindowsPerformanceCounterFeed` was moved to brooklyn-software-winrm in the commit 45308895d59c3e1ddf7828f1d63d6723cc671f0c, but `WindowsPerformanceCounterPollConfig` was left where it was. This has been broken in OSGi ever since (because of the split package).

My guess is that `WindowsPerformanceCounterPollConfig` was not moved because it doesn't strictly have a dependency on winrm code, because it's really just config. In hindsight, I think that was a mistake.

bostko:
That worked, thank you!

aledsage:
The jenkins failure is unrelated - it looks like an environment problem:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:2.19.1:test (default-test) on project camp-base: Execution default-test of goal org.apache.maven.plugins:maven-surefire-plugin:2.19.1:test failed: The forked VM terminated without properly saying goodbye. VM crash or System.exit called?
```

aledsage:
LGTM; merging.

PR #596, opened 2017-03-15
WinRmCommandSensor default executionDir

bostko:
Use default dir obtained on the winrm sesion
until Apache Brooklyn has working concept for RUN_DIR on VanillaWindowsProcess

### DIFF
--- a/software/winrm/src/main/java/org/apache/brooklyn/core/sensor/windows/WinRmCommandSensor.java
+++ b/software/winrm/src/main/java/org/apache/brooklyn/core/sensor/windows/WinRmCommandSensor.java
@@ -144,15 +144,7 @@ public final class WinRmCommandSensor<T> extends AddSensor<T> {
     public static String makeCommandExecutingInDirectory(String command, String executionDir, Entity entity) {
         String finalCommand = command;
         String execDir = executionDir;
-        if (Strings.isBlank(execDir)) {
-            // default to run dir
-            execDir = entity.getAttribute(BrooklynConfigKeys.RUN_DIR);
-            // if no run dir, default to home
-            if (Strings.isBlank(execDir)) {
-                execDir = "%USERPROFILE%";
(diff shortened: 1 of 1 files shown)


software/winrm/src/main/java/org/apache/brooklyn/core/sensor/windows/WinRmCommandSensor.java
```diff
@@ -144,15 +144,7 @@ public T apply(String input) {
...
-        if (Strings.isBlank(execDir)) {
-            // default to run dir
-            execDir = entity.getAttribute(BrooklynConfigKeys.RUN_DIR);
-            // if no run dir, default to home
-            if (Strings.isBlank(execDir)) {
-                execDir = "%USERPROFILE%";
-            }
-        }
-        if (!"~".equals(execDir)) {
+        if (Strings.isNonBlank(execDir) && !"~".equals(execDir)) {
```
neykov: That's nice as a temporary patch around the fact that `RUN_DIR` is set to a UNIX-y path. What's the default folder we get on login? Is there a risk in executing in a protected folder (`C:\Windows\system32`) which will fail the code if it writes to the file system?
What if we keep the `%USERPROFILE%` default?
I thought this is executed as powershell, which makes the next line look like it would fail - how does it work then?

bostko: Default directory is `%USERPROFILE%`.
You can try by setting a sensor from `command: dir`.
We already use the same default directory for launch.command and other commands in `VanillaWindowsProcess`.

bostko: That's why empty has the same meaning as `~`.

neykov:
LGTM

PR #598, opened 2017-03-16
BROOKLYN-453: Fix rebind for classrename with bundle prefix

aledsage:
The historic persisted state added here for testing comes from a clocker.io blueprint.

The "Fix classRenames for SshPollConfig anonymous classes" is me playing it safe, to re-add the anonymous inner classes that might be referenced by customers out there. As the comment in the code says, we can't just use `deserializingClassRenames.properties`, because persistence format for a static versus anonymous (non-static) inner class is different.

For the "Fix xstream deserialising osgi class renames", I discussed this with @neykov first. We both agreed that we don't particularly like the fix, but that at least it's localised and it fixes the problem (which is causing rebind to older versions of Brooklyn to fail). We'd like to revisit this, and see about pushing the calls to the class-renames into `ClassLoaderUtils` perhaps. But that's a separate discussion, not for this PR.

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/util/core/xstream/ClassRenamingMapper.java
+++ b/core/src/main/java/org/apache/brooklyn/util/core/xstream/ClassRenamingMapper.java
@@ -22,32 +22,168 @@ import static com.google.common.base.Preconditions.checkNotNull;
 
 import java.util.Map;
 
+import org.apache.brooklyn.core.mgmt.persist.OsgiClassPrefixer;
 import org.apache.brooklyn.util.guava.Maybe;
 import org.apache.brooklyn.util.javalang.Reflections;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.google.common.base.Supplier;
+import com.thoughtworks.xstream.mapper.CannotResolveClassException;
 import com.thoughtworks.xstream.mapper.Mapper;
 import com.thoughtworks.xstream.mapper.MapperWrapper;
 
+/**
+ * An xstream mapper that handles class-renames, so we can rebind to historic persisted state.
+ */
 public class ClassRenamingMapper extends MapperWrapper {
+    
+    /*
+     * TODO There is a strange relationship between this and XmlMementoSerializer$OsgiClassnameMapper.
+     * Should these be perhaps merged?
+     * 
+     * TODO For class-loading on deserialzation, should we push the class-rename logic into 
+     * org.apache.brooklyn.util.core.ClassLoaderUtils instead? Does the xstream mapper do
+     * anything else important, beyond that class-loading responsibility? It's registration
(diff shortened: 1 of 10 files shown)


aledsage:
jenkins test failure looked unrelated - we should investigate that separately. I've amended the commits to force jenkins to build again.

For the record, here's the test failure:
```
2017-03-16 00:39:32,545 INFO  TESTNG INVOKING: "Surefire test" - org.apache.brooklyn.entity.software.base.SoftwareProcessStopsDuringStartTest.testStopDuringProvisionWaitsForCompletion()
2017-03-16 00:39:32,549 INFO  Stopping EmptySoftwareProcessImpl{id=ax1j5mlygp} in []
2017-03-16 00:39:32,550 INFO  When stopping EmptySoftwareProcessImpl{id=ax1j5mlygp}, waiting for up to 10m for the machine to finish provisioning, before terminating it
2017-03-16 00:39:32,554 WARN  Deprecated use of unmanaged location (SshMachineLocation[SshMachineLocation:y2op:null@localhost/127.0.0.1:22(id=y2ope2zmax)]); will be managed automatically now but not supported in future versions
2017-03-16 00:39:32,555 INFO  Simulated obtain of machine SshMachineLocation[SshMachineLocation:y2op:null@localhost/127.0.0.1:22(id=y2ope2zmax)]
2017-03-16 00:39:32,574 WARN  Request for bundle 'org.apache.brooklyn.software-winrm' (0.11.0-SNAPSHOT) will be ignored, loading 'org.apache.brooklyn.location.winrm.WinRmMachineLocation' as no framework available
2017-03-16 00:39:32,730 INFO  TESTNG FAILED: "Surefire test" - org.apache.brooklyn.entity.software.base.SoftwareProcessStopsDuringStartTest.testStopDuringProvisionWaitsForCompletion() finished in 183 ms
java.lang.AssertionError: lists don't have the same size expected [2] but found [1]
	at org.apache.brooklyn.entity.software.base.SoftwareProcessStopsDuringStartTest.testStopDuringProvisionWaitsForCompletion(SoftwareProcessStopsDuringStartTest.java:185)
```

core/src/main/java/org/apache/brooklyn/util/core/xstream/ClassRenamingMapper.java
```diff
@@ -22,32 +22,168 @@
...
+    
+    public ClassRenamingMapper(Mapper wrapped, Map<String, String> nameToType, Supplier<? extends ClassLoader> classLoaderSupplier) {
         super(wrapped);
         this.nameToType = checkNotNull(nameToType, "nameToType");
+        this.classLoaderSupplier = checkNotNull(classLoaderSupplier, "classLoaderSupplier");
     }
     
     @Override
     public Class<?> realClass(String elementName) {
+        String elementNamOrig = elementName;
```
neykov: Typo elementNam**e**Orig

PR #599, opened 2017-03-20
use http://httpbin instead of https for tests

andreaturli:
this should fix the build failures we are seeing after 16th March

cc @geomacy please have a look at this

### DIFF
--- a/camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/CompositeEffectorYamlTest.java
+++ b/camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/CompositeEffectorYamlTest.java
@@ -37,7 +37,7 @@ import com.google.common.collect.Iterables;
 public class CompositeEffectorYamlTest extends AbstractYamlTest {
     private static final Logger log = LoggerFactory.getLogger(CompositeEffectorYamlTest.class);
 
-    @Test
+    @Test(groups="Integration")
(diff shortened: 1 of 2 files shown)


aledsage:
LGTM; waiting for jenkins to finish and then good to merge.

andreaturli:
looks there is another problem with ssl in another test. Wonder if it's a jenkins issue?

ahgittin:
Make sure unit tests run with no internet connectivity (no DNS lookup etc).  That might not be the Jenkins problem but it is a requirement of our unit tests.  If you need to do an http call, run a local mock server, or make it an integration test.

andreaturli:
I agree `@Test(groups="Integration")` is better for this test.

Notice, though, the failing tests is different this time:

org.apache.brooklyn.camp.brooklyn.CompositeEffectorYamlRebindTest.testRebindWhenHealthy (from TestSuite)

geomacy:
@andreaturli CompositeEffectorYamlRebindTest also uses httpbin.org, so it would need to be an integration test too.  The full list looks like 
- CompositeEffectorYamlRebindTest
- CompositeEffectorYamlTest
- HttpCommandEffectorYamlRebindTest
- HttpCommandEffectorYamlTest
and HttpCommandEffectorIntegrationTest which already has its tests as integration tests.

geomacy:
Looks good to me but will need a couple more tests marked integration. I will merge this and do those tests.

PR #601, opened 2017-03-20
Make a couple more tests be integration tests.

geomacy:
See https://github.com/apache/brooklyn-server/pull/599


geomacy:
That looks a bit better. Will merge this.

PR #602, opened 2017-03-20
Add test illustrating SNI connection problem.

geomacy:
See https://issues.apache.org/jira/browse/BROOKLYN-456.


sjcorbett:
Looks good.

PR #603, opened 2017-03-21
WinRmMachineLocation obtaining WinRmTool parameters via its config().get

bostko:
- This allows WinrmTool props to be supplied via external configuration providers.
- WinRmTool instantiation is made to be the same as in SshMachineLocation#connectSsh

### DIFF
--- a/camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/WinRmMachineLocationExternalConfigYamlTest.java
+++ b/camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/WinRmMachineLocationExternalConfigYamlTest.java
@@ -1,0 +1,99 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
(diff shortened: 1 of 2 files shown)


neykov:
LGTM. @bostko can you add a test to verify the changes work.

bostko:
@neykov as far as I can see test for this requires two things:
- BrooklynCampPlatformLauncherNoServer in order external config to work https://github.com/apache/brooklyn-server/blob/master/camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/BrooklynCampPlatform.java#L90
- WinRmMachineLocation in order to test it

Could you advice on how to approach writing such test?

neykov:
* You'd need your test living in `brooklyn-camp`, see `ExternalConfigBrooklynPropertiesTest` for an example on how to set it up
* Have a look at tests using `RecordingWinRmTool` for examples of a unit-testable winrm location. See `AbstractJcloudsStubYamlTest` for a yaml-accessible location
* `JcloudsExternalConfigYamlTest` might be just what you need as a starting point

bostko:
I missed to add that my concern is that I do not know find suitable place to put such test.
Adding test in brooklyn-camp or brooklyn-software-winrm will introduce circular dependency.

neykov:
`AbstractJcloudsStubYamlTest` in `brooklyn-camp` already depends on `brooklyn-software-winrm`. Adding the test in `brooklyn-camp` would be fine.

neykov:
Thanks @bostko, merging.

aledsage:
This fixes https://issues.apache.org/jira/browse/BROOKLYN-476

PR #605, opened 2017-03-22
Add series of meta effectors for composite operations

grkvlt:
Adds `ComposeEffector`, `SequenceEffector`, `TransformEffector` and `LoopEffector` for orchestrating effector calls in blueprints.

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/effector/composite/AbstractCompositeEffector.java
+++ b/core/src/main/java/org/apache/brooklyn/core/effector/composite/AbstractCompositeEffector.java
@@ -1,0 +1,184 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.brooklyn.core.effector.composite;
+
+import java.util.Map;
+import java.util.Set;
+
+import org.apache.brooklyn.api.effector.Effector;
+import org.apache.brooklyn.api.entity.Entity;
+import org.apache.brooklyn.api.entity.EntityLocal;
+import org.apache.brooklyn.core.effector.AddEffector;
+import org.apache.brooklyn.core.effector.EffectorBody;
+import org.apache.brooklyn.util.collections.CollectionFunctionals;
+import org.apache.brooklyn.util.core.config.ConfigBag;
+import org.apache.brooklyn.util.guava.Maybe;
+import org.apache.brooklyn.util.text.Strings;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.common.annotations.Beta;
+import com.google.common.collect.Iterables;
+
+@Beta
+public abstract class AbstractCompositeEffector extends AddEffector {
+
+    private static final Logger LOG = LoggerFactory.getLogger(AbstractCompositeEffector.class);
+
+    public AbstractCompositeEffector(Effector<?> effector) {
+        super(effector);
+    }
+
+    @Override
+    public void apply(EntityLocal entity) {
+        Maybe<Effector<?>> effectorMaybe = entity.getEntityType().getEffectorByName(effector.getName());
+        if (!effectorMaybe.isAbsentOrNull()) {
+//            Effector<?> original = Effectors.effector(effectorMaybe.get()).name(ORIGINAL_PREFIX + effector.getName()).build();
+//            ((EntityInternal) entity).getMutableEntityType().addEffector(original);
+        }
+        super.apply(entity);
+    }
+
+    protected static abstract class Body<T> extends EffectorBody<T> {
+        protected final Effector<?> effector;
+        protected final ConfigBag params;
+
+        public Body(Effector<?> eff, ConfigBag params) {
+            this.effector = eff;
+            this.params = params;
+        }
+
+        @Override
+        public abstract T call(final ConfigBag params);
+
+        protected String getEffectorName(Object effectorDetails) {
+            String effectorName = null;
+            if (effectorDetails instanceof String) {
+                effectorName = (String) effectorDetails;
+            } else if (effectorDetails instanceof Map) {
+                Map<String,Object> effectorMap = (Map) effectorDetails;
+                Set<String> keys = effectorMap.keySet();
+                if (keys.size() != 1) {
+                    throw new IllegalArgumentException("Effector parameter cannot be parsed: " + effectorDetails);
+                }
+                effectorName = Iterables.getOnlyElement(keys);
+            } else {
+                effectorName = Strings.toString(effectorDetails);
+            }
+            return effectorName;
+        }
+
+        protected Entity getTargetEntity(Object effectorDetails) {
+            Entity targetEntity = entity();
+            if (effectorDetails instanceof Map) {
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/apache/brooklyn/core/effector/composite/AbstractCompositeEffector.java
```diff
@@ -0,0 +1,184 @@
...
+    private static final Logger LOG = LoggerFactory.getLogger(AbstractCompositeEffector.class);
+
+    public AbstractCompositeEffector(Effector<?> effector) {
+        super(effector);
+    }
+
+    @Override
+    public void apply(EntityLocal entity) {
+        Maybe<Effector<?>> effectorMaybe = entity.getEntityType().getEffectorByName(effector.getName());
+        if (!effectorMaybe.isAbsentOrNull()) {
```
robertgmoss: All of the logic in this has been commented out! If none is required, makes overriding the entire method superfluous.

core/src/main/java/org/apache/brooklyn/core/effector/composite/AbstractCompositeEffector.java
```diff
@@ -0,0 +1,184 @@
...
+            if (effectorDetails instanceof String) {
+                effectorName = (String) effectorDetails;
+            } else if (effectorDetails instanceof Map) {
+                Map<String,Object> effectorMap = (Map) effectorDetails;
+                Set<String> keys = effectorMap.keySet();
+                if (keys.size() != 1) {
+                    throw new IllegalArgumentException("Effector parameter cannot be parsed: " + effectorDetails);
+                }
+                effectorName = Iterables.getOnlyElement(keys);
+            } else {
```
robertgmoss: what about `instanceof List`?

geomacy:
@grkvlt there should be some unit tests at least along with these, can you add some?

core/src/main/java/org/apache/brooklyn/core/effector/composite/ComposeEffector.java
```diff
@@ -0,0 +1,109 @@
...
+        EffectorBody<Object> body = new Body(eff.buildAbstract(), params);
+        eff.impl(body);
+        return eff;
+    }
+
+    protected static class Body extends AbstractCompositeEffector.Body<Object> {
+
+        public Body(Effector<?> eff, ConfigBag params) {
+            super(eff, params);
+            Preconditions.checkNotNull(params.getAllConfigRaw().get(COMPOSE.getName()), "Effector names must be supplied when defining this effector");
```
robertgmoss: Should this be `Preconditions.checkNotNull(params.get(COMPOSE), "compose must be supplied when defining a compose effector: %s", params)`?

core/src/main/java/org/apache/brooklyn/core/effector/composite/LoopEffector.java
```diff
@@ -0,0 +1,108 @@
...
+        EffectorBody<List> body = new Body(eff.buildAbstract(), params);
+        eff.impl(body);
+        return eff;
+    }
+
+    protected static class Body extends AbstractCompositeEffector.Body<List> {
+
+        public Body(Effector<?> eff, ConfigBag params) {
+            super(eff, params);
+            Preconditions.checkNotNull(params.getAllConfigRaw().get(LOOP.getName()), "Effector names must be supplied when defining this effector");
```
robertgmoss: Should this be `Preconditions.checkNotNull(params.get(LOOP), "loop must be supplied when defining a loop effector: %s", params)`?

core/src/main/java/org/apache/brooklyn/core/effector/composite/SequenceEffector.java
```diff
@@ -0,0 +1,100 @@
...
+        EffectorBody<Object> body = new Body(eff.buildAbstract(), params);
+        eff.impl(body);
+        return eff;
+    }
+
+    protected static class Body extends AbstractCompositeEffector.Body<Object> {
+
+        public Body(Effector<?> eff, ConfigBag params) {
+            super(eff, params);
+            Preconditions.checkNotNull(params.getAllConfigRaw().get(SEQUENCE.getName()), "Effector names must be supplied when defining this effector");
```
robertgmoss: Should this be `Preconditions.checkNotNull(params.get(SEQUENCE), "sequence must be supplied when defining a sequence effector: %s", params)`?

core/src/main/java/org/apache/brooklyn/core/effector/composite/TransformEffector.java
```diff
@@ -0,0 +1,90 @@
...
+        EffectorBody<Object> body = new Body(eff.buildAbstract(), params);
+        eff.impl(body);
+        return eff;
+    }
+
+    protected static class Body extends AbstractCompositeEffector.Body<Object> {
+
+        public Body(Effector<?> eff, ConfigBag params) {
+            super(eff, params);
+            Preconditions.checkNotNull(params.getAllConfigRaw().get(FUNCTION.getName()), "Function must be supplied when defining this effector");
```
robertgmoss: Should this be `Preconditions.checkNotNull(params.get(FUNCTION), "function must be supplied when defining a function effector: %s", params)`?

Also what about `INPUT`?

core/src/main/java/org/apache/brooklyn/core/effector/composite/AbstractCompositeEffector.java
grkvlt: This is the individual effector elements of the list of effectors, they can only be a `String` or a `Map`

core/src/main/java/org/apache/brooklyn/core/effector/composite/ComposeEffector.java
grkvlt: Yes, I'll fix all of these to have better error messages.

core/src/main/java/org/apache/brooklyn/core/effector/composite/AbstractCompositeEffector.java
robertgmoss: I mean what if someone passes a list - `effectorName = Strings.toString(effectorDetails);` does not feel like the right way to handle that

policy/src/main/java/org/apache/brooklyn/policy/action/PeriodicEffectorPolicy.java
```diff
@@ -0,0 +1,105 @@
...
+            .constraint(Predicates.notNull())
+            .defaultValue(Duration.hours(1))
+            .build();
+
+    public static final AttributeSensor<Void> INVOKE_IMMEDIATELY = Sensors.newSensor(Void.TYPE, "scheduler.invoke");
+    public static final AttributeSensor<Boolean> START_SCHEDULER = Sensors.newBooleanSensor("scheduler.start");
+
+    protected long delay;
+
+    private final ScheduledExecutorService executor = Executors.newSingleThreadScheduledExecutor();
```
robertgmoss: Any reason not to inherit the `ScheduledExecutorService` of the parent class and use that?

core/src/main/java/org/apache/brooklyn/core/effector/composite/AbstractCompositeEffector.java
grkvlt: But why would they do that? They would be creating a list of lists, as in the third option below, and if someone does that, then I'm happy for things to just break...

```YAML
compose:
  - name1 # string
  - name2: # map
      key: value
  - [ whyWouldYouDoThis ] # list
```

robertgmoss: Indeed, it would clearly be a mistake - prefer to fail fast, so perhaps throwing an exception rather than allowing it to pass by calling `toString()`

Graeme-Miller:
@grkvlt this PR is rather large, and missing tests. I suggest it is closed for now and we work out a way to break this down.

Some ideas for how to break it down:
*) Separate PR's per effector would be nice
*) New sensor types as separate PRs (e.g. AbstractCompositeEffector/PeriodicEffectorPolicy.java)
*) Refactors as separate PRs (Moving AddSensor)

This should make it easier to review and merge.

What are your thoughts on this?

grkvlt:
@Graeme-Miller I think once the tests are added it should be OK, as the individual effectors are all reliant on a single parent abstract class, so they fit together. I will work with @robertgmoss on getting this done and reviewed.

nakomis:
@grkvlt Unfortunately it looks like this branch has gone stale since you updated it. Can you have a look to see if it is still relevant (I suspect it is), and rebase against master

Thanks

PR #607, opened 2017-03-22
Improve test location rebind stub

neykov:
* Makes the stubbed location more functional by implementing the getNodeMetadata, setting the location field on the NodeMetadata;
* Disabled aws hostname lookup - nothing to ssh into;
* Sets the HA manager for the new rebind management context to disabled. This fixes it to return MASTER state. That's what the launcher does when no HA configured.


sjcorbett:
Looks good.

PR #608, opened 2017-03-23
`BundleMaker` REST call allowing to add ZIP/JAR to catalog - Part #2

tbouron:
Builds on top of #485. It implements the missing bits and pieces, such as:
- fields `version` and `bundle` are required under `brooklyn.catalog` when uploading a ZIP/JAR archive
- returns 400 HTTP status code is anything goes wrong during the process of the archive (as this is a bad input, therefore a bad request)
- avoid using the `CatalogPopulator` as it reloads all bundles. Instead, it uses the new `CatalogBundleLoader` that loads only a given bundle
- if the bundle has been installed, but the CAMP parser fails, the bundle is uninstalled
- adds more unit tests, especially to check error handling

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/catalog/internal/CatalogBomScanner.java
+++ b/core/src/main/java/org/apache/brooklyn/core/catalog/internal/CatalogBomScanner.java
@@ -20,15 +20,19 @@ package org.apache.brooklyn.core.catalog.internal;
 
 import java.util.List;
 
+import javax.annotation.Nullable;
+
 import org.apache.brooklyn.api.mgmt.ManagementContext;
 import org.apache.brooklyn.core.BrooklynFeatureEnablement;
 import org.apache.brooklyn.util.text.Strings;
 import org.osgi.framework.Bundle;
+import org.osgi.framework.BundleContext;
 import org.osgi.framework.ServiceReference;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.google.common.annotations.Beta;
+import com.google.common.base.Predicate;
 import com.google.common.collect.ImmutableList;
 
 /** Scans bundles being added, filtered by a whitelist and blacklist, and adds catalog.bom files to the catalog.
@@ -45,16 +49,22 @@ public class CatalogBomScanner {
 
     private CatalogBundleTracker catalogBundleTracker;
 
-    public void bind(ServiceReference<ManagementContext> managementContext) throws Exception {
+    public void bind(ServiceReference<ManagementContext> mgmtContextReference) throws Exception {
         if (isEnabled()) {
             LOG.debug("Binding management context with whiteList [{}] and blacklist [{}]",
                 Strings.join(getWhiteList(), "; "),
                 Strings.join(getBlackList(), "; "));
-            catalogBundleTracker = new CatalogBundleTracker(this, managementContext);
+
+            final BundleContext bundleContext = mgmtContextReference.getBundle().getBundleContext();
+            ManagementContext mgmt = bundleContext.getService(mgmtContextReference);
+            CatalogBundleLoader bundleLoader = new CatalogBundleLoader(new SymbolicNameAccessControl(), mgmt);
+
+            catalogBundleTracker = new CatalogBundleTracker(bundleContext, bundleLoader);
+            catalogBundleTracker.open();
         }
     }
 
-    public void unbind(ServiceReference<ManagementContext> managementContext) throws Exception {
+    public void unbind(ServiceReference<ManagementContext> mgmtContextReference) throws Exception {
         if (isEnabled()) {
             LOG.debug("Unbinding management context");
             if (null != catalogBundleTracker) {
@@ -62,17 +72,12 @@ public class CatalogBomScanner {
                 catalogBundleTracker = null;
                 temp.close();
             }
+            mgmtContextReference.getBundle().getBundleContext().ungetService(mgmtContextReference);
         }
     }
 
     private boolean isEnabled() {
         return BrooklynFeatureEnablement.isEnabled(BrooklynFeatureEnablement.FEATURE_LOAD_BUNDLE_CATALOG_BOM);
-    }
-
-    public String[] bundleIds(Bundle bundle) {
-        return new String[] {
-            String.valueOf(bundle.getBundleId()), bundle.getSymbolicName(), bundle.getVersion().toString()
-        };
     }
 
     public List<String> getWhiteList() {
@@ -99,6 +104,27 @@ public class CatalogBomScanner {
     public void setBlackList(String blackListText) {
         LOG.debug("Setting blackList to ", blackListText);
         this.blackList = Strings.parseCsv(blackListText);
+    }
+
+    public class SymbolicNameAccessControl implements Predicate<Bundle> {
+        @Override
+        public boolean apply(@Nullable Bundle input) {
+            return passesWhiteAndBlacklists(input);
+        }
+    }
+
+    private boolean passesWhiteAndBlacklists(Bundle bundle) {
+        return on(bundle, getWhiteList()) && !on(bundle, getBlackList());
+    }
+
+    private boolean on(Bundle bundle, List<String> list) {
+        for (String candidate : list) {
+            final String symbolicName = bundle.getSymbolicName();
+            if (symbolicName.matches(candidate.trim())) {
+                return true;
+            }
+        }
+        return false;
     }
 
 }
--- a/core/src/main/java/org/apache/brooklyn/core/catalog/internal/CatalogBomScanner.java
+++ b/core/src/main/java/org/apache/brooklyn/core/catalog/internal/CatalogBomScanner.java
@@ -18,37 +18,18 @@
  */
 package org.apache.brooklyn.core.catalog.internal;
 
-import static org.apache.brooklyn.api.catalog.CatalogItem.CatalogItemType.TEMPLATE;
-
-import java.io.IOException;
-import java.io.InputStream;
-import java.net.URL;
 import java.util.List;
-import java.util.Map;
 
-import org.apache.brooklyn.api.catalog.CatalogItem;
 import org.apache.brooklyn.api.mgmt.ManagementContext;
 import org.apache.brooklyn.core.BrooklynFeatureEnablement;
-import org.apache.brooklyn.util.collections.MutableList;
-import org.apache.brooklyn.util.exceptions.Exceptions;
-import org.apache.brooklyn.util.stream.Streams;
 import org.apache.brooklyn.util.text.Strings;
-import org.apache.brooklyn.util.yaml.Yamls;
 import org.osgi.framework.Bundle;
-import org.osgi.framework.BundleContext;
-import org.osgi.framework.BundleEvent;
 import org.osgi.framework.ServiceReference;
-import org.osgi.util.tracker.BundleTracker;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import org.yaml.snakeyaml.DumperOptions;
-import org.yaml.snakeyaml.Yaml;
 
 import com.google.common.annotations.Beta;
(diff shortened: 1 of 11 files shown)


core/src/main/java/org/apache/brooklyn/core/catalog/internal/CatalogBomScanner.java
```diff
@@ -195,22 +210,29 @@ public void removedBundle(Bundle bundle, BundleEvent bundleEvent, Iterable<? ext
...
-            try {
-                getManagementContext().getCatalog().deleteCatalogItem(item.getSymbolicName(), item.getVersion());
-            } catch (Exception e) {
-                Exceptions.propagateIfFatal(e);
-                LOG.warn(Strings.join(new String[] {
-                    "Failed to remove", item.getSymbolicName(), item.getVersion(), "from catalog"
-                }, " "), e);
-            }
+    @Beta
+    public class CatalogBundleLoader {
```
geomacy: I'd say it's worth splitting this out into a standalone class. Same for CatalogPopulator; which I think we should rename to something like 'CatalogLoadingBundleTracker'.

geomacy: In CatalogBomScanner, now that this refactorization has being done we should remove the constructor 
```
CatalogPopulator(BundleContext bundleContext, ManagementContext managementContext)
``` 
which was introduced for this purpose, including deleting the `bundleContext` and `managementContext` members and re-simplifying any methods that use them.

rest/rest-api/src/main/java/org/apache/brooklyn/rest/api/CatalogApi.java
```diff
@@ -38,27 +38,60 @@
...
+                    required = true)
+            byte[] item);
+
+    @POST
+    @Beta
+    @Consumes({"application/x-zip", "application/x-jar"})
+    @ApiOperation(value = "Add a catalog item (e.g. new type of entity, policy or location) by uploading OSGi bundle JAR, or ZIP which will be turned into bundle JAR, "
+            + "containing catalog.bom containing bundle name and version. "
+            + "Return value is 201 CREATED if bundle could be added.", response = String.class)
+    public Response createFromArchive(
```
geomacy: Would it be worth deleting this and `createFromYaml` so that Swagger on the UI had only one endpoint to present?

core/src/main/java/org/apache/brooklyn/core/catalog/internal/CatalogBomScanner.java
geomacy: We should also remove the changes introduced in 33f7940c18ba8cd17fd83500486e02376d3b8fd8 to CatalogPopulator, i.e. adding the `CatalogPopulator(BundleContext bundleContext, ManagementContext managementContext) ` constructor and `bundleContext` member variable, and re-simplifying `open()` and `close()` and anything else that uses them.  Those changes were introduced for the purposes of this PR, but are not needed now that the new class has been factored out.

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/resources/CatalogResource.java
```diff
@@ -129,6 +176,97 @@ public Response create(String yaml) {
...
+            }
+            
+            f2 = bm.copyAddingManifest(f, mf);
+            
+            Bundle bundle = bm.installBundle(f2, true);
+            
+            if (!BrooklynFeatureEnablement.isEnabled(BrooklynFeatureEnablement.FEATURE_LOAD_BUNDLE_CATALOG_BOM)) {
+                // if the above feature is not enabled, let's do it manually (as a contract of this method)
+                try {
+                    new CatalogBomScanner().new CatalogBundleLoader(mgmt()).scanForCatalog(bundle);
```
geomacy: This should just be `new CatalogBundleLoader(mgmt()).scanForCatalog(bundle);` (with it as a standalone class)

tbouron:
@geomacy I addressed your comments and separated out the `CatalogBomScanner`, `CatalogBundleTracker` and `CatalogBundleLoader`. I also dug into the swagger docs and found out how to hide specific `@ApiOperation`. I had to bump the version of the `swagger-maven-plugin` because it was throwing an NPE with the new `hidden` parameter.

I did a full build locally, with the latest master of the other submodules, everything went fine so I'm guessing the error reported by Jenkins was due to something else

neykov:
@tbouron Looks like #485 broke the `brooklyn-client` build, can you try fixing it here.
See https://builds.apache.org/view/Brooklyn/job/brooklyn-client-master/41/console.

tbouron:
@neykov Yep, I'll look into it

neykov:
Thanks @tbouron.

tbouron:
@neykov Should be fixed now. I also created https://github.com/apache/brooklyn-client/pull/41 to use the non-deprecated call

geomacy:
hi @tbouron I get an error trying to build this
```
[ERROR] Failed to execute goal com.github.kongchen:swagger-maven-plugin:3.1.4:generate (default) on project brooklyn-rest-api: Unable to parse configuration of mojo com.github.kongchen:swagger-maven-plugin:3.1.4:generate for parameter locations: Cannot assign configuration entry 'locations' with value 'org.apache.brooklyn.rest.api' of type java.lang.String to property of type java.util.List -> [Help 1]
```
I just synced and built this as-is, should I be merging pre-requisite PRs or something?

tbouron:
@geomacy Hum odd. I have build this multiple time locally with the latest master of all the sub-modules. Jenkins also seems happy.

I did [another PR](https://github.com/apache/brooklyn-client/pull/41) in `brooklyn-client` to fix the tests in the client (related to the changes I did here) but I don't see how that could be related.

geomacy:
aha - Maven version issue.   Upgrading 3.3.3 -> 3.3.9 fixed the problem.

neykov:
Sounds like we should bump the maven version requirement - https://github.com/apache/brooklyn-server/blob/master/parent/pom.xml#L948. Or perhaps downgrade `swagger-maven-plugin`?

tbouron:
@neykov we cannot downgrade `swagger-maven-plugin` because of the issue I reported in [my comment](https://github.com/apache/brooklyn-server/pull/608#issuecomment-289000101).

We can either update the plugin configuration to properly use list instead of comma separated string or bump the maven requirement. Up to you.

neykov:
Not sure what you mean by "update the plugin configuration". Couldn't see any related configuration in the pom.xml. No strong opinion though, either one is fine by me.

tbouron:
@neykov I'm talking [about this](https://github.com/apache/brooklyn-server/pull/608/commits/5fb40ff8ae077da061aee927a773ec7a232df97d)

neykov:
Thanks for the pointer, missed that in the changes.

tbouron:
@neykov You didn't miss it, I added it after your comment to solve the issue ;)

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/resources/CatalogResource.java
```diff
@@ -245,13 +246,21 @@ public Response createFromArchive(byte[] zipInput) {
             
             if (!BrooklynFeatureEnablement.isEnabled(BrooklynFeatureEnablement.FEATURE_LOAD_BUNDLE_CATALOG_BOM)) {
                 // if the above feature is not enabled, let's do it manually (as a contract of this method)
-                new CatalogBomScanner().new CatalogPopulator(
-                        ((LocalManagementContext) mgmt()).getOsgiManager().get().getFramework().getBundleContext(),
-                        mgmt()).addingBundle(bundle, null);
+                try {
+                    new CatalogBundleLoader(new CatalogBomScanner(), mgmt()).scanForCatalog(bundle);
```
geomacy: This is a bit of a problem; you shouldn't have to pass a `CatalogBomScanner()` in here; the point of doing this decoupling was so that the `CatalogBundleLoader` could avoid knowing about the `CatalogBomScanner`. 

I see that the use that is made of this is handling the [whitelist/blacklist](https://github.com/apache/brooklyn-server/commit/7046ba59cefd7e82ffad3f97057b3646915d1f08#diff-c46e6ef1f0a91ac1f58e7a340ecc0f34) functionality.  (Well there is also `CatalogBomScanner.bundleIds()` but that can be made a static method,
and maybe moved to `CatalogUtils`.)  

This whitelist stuff is configured [here](https://github.com/apache/brooklyn-server/blob/master/karaf/init/src/main/resources/OSGI-INF/blueprint/blueprint.xml#L123), in the definition of the singleton scanner object. (Only supported in Karaf launcher at the moment, as you see.)

Doing `new CatalogBundleLoader(new CatalogBomScanner(), mgmt()).scanForCatalog(bundle);` means you are supplying an object that hasn't been configured with the right white/blacklists, so it bypasses this restriction. 

I'm going to suggest this is something we can live with for now, but mark as a "TODO" for future action. 

However, it is still ugly to have the dependency on the `CatalogBomScanner` in the tracker and the loader, so I suggest we replace it with a simple predicate that wraps the whitelist checks and inject the predicate into the loader. That way the CatalogBomScanner can inject the correct whitelist configuration in the blueprint.xml, while in the REST API here we just live with an "always true" for the moment. 

The loader could also be created in the scanner and injected into the tracker, which will simplify the latter a bit.

That's all a bit vague; I have to play around with the code a bit to help me think, so here's a PR that 
illustrates what I mean - https://github.com/tbouron/brooklyn-server/pull/2

What do you think?

tbouron: @geomacy Yup, the dependency on `CatalogBomScanner` is ugly. I knew it had to be changed but wanted this to be merged as soon as possible so I just moved the code around rather than really refactor.

Re the whitelist injection for the REST API: I was wondering what if we add the `CatalogBomScanner` to the new `ScratchPad` area? It's supposed to be a singleton isn't it? We can then get back the correctly setup predicate from it.

WDYT?

geomacy:
I hope that PR of mine could just be used as-is to finish off that refactoring.  Personally I wouldn't worry too much about getting this integrated just now with a singleton `CatalogBomScanner`, especially as this is disabled by default at the minute anyway (see #233), until it can be made to work well with rebind.

geomacy:
I think this looks good; further work as discussed earlier can be done as a separate PR. Will merge.

neykov:
@tbouron started getting test failures in [VanillaSoftwareProcessTest.testVanillaSoftwareProcessCanUseResourcesInBundles](https://builds.apache.org/view/Brooklyn/job/brooklyn-integration-tests/70/testngreports/org.apache.brooklyn.rest.test.entity/VanillaSoftwareProcessTest/testVanillaSoftwareProcessCanUseResourcesInBundles/) around the catalog changes. Could you have a look.

tbouron:
@neykov Fixed in #608

neykov:
That was quick, thanks @tbouron.

core/src/main/java/org/apache/brooklyn/core/catalog/internal/CatalogBundleLoader.java
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    /**
+     * Scan the given bundle for a catalog.bom and adds it to the catalog.
+     *
+     * @param bundle The bundle to add
+     * @return A list of items added to the catalog
+     * @throws RuntimeException if the catalog items failed to be added to the catalog
+     */
+    public Iterable<? extends CatalogItem<?, ?>> scanForCatalog(Bundle bundle) {
```
aledsage: I'd prefer a name that makes clear the method will add the catalog items (as well as scanning + returning the items found). For example `addCatalogItemsFromBundle(...)`?

(This method name was originally in `CatalogBomScanner`, and was refactored to here - but in CatalogBomScanner it was a private method so the name didn't need to be as clear.)

ahgittin:
Thanks for polishing this off @tbouron !

PR #609, opened 2017-03-24
HttpCommandEffector allows string payload

duncangrant:
Also fixed integration tests that failing due to https://httpbin


neykov:
@duncangrant You beat me to fixing the integration tests. I went for replacing `httpbin.org` with a local server, changing the integration tests to unit tests. See https://github.com/apache/brooklyn-server/pull/610.

duncangrant:
@neykov I thought about doing something similar but it looked too much like hard work. Agreed that you approach is the way to go especially as it allows more of those tests to be run as unit tests.

PR #610, opened 2017-03-24
Various integration test fixes

neykov:
See commits for details

### DIFF
--- a/core/src/test/java/org/apache/brooklyn/core/effector/http/HttpCommandEffectorHttpBinTest.java
+++ b/core/src/test/java/org/apache/brooklyn/core/effector/http/HttpCommandEffectorHttpBinTest.java
@@ -18,6 +18,10 @@
  */
 package org.apache.brooklyn.core.effector.http;
 
+import java.io.IOException;
+import java.net.URI;
+import java.util.List;
+
 import org.apache.brooklyn.api.effector.Effector;
 import org.apache.brooklyn.api.entity.EntityLocal;
 import org.apache.brooklyn.api.entity.EntitySpec;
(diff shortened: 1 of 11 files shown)


core/src/test/java/org/apache/brooklyn/core/effector/http/HttpCommandEffectorHttpBinTest.java
```diff
@@ -38,42 +54,93 @@
...
+public class HttpCommandEffectorHttpBinTest {
 
     final static Effector<String> EFFECTOR_HTTPBIN = Effectors.effector(String.class, "Httpbin").buildAbstract();
 
+    private TestHttpServer server;
+    private String serverUrl;
     private TestApplication app;
     private EntityLocal entity;
     
+    public static final class HttpBinRequestHandler implements HttpRequestHandler {
```
duncangrant: Would it be simpler to use BetterMockWebServer here as in https://github.com/duncangrant/brooklyn-server/blob/1c064b284757ed0c794cbf8f3a6d65e34564af3a/core/src/test/java/org/apache/brooklyn/core/effector/http/HttpCommandEffectorTest.java#L68-L70

neykov:
I considered that but decided against it for two reasons:
  * would be simpler to implement the behaviour once, rather then to setup the mock over and over in each test
  * some of the tests rely on the server returning different content, based on the url (say to check if an override url/parameter is applied). This would be harder to express with a fixed mock.

software/base/src/test/java/org/apache/brooklyn/entity/software/base/lifecycle/ScriptHelperIntegrationTest.java
```diff
@@ -54,14 +56,25 @@
...
+//        Stacktrace:
+//
+//        at org.testng.Assert.fail(Assert.java:94)
+//        at org.testng.Assert.failNotEquals(Assert.java:513)
+//        at org.testng.Assert.assertEqualsImpl(Assert.java:135)
+//        at org.testng.Assert.assertEquals(Assert.java:116)
+//        at org.testng.Assert.assertEquals(Assert.java:389)
+//        at org.apache.brooklyn.entity.software.base.lifecycle.ScriptHelperIntegrationTest.testStopCommandWaitsToStopWithSigTerm(ScriptHelperIntegrationTest.java:83)
+//        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
+    @Test(groups={"Integration","Broken"})
```
neykov: cc @bostko - tried to figure out why kill doesn't work but didn't find an obvious reason

sjcorbett:
Looks good.

PR #611, opened 2017-03-24
Deprecate groovy

aledsage:
As discussed on the dev@brooklyn mailing list (subject "[PROPOSAL] Remove groovy dependency/support").

Deprecate methods that take explicit groovy types (e.g. `groovy.lang.Closure`, `groovy.time.TimeDuration`, etc).
    
Also updates some javadoc that mentioned `Closure` to say that its usage is deprecated.

In some places, it does a `LOG.warn()` about the deprecation when a groovy Closure etc is passed in - this is when we can't just deprecate the method, e.g. for use of `TypeCoercion` on when the method takes `Object` and does an ugly `instanceof`.

This does not touch:
* The `org.apache.brooklyn.rest.resources.ScriptResource` (for groovy-script execution)
* Internal usage of groovy (e.g. `AbstractManagementContext.invokeEffectorMethodLocal()` calling `GroovyJavaMethods.invokeMethodOnMetaClass`)

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/effector/ExplicitEffector.java
+++ b/core/src/main/java/org/apache/brooklyn/core/effector/ExplicitEffector.java
@@ -18,18 +18,30 @@
  */
 package org.apache.brooklyn.core.effector;
 
-import groovy.lang.Closure;
-
 import java.util.List;
 import java.util.Map;
 
 import org.apache.brooklyn.api.effector.ParameterType;
 import org.apache.brooklyn.api.entity.Entity;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import com.google.common.base.Objects;
 import com.google.common.collect.ImmutableList;
 
+import groovy.lang.Closure;
+
+/**
+ * @deprecated since 0.11.0; explicit groovy utilities/support will be deleted 
+ *             (currently only used via {@link #create(String, Class, List, String, Closure)}, 
+ *             so class is not deemed useful enough - extend {@link AbstractEffector} directly 
+ *             if really required).
+ */
+@Deprecated
 public abstract class ExplicitEffector<I,T> extends AbstractEffector<T> {
+    
+    private static final Logger LOG = LoggerFactory.getLogger(ExplicitEffector.class);
+
     public ExplicitEffector(String name, Class<T> type, String description) {
         this(name, type, ImmutableList.<ParameterType<?>>of(), description);
     }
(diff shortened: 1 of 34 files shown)


core/src/main/java/org/apache/brooklyn/core/effector/ExplicitEffector.java
```diff
@@ -46,12 +58,20 @@ public T call(Entity entity, Map parameters) {
...
     /** convenience to create an effector supplying a closure; annotations are preferred,
      * and subclass here would be failback, but this is offered as 
-     * workaround for bug GROOVY-5122, as discussed in test class CanSayHi 
+     * workaround for bug GROOVY-5122, as discussed in test class CanSayHi.
+     * 
+     * @deprecated since 0.11.0; explicit groovy utilities/support will be deleted.
      */
+    @Deprecated
     public static <I,T> ExplicitEffector<I,T> create(String name, Class<T> type, List<ParameterType<?>> parameters, String description, Closure body) {
+        LOG.warn("Use of groovy.lang.Closure is deprecated, in ExplicitEffector.create()");
```
aledsage: I've added this to javadoc on the class - I think that's good enough, as there's not a direct replacement (i.e. one would extend `AbstractEffector` yourself instead, as we don't need to workaround for GROOVY-5122 if you're not using groovy - I'm not sure why anyone would want to do that).

I agree in general that if deprecating a method/class, one should point at the alternative - but if deprecating a class, I don't think we need it on every method.

core/src/main/java/org/apache/brooklyn/core/effector/MethodEffector.java
```diff
@@ -133,6 +133,10 @@ public MethodEffector(Method method) {
         this(new AnnotationsOnMethod(method.getDeclaringClass(), method), null);
     }
 
+    /**
+     * @deprecated since 0.11.0; explicit groovy utilities/support will be deleted.
+     */
+    @Deprecated
```
aledsage: My rule-of-thumb... if one might get at the code via a yaml blueprint (so never seeing a deprecation warning in your IDE/compiler), then we should include a log.warn. But if we can assume that the user will get compiler warnings, then that is sufficient.

I believe that this will only get called directly in code, so one should see a compiler warning.

software/base/src/main/java/org/apache/brooklyn/entity/software/base/SoftwareProcessImpl.java
```diff
@@ -476,6 +476,11 @@ public void waitForServiceUp() {
     public void waitForServiceUp(Duration duration) {
         Entities.waitForServiceUp(this, duration);
     }
+    
+    /**
+     * @deprecated since 0.11.0; explicit groovy utilities/support will be deleted.
+     */
+    @Deprecated
```
aledsage: Yes, the parameter is of type `groovy.time.TimeDuration`.

aledsage:
@tbouron I've responded to your comments - do you agree with those, or do you think I should make further changes before merging the PR?

core/src/main/java/org/apache/brooklyn/core/effector/ExplicitEffector.java
tbouron: Fair enough

tbouron:
@aledsage Yes, I think that covers everything so LGTM 👍

aledsage:
Thanks @tbouron - merging now.

PR #612, opened 2017-03-27
BROOKLYN-460: Brooklyn Camp syntax for adding tags to an entity spec

bostko:
```yaml
services:
- type: BasicApplication
  brooklyn.tags: [tag1, tag2]
```

`brooklyn.tags` adds NamedStringTag to an entity spec.
_Tags can be inspected via REST API `/v1/applications/<appId>/entities/<entityId>/tags`_

### DIFF
--- a/camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/creation/BrooklynComponentTemplateResolver.java
+++ b/camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/creation/BrooklynComponentTemplateResolver.java
@@ -257,6 +257,7 @@ public class BrooklynComponentTemplateResolver {
         new BrooklynEntityDecorationResolver.EnricherSpecResolver(yamlLoader).decorate(spec, attrs, encounteredRegisteredTypeIds);
         new BrooklynEntityDecorationResolver.InitializerResolver(yamlLoader).decorate(spec, attrs, encounteredRegisteredTypeIds);
         new BrooklynEntityDecorationResolver.SpecParameterResolver(yamlLoader).decorate(spec, attrs, encounteredRegisteredTypeIds);
+        new BrooklynEntityDecorationResolver.TagsResolver(yamlLoader).decorate(spec, attrs, encounteredRegisteredTypeIds);
 
         configureEntityConfig(spec, encounteredRegisteredTypeIds);
     }
--- a/camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/creation/BrooklynEntityDecorationResolver.java
+++ b/camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/creation/BrooklynEntityDecorationResolver.java
@@ -18,14 +18,18 @@
  */
 package org.apache.brooklyn.camp.brooklyn.spi.creation;
 
+import java.io.Serializable;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
 import com.google.common.base.Function;
+import com.google.common.base.Predicate;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 
+import com.google.common.collect.Iterables;
+import com.sun.org.apache.xpath.internal.operations.Bool;
 import org.apache.brooklyn.api.entity.EntityInitializer;
 import org.apache.brooklyn.api.entity.EntitySpec;
 import org.apache.brooklyn.api.mgmt.ManagementContext;
@@ -37,12 +41,20 @@ import org.apache.brooklyn.api.sensor.EnricherSpec;
 import org.apache.brooklyn.api.typereg.RegisteredType;
 import org.apache.brooklyn.camp.brooklyn.BrooklynCampReservedKeys;
 import org.apache.brooklyn.camp.brooklyn.spi.creation.BrooklynYamlTypeInstantiator.InstantiatorFromKey;
+import org.apache.brooklyn.camp.brooklyn.spi.dsl.DslAccessible;
+import org.apache.brooklyn.camp.brooklyn.spi.dsl.methods.BrooklynDslCommon;
+import org.apache.brooklyn.core.mgmt.BrooklynTags;
 import org.apache.brooklyn.core.objs.BasicSpecParameter;
 import org.apache.brooklyn.core.typereg.RegisteredTypeLoadingContexts;
 import org.apache.brooklyn.core.typereg.RegisteredTypes;
 import org.apache.brooklyn.util.collections.MutableList;
 import org.apache.brooklyn.util.core.config.ConfigBag;
+import org.apache.brooklyn.util.core.task.DeferredSupplier;
 import org.apache.brooklyn.util.guava.Maybe;
+
+import javax.annotation.Nullable;
+
+import static com.google.common.base.Preconditions.checkArgument;
 
 /**
  * Pattern for resolving "decorations" on service specs / entity specs, such as policies, enrichers, etc.
@@ -227,7 +239,49 @@ public abstract class BrooklynEntityDecorationResolver<DT> {
 
         @Override
         protected void addDecorationFromJsonMap(Map<?, ?> decorationJson, List<SpecParameter<?>> decorations) {
-            throw new IllegalStateException("Not called");
+            throw new UnsupportedOperationException("SpecParameterResolver.addDecorationFromJsonMap should never be called.");
+        }
+    }
+
+    public static class TagsResolver extends BrooklynEntityDecorationResolver<Iterable<Object>> {
+        protected TagsResolver(BrooklynYamlTypeInstantiator.Factory instantiator) {
+            super(instantiator);
+        }
+
+        @Override
+        public void decorate(EntitySpec<?> entitySpec, ConfigBag attrs, Set<String> encounteredRegisteredTypeIds) {
+            Iterable<Object> decorationAttributeJsonValue = getDecorationAttributeJsonValue(attrs);
+            if (decorationAttributeJsonValue != null) {
+                entitySpec.tagsAdd(decorationAttributeJsonValue);
+            }
+        }
+
+        @Override
+        protected String getDecorationKind() {
+            return "Brooklyn Tags";
+        }
+
+        @Override
+        protected Iterable<Object> getDecorationAttributeJsonValue(ConfigBag attrs) {
+            Object brooklynTags = attrs.getStringKey(BrooklynCampReservedKeys.BROOKLYN_TAGS);
+            if (brooklynTags == null) {
+                return null;
+            } else if (!(brooklynTags instanceof List)) {
+                throw new IllegalArgumentException(BrooklynCampReservedKeys.BROOKLYN_TAGS + " should be a List of String elements. You supplied " + brooklynTags);
+            } else {
+                checkArgument(Iterables.all((List) brooklynTags, new Predicate() {
+                    @Override
+                    public boolean apply(Object input) {
+                        return !(input instanceof DeferredSupplier);
+                    }
+                }), BrooklynCampReservedKeys.BROOKLYN_TAGS + " should not contain DeferredSupplier. A DeferredSupplier is made when using $brooklyn:attributeWhenReady. You supplied " + brooklynTags);
+                return (List)brooklynTags;
+            }
+        }
+
+        @Override
+        protected void addDecorationFromJsonMap(Map<?, ?> decorationJson, List<Iterable<Object>> decorations) {
+            throw new UnsupportedOperationException("TagsResolver.addDecorationFromJsonMap should never be called.");
         }
     }
 }
--- a/camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/TagsYamlTest.java
+++ b/camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/TagsYamlTest.java
@@ -1,0 +1,135 @@
+/*
+ * Copyright 2016 The Apache Software Foundation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.brooklyn.camp.brooklyn.spi.dsl;
+
+import com.google.common.base.Predicate;
+import com.google.common.collect.Iterables;
+import org.apache.brooklyn.api.entity.Entity;
+import org.apache.brooklyn.camp.brooklyn.AbstractYamlTest;
+import org.apache.brooklyn.entity.stock.BasicApplication;
+import org.apache.brooklyn.test.Asserts;
+import org.apache.brooklyn.util.exceptions.CompoundRuntimeException;
+import org.apache.brooklyn.util.exceptions.Exceptions;
+import org.testng.annotations.Test;
+
+import javax.annotation.Nullable;
+
+import static org.testng.Assert.assertTrue;
+import static org.testng.Assert.fail;
+
+public class TagsYamlTest extends AbstractYamlTest {
+    @Test
+    public void testBrooklynCampSingleTag() throws Exception {
+        final Entity app = createAndStartApplication(
+                "services:",
+                "- type: " + BasicApplication.class.getName(),
+                "  brooklyn.tags:",
+                "    - hi");
+        assertTrue(app.tags().getTags().contains("hi"));
+    }
+
+    @Test
+    public void testBrooklynCampMultipleTags() throws Exception {
+        final Entity app = createAndStartApplication(
+                "services:",
+                "- type: " + BasicApplication.class.getName(),
+                "  brooklyn.tags:",
+                "  - tag1",
+                "  - \"2\"",
+                "  - \"- 3\"");
+        assertTrue(app.tags().getTags().contains("tag1"));
+        assertTrue(app.tags().getTags().contains("2"));
+        assertTrue(app.tags().getTags().contains("- 3"));
+    }
+
+    @Test
+    public void testBrooklynCampTagsFailNonList() throws Exception {
+        try {
+            final Entity app = createAndStartApplication(
+                    "services:",
+                    "- type: " + BasicApplication.class.getName(),
+                    "  brooklyn.tags:",
+                    "    tag1: true",
+                    "    tag2: 2");
+            fail("Should throw IllegalArgumentException exception.");
+        } catch (CompoundRuntimeException e) {
+            Asserts.assertStringContainsAtLeastOne(Exceptions.getFirstInteresting(e).getMessage(),"brooklyn.tags must be a list, is: ");
+        }
+    }
+
+    @Test
+    public void testBrooklynCampKnowsIntegerTags() throws Exception {
+        final Entity app = createAndStartApplication(
+                "services:",
+                "- type: " + BasicApplication.class.getName(),
+                "  brooklyn.tags:",
+                "  - tag1",
+                "  - 3");
+        assertTrue(app.tags().getTags().contains(3));
+        assertTrue(app.tags().getTags().contains("tag1"));
+    }
+
+    @Test
+    public void testBrooklynCampObjectTags() throws Exception {
+        final Entity app = createAndStartApplication(
+                "services:",
+                "- type: " + BasicApplication.class.getName(),
+                "  brooklyn.tags:",
+                "  - tag1",
+                "  - $brooklyn:object:",
+                "      type: " + TagsTestObject.class.getName());
+        assertTrue(Iterables.any(app.tags().getTags(), new Predicate<Object>() {
+            @Override
+            public boolean apply(@Nullable Object input) {
+                return input instanceof TagsTestObject;
+            }
+        }));
+        assertTrue(app.tags().getTags().contains("tag1"));
+    }
+
+    @Test
+    public void testBrooklynCampFailDslTags() throws Exception {
+        try {
+            final Entity app = createAndStartApplication(
+                    "services:",
+                    "- type: " + BasicApplication.class.getName(),
+                    "  brooklyn.tags:",
+                    "  - tag1",
+                    "  - $brooklyn:object:",
+                    "      type: "+TagsTestObject.class.getName(),
+                    "      constructor.args:",
+                    "      - $brooklyn:attributeWhenReady(\"host.name\")");
+            fail("Should throw IllegalArgumentException exception.");
+        } catch (CompoundRuntimeException e) {
+            Asserts.assertStringContainsAtLeastOne(Exceptions.getFirstInteresting(e).getMessage(),"brooklyn.tags should not contain DeferredSupplier. A DeferredSupplier is made when using $brooklyn:attributeWhenReady");
+        }
+    }
+
+    @Test
+    public void testTagWithDslValue() throws Exception {
+        Entity app = createAndStartApplication(
+                "services:",
+                "- type: " + BasicApplication.class.getName(),
+                "  brooklyn.tags:",
+                "  - $brooklyn:literal(\"myval\")");
+        assertTrue(app.tags().getTags().contains("myval"));
+    }
+
+    public static class TagsTestObject {
+        public TagsTestObject() {}
+        public TagsTestObject(Object arg1) {}
+    }
+}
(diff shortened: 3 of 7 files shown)


camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/creation/BrooklynComponentTemplateResolver.java
```diff
@@ -257,6 +257,7 @@ public boolean canResolve() {
         new BrooklynEntityDecorationResolver.EnricherSpecResolver(yamlLoader).decorate(spec, attrs, encounteredRegisteredTypeIds);
         new BrooklynEntityDecorationResolver.InitializerResolver(yamlLoader).decorate(spec, attrs, encounteredRegisteredTypeIds);
         new BrooklynEntityDecorationResolver.SpecParameterResolver(yamlLoader).decorate(spec, attrs, encounteredRegisteredTypeIds);
+        new BrooklynEntityDecorationResolver.TagsResolver(yamlLoader).decorate(spec, attrs, encounteredRegisteredTypeIds);
```
tbouron: Do we want to promote `brooklyn.tags` as first-class citizen?

camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/creation/BrooklynEntityDecorationResolver.java
```diff
@@ -230,4 +234,45 @@ protected void addDecorationFromJsonMap(Map<?, ?> decorationJson, List<SpecParam
...
+                    public BrooklynTags.NamedStringTag apply(@Nullable String input) {
+                        return BrooklynTags.newNotesTag(input);
+                    }
+                });
+            }
+        }
+
+        @Override
+        protected void addDecorationFromJsonMap(Map<?, ?> decorationJson, List<Iterable<Object>> decorations) {
+            throw new IllegalStateException("Not called");
```
tbouron: Is it right to not implement this? Or at least, shouldn't we have a better exception message?

camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/creation/BrooklynEntityDecorationResolver.java
```diff
@@ -230,4 +237,51 @@ protected void addDecorationFromJsonMap(Map<?, ?> decorationJson, List<SpecParam
             throw new IllegalStateException("Not called");
```
aledsage: As per @tbouron's comment about whether to implement this - throwing an exception saying "Not called" is very misleading. If it is ever called, then telling the user that it wasn't called in the exception message will look weird!

Better to have a comment saying that it won't be called, and to throw an `UnsupportedOperationException` I think.

But it looks like this code already existed, rather than being modified in this PR (so doesn't need to be fixed here).

core/src/main/java/org/apache/brooklyn/core/mgmt/BrooklynTags.java
```diff
@@ -59,6 +59,21 @@ public String getKind() {
         public String getContents() {
             return contents;
         }
+
+        @Override
+        public boolean equals(Object other) {
+            if (other == null || !(other instanceof NamedStringTag)) {
```
aledsage: Very minor: don't need the `other == null`: `other instanceof NamedStringTag` is false when other is null.

core/src/main/java/org/apache/brooklyn/core/mgmt/BrooklynTags.java
```diff
@@ -59,6 +59,21 @@ public String getKind() {
...
+                return false;
+            }
+
+            return kind != null && kind.equals(((NamedStringTag)other).kind)
+                    && contents != null && contents.equals(((NamedStringTag)other).contents);
+        }
+
+        @Override
+        public int hashCode() {
+            return (kind == null ? 0 : kind.hashCode()) + (contents == null ? 0 : contents.hashCode());
```
aledsage: Personal preference is to use guava's utilities: `Objects.hashCode(kind, contents)`

core/src/main/java/org/apache/brooklyn/core/mgmt/BrooklynTags.java
```diff
@@ -59,6 +59,21 @@ public String getKind() {
...
             return contents;
         }
+
+        @Override
+        public boolean equals(Object other) {
+            if (other == null || !(other instanceof NamedStringTag)) {
+                return false;
+            }
+
+            return kind != null && kind.equals(((NamedStringTag)other).kind)
```
aledsage: This will return not equals if both have `kind == null` and/or `contents == null`. Better to do:
```
if (!(other instanceof NamedStringTag)) {
    return false;
}
NamedStringTag o = (NamedStringTag) other;
return Objects.equal(kind, o.kind) && Objects.equal(contents, o.contents);
```

aledsage: Just realised this is in test code, so I really don't care that much! But comments are still worth taking on board for next time!

camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/creation/BrooklynEntityDecorationResolver.java
```diff
@@ -230,4 +237,51 @@ protected void addDecorationFromJsonMap(Map<?, ?> decorationJson, List<SpecParam
...
     }
+
+    public static class TagsResolver extends BrooklynEntityDecorationResolver<Iterable<Object>> {
+        protected TagsResolver(BrooklynYamlTypeInstantiator.Factory instantiator) {
+            super(instantiator);
+        }
+
+        @Override
+        public void decorate(EntitySpec<?> entitySpec, ConfigBag attrs, Set<String> encounteredRegisteredTypeIds) {
+            if (getDecorationAttributeJsonValue(attrs) != null) {
```
aledsage: Minor personal preference: I'd store this as a variable rather than evaluating it twice.

camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/creation/BrooklynEntityDecorationResolver.java
```diff
@@ -227,7 +234,54 @@ protected Object getDecorationAttributeJsonValue(ConfigBag attrs) {
...
+            }
+        }
+
+        @Override
+        protected String getDecorationKind() {
+            return "Brooklyn Tags";
+        }
+
+        @Override
+        protected Object getDecorationAttributeJsonValue(ConfigBag attrs) {
```
aledsage: You can declare the return type as `Iterable<Object>`, and therefore avoid the cast in `decorate()`.

As an aside, the `entitySpec.tagsAdd() should really take an `Iterable<?>` rather than an `Iterable<Object>`, but don't worry about that for this PR.

camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/creation/BrooklynComponentTemplateResolver.java
m4rkmckenna: @bostko IMO any API changes should be discussed on the mailing list even if they are additive

camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/creation/BrooklynEntityDecorationResolver.java
```diff
@@ -227,7 +234,54 @@ protected Object getDecorationAttributeJsonValue(ConfigBag attrs) {
...
+            Object brooklynTags = attrs.getStringKey(BrooklynCampReservedKeys.BROOKLYN_TAGS);
+            if (brooklynTags == null) {
+                return null;
+            } else if (brooklynTags != null && !(brooklynTags instanceof List)) {
+                throw new IllegalArgumentException(BrooklynCampReservedKeys.BROOKLYN_TAGS + " should be a List of String elements. You supplied " + attrs);
+            } else {
+                checkArgument(Iterables.all((List) brooklynTags, new Predicate() {
+                    @Override
+                    public boolean apply(@Nullable Object input) {
+                        return input instanceof String;
```
aledsage: Is this too strict? What about primitives? I tried the yaml below, and it failed at this assertion (because it has the int `2`):
```
services:
  brooklyn.tags:
  - tag1
  - 2
```

camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/creation/BrooklynEntityDecorationResolver.java
```diff
@@ -227,7 +234,54 @@ protected Object getDecorationAttributeJsonValue(ConfigBag attrs) {
...
+            return "Brooklyn Tags";
+        }
+
+        @Override
+        protected Object getDecorationAttributeJsonValue(ConfigBag attrs) {
+            Object brooklynTags = attrs.getStringKey(BrooklynCampReservedKeys.BROOKLYN_TAGS);
+            if (brooklynTags == null) {
+                return null;
+            } else if (brooklynTags != null && !(brooklynTags instanceof List)) {
+                throw new IllegalArgumentException(BrooklynCampReservedKeys.BROOKLYN_TAGS + " should be a List of String elements. You supplied " + attrs);
```
aledsage: Perhaps instead log `brooklynTags` rather than `attrs`, and also include in the log message `" of type " + brooklynTags.getClass().getName()`.

Don't need to repeat the null check for `brooklynTags != null`

camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/creation/BrooklynComponentTemplateResolver.java
aledsage: @tbouron what do you mean by a "first-class citizen" here? Do you mean are we right to be promoting it as a first-class citizen, or should we be promoting it more?

It should be a first-class concept - it's not a part of config or anything like that. It needs to end up being added to the `EntitySpec` via `spec.tagsAdd(...)`.

bostko: Thanks @m4rkmckenna . I opened BROOKLYN-460 as described in https://brooklyn.apache.org/developers/how-to-contribute.html

camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/DslYamlTest.java
```diff
@@ -66,6 +71,63 @@ public void testDslSelf() throws Exception {
...
+    }
+
+    @Test
+    public void testDslMultipleTags() throws Exception {
+        final Entity app = createAndStartApplication(
+                "services:",
+                "- type: " + BasicApplication.class.getName(),
+                "  brooklyn.tags:",
+                "  - tag1",
+                "  - tAg2");
```
aledsage: Include a test for a tag of a primitive type, e.g. `- 3`? (see earlier comment).

camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/DslYamlTest.java
```diff
@@ -66,6 +71,63 @@ public void testDslSelf() throws Exception {
     }
 
     @Test
+    public void testDslSingleTag() throws Exception {
```
aledsage: Why are these tests in the `DslYamlTest` class? This class is for testing `$brooklyn:...`.

I'd probably create a new test class for this: `org.apache.brooklyn.camp.brooklyn.TagsYamlTest`.

camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/DslYamlTest.java
```diff
@@ -66,6 +71,63 @@ public void testDslSelf() throws Exception {
...
+
+    @Test
+    public void testDslFailObjectTags() throws Exception {
+        try {
+            final Entity app = createAndStartApplication(
+                    "services:",
+                    "- type: " + BasicApplication.class.getName(),
+                    "  brooklyn.tags:",
+                    "  - tag1",
+                    "  - $brooklyn:object:",
```
aledsage: Does this fail because it can't parse the DSL, or because it has evaluated it as a `TagsTestObject` instance, which is not a string?

If the latter, why forbid it? Tags can be any object in the java API.

core/src/main/java/org/apache/brooklyn/core/mgmt/BrooklynTags.java
bostko: Thanks @aledsage ! I was looking for thing like `Objects.equal` but didn't find at the time.

camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/creation/BrooklynComponentTemplateResolver.java
tbouron: @aledsage By first-class citizen, I mean: do we want to create this specific `brooklyn.tags` key, like we did for `brooklyn.initializers` or `brooklyn.enrichers`? Is this the right thing to do? I don't have the answer to that hence why I'm asking the question

camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/DslYamlTest.java
bostko: I introduced only String tags since normally only types defined in `org.apache.brooklyn.core.mgmt.BrooklynTags` are used for tags.

aledsage:
@bostko (cc @tbouron) Thinking about this more, I'm not sure what types we should support for tags. Maybe accepting any object is a bad idea, as it might make subsequent features harder to support (e.g. support for looking up / processing tags).

On the other hand, locking down the types to be just strings without a good reason seems too extreme.

Looking at the REST api (`org.apache.brooklyn.rest.resources.EntityResource.listTags()`), it turns the tags into json via the standard mechanism. It can therefore handle non-string types. if you use something that does not have a good json representation, then it becomes a lot less useful. But arguably that's the user's problem (presumably they had a reason for putting the object in the tags in the first place).

aledsage:
@bostko it's worth adding this test (which passes), for use of a DSL expression in the tags:
```
    @Test
    public void testTagWithDslValue() throws Exception {
        Entity app = createAndStartApplication(
                "services:",
                "- type: " + BasicApplication.class.getName(),
                "  brooklyn.tags:",
                "  - $brooklyn:literal(\"myval\")");
        assertTrue(getTagsEventually(app).contains(BrooklynTags.newNotesTag("myval")));
    }
```

camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/creation/BrooklynComponentTemplateResolver.java
neykov: @tbouron is onto something here. For example for effectors we go through an initializer and try not to expand the keywords too much.

I think that anything that is a "first class concept" on an `EntitySpec` should have the corresponding way to express in yaml. There are tags on specs, so `brooklyn.tags` in CAMP makes sense. Similarly for `brooklyn.parameters`, `brooklyn.config`. There are no effectors on the `EntitySpec` so there are no `brooklyn.effectors` in CAMP.

camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/creation/BrooklynEntityDecorationResolver.java
bostko: I introduced only String tags in the PR. I suggest documenting that numbers should be in quotes at this stage.

neykov:
+1 for supporting arbitrary objects, but making strings very easy to add.
What about DSL in tags? This is a prerequisite for supporting objects, but on the other hand tags need to be evaluated at the time the entity is constructed which restricts DSL quite a bit. Need to be careful not to set the DSL object itself as a tag.

geomacy:
Given that there's lots of discussion above, can I +1 the suggestion by @m4rkmckenna to take this to the mailing list as a proposal and invite discussion there?

camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/creation/BrooklynEntityDecorationResolver.java
```diff
@@ -227,7 +234,55 @@ protected Object getDecorationAttributeJsonValue(ConfigBag attrs) {
...
+        protected Object getDecorationAttributeJsonValue(ConfigBag attrs) {
+            Object brooklynTags = attrs.getStringKey(BrooklynCampReservedKeys.BROOKLYN_TAGS);
+            if (brooklynTags == null) {
+                return null;
+            } else if (!(brooklynTags instanceof List)) {
+                throw new IllegalArgumentException(BrooklynCampReservedKeys.BROOKLYN_TAGS + " should be a List of String elements. You supplied " + brooklynTags);
+            } else {
+                checkArgument(Iterables.all((List) brooklynTags, new Predicate() {
+                    @Override
+                    public boolean apply(@Nullable Object input) {
```
geomacy: Why is this marked nullable?

camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/creation/BrooklynComponentTemplateResolver.java
aledsage: Thanks @neykov - makes sense. However, a much bigger topic is how we really should be defining effectors/feeds in yaml. To me, the `brooklyn.initializers` feels like a powerful hack - re-using a generic feature to add things that are fundamental to blueprints. I think it's confusing for users to have to use `brooklyn.initializers` rather than us having first-class `brooklyn.effectors`. But let's not discuss that here - it can be a dev@brooklyn discussion when we're ready!

rest/rest-api/src/main/java/org/apache/brooklyn/rest/api/EntityApi.java
```diff
@@ -141,7 +141,7 @@ public Response getIcon(
 
     @GET
     @Path("/{entity}/tags")
-    @ApiOperation(value = "Fetch list of tags on this entity")
+    @ApiOperation(value = "Fetch list of tags on this entity. For retrieving brooklyn.tags filter only elements such that: an element is a map AND an element has kind: camp_brooklyn_tags.")
```
aledsage: Not sure we want to include this in the rest call's description. It's very technical, and doesn't tell people how to use the information (e.g. "filter" is not something that the rest api supports, it has to be done by the user after receiving the entire response).

I think this would be better to just include in the docs, with a more complete example of how to do it.

core/src/main/java/org/apache/brooklyn/core/mgmt/BrooklynTags.java
```diff
@@ -108,6 +123,39 @@ public TraitsTag(List<Class<?>> interfaces) {
         }
     }
 
+    public static class CampBrooklynTag implements Serializable {
```
aledsage: I'm not convinced we want to wrap the user's tag in a `CampBrooklynTag`, rather than just adding their tag directly. I'm guessing that things behave very differently if you do `brooklyn.tags: ["hi"]` then the result is very different from doing Java `entity.tags()addTag("hi")` - i.e. the latter has the tag "hi", whereas the former has the tag of type "CampBrooklynTag" that wraps the string "hi".

Other than being able to tell where the tag came from, what is the advantage of using `CampBrooklynTag`?

camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/TagsYamlTest.java
```diff
@@ -0,0 +1,171 @@
...
+
+        public TagsTestSerializableObject() {}
+        public TagsTestSerializableObject(Object arg1) {}
+    }
+
+    private static Set<Object> getTagsEventually(final Entity entity) throws Exception {
+        Task<Set<Object>> result = ((EntityInternal)entity).getExecutionContext().submit(new Callable<Set<Object>>() {
+            @Override
+            public Set<Object> call() throws Exception {
+                return entity.tags().getTags();
```
aledsage: Why do we need to call this inside a task? Does it fail without that? Under what circumstances would `entity.tags().getTags()` block, and timeout?

Why do we need "eventually" - is the tag potentially not there initially?

camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/TagsYamlTest.java
```diff
@@ -0,0 +1,171 @@
...
+            final Entity app = createAndStartApplication(
+                    "services:",
+                    "- type: " + BasicApplication.class.getName(),
+                    "  brooklyn.tags:",
+                    "  - tag1",
+                    "  - $brooklyn:object:",
+                    "      type: "+NonSerializable.class.getName());
+            fail("Should throw IllegalArgumentException exception.");
+        } catch (CompoundRuntimeException e) {
+            Asserts.assertStringContainsAtLeastOne(Exceptions.getFirstInteresting(e).getMessage(),"brooklyn.tags should not contain DeferredSupplier objects, such example is using $brooklyn:attributeWhenReady");
```
aledsage: This is a strange error message to assert - the test has passed in a non-serializable object, rather than a `DeferredSupplier`.

camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/creation/BrooklynEntityDecorationResolver.java
aledsage: As said earlier, I'd declare the return type as `Iterable<Object>` to avoid the cast in the caller.

neykov:
@aledsage @bostko I think we should insert the object as is, without wrapping it in `CampBrooklynTag`. The wrapper doesn't provide any useful functionality and in fact could lead to unexpected behaviour - the user is no longer able to control the type of the tag which defeats the purpose.

camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/TagsYamlTest.java
neykov: Agree, should be there since app creation.

core/src/main/java/org/apache/brooklyn/core/mgmt/BrooklynTags.java
```diff
@@ -38,7 +38,8 @@
     public static final String YAML_SPEC_KIND = "yaml_spec";
     public static final String NOTES_KIND = "notes";
     public static final String OWNER_ENTITY_ID = "owner_entity_id";
-    
+    public static final String BROOKLYN_CAMP_TAG_KIND = "camp_brooklyn_tags";
```
neykov: unused

camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/creation/BrooklynEntityDecorationResolver.java
```diff
@@ -227,7 +239,49 @@ protected Object getDecorationAttributeJsonValue(ConfigBag attrs) {
...
+            Object brooklynTags = attrs.getStringKey(BrooklynCampReservedKeys.BROOKLYN_TAGS);
+            if (brooklynTags == null) {
+                return null;
+            } else if (!(brooklynTags instanceof List)) {
+                throw new IllegalArgumentException(BrooklynCampReservedKeys.BROOKLYN_TAGS + " should be a List of String elements. You supplied " + brooklynTags);
+            } else {
+                checkArgument(Iterables.all((List) brooklynTags, new Predicate() {
+                    @Override
+                    public boolean apply(Object input) {
+                        return !(input instanceof DeferredSupplier) && (input instanceof Serializable);
```
neykov: We are not actually using the `Serializable` mechanism to persist objects so they don't need to implement the interface. Do we really need this check?

neykov:
LGTM. Haven't checked if all of @aledsage's comments are addressed - there's at least one re `getTagsEventually` so can you double check that. A couple of minor comments from me as well.

camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/creation/BrooklynEntityDecorationResolver.java
bostko: Still `NamedStringTag` is marked as `Serializable`.

neykov: That's not the only tag that we are setting. `WrappedEntity` and `ListTag` for example are not `Serializable`. As I said it's arbitrary/not required for the tags to be `Serializable`.

geomacy:
+1 LGTM; have run some tests successfully, and have checked back through all of @aledsage's and @neykov's comments, which are all addressed. 

Will merge.

PR #613, opened 2017-03-27
Returns catalog items created via REST API when coming from an archive

tbouron:
There is still an issue when the `FEATURE_LOAD_BUNDLE_CATALOG_BOM` is enabled; the REST API won't return the added items as it is currently handled by the `CatalogBomScanner`

But, this feature is disabled by default so we can probably live with that for now

### DIFF
--- a/rest/rest-resources/src/main/java/org/apache/brooklyn/rest/resources/CatalogResource.java
+++ b/rest/rest-resources/src/main/java/org/apache/brooklyn/rest/resources/CatalogResource.java
@@ -246,8 +246,12 @@ public class CatalogResource extends AbstractBrooklynRestResource implements Cat
                 }
             }
 
-            // TODO improve on this - If the FEATURE_LOAD_BUNDLE_CATALOG_BOM is enabled, the REST API won't return the
-            // added items which breaks the contract on the API endpoint.
+            // TODO improve on this - Currently, the added items are returned ONLY if the FEATURE_LOAD_BUNDLE_CATALOG_BOM
+            // is disabled. When enabled, the above code is not executed and the catalog items addition is delegated
+            // to the CatalogBomScanner. The REST API will therefore not know what are the added catalog items and won't
+            // return them.
+            // One way to improved this would be couple agoin the CatalogBomScanner with the CatalogBundleLoader to
+            // retrieve the list of added catalog items per bundle.
             return buildCreateResponse(catalogItems);
         } catch (RuntimeException ex) {
             throw WebResourceUtils.badRequest(ex);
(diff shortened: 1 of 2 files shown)


rest/rest-resources/src/main/java/org/apache/brooklyn/rest/resources/CatalogResource.java
```diff
@@ -261,15 +245,37 @@ public Response createFromArchive(byte[] zipInput) {
                     throw new IllegalArgumentException("Error installing catalog items", ex);
                 }
             }
-            
-            return Response.status(Status.CREATED).build();
+
+            // TODO improve on this - If the FEATURE_LOAD_BUNDLE_CATALOG_BOM is enabled, the REST API won't return the
```
geomacy: Could you expand this section a bit so it explains the issue in more detail

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/resources/CatalogResource.java
```diff
@@ -261,15 +245,37 @@ public Response createFromArchive(byte[] zipInput) {
...
     }
+
+    private Response buildCreateResponse(Iterable<? extends CatalogItem<?, ?>> catalogItems) {
+        log.info("REST created catalog items: "+catalogItems);
+
+        Map<String,Object> result = MutableMap.of();
+
+        for (CatalogItem<?,?> catalogItem: catalogItems) {
+            try {
+                result.put(catalogItem.getId(), CatalogTransformer.catalogItemSummary(brooklyn(), catalogItem, ui.getBaseUriBuilder()));
```
geomacy: Can you break this line so it's readable on github?

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/resources/CatalogResource.java
```diff
@@ -261,15 +245,37 @@ public Response createFromArchive(byte[] zipInput) {
...
+            } catch (Throwable t) {
+                log.warn("Error loading catalog item '"+catalogItem+"' (rethrowing): "+t);
+                // unfortunately items are already added to the catalog and hard to remove,
+                // but at least let the user know;
+                // happens eg if a class refers to a missing class, like
+                // loading nosql items including mongo without the mongo bson class on the classpath
+                throw Exceptions.propagateAnnotated("At least one unusable item was added ("+catalogItem.getId()+")", t);
```
geomacy: I suggest you collect the exception in here rather than throwing it - 
- let the loop run through all items, collect any failures that occur
- if there are any failures, loop through the catalog items that _were_ added and remove them again, before
- create a failure response with the details of each item that could not be added.

Having said that, I'm not sure there's much that really can go wrong above, as this is just creating the summary objects.  It would be good to apply the logic above to the `CatalogItemLoader.scanForCatalog`, but I don't think that is within the scope of this PR.

tbouron: @geomacy I move this code from the `createFromYaml` function, agree that it could be refactored but It is not in the scope of this PR. I would leave it for future improvement

geomacy:
LGTM, will merge this.

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/resources/CatalogResource.java
```diff
@@ -146,33 +146,15 @@ public Response createFromYaml(String yaml) {
...
         }
-        
-        Iterable<? extends CatalogItem<?, ?>> items; 
+
         try {
-            items = brooklyn().getCatalog().addItems(yaml);
+            final Iterable<? extends CatalogItem<?, ?>> items = brooklyn().getCatalog().addItems(yaml);
+            return buildCreateResponse(items);
         } catch (Exception e) {
             e.printStackTrace();
```
aledsage: We should never use `e.printStackTrace()` (not introduced in this PR; I just spotted it while looking over these changes)

PR #614, opened 2017-03-29
Fix Transformer’s DSL resolving for function vals

aledsage:


geomacy:
Looks good to me from visual review- nice tests! I shall take Duncan at his word that he has tested this. Will merge.

PR #616, opened 2017-03-31
Fixed BlobStoreContextFactoryImpl. It can now have dsl in its config

Graeme-Miller:
Fixes a bug introduced in this PR https://github.com/apache/brooklyn-server/commit/013b38bf0f146114d3a67423c0dcc3661b50f88c#diff-30cb47bc0f055776b4a246abeddd4ad7L99

Before that change we would ask the JCloudsLocation directly for config. JCloudsLocation uses a ResolvingConfigBag which will resolve dsl (e.g. "$brooklyn:external").

After the above change, we get the ConfigBag from the location first and ask the ConfigBag for the result. When we get the ConfigBag from the location we create a new ConfigBag to return, but we do not create a resolving config bag.

This PR changes back to asking the location directly for config

### DIFF
--- a/launcher-common/src/main/java/org/apache/brooklyn/launcher/command/support/CloudExplorerSupport.java
+++ b/launcher-common/src/main/java/org/apache/brooklyn/launcher/command/support/CloudExplorerSupport.java
@@ -329,7 +329,7 @@ public abstract class CloudExplorerSupport implements Callable<Void> {
 
         @Override
         protected void doCall(JcloudsLocation loc, String indent) throws Exception {
-            BlobStoreContext context = BlobStoreContextFactoryImpl.INSTANCE.newBlobStoreContext(loc.config().getBag());
+            BlobStoreContext context = BlobStoreContextFactoryImpl.INSTANCE.newBlobStoreContext(loc);
             try {
                 org.jclouds.blobstore.BlobStore blobStore = context.getBlobStore();
                 doCall(blobStore, indent);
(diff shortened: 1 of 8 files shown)


aledsage:
LGTM; I also ran the live tests in `org.apache.brooklyn.core.mgmt.persist.jclouds`: `BlobStoreTest` and `BlobStoreExpiryTest`.

It's a shame about the signature `BlobStoreContextFactory.newBlobStoreContext(Location)` - it taking a `ConfigBag` felt nicer, but I see how that doesn't work (without resorting to the `ResolvingConfigBag`, which feels overly complicated). It also means there is less symmetry with `ComputeServiceRegistry.findComputeService(ConfigBag conf, boolean allowReuse)`, but we can live with that as well.

I think we can treat `org.apache.brooklyn.api.location.Location.BlobStoreContextFactory` as being like an internal interface so don't need to worry about backwards compatibility.

A shame we don't have any non-live tests for this stuff. Longer term, we could add something akin to the `StubbedComputeServiceRegistry` but it's certainly not the place of this PR to add that.

Merging now.

sjcorbett:
@aledsage Why should `BlobStoreContextFactory` be regarded as an internal interface? It's already being used in a downstream project that this PR has broken. Should the project use something else?

aledsage:
@sjcorbett oops, I didn't realise it was in use. In hindsight, it's understandable that power-users in downstream projects would implement this so they can get the exact (configuration of) `BlobStoreContext` that they need.

Other than the hassle of updating your impl, do the changes here make things particularly difficult for you?


---
I previously had in my head that we'd really only have two impls: the `BlobStoreContextFactoryImpl` and a stub test impl for unit testing (which isn't written yet).

That's how `ComputeServiceRegistry` is used, I believe (but who knows what folk have done in their downstream projects!).

I hope that `BlobStoreContextFactoryImpl` can become configurable enough that folk don't need to create their own implementations, but maybe that's unrealistic.

sjcorbett:
There's not a good parallel in the downstream project at the moment. It only has the config bag to use, not a location. I think I'll just take a copy of the previous version of the class.

PR #617, opened 2017-04-03
Persist management plane ID

neykov:
Some considerations on where to place the planeId initialisation:
 * BrooklynMemento (RebindManager). `planeId` will be available only after rebind, so it will
   be available to MASTER nodes only (and potentially HOT_STANDBY)
 * ManagementPlaneSyncRecord (HighAvailabilityManager). Used only when HA is enabled. When HA is disabled
   HighAvailabilityManager is not initialised, doesn't access the store at all, no
   records are getting written.
 * Separate init step in BrooklynLauncher. Early in the lifecycle so users need not to care
   too much when they are allowed to read it. On the other hand it's not encapsulated so
   need to touch all places that use persistence (i.e. launcher, copy-state, backup). Can
   only read from the store at this point, write access is not enabled yet. If HA is configured
   store writes are enabled only after becoming master. Writing to the store in this case would
   be risky, even if we know that it's a clean state (no planeId file). Better leave only the
   master update the store.

Because of the above I decided to go for a mixed approach:
  * persist the `planeId` immediately after starting persistence (i.e. MASTER when HA) and repeat hourly
  * don't overwrite persisted store `planeId` if different
  * init `planeId` as part of the rebind sequence; if no `planeId` exists in store generate one
  * init `planeId` when reading the management records so we have it initialized for non-`MASTER` nodes as well; don't generate one if not in persisted store

### DIFF
--- a/api/src/main/java/org/apache/brooklyn/api/mgmt/ManagementContext.java
+++ b/api/src/main/java/org/apache/brooklyn/api/mgmt/ManagementContext.java
@@ -67,18 +67,33 @@ public interface ManagementContext {
      * In other words the value of {@link Application#getManagementContext()#getManagementPlaneId()} 
      * will generally be constant (in contrast to {@link #getManagementNodeId()}).
      * <p>
-     * Throws an {@link NullPointerException} if the value hasn't been initialized yet. The value is set:
+     * This value should not be null unless the management context is still initialising. The value is set:
      * <ul>
      *   <li>no persistence - during launch
      *   <li>persistence enabled, HA disabled - on rebind (during launch)
      *   <li>persistence enabled, HA enabled - on the first HA state check (async to launch)
      * </ul>
+     * 
+     * @deprecated since 0.11.0, use {@link #getOptionalManagementPlaneId()} instead.
      */
+    @Deprecated
     String getManagementPlaneId();
 
     /**
-     * Same as {@link #getManagementPlaneId()}, but will return {@link Optional#absent()} if the
-     * {@code managementPlaneId} hasn't been initialized yet.
+     * UID for the Brooklyn management plane which this {@link ManagementContext} node is a part of.
+     * <p>
+     * Each Brooklyn entity is actively managed by a unique management plane 
+     * whose ID which should not normally change for the duration of that entity, 
+     * even though the nodes in that plane might, and the plane may go down and come back up. 
+     * In other words the value of {@link Application#getManagementContext()#getManagementPlaneId()} 
+     * will generally be constant (in contrast to {@link #getManagementNodeId()}).
+     * <p>
+     * Returns absent while the management context is still initialising. The value is set:
+     * <ul>
+     *   <li>no persistence - during launch
+     *   <li>persistence enabled, HA disabled - on rebind (during launch)
+     *   <li>persistence enabled, HA enabled - on the first HA state check (async to launch)
+     * </ul>
      */
     Optional<String> getOptionalManagementPlaneId();
     
@@ -87,7 +102,7 @@ public interface ManagementContext {
      * <p>
      * No two instances of {@link ManagementContext} should ever have the same node UID. 
      * The value of {@link Application#getManagementContext()#getManagementNodeId()} may
-     * change many times (in contrast to {@link #getManagementPlaneId()}). 
+     * change many times (in contrast to {@link #getOptionalManagementPlaneId()}). 
      * <p>
      * This value should not be null unless the management context is a non-functional
      * (non-deployment) instance. */
--- a/api/src/main/java/org/apache/brooklyn/api/mgmt/ManagementContext.java
+++ b/api/src/main/java/org/apache/brooklyn/api/mgmt/ManagementContext.java
@@ -39,6 +39,7 @@ import org.apache.brooklyn.config.StringConfigMap;
 import org.apache.brooklyn.util.guava.Maybe;
 
 import com.google.common.annotations.Beta;
+import com.google.common.base.Optional;
 
 /**
  * This is the entry point for accessing and interacting with a realm of applications and their entities in Brooklyn.
@@ -66,9 +67,20 @@ public interface ManagementContext {
      * In other words the value of {@link Application#getManagementContext()#getManagementPlaneId()} 
      * will generally be constant (in contrast to {@link #getManagementNodeId()}).
      * <p>
-     * This value should not be null unless the management context is a non-functional
-     * (non-deployment) instance. */
+     * Throws an {@link NullPointerException} if the value hasn't been initialized yet. The value is set:
+     * <ul>
+     *   <li>no persistence - during launch
+     *   <li>persistence enabled, HA disabled - on rebind (during launch)
+     *   <li>persistence enabled, HA enabled - on the first HA state check (async to launch)
+     * </ul>
+     */
     String getManagementPlaneId();
+
+    /**
+     * Same as {@link #getManagementPlaneId()}, but will return {@link Optional#absent()} if the
+     * {@code managementPlaneId} hasn't been initialized yet.
+     */
+    Optional<String> getOptionalManagementPlaneId();
     
     /** 
      * UID for this {@link ManagementContext} node (as part of a single management plane).
--- a/core/src/main/java/org/apache/brooklyn/core/mgmt/ha/HighAvailabilityManagerImpl.java
+++ b/core/src/main/java/org/apache/brooklyn/core/mgmt/ha/HighAvailabilityManagerImpl.java
@@ -284,7 +284,7 @@ public class HighAvailabilityManagerImpl implements HighAvailabilityManager {
         boolean weAreRecognisedAsMaster = existingMaster!=null && ownNodeId.equals(existingMaster.getNodeId());
         boolean weAreMasterLocally = getInternalNodeState()==ManagementNodeState.MASTER;
         
-        updatePlaneId(planeRec);
+        updateLocalPlaneId(planeRec);
         
         // catch error in some tests where mgmt context has a different HA manager
         if (managementContext.getHighAvailabilityManager()!=this)
@@ -460,7 +460,7 @@ public class HighAvailabilityManagerImpl implements HighAvailabilityManager {
             registerPollTask();
     }
 
-    protected void updatePlaneId(ManagementPlaneSyncRecord existingMaster) {
+    protected void updateLocalPlaneId(ManagementPlaneSyncRecord existingMaster) {
         if (existingMaster.getPlaneId() != null) {
             ((LocalManagementContext)managementContext).setManagementPlaneId(existingMaster.getPlaneId());
         }
@@ -722,7 +722,7 @@ public class HighAvailabilityManagerImpl implements HighAvailabilityManager {
             return;
         }
         
-        updatePlaneId(memento);
+        updateLocalPlaneId(memento);
         
         String currMasterNodeId = memento.getMasterNodeId();
         ManagementNodeSyncRecord currMasterNodeRecord = memento.getManagementNodes().get(currMasterNodeId);
--- a/core/src/main/java/org/apache/brooklyn/core/mgmt/ha/HighAvailabilityManagerImpl.java
+++ b/core/src/main/java/org/apache/brooklyn/core/mgmt/ha/HighAvailabilityManagerImpl.java
@@ -56,6 +56,7 @@ import org.apache.brooklyn.core.mgmt.ha.dto.ManagementPlaneSyncRecordImpl;
 import org.apache.brooklyn.core.mgmt.ha.dto.ManagementPlaneSyncRecordImpl.Builder;
 import org.apache.brooklyn.core.mgmt.internal.BrooklynObjectManagementMode;
 import org.apache.brooklyn.core.mgmt.internal.LocalEntityManager;
+import org.apache.brooklyn.core.mgmt.internal.LocalManagementContext;
 import org.apache.brooklyn.core.mgmt.internal.LocationManagerInternal;
 import org.apache.brooklyn.core.mgmt.internal.ManagementContextInternal;
 import org.apache.brooklyn.core.mgmt.internal.ManagementTransitionMode;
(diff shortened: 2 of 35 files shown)


api/src/main/java/org/apache/brooklyn/api/mgmt/ManagementContext.java
```diff
@@ -66,9 +67,20 @@
      * In other words the value of {@link Application#getManagementContext()#getManagementPlaneId()} 
      * will generally be constant (in contrast to {@link #getManagementNodeId()}).
      * <p>
-     * This value should not be null unless the management context is a non-functional
-     * (non-deployment) instance. */
+     * Throws an {@link NullPointerException} if the value hasn't been initialized yet. The value is set:
```
neykov: Wonder whether we should leave this to return `null` when not initialised. It's been useful to find instances of wrong (too early) usage but on the other hand changes the method behaviour.

core/src/main/java/org/apache/brooklyn/core/mgmt/persist/BrooklynMementoPersisterToObjectStore.java
```diff
@@ -674,6 +685,22 @@ private void delete(String subPath, String id, PersistenceExceptionHandler excep
...
 
+    private void updatePlaneId(String planeId, PersistenceExceptionHandler exceptionHandler) {
+        try {
+            if (planeId==null) {
+                LOG.warn("Null content for planeId");
+            }
+
+            String persistedPlaneId = read(PLANE_ID_FILE_NAME);
+            if (persistedPlaneId != null && !persistedPlaneId.equals(planeId)) {
+                throw new IllegalStateException("Persisted planeId found (" + persistedPlaneId + ") but instance planeId is different (" + planeId + "). Overwriting!");
```
geomacy: "Overwriting!"?  does throwing the exception here not in fact skip the overwrite?  As per your comment on the description of the PR, "don't overwrite persisted store planeId if different"?

core/src/main/java/org/apache/brooklyn/core/mgmt/rebind/PeriodicDeltaChangeListener.java
```diff
@@ -382,6 +401,8 @@ protected void persistNowInternal(boolean alreadyHasMutex) {
             if (!alreadyHasMutex) persistingMutex.acquire();
             if (!isActive() && state != ListenerState.STOPPING) return;
             
+            updatePlaneIdIfTimedOut();
```
geomacy: What's the motivation for updating the plane id like this?  When would you expect this to be required?

core/src/test/java/org/apache/brooklyn/core/mgmt/rebind/ManagementPlaneIdTest.java
```diff
@@ -0,0 +1,251 @@
...
+
+    private void markForTermination(ManagementContext mgmt) {
+        managementContextForTermination.add(mgmt);
+    }
+
+    protected static String readFile(File planeIdFile) throws IOException {
+        return new String(Files.readAllBytes(planeIdFile.toPath()), StandardCharsets.UTF_8);
+    }
+
+    protected void checkRebindIdPersisted(final ManagementContext mgmt) {
```
geomacy: `checkPlaneIdPersisted` surely?

core/src/test/java/org/apache/brooklyn/core/mgmt/rebind/ManagementPlaneIdTest.java
```diff
@@ -0,0 +1,251 @@
...
+        return createManagementContext(persistMode, haMode, true);
+    }
+    
+    protected LocalManagementContext createManagementContext(PersistMode persistMode, HighAvailabilityMode haMode, boolean backedUp) {
+        BrooklynProperties props = BrooklynProperties.Factory.newEmpty();
+        props.put(BrooklynServerConfig.PERSISTENCE_BACKUPS_DIR, mementoDir);
+        LocalManagementContext mgmt = RebindTestUtils.managementContextBuilder(mementoDir, classLoader)
+                .persistPeriodMillis(1)
+                .persistMode(persistMode)
+                .haMode(HighAvailabilityMode.DISABLED)
```
geomacy: `.haMode(haMode)` ?   Currently `haMode` parameter is unused.

core/src/main/java/org/apache/brooklyn/core/mgmt/ha/HighAvailabilityManagerImpl.java
```diff
@@ -456,6 +460,12 @@ else if (startMode==HighAvailabilityMode.HOT_STANDBY && getNodeState()==Manageme
             registerPollTask();
     }
 
+    protected void updatePlaneId(ManagementPlaneSyncRecord existingMaster) {
```
neykov: Renaming to `updateLocalPlaneId`. It would be surprising to update with a different planeId, but I decided to support the use case where someone goes and edits the persisted state planeId. In case the planeId is different it will log a warning, but will accept the change.

core/src/main/java/org/apache/brooklyn/core/mgmt/persist/BrooklynMementoPersisterToObjectStore.java
neykov: Good catch - it used to overwrite but then I changed my mind. The persisted state is considered the "ground truth" now, so won't overwrite the planeId if different.

core/src/main/java/org/apache/brooklyn/core/mgmt/rebind/PeriodicDeltaChangeListener.java
neykov: Writes to the datastore are lossy. We'll just log failures and move on. (Most) entities will get updated multiple times in their lifecycle so not a huge deal. planeId does not get updated so if the first write fails it's not available to the HA cluster at all. That's why I decided to periodically write it to the datastore. Will add a comment.

geomacy: Hm; if they are lossy then for planeId would we need to write it using a "keep trying and double-check until successful" approach?  I.e. don't progress until write is confirmed?

neykov: I agree it makes sense but would do it for all writes, not just the planeId. Not comfortable with doing it just for planeId - would break the existing assumption about writes.

neykov:
Thanks @geomacy. Addressed comments.

core/src/main/java/org/apache/brooklyn/core/mgmt/persist/BrooklynMementoPersisterToObjectStore.java
```diff
@@ -674,6 +685,22 @@ private void delete(String subPath, String id, PersistenceExceptionHandler excep
...
+        try {
+            if (planeId==null) {
+                LOG.warn("Null content for planeId");
+            }
+
+            String persistedPlaneId = read(PLANE_ID_FILE_NAME);
+            if (persistedPlaneId != null && !persistedPlaneId.equals(planeId)) {
+                throw new IllegalStateException("Persisted planeId found (" + persistedPlaneId + ") but instance planeId is different (" + planeId + ")");
+            }
+            getWriter(PLANE_ID_FILE_NAME).put(planeId);
```
aledsage: Can we skip the `write` if the `read` returned the expected plane id?

core/src/main/java/org/apache/brooklyn/core/mgmt/rebind/RebindIteration.java
```diff
@@ -467,6 +469,15 @@ protected void instantiateMementos() throws IOException {
         memento = persistenceStoreAccess.loadMemento(mementoRawData, rebindContext.lookup(), exceptionHandler);
     }
 
+    protected void initPlaneId() {
+        String persistedPlaneId = mementoRawData.getPlaneId();
+        if (persistedPlaneId == null) {
+            ((LocalManagementContext)managementContext).generateManagementPlaneId();
```
aledsage: On balance, I agree with this code. But it feels like we're relying on null too much. We don't distinguish between persisted state that has no entities etc, and an unpopulated persisted state directory - that makes it hard to write code that asserts/warns it's in the state we expect.

core/src/test/java/org/apache/brooklyn/core/mgmt/rebind/ManagementPlaneIdTest.java
```diff
@@ -0,0 +1,247 @@
...
+        assertFalse(mgmt.getOptionalManagementPlaneId().isPresent(), "expected managementPlaneId to be absent");
+    }
+    
+    @Test
+    public void testPlaneIdPersists() throws Exception {
+        final ManagementContext mgmt = createManagementContext(PersistMode.AUTO, HighAvailabilityMode.DISABLED);
+        checkPlaneIdPersisted(mgmt);
+    }
+
+    @Test(enabled=false) //Need to wait for 1hr to verify the update will not overwrite
```
aledsage: Perhaps a sign that we should make that period configurable.

core/src/test/java/org/apache/brooklyn/core/mgmt/rebind/ManagementPlaneIdTest.java
```diff
@@ -0,0 +1,247 @@
...
+import org.apache.brooklyn.core.server.BrooklynServerPaths;
+import org.apache.brooklyn.test.Asserts;
+import org.apache.brooklyn.util.os.Os;
+import org.apache.brooklyn.util.text.Strings;
+import org.testng.annotations.AfterMethod;
+import org.testng.annotations.BeforeMethod;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+public class ManagementPlaneIdTest {
```
aledsage: Can we extend `RebindTestFixture`? Would that save us much code, e.g. simplifying the `createManagementContext`? I'm guessing not, because it will have already created the managementContext in setUp. No strong feelings from me.

core/src/main/java/org/apache/brooklyn/core/mgmt/persist/BrooklynMementoPersisterToObjectStore.java
neykov: +1

core/src/main/java/org/apache/brooklyn/core/mgmt/rebind/RebindIteration.java
neykov: Added a warning here if there's already existing state, but no planeId.

core/src/test/java/org/apache/brooklyn/core/mgmt/rebind/ManagementPlaneIdTest.java
neykov: That's how the test started, but I needed some functionality which wasn't available in `RebindTestFixture` (can't remember the details now)

core/src/test/java/org/apache/brooklyn/core/mgmt/rebind/ManagementPlaneIdTest.java
neykov: Reversed the test to verify that the local planeId is "fixed" by the persisted state so it's quick now.

neykov:
Thanks @aledsage addressed the comments.

> I'd be interested if this has highlighted any areas of Brooklyn that we should improve, 
There's definitely room for improvement around HA + persistence. One example is we could clean up  the code around that's responsible for starting persistence/HA in the various cases (ha, noha * persistence, no persistence matrix). It's something I had to replicate in the [test - see TODO](https://github.com/apache/brooklyn-server/pull/617/files#diff-37574fbba472cb8c35ca723a3118ed93R268). All interactions should go through `HighAvailabilityManager` enabled or not.

core/src/main/java/org/apache/brooklyn/core/mgmt/internal/LocalManagementContext.java
```diff
@@ -224,6 +224,7 @@ public String getManagementPlaneId() {
     public void setManagementPlaneId(String newPlaneId) {
         if (managementPlaneId != null && !managementPlaneId.equals(newPlaneId)) {
             log.warn("Management plane ID changed from {} to {}", managementPlaneId, newPlaneId);
+            log.debug("Management plane ID changed from {} to {}", new Object[] {managementPlaneId, newPlaneId}, new RuntimeException("Stack trace for setManagementPlaneId"));
```
geomacy: I think this actually gives you `debug(String, Object, Object)` rather than an overload that knows about `Throwable`, so you'll not get the right parameter matching here.

neykov: Good spot @geomacy. The fix might surprise you though (it surely did surprise me):
```
log.debug("Management plane ID changed from {} to {}", new Object[] {managementPlaneId, newPlaneId, new RuntimeException("Stack trace for setManagementPlaneId")});
```

Turns out slf4j will [do the right thing](https://www.slf4j.org/faq.html#paramException) in this case and log the last argument as an exception. Verified the debug log contains:

```
2017-04-19 15:11:34,011 DEBUG o.a.b.c.m.i.LocalManagementContext [brooklyn-execmanager-KWNkyI3R-3]: Management plane ID changed from X7dpWzuF to wJl3j3K1
java.lang.RuntimeException: Stack trace for setManagementPlaneId
        at org.apache.brooklyn.core.mgmt.internal.LocalManagementContext.setManagementPlaneId(LocalManagementContext.java:227) [classes/:na]
        at org.apache.brooklyn.core.mgmt.ha.HighAvailabilityManagerImpl.updateLocalPlaneId(HighAvailabilityManagerImpl.java:465) [classes/:na]
        at org.apache.brooklyn.core.mgmt.ha.HighAvailabilityManagerImpl.checkMaster(HighAvailabilityManagerImpl.java:725) [classes/:na]
        at org.apache.brooklyn.core.mgmt.ha.HighAvailabilityManagerImpl.publishAndCheck(HighAvailabilityManagerImpl.java:602) [classes/:na]
        at org.apache.brooklyn.core.mgmt.ha.HighAvailabilityManagerImpl$2.run(HighAvailabilityManagerImpl.java:557) [classes/:na]
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:471) [na:1.7.0_80]
        at org.apache.brooklyn.util.core.task.BasicExecutionManager$ScheduledTaskCallable$1.call(BasicExecutionManager.java:451) [classes/:na]
        at org.apache.brooklyn.util.core.task.BasicExecutionManager$SubmissionCallable.call(BasicExecutionManager.java:529) [classes/:na]
        at java.util.concurrent.FutureTask.run(FutureTask.java:262) [na:1.7.0_80]
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1145) [na:1.7.0_80]
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:615) [na:1.7.0_80]
        at java.lang.Thread.run(Thread.java:745) [na:1.7.0_80]

```

geomacy: Hm, interesting!  That's new to me - a good one to know!

aledsage:
This fixes https://issues.apache.org/jira/browse/BROOKLYN-202

PR #619, opened 2017-04-05
Add greaterThan and lessThan test framework assertions

sjcorbett:
Each relies on the two objects being compared being instances of the same `Comparable<T>`.

I also took the opportunity to simplify the logic in `TestAssertions`.

### DIFF
--- a/test-framework/src/main/java/org/apache/brooklyn/test/framework/TestFrameworkAssertions.java
+++ b/test-framework/src/main/java/org/apache/brooklyn/test/framework/TestFrameworkAssertions.java
@@ -23,7 +23,9 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Set;
 import java.util.concurrent.Callable;
+import javax.annotation.Nullable;
 
 import org.apache.brooklyn.api.entity.Entity;
 import org.apache.brooklyn.config.ConfigKey;
@@ -44,6 +46,7 @@ import com.google.common.base.Predicate;
 import com.google.common.base.Predicates;
 import com.google.common.base.Supplier;
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
 import com.google.common.reflect.TypeToken;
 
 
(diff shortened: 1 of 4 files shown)


test-framework/src/main/java/org/apache/brooklyn/test/framework/TestFrameworkAssertions.java
```diff
@@ -287,140 +292,86 @@ public Boolean call() {
...
+            return false;
+        }
+    }
+
+    private static boolean knownCondition(String condition) {
+        // The conditions should really be an enum!
+        Set<String> allConditions = ImmutableSet.of(
+                IS_NULL, NOT_NULL, IS_EQUAL_TO, EQUAL_TO, EQUALS, NOT_EQUAL,
+                MATCHES, CONTAINS, IS_EMPTY, NOT_EMPTY, HAS_TRUTH_VALUE,
+                GREATER_THAN, LESS_THAN, UNKNOWN_CONDITION);
```
aledsage: I don't think `UNKNOWN_CONDITION` should be included here in the `knownCondition`s.

test-framework/src/main/java/org/apache/brooklyn/test/framework/TestFrameworkAssertions.java
```diff
@@ -287,140 +292,86 @@ public Boolean call() {
...
+        return allConditions.contains(condition);
+    }
+
+    /** @return True if actual and expected are both non-null instances of {@code Comparable<T>}. */
+    private static boolean canCompare(@Nullable Object actual, @Nullable Object expected) {
+        return actual != null
+                && expected != null
+                && actual instanceof Comparable
+                && actual.getClass().equals(expected.getClass());
```
aledsage: I wonder if we should do any coercion, or perhaps even just give it a go to compare! For example, comparing `(Integer)1` with `(Long)2` might well be possible. However, on balance I think your code is right as it is the simplest (and avoid `ClassCastException`).

geomacy:
LGTM,  & @aledsage's comment has been addressed.  Will merge.

test-framework/src/main/java/org/apache/brooklyn/test/framework/TestFrameworkAssertions.java
sjcorbett: Interesting one. This implementation has already bitten me: "webapp.reqs.perSec.last expected greaterThan 80 but found 96.8063872255489". I'm inclined to leave this as-is until there are more complaints. I do think that this is to restrictive though - `actual` and `expected` could be different types if `Comparable` is implemented by a shared superclass. Don't think it'll be a common problem though.

PR #624, opened 2017-04-07
Adds authSudo support (for sudo’ing with password)

aledsage:
Required for some clouds (e.g. azure-arm), when using `useJcloudsSshInit: false`

Tested on `azure-arm` using a location like that below (which previously failed):

```
brooklyn.catalog:
  id: azure-arm-lon-centos7
  name: "Azure ARM UK South Centos 7.3"
  itemType: location
  item:
    type: jclouds:azurecompute-arm
    brooklyn.config:
      identity: xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
      credential: xxxxxxxx
      endpoint: https://management.azure.com/subscriptions/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
      oauth.endpoint: https://login.microsoftonline.com/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx/oauth2/token
      jclouds.azurecompute.arm.publishers: OpenLogic
      jclouds.azurecompute.operation.timeout: 120000
  
      jclouds.compute.resourcename-prefix: myprefix
      useJcloudsSshInit: false
      destroyOnFailure: false
      machineCreateAttempts: 1
      logCredentials: true

      installDevUrandom: true
      osVersionRegex: 7.3
      region: uksouth
      minRam: 2000
      waitForSshable: 10m
      waitForConnectable: 10m
      user: myname
```

The reason this works is that `authSudo` is set in `JcloudsLocation` based on the jclouds `LoginCredentialsshouldAuthenticateSudo()`. jclouds gets this from a combination of the `jclouds.image.authenticate-sudo` property and the equivalent template option (https://github.com/jclouds/jclouds/blob/rel/jclouds-2.0.1/compute/src/main/java/org/jclouds/compute/options/TemplateOptions.java#L196). For azure-arm, jclouds hard-codes the default to true so it works out-of-the-box (see https://github.com/jclouds/jclouds-labs/blob/rel/jclouds-labs-2.0.1/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/AzureComputeProviderMetadata.java#L96).

### DIFF
--- a/utils/common/src/main/java/org/apache/brooklyn/util/ssh/BashCommands.java
+++ b/utils/common/src/main/java/org/apache/brooklyn/util/ssh/BashCommands.java
@@ -91,10 +91,21 @@ public class BashCommands {
      * If null is supplied, it is returned (sometimes used to indicate no command desired).
      */
     public static String sudo(String command) {
+        if (command==null) return null;
         if (command.startsWith("( ") || command.endsWith(" &"))
             return sudoNew(command);
         else
             return sudoOld(command);
+    }
+
+    public static String authSudo(String command, String password) {
+        checkNotNull(password, "password must not be null");
+        if (command==null) return null;
+        if (command.startsWith("( ") || command.endsWith(" &")) {
+            throw new UnsupportedOperationException("authSudo supports only simple commands, not those wrapped in parentheses or backgrounded: cmd="+command);
+        }
+        // some OS's (which?) fail if you try running sudo when you're already root (dumb but true)
+        return format("( if test \"$UID\" -eq 0; then ( %s ); else echo -e '%s\\n' | sudo -E -S -- %s; fi )", command, password, command);
     }
 
     // TODO would like to move away from sudoOld -- but needs extensive testing!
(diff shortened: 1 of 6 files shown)


utils/common/src/main/java/org/apache/brooklyn/util/ssh/BashCommands.java
```diff
@@ -91,12 +91,23 @@ public static String ok(String command) {
...
     }
 
+    public static String authSudo(String command, String password) {
+        checkNotNull(password, "password must not be null");
+        if (command==null) return null;
+        if (command.startsWith("( ") || command.endsWith(" &")) {
+            throw new UnsupportedOperationException("authSudo supports only simple commands, not those wrapped in parentheses or backgrounded: cmd="+command);
+        }
+        // some OS's (which?) fail if you try running sudo when you're already root (dumb but true)
+        return format("( if test \"$UID\" -eq 0; then ( %s ); else echo -e '%s\\n' | sudo -E -S -- %s; fi )", command, password, command);
```
neykov: Will fail if `password` contains single quotes.

neykov:
+1 apart from password escape concern.

PR #625, opened 2017-04-10
Updates BasicExternalConfigSupplierRegistry to use ClassLoaderUtils

nakomis:


core/src/main/java/org/apache/brooklyn/core/mgmt/internal/BasicExternalConfigSupplierRegistry.java
```diff
@@ -103,12 +103,13 @@ private void updateFromBrooklynProperties(ManagementContext mgmt) {
             BrooklynProperties config = ConfigUtils.filterForPrefixAndStrip(externalProviderProperties, key + ".");
 
             try {
-                Maybe<ExternalConfigSupplier> configSupplier = Reflections.invokeConstructorFromArgs(classloader, ExternalConfigSupplier.class, providerClassname, mgmt, name, config);
+                Class<ExternalConfigSupplier> supplierClass = (Class<ExternalConfigSupplier>)new ClassLoaderUtils(this, mgmt).loadClass(providerClassname);
+                Maybe<ExternalConfigSupplier> configSupplier = Reflections.invokeConstructorFromArgs(supplierClass, new Object[] {mgmt, name, config});
                 if (!configSupplier.isPresent()) {
-                    configSupplier = Reflections.invokeConstructorFromArgs(classloader, ExternalConfigSupplier.class, providerClassname, mgmt, name, config.asMapWithStringKeys());
+                    configSupplier = Reflections.invokeConstructorFromArgs(supplierClass, new Object[] {mgmt, name, config.asMapWithStringKeys()});
```
neykov: No need to create the array, you can simplify to: 
```
Reflections.invokeConstructorFromArgs(supplierClass, mgmt, name, config.asMapWithStringKeys());
```

utils/common/src/main/java/org/apache/brooklyn/util/javalang/Reflections.java
```diff
@@ -204,6 +209,7 @@ public Object loadInstance(String classname) throws ReflectionNotFoundException,
     }
 
     /** does not look through ancestors of outer class */
+    /** @deprecated since 0.11.0, use {@link org.apache.brooklyn.util.core.ClassLoaderUtils} in a combination with {@link #invokeConstructorFromArgs(Class, Object...)} instead */
```
neykov: Can you deprecate the corresponding `invokeConstructorFromArgs`  methods as well.

neykov:
Looks good.
Can you add test coverage in `ExternalConfigBrooklynPropertiesTest` by implementing a dummy `ExternalConfigSupplier` in 
 * `brooklyn-server/utils/common/dependencies/osgi/com-example-entities/src/main/java/com/example/brooklyn/test/osgi` and trying to load it with a bundle prefix
  * `brooklyn-server/utils/common/dependencies/osgi/entities/src/main/java/org/apache/brooklyn/test/osgi` and trying to load it without prefix.

To add the bundle to the osgi cache follow the example in `ClassLoaderUtilsTest.installBundle(ManagementContext, String)` to load the bundle from the classpath.

core/src/main/java/org/apache/brooklyn/core/mgmt/internal/BasicExternalConfigSupplierRegistry.java
```diff
@@ -103,12 +103,13 @@ private void updateFromBrooklynProperties(ManagementContext mgmt) {
             BrooklynProperties config = ConfigUtils.filterForPrefixAndStrip(externalProviderProperties, key + ".");
 
             try {
-                Maybe<ExternalConfigSupplier> configSupplier = Reflections.invokeConstructorFromArgs(classloader, ExternalConfigSupplier.class, providerClassname, mgmt, name, config);
+                Class<ExternalConfigSupplier> supplierClass = (Class<ExternalConfigSupplier>)new ClassLoaderUtils(this, mgmt).loadClass(providerClassname);
+                Maybe<ExternalConfigSupplier> configSupplier = Reflections.invokeConstructorFromArgs(supplierClass, mgmt, name, config);
                 if (!configSupplier.isPresent()) {
```
aledsage: I think we should only try to load `supplierClass` if `configSupplier` is absent (i.e. move that line to inside the if below.

utils/common/dependencies/osgi/com-example-entities/src/main/java/com/example/brooklyn/test/osgi/PrefixedDummyExternalConfigSupplier.java
```diff
@@ -0,0 +1,33 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package com.example.brooklyn.test.osgi;
+
+import org.apache.brooklyn.core.config.external.ExternalConfigSupplier;
+
+public class DummyExternalConfigSupplier implements ExternalConfigSupplier {
```
aledsage: This classname doesn't match its filename.

utils/common/dependencies/osgi/entities/src/main/java/org/apache/brooklyn/test/osgi/UnprefixedDummyExternalConfigSupplier.java
```diff
@@ -0,0 +1,33 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package com.example.brooklyn.test.osgi;
+
+import org.apache.brooklyn.core.config.external.ExternalConfigSupplier;
+
+public class DummyExternalConfigSupplier implements ExternalConfigSupplier {
```
aledsage: This classname doesn't match its filename.

utils/common/src/main/java/org/apache/brooklyn/util/javalang/Reflections.java
aledsage: Alternatively we could allow `Reflections to be constructed with a `ClassLoaderUtils`, and deprecate the method that creates it with a `ClassLoader`. We'd also deprecate the static methods that take a `ClassLoader`?

neykov: `Reflections` is in `utils-common`, while `ClassLoaderUtils` is in `core`. Could move the string based methods to another utility class into `core`.  If the caller is going to be constructing the `ClassLoaderUtils` there's not much to gain from it though.

core/src/main/java/org/apache/brooklyn/core/mgmt/internal/BasicExternalConfigSupplierRegistry.java
nakomis: `supplierClass` is passed to `Reflections.invokeConstructorFromArgs` so it needs to be loaded first

nakomis:
Tests added at https://github.com/apache/brooklyn-dist/pull/89

neykov:
LGTM, merging.

PR #632, opened 2017-04-12
Resolve release build failures due to optional JARs

rdowner:
Some testing resources are JARs and cannot be included in the release build. Tests must be written to gracefully handle the absence of those test resources. Some new tests failed to include the check for the presence of the resources and therefore fail the release build.


tbouron:
Tested, LGTM 👍

PR #634, opened 2017-04-13
BROOKLYN-474: rebind to policy with private constructor

aledsage:


Graeme-Miller:
lgtm

bostko:
Are there other use cases for using private classes besides BROOKLYN-474?

aledsage:
@bostko I don't know of any other such private classes. Note that these are private policy/enricher/location/entity classes.  Other classes, such as those used as config key values, will be deserialized by xstream which will handle private constructors.

PR #635, opened 2017-04-13
Allow passing "latest" as version for catalog endpoints

tbouron:
This allow passing `latest` as the version for the following endpoints:
- `GET /v1/catalog/applications/<symbolicName>/<version>`
- `GET /v1/catalog/entities/<symbolicName>/<version>`
- `GET /v1/catalog/policies/<symbolicName>/<version>`
- `GET /v1/catalog/locations/<symbolicName>/<version>`
- `DELETE /v1/catalog/applications/<symbolicName>/<version>`
- `DELETE /v1/catalog/entities/<symbolicName>/<version>`
- `DELETE /v1/catalog/policies/<symbolicName>/<version>`
- `DELETE /v1/catalog/locations/<symbolicName>/<version>`

For the `GET` requests, the latest catalog item matching the `symbolicName` will be returned. Similarly, for the `DELETE` requests, the latest catalog item matching the `symbolicName` will be deleted.

This is very useful to get the information back of an entity if one don't currently know the version. Most of the times, one will be interested in the latest version.

### DIFF
--- a/rest/rest-resources/src/main/java/org/apache/brooklyn/rest/resources/CatalogResource.java
+++ b/rest/rest-resources/src/main/java/org/apache/brooklyn/rest/resources/CatalogResource.java
@@ -114,6 +114,13 @@ public class CatalogResource extends AbstractBrooklynRestResource implements Cat
         };
     };
 
+    private String processVersion(String version) {
+        if (version != null && LATEST.equals(version.toLowerCase())) {
+            version = null;
+        }
+        return version;
+    }
+
     static Set<String> missingIcons = MutableSet.of();
 
     @Override
@@ -333,9 +340,7 @@ public class CatalogResource extends AbstractBrooklynRestResource implements Cat
                 Entitlements.getEntitlementContext().user());
         }
 
-        if (LATEST.equals(version)) {
-            version = null;
-        }
+        version = processVersion(version);
         
         RegisteredType item = mgmt().getTypeRegistry().get(symbolicName, version);
         if (item == null) {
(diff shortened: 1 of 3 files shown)


rest/rest-resources/src/main/java/org/apache/brooklyn/rest/resources/CatalogResource.java
```diff
@@ -331,6 +332,10 @@ public void deleteEntity(String symbolicName, String version) throws Exception {
             throw WebResourceUtils.forbidden("User '%s' is not authorized to modify catalog",
                 Entitlements.getEntitlementContext().user());
         }
+
+        if (LATEST.equals(version)) {
```
m4rkmckenna: `version.toLowercase()`

tbouron: Good catch. Addressed in https://github.com/apache/brooklyn-server/pull/635/commits/361a0fd47377d012d8b484dc925b4da3695887d3

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/resources/CatalogResource.java
```diff
@@ -481,6 +498,10 @@ public CatalogPolicySummary getPolicy(String policyId, String version) throws Ex
                 Entitlements.getEntitlementContext().user());
         }
 
+        if (LATEST.equals(version)) {
+            version = null;
```
tbouron: I replaced all `if` statement, see the latest commit

PR #636, opened 2017-04-14
Enable older Java versions

drigodwin:
Versions of Java prior to 1.8 are not available by standard on Ubuntu 16.x+, this adds the repo for older versions required by some software.

### DIFF
--- a/utils/common/src/main/java/org/apache/brooklyn/util/ssh/BashCommands.java
+++ b/utils/common/src/main/java/org/apache/brooklyn/util/ssh/BashCommands.java
@@ -621,6 +621,7 @@ public class BashCommands {
     public static String installJava(int version) {
         Preconditions.checkArgument(version == 6 || version == 7 || version == 8, "Supported Java versions are 6, 7, or 8");
         List<String> commands = new LinkedList<String>();
+        commands.add(ok(addOpenJDKPPK()));
         commands.add(installPackageOr(MutableMap.of("apt", "openjdk-" + version + "-jdk","yum", "java-1." + version + ".0-openjdk-devel"), null,
                 ifExecutableElse1("zypper", chainGroup(
                         ok(sudo("zypper --non-interactive addrepo http://download.opensuse.org/repositories/Java:/openjdk6:/Factory/SLE_11_SP3 java_sles_11")),
@@ -668,6 +669,15 @@ public class BashCommands {
 
     public static String installJavaLatestOrWarn() {
         return alternatives(installJava8(), installJava7(), installJava6(), warn("java latest install failed, entity may subsequently fail"));
+    }
+
+    /**
+     * Adds the PPA for OpenJDK for older JDK versions (7 and lower) required by some software (e.g. JBoss)
+     */
+    public static String addOpenJDKPPK(){
+        return chainGroup(
+                sudo("sudo add-apt-repository -y ppa:openjdk-r/ppa"),
+                sudo("sudo apt-get update"));
     }
 
     /**


utils/common/src/main/java/org/apache/brooklyn/util/ssh/BashCommands.java
```diff
@@ -621,6 +621,7 @@ private static Object getFlag(Map<?,?> flags, String flagName, Object defaultVal
     public static String installJava(int version) {
         Preconditions.checkArgument(version == 6 || version == 7 || version == 8, "Supported Java versions are 6, 7, or 8");
         List<String> commands = new LinkedList<String>();
+        commands.add(ok(addOpenJDKPPK()));
```
tbouron: Shouldn't you wrap this in a `ifExecutableElse1` instead of `ok`?

duncangrant: Should this only happen if the version is 6 or 7?

utils/common/src/main/java/org/apache/brooklyn/util/ssh/BashCommands.java
```diff
@@ -671,6 +672,15 @@ public static String installJavaLatestOrWarn() {
     }
 
     /**
+     * Adds the PPA for OpenJDK for older JDK versions (7 and lower) required by some software (e.g. JBoss)
+     */
+    public static String addOpenJDKPPK(){
```
tbouron: Not sure that is the right was of doing this. I'm wondering if it would be better to handle this case by case, i.e. in the blueprints themselves

utils/common/src/main/java/org/apache/brooklyn/util/ssh/BashCommands.java
duncangrant: sorry possibly I mean the other way round so you can ignore me

drigodwin: Does it matter if the repo is enabled for the other versions too?

utils/common/src/main/java/org/apache/brooklyn/util/ssh/BashCommands.java
drigodwin: I would agree as will be the case in future YAML based blueprints, but this is the pattern used elsewhere, see JMX or NSS

utils/common/src/main/java/org/apache/brooklyn/util/ssh/BashCommands.java
drigodwin: OK ignores if the command fails (so if the OS isn't Ubuntu for instance) so this should be fine. It's the pattern used elsewhere for similar things - see the [lines below](https://github.com/drigodwin/brooklyn-server/blob/b6c8566bbbc4766dd5e332a9c0db7f992f5f9ed4/utils/common/src/main/java/org/apache/brooklyn/util/ssh/BashCommands.java#L633).

Graeme-Miller: I can't see any harm in it

tbouron: Fair enough

utils/common/src/main/java/org/apache/brooklyn/util/ssh/BashCommands.java
tbouron: Fair enough too, let's go for that now but I would like to see a better pattern implemented in the future (if possible)

PR #637, opened 2017-04-14
Remove deprecated REST API endpoints

tbouron:
This removes the deprecated REST API endpoints from Brooklyn 0.7.0. It also fixes the launcher configuration (as the `VersionApi` has been entirely removed) and unit tests.

Please note that I didn't touch the `LocationApi` for a very specific reason: currently, this is the only of getting the locations defined within `brooklyn.properties`/`brooklyn.cfg` files. We need to figure out first how to handle those locations before removing the endpoints. As you can see, this is not in the scope of this PR.


tbouron:
BTW, I also checked the JSGUI and it does not use any of the deprecated call so you are good on this part

geomacy:
👍  lgtm; have also tested a variety of `br` command line operations against this without problems. Will merge.

tbouron:
Awesome, thanks @geomacy! Could you also take a look at https://github.com/apache/brooklyn-client/issues/45 please? That will avoid the build to fail

geomacy:
@tbouron that's done thanks

PR #638, opened 2017-04-14
Add support for managing enrichers within the catalog

tbouron:
This allows enrichers to be registered within the Brooklyn catalog, same as entities or policies. It allows enrichers' registration via:
- `@Catalog` annotation on the java class
- YAML with an `itemType: enricher`

This also adds 3 endpoints to the `/v1/catalog` REST API:
- `GET /v1/catalog/enrichers` returns the list of all registered enrichers
- `GET /v1/catalog/enrichers/<enricherId>/<version>` returns the detail of a specific enricher
- `DELETE /v1/catalog/enrichers/<enricherId>/<version>` details a specific enricher

_PS: If the PR is too big, I can split up to multiples PRs. Hopefully the commit's messages are pretty clear_


aledsage:
@tbouron see brooklyn-server/policy/src/main/resources/catalog.bom - I think we'd need to list the enrichers in a .bom file, otherwise we won't list them when running in karaf (because we won't classpath-scan for annotated enrichers across all bundles, I believe).

tbouron:
@aledsage Tried to add enricher definitions in the `/src/main/resources/catalog.bom` of each bundle where the enricher is located, plus in `karaf/init/src/main/resources/catalog-classes.bom` but so far, they are not registered when I start Brooklyn in Karaf mode.

Am I doing something wrong?

tbouron:
Thanks for the review @geomacy! I addressed your comments and rebased on the latest master

geomacy:
Thanks @tbouron, that's great, merging now.

geomacy:
Actually can you squash the commits?

tbouron:
@geomacy done!

geomacy:
Thanks, merging now

PR #639, opened 2017-04-18
BROOKLYN-475: fix machine-details parsing

aledsage:


bostko:
Verified without the fix test fails. I think the fix is sufficient.
LGTM.

aledsage:
Thanks @bostko - merging.

PR #640, opened 2017-04-18
Fix address publish bug

drigodwin:
When a server provisions but JClouds fails to connect to it, the server is killed and re-provisioned. When this happens the private and public address of the old server remain in the `host.address` sensors because of a null check in `DefaultConnectivityResolver`. This PR removes that so the machine correctly obtains the address of the re-provisioned machine.


aledsage:
LGTM.

@sjcorbett any thoughts on the implications of removing this null check? It looks safe to me, but then would it interfere with any downstream projects?

aledsage:
Via separate comms, @sjcorbett said to @drigodwin that ... the null-check was added because he was worried about triggering an event upon rebind. But probably being too cautious and can probably remove it. Nothing in downstream projects is relying on it at the moment, as far as we know.

So I think this is good to merge.

Merging now.

sjcorbett:
+1 to the change. I was probably being overly cautious about not triggering duplicate sensor change events (since the block will run during rebind). I'm not aware of any downstream projects reliant on this.

PR #643, opened 2017-04-20
Don't fail white list loading if invalid bundles present

neykov:
If we add a non-bundle jar through `brooklyn.libraries` subsequent white list loads fail.


aledsage:
LGTM - confirmed the test fails before the fix as well. Merging.

PR #645, opened 2017-04-21
Persist OSGi bundles

ahgittin:
This works to persist bundles and is ready for review.  However it has conflicts and needs to be reconciled with #338 -- both coming soon.


ahgittin:
Merged with other OSGi PRs, so ready for review and merging /cc @geomacy @neykov .

ahgittin:
have done some extra manual testing, including running multiple servers, doing a `br catalog add` (using the luvely https://github.com/apache/brooklyn-client/pull/44 ), and checking that bundles are transferred across nicely.  all good.

as noted in the PR above, one thing i find is that an attempt to re-upload a bundle is silently.  it looks as though it is working but internally the first things `OsgiManager.installUploadedBundle()` does is find a cache hit and ignores the input.  this is fine, good even, if a user is uploading an identical bundle.  however we don't actually check whether the bundle is identical, and if the bundle is changed:

* it should fail if the version is not snapshot
* it should install the new bundle and work towards abandoning the old bundle

it's actually not easy to check whether the bundle is changed so i'm thinking we'll apply the above logic whether or not the bundle has changed.

by "work towards abandoning the old bundle" i mean:

* remove it from persistence
* deregister the old bundle in osgi ... java classes loaded from it may force the bundle to be loaded but inaccessible until a rebind or explicit "upgrade to same version" (when available)
* remove any catalog items from the previous bundle not present in the new bundle

however i think this can all be done in another PR (and probably it will want a record of the originating bundle, at least on the catalog item / registered type, so it will be a bigger PR).

(this PR doesn't support uninstall yet, but we can live with that for a wee while i think.)

--

we should also _not_ persist the catalog items for things coming from persisted bundles, and when a bom is uploaded we should make it in to a bundle, so net net we stop creating catalog items altogether.

--

can i get a review?  cc @geomacy @neykov @aledsage

geomacy:
Will do

geomacy:
@ahgittin have made a start on reviewing this but will have to carry on tomorrow

geomacy:
Got some time on this today, should be able to finish going through it tomorrow.

core/src/main/java/org/apache/brooklyn/core/typereg/BasicManagedBundle.java
```diff
@@ -0,0 +1,175 @@
...
+        this.localFileWhenJustUploaded = localFileWhenJustUploaded;
+    }
+    @Beta
+    public File getTempLocalFileWhenJustUploaded() {
+        return localFileWhenJustUploaded;
+    }
+    
+    @Override
+    public String getOsgiUniqueUrl() {
+        return "brooklyn:"+getId();
```
geomacy: Would it be worth also encoding the name and version in here so that  the URL is more meaningful, at least to a person who should ever happen to read it, even if it's not strictly needed from Karaf's point of view?

core/src/main/java/org/apache/brooklyn/core/typereg/BasicManagedBundle.java
```diff
@@ -0,0 +1,175 @@
...
+        if (!Objects.equal(version, other.getVersion())) return false;
+        if (!Objects.equal(url, other.getUrl())) return false;
+        return true;
+    }
+
+    // ---
+    
+    @Override
+    public String getDisplayName() {
+        return null;
```
geomacy: `return toString()`?

core/src/test/java/org/apache/brooklyn/core/mgmt/persist/ListeningObjectStore.java
```diff
@@ -209,6 +213,20 @@ public void put(String val) {
             delegate.put(val);
         }
         @Override
+        public void put(ByteSource val) {
+            if (writesFailSilently)
+                return;
+
+            for (ObjectStoreTransactionListener listener: listeners) {
+                try {
+                    listener.recordDataOut("writing (binary) "+path, Streams.readFullyAndClose(val.openStream()).length);
```
geomacy: Is it really worth the extra read and close just to be able to write the length here?

locations/jclouds/src/main/java/org/apache/brooklyn/core/mgmt/persist/jclouds/JcloudsStoreObjectAccessor.java
```diff
@@ -57,11 +55,13 @@ public boolean exists() {
     @Override
     public void put(String val) {
         if (val==null) val = "";
```
geomacy: In `FileBasedObjectStoreAccessor` you throw an exception on trying to write a null value, should be consistent here.

utils/common/src/test/java/org/apache/brooklyn/util/osgi/OsgiTestResources.java
```diff
@@ -103,5 +103,14 @@
      * at runtime.
      */
     public static final String BROOKLYN_TEST_MORE_ENTITIES_V2_EVIL_TWIN_PATH = "/brooklyn/osgi/" + BROOKLYN_TEST_MORE_ENTITIES_SYMBOLIC_NAME_FINAL_PART + "_evil-twin_0.2.0.jar";
+
+
+    public static final String BROOKLYN_TEST_OSGI_ENTITIES_URL = "classpath:"+OsgiTestResources.BROOKLYN_TEST_OSGI_ENTITIES_PATH;
```
geomacy: Could remove the `public static final` now.

api/src/main/java/org/apache/brooklyn/api/mgmt/rebind/mementos/BrooklynMementoRawData.java
```diff
@@ -188,14 +211,22 @@ public String getBrooklynVersion() {
...
+    
     // to handle reset catalog
     @Beta
     public void clearCatalogItems() {
         catalogItems.clear();
     }
     
     public boolean isEmpty() {
-        return planeId == null && entities.isEmpty() && locations.isEmpty() && policies.isEmpty() && enrichers.isEmpty() && feeds.isEmpty() && catalogItems.isEmpty();
+        return planeId == null && entities.isEmpty() && locations.isEmpty() && policies.isEmpty() && enrichers.isEmpty() && feeds.isEmpty() && catalogItems.isEmpty() && bundles.isEmpty();
```
ahgittin: would be misleading -- better would be an assertion bundleJars isn't empty if the bundles is, but that's also not worth it i think

core/src/main/java/org/apache/brooklyn/core/mgmt/persist/BrooklynMementoPersisterToObjectStore.java
```diff
@@ -658,6 +701,33 @@ private Stopwatch deltaImpl(Delta delta, PersistenceExceptionHandler exceptionHa
...
+                throw new IllegalStateException("Cannot persist bundles without a mangaement context");
+            }
+            final ManagedBundle mb = ((ManagementContextInternal)mgmt).getOsgiManager().get().getManagedBundles().get(id);
+            if (mb==null) {
+                LOG.warn("Cannot find managed bundle for added bundle "+id+"; ignoring");
+                return;
+            }
+            
+            if (mb instanceof BasicManagedBundle) {
+                final File f = ((BasicManagedBundle)mb).getTempLocalFileWhenJustUploaded();
```
ahgittin: added an extra check to bail out in the executor.  the executor is single-threaded so that is fine.  (just synching wouldn't guarantee it doesn't submit multiple times i don't htink.)

core/src/main/java/org/apache/brooklyn/core/mgmt/persist/BrooklynPersistenceUtils.java
```diff
@@ -171,21 +173,31 @@ private static BrooklynMementoRawData newStateMementoFromLocal(ManagementContext
...
+            }
         }
-        for (CatalogItem<?,?> instance: mgmt.getCatalog().getCatalogItems())
+        for (CatalogItem<?,?> instance: mgmt.getCatalog().getCatalogItems()) {
             result.catalogItem(instance.getId(), serializer.toString(newObjectMemento(instance)));
+        }
+        OsgiManager osgi = ((LocalManagementContext)mgmt).getOsgiManager().orNull();
+        if (osgi!=null) {
+            for (ManagedBundle instance: osgi.getManagedBundles().values()) {
+                result.catalogItem(instance.getId(), serializer.toString(newObjectMemento(instance)));
```
ahgittin: good spot

core/src/main/java/org/apache/brooklyn/core/mgmt/persist/StoreObjectAccessorLocking.java
```diff
@@ -146,6 +148,28 @@ public void put(String val) {
     }
     
     @Override
+    public void put(ByteSource val) {
```
ahgittin: i like the "rule of 3" then refactor

core/src/main/java/org/apache/brooklyn/core/mgmt/rebind/RebindContextImpl.java
```diff
@@ -85,6 +88,14 @@ public void registerFeed(String id, Feed feed) {
     public void registerCatalogItem(String id, CatalogItem<?, ?> catalogItem) {
         catalogItems.put(id, catalogItem);
     }
+
+    public void registerBundle(String id, ManagedBundle bundle) {
```
ahgittin: another good spot, removed.  and `unregister`.

core/src/main/java/org/apache/brooklyn/core/typereg/BasicManagedBundle.java
ahgittin: it's only ever seen in karaf i think

core/src/main/java/org/apache/brooklyn/core/typereg/BasicManagedBundle.java
ahgittin: no, display name is null and caller should find a sensible fallback to display, that's the convention elsewhere i think

core/src/test/java/org/apache/brooklyn/core/mgmt/persist/ListeningObjectStore.java
ahgittin: this is test code -- yes absolutely we want this to assert lengths

locations/jclouds/src/main/java/org/apache/brooklyn/core/mgmt/persist/jclouds/JcloudsStoreObjectAccessor.java
ahgittin: restoring previous behaviour which replaces null with `""`

utils/common/src/test/java/org/apache/brooklyn/util/osgi/OsgiTestResources.java
ahgittin: personal preference to keep it in case you don't notice it's an interface; don't feel strongly but not going to rewrite this class in this PR

ahgittin:
@geomacy thanks -- all updated.  let's see what automated tests say, and how your tests go.  thx for the review.

geomacy:
Still need to test this in HA mode but have tried the test below to verify that resources included in the zip can be used and survive a rebind.

The script here creates a zip with a couple of boms and an icon, then deploys an app using the  catalog.bom entity, which is defined in terms of an entity from testent.bom.  The app deploys ok and the icon appears on the UI. However, upon Karaf shutdown and restart, although the app rebinds ok, its icon has disappeared, so it's not being fetched from the persisted bundle.  

```
cat > testent.bom <<EOF
brooklyn.catalog:
  version: 0.12.0.SNAPSHOT
  items:
  - id: myTestEnt
    item:
      type: org.apache.brooklyn.entity.stock.BasicEntity

  - id: myOtherTest
    item:
      type: org.apache.brooklyn.entity.stock.BasicEntity
EOF

cat > catalog.bom <<EOF
brooklyn.catalog:
  bundle: io.cloudsoft.cli44
  version: 0.12.0.SNAPSHOT
  iconUrl: classpath://io.cloudsoft.cli44:plane.png

  items:
  - classpath://testent.bom
  - id: testent1
    item:
      type: myTestEnt
EOF

cat > test.yaml <<EOF
location: localhost
services:
- type: testent1
  id: testent1
EOF

wget https://upload.wikimedia.org/wikipedia/commons/thumb/7/7d/Plane_icon.svg/200px-Plane_icon.svg.png -O plane.png

zip test.zip *.bom *.png

br catalog add test.zip

br deploy test.yaml

```

Can you check the above to verify that this is a valid test? Shouldn't the icon resource be read from the persisted bundle upon rebind?

ahgittin:
@geomacy Great test.  I've been trying it and I find one good thing and several problems.

The good:  deployment, persistence, versioning, etc all work smoothly in this fairly complex example, with the exception of icons.

The bad:
* the OSGi syntax `classpath://io.cloudsoft.cli44:plane.png` does not seem to work, neither before nor after rebind -- not sure why
* local bundle syntax `classpath://plane.png` (which should be preferred here) also does not work, not even before rebind -- easy to see that the `getIconUrl()` isn't using the entity's classpath, and _I have a confirmed fix for this_
* in above case, icon is not available after rebind -- it isn't being persisted
* if there is no icon on the entity instance possibly it should defer to the catalog item (not sure) which would allow it to work even if not persisted -- need to look more at this

I suggest we handle this as a bug in icons as I suspect these were mostly if not all true prior to this PR.

Merge?  I'll see if I can get an icon PR out today.

geomacy:
hi @ahgittin the `classpath://io.cloudsoft.cli44:plane.png` seems to work for me
screen shot 2017-04-28 at 11 01 21

but I agree that this is something that can be treated separately.

Think this is good to merge and handle any issues subsequently.

Graeme-Miller:
I have briefly reviewed the code and I am in the process of testing it.
If you guys are happy to merge go ahead- I'll keep on with testing just in case

ahgittin:
@geomacy curious about the osgi url working for you.  are you running karaf or `o.a.brooklyn.Main`?

i've found the relevant code for all the other behaviours and i don't think they'll be hard to fix, and i'm sure i can find what's going on with the osgi url.  (@neykov if you know off the top of your head where the osgi syntax is handled that would be a helpful pointer but it won't be too hard to find.)

geomacy:
@ahgittin I was running in Karaf

geomacy:
@Graeme-Miller agreed, will merge now.

PR #604, opened 2017-03-22
WIP DO NOT MERGE Fixes to testManuallyAdd... 

geomacy:
Make initial changes as suggested by Alex in #573.
To be merged after #338.  Only the last (a5cdc19) of the changes in this PR is actually not part of 338.

Add addToCatalogSearchPathOnAddition and make it set search paths.
Update management context classloader provider
Add getCatalogItemIdAndSearchPathFromContext.
Set context details as search path in AbstractBrooklynObject constructor.
Update testManuallyAddWithParentFromCatalog assertion as suggested.
Update testManuallyAddInTaskOfOtherEntity assertion as suggested.

The changes above allow the tests just named to succeed, however
they have knock-on effects in other tests, as they change the
behaviour of DSL "scopeRoot", which effectively relies on
the catalogItemId of children being copied from that of their
parents. This means the following tests fail:

Failed tests:
  EntitiesYamlTest.testScopeReferences:569->assertScopes:581 expected [ReferencingYamlTestEntityImpl{id=f3z5pyzt8r}] but found [ReferencingYamlTestEntityImpl{id=witrlgcuo1}]
  EntityRefsYamlTest.testRefToScopeRoot:101 expected [TestEntityImpl{id=fc9nfhoj8x}] but found [TestEntityImpl{id=g2ai718knw}]
  DslYamlTest.testDslScopeRoot:253->assertScopeRoot:261 expected [BasicEntityImpl{id=a860h6w3at}] but found [BasicEntityImpl{id=aoxrtylbv2}]

PR #646, opened 2017-04-21
Cleanup search paths

ahgittin:
minor cleanups to #338 -- see the last commit

this seems to fix the test failures by deduping the path, and some other very minor things.  opened PR mainly to confirm it's okay on jenkins also.  /cc @geomacy

merging this should also close #338.

### DIFF
--- a/camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/catalog/CatalogYamlEntityTest.java
+++ b/camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/catalog/CatalogYamlEntityTest.java
@@ -40,15 +40,26 @@ import org.apache.brooklyn.core.typereg.RegisteredTypes;
 import org.apache.brooklyn.entity.stock.BasicApplication;
 import org.apache.brooklyn.entity.stock.BasicEntity;
 import org.apache.brooklyn.util.exceptions.Exceptions;
+import org.apache.brooklyn.util.osgi.OsgiTestResources;
 import org.testng.Assert;
 import org.testng.annotations.Test;
 
 import com.google.common.base.Joiner;
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Iterables;
 
 
 public class CatalogYamlEntityTest extends AbstractYamlTest {
     
+    private static final String SIMPLE_ENTITY_TYPE = OsgiTestResources.BROOKLYN_TEST_OSGI_ENTITIES_SIMPLE_ENTITY;
+    private static final String MORE_ENTITIES_POM_PROPERTIES_PATH =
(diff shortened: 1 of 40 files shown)


ahgittin:
build failure because `ubuntu-us1` offline -- forcing recheck

camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/catalog/CatalogYamlEntityTest.java
```diff
@@ -50,15 +48,6 @@
...
 
 public class CatalogYamlEntityTest extends AbstractYamlTest {
-    
-    private static final String SIMPLE_ENTITY_TYPE = OsgiTestResources.BROOKLYN_TEST_OSGI_ENTITIES_SIMPLE_ENTITY;
-    private static final String MORE_ENTITIES_POM_PROPERTIES_PATH =
-        "META-INF/maven/org.apache.brooklyn.test.resources.osgi/brooklyn-test-osgi-more-entities/pom.properties";
-
-    @Override
-    protected boolean disableOsgi() {
-        return false;
```
ahgittin: @geomacy This appears to be unnecessary, and it slows down testing dramatically, and causes a lot more GC messages. It might be the reason for the testing failures in #338 (possibly compounded by not de-duping the search paths, or possibly not).

cc @aledsage @neykov As we do more with OSGi we should be aware that starting tests that use OSGi each take almost 1s.  Also we're using the "fake" OSGi I think (felix not karaf); maybe it makes sense for some modules to launch in OSGi when they run their tests, and/or maybe there are some other optimizations?

ahgittin:
@geomacy Worth your casting an eye over my changes just to be sure but they're tiny compared to your good work in #338 so merging this now to have them all.  Mainly the deduping search path is what I've added, plus using some minor utility methods I think are cleaner.  BTW I really like the `setCatalogItemIdAndSearchPath` -- it wasn't obvious how right that was until I reviewed it more, but good shout there.

geomacy:
Thanks @ahgittin will certainly look at your changes, thanks for that!  

Maybe soon we can make some time to discuss the changes needed as you suggested in #573 for ConfigParametersYamlTest. testManuallyAddWithParentFromCatalog and friends.  

I made a start on this in #604 but found that the changes altered the behaviour of DSL's 'scopeRoot', which effectively relies on the catalogItemId of children being copied from that of their parents, as noted here: https://github.com/apache/brooklyn-server/pull/604/commits/a5cdc192816244e30e6b2b2163fd4160c9f3b252

geomacy:
Those changes all look good @ahgittin, especially the update to set the search path in the `AbstractBrooklynObject` constructor.  Good point also about the synchronization of search path modifications!

PR #647, opened 2017-04-24
better logging of REST exceptions

ahgittin:
include the trace on the first encounter of a new type or simply a new place where a type is thrown from;
subsequent instances won't get the trace but handy when debugging if the first one does

### DIFF
--- a/rest/rest-resources/src/main/java/org/apache/brooklyn/rest/util/DefaultExceptionMapper.java
+++ b/rest/rest-resources/src/main/java/org/apache/brooklyn/rest/util/DefaultExceptionMapper.java
@@ -39,12 +39,15 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.yaml.snakeyaml.error.YAMLException;
 
+import com.google.common.annotations.Beta;
+
 @Provider
 public class DefaultExceptionMapper implements ExceptionMapper<Throwable> {
 
     private static final Logger LOG = LoggerFactory.getLogger(DefaultExceptionMapper.class);
 
-    static Set<Class<?>> warnedUnknownExceptions = MutableSet.of();
+    static Set<Class<?>> encounteredUnknownExceptions = MutableSet.of();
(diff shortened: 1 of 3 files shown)


rest/rest-resources/src/main/java/org/apache/brooklyn/rest/util/DefaultExceptionMapper.java
```diff
@@ -120,6 +121,32 @@ public Response toResponse(Throwable throwable1) {
...
+                LOG.debug("Full details of "+Entitlements.getEntitlementContext()+" "+t+" (logging debug on first encounter; subsequent instances will be logged trace)", t);
+            } else if (LOG.isTraceEnabled()) {
+                LOG.trace("Full details of "+Entitlements.getEntitlementContext()+" "+t, t);
+            }
+        }
+    }
+        
+    private static boolean firstEncounter(Throwable t) {
+        Set<Object> record = MutableSet.of();
+        while (true) {
```
geomacy: `while (true)` always looks alarming; in this case it may well be fine, as `t = t.getCause()` won't last forever. But why not simply write `while (t != null)` and avoid the `if (t==null) break;`?

geomacy:
This is fine as-is, will merge.

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/util/DefaultExceptionMapper.java
ahgittin: there is the additional `break` when we find a stack trace element we've seen before.

i've been caught out by perverse `Throwable` instances with cyclic causal chains which that catches.

i don't like using `break` but there are times it is simpler.  in those cases i prefer not to mix it with `while (interestingCondition)` syntax to make it clear that `break` semantics are being used.

geomacy: Interesting, that's a nice guideline, now that you mention it.

PR #648, opened 2017-04-25
Changes Application API behaviour so that returned location id is the catalog item id 

m4rkmckenna:
Updates ApplicationTransformer to return the catalogItemId (if available)
* If unavailable ID is returned.

**NOTE** This changes the behaviour of the Application API

### DIFF
--- a/rest/rest-resources/src/main/java/org/apache/brooklyn/rest/transform/ApplicationTransformer.java
+++ b/rest/rest-resources/src/main/java/org/apache/brooklyn/rest/transform/ApplicationTransformer.java
@@ -100,7 +100,7 @@ public class ApplicationTransformer {
             @Override
             @Nullable
             public String apply(@Nullable Location input) {
-                return input.getId();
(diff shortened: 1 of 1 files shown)


ahgittin:
As I wrote on mailing list:

```
I'd like to keep backwards compatibility.  We don't know where this might be used.

Also the catalog item ID isn't sufficient I don't think.  Returning just the catalog item ID means you lose any custom configuration on the location in the blueprint (ie on top of the catalog item ID); I'm not sure it even works at all if you use legacy "jclouds:aws-ec2" syntax; and you won't get any added-after-deployed information.  It isn't a problem in many cases but it is a weaker API.

Could we add a new "locationsDetail" field which contains eg "{ [ id: <old value>, catalogItemId: <what you want>, displayName: <optional>, ... ] }" and we could add any further metadata we want?

I'd also be in favour of doing this for all entities, not just applications.
```

m4rkmckenna:
@ahgittin Agree with the above ... closing now ill try and get a better solution

PR #649, opened 2017-04-25
Reuse embedded OSGi container in tests

ahgittin:
This speeds up build time from 17m to 11m on my machine by reusing OSGi framework containers where possible.  That also fixes an OSGi container which is causing lots of build failures on Jenkins.

See the last commit which is the only thing new, and the `OsgiManager` class is the main significant difference in that.  Other things are just wiring to control bundle reuse or not.

This includes #645 and #647 so if this build passes we can assume the failures in those PR's are ignorable (they seem to be due to memory leaks).  But merge those before merging this (and I can rebase this if needed once they are merged.)

/cc @neykov @aledsage @geomacy


geomacy:
This all looks good to me; tests all pass, with speedup (just for brooklyn-server build) on my machine from 12:20 min to 09:05 min.

Will merge.

PR #651, opened 2017-04-25
Test speedups

ahgittin:
Lots of little things to make tests run faster.

Builds on #649 (so we get decent test results) so it's only the last several commits which are relevant.  Will rebase once that is merged.

### DIFF
--- a/locations/jclouds/src/test/java/org/apache/brooklyn/location/jclouds/DefaultConnectivityResolverTest.java
+++ b/locations/jclouds/src/test/java/org/apache/brooklyn/location/jclouds/DefaultConnectivityResolverTest.java
@@ -68,7 +68,7 @@ public class DefaultConnectivityResolverTest extends AbstractJcloudsStubbedUnitT
                 JcloudsLocationConfig.LOOKUP_AWS_HOSTNAME, true));
         ConnectivityResolverOptions options = newResolveOptions()
                 .waitForConnectable(true)
-                .pollForReachableAddresses(Predicates.<HostAndPort>alwaysTrue(), Duration.ONE_SECOND, true)
+                .pollForReachableAddresses(Predicates.<HostAndPort>alwaysTrue(), Duration.millis(1), true)
                 .userCredentials(credential)
                 .build();
         DefaultConnectivityResolver customizer = new DefaultConnectivityResolver();
@@ -92,7 +92,7 @@ public class DefaultConnectivityResolverTest extends AbstractJcloudsStubbedUnitT
         initNodeCreatorAndJcloudsLocation(newNodeCreator(), ImmutableMap.of());
         DefaultConnectivityResolver customizer = new DefaultConnectivityResolver();
         final ConfigBag config = ConfigBag.newInstance(ImmutableMap.of(
-                JcloudsLocationConfig.WAIT_FOR_SSHABLE, "1ms"));
+                JcloudsLocationConfig.POLL_FOR_FIRST_REACHABLE_ADDRESS, "1ms"));
         assertTrue(customizer.checkCredential(
                 jcloudsLocation, HostAndPort.fromParts("10.0.0.234", 22),
                 LoginCredentials.builder().user(allowedUser).password("password1").build(), config, false));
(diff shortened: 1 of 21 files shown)


ahgittin:
Build time was 17m.  #649 got it down to 11m.  This gets it down to 9m.  Die slow tests!!!

aledsage:
Please mark https://issues.apache.org/jira/browse/BROOKLYN-241 as fixed once this is merged.

ahgittin:
Dependent PRs have all been merged so this can be evaluated and merged afterwards.

geomacy:
These all look good; test speed up for build of brooklyn-server for me is from 09:05 min to 07:41 min. Will merge.

locations/jclouds/src/test/java/org/apache/brooklyn/location/jclouds/DefaultConnectivityResolverTest.java
```diff
@@ -147,10 +149,9 @@ public void testResolveChecksCredentials() throws Exception {
         ConfigBag config = ConfigBag.newInstance(ImmutableMap.of(
                 JcloudsLocationConfig.LOOKUP_AWS_HOSTNAME, false,
                 JcloudsLocationConfig.WAIT_FOR_SSHABLE, "1ms",
+                JcloudsLocationConfig.POLL_FOR_FIRST_REACHABLE_ADDRESS, "1ms",
                 JcloudsLocation.CUSTOM_CREDENTIALS, credential));
-        ConnectivityResolverOptions options = newResolveOptions()
-                .pollForReachableAddresses(Predicates.in(reachableIps), Duration.ONE_SECOND, true)
```
sjcorbett: The reason the timeouts were so large was to ensure the tests passed on Jenkins (something [other tests do the same thing for](https://github.com/apache/brooklyn-server/pull/663#discussion_r114291567)). With a one millisecond timeout I think these tests will sporadically fail. Here's one that affected PR 668:
```
java.lang.AssertionError: expected [10.0.0.2:22] but found [10.0.0.1:22]
	at org.apache.brooklyn.location.jclouds.DefaultConnectivityResolverTest.testResolveChecksCredentials(DefaultConnectivityResolverTest.java:161)
Standard Output

2017-05-04 19:54:38,459 INFO  TESTNG PASSED: "Surefire test" - org.apache.brooklyn.location.jclouds.DefaultConnectivityResolverTest.testObtainsHostnameFromAwsMachine() finished in 52 ms
2017-05-04 19:54:38,460 INFO  TESTNG INVOKING CONFIGURATION: "Surefire test" - @AfterMethod org.apache.brooklyn.location.jclouds.AbstractJcloudsStubbedUnitTest.tearDown()
2017-05-04 19:54:38,462 INFO  TESTNG PASSED CONFIGURATION: "Surefire test" - @AfterMethod org.apache.brooklyn.location.jclouds.AbstractJcloudsStubbedUnitTest.tearDown() finished in 2 ms
2017-05-04 19:54:38,463 INFO  TESTNG INVOKING CONFIGURATION: "Surefire test" - @BeforeMethod org.apache.brooklyn.location.jclouds.AbstractJcloudsStubbedUnitTest.setUp()
2017-05-04 19:54:38,464 INFO  TESTNG PASSED CONFIGURATION: "Surefire test" - @BeforeMethod org.apache.brooklyn.location.jclouds.AbstractJcloudsStubbedUnitTest.setUp() finished in 2 ms
2017-05-04 19:54:38,465 INFO  TESTNG INVOKING: "Surefire test" - org.apache.brooklyn.location.jclouds.DefaultConnectivityResolverTest.testResolveChecksCredentials()
2017-05-04 19:54:38,474 INFO  No Camp-YAML parser registered for parsing catalog item DSL; skipping DSL-parsing
2017-05-04 19:54:38,480 WARN  No context entity found in config or current task
```

ahgittin: @sjcorbett in these unit tests behaviour shouldn't depend on whether addresses are reachable or not, no?

sjcorbett: No, and they don't. The timeout is used in a handful of places. In the case of the failure above I reckon the flow was:

* `DefaultConnectivityResolver` favours address and credential pairs that can be used for SSH/WinRM.†
* It tests connectivity with `JcloudsLocation.waitForReachable` via `waitForSshable`.
* `waitForReachable` uses a `Repeater` whose timeout is configured by this test to 1ms.
* The Repeater calls `future.get(1ms)`. The task does not complete in time on slow Jenkins so a `TimeoutException` is thrown. (For reference, the Repeater's task uses `RecordingSshTool` which uses the response generator defined in the test case.)
* DefaultConnectivityResolver falls back to the first available address.
* The test fails.

I see https://github.com/apache/brooklyn-server/commit/f9cb0b2447b02d93fe2b39496320cb1c920be3ff changed Repeater to respect timeouts and imagine the tighter constraints have forced the failure. Before that commit the repeater would have waited the full initial delay, which in this case is configured by `waitForReachable` to be one second.

† e.g. if jclouds gives it `addresses=[10.0.0.1, 10.0.0.2]` and `credentials=[key1, key2]` and the instance is only accessible on `key2@10.0.0.2` it will do its best to find that out. It checks addresses in order so in this example would have tested and rejected the other three combinations of address and key.

PR #654, opened 2017-04-26
Fix BROOKLYN-493 Rebind fails when using WinRmCommandSensor

bostko:
Tested with restart with Aled's blueprint.


neykov:
LGTM, merging.

PR #657, opened 2017-04-27
BROOKLYN-334: ascii art banner says “Apache Brooklyn”

aledsage:
See https://issues.apache.org/jira/browse/BROOKLYN-334

Switching to:
```
    _                     _            ____                  _    _             
   / \   _ __   __ _  ___| |__   ___  | __ ) _ __ ___   ___ | | _| |_   _ _ __ (R)
  / _ \ | '_ \ / _` |/ __| '_ \ / _ \ |  _ \| '__/ _ \ / _ \| |/ / | | | | '_ \ 
 / ___ \| |_) | (_| | (__| | | |  __/ | |_) | | | (_) | (_) |   <| | |_| | | | |
/_/   \_\ .__/ \__,_|\___|_| |_|\___| |____/|_|  \___/ \___/|_|\_\_|\__, |_| |_|
        |_|                                                         |___/      0.11.0
```

That is unfortunately 85 characters long (including the 0.11.0 subscript), which is more than the standard 80 characters (http://stackoverflow.com/questions/4651012/why-is-the-default-terminal-width-80-characters).

Alternatively, we could use a smaller format such as:
```
    _                 _          ___              _   _           
   /_\  _ __  __ _ __| |_  ___  | _ )_ _ ___  ___| |_| |_  _ _ _  
  / _ \| '_ \/ _` / _| ' \/ -_) | _ \ '_/ _ \/ _ \ / / | || | ' \ 
 /_/ \_\ .__/\__,_\__|_||_\___| |___/_| \___/\___/_\_\_|\_, |_||_|
       |_|                                              |__/      
```

See http://patorjk.com/software/taag/#p=display&f=Standard&t=Apache%20Brooklyn and http://patorjk.com/software/taag/#p=display&f=Small&t=Apache%20Brooklyn

Personally, I like the upper case for A and B.

@rdowner what do you think?

### DIFF
--- a/server-cli/src/main/java/org/apache/brooklyn/cli/AbstractMain.java
+++ b/server-cli/src/main/java/org/apache/brooklyn/cli/AbstractMain.java
@@ -72,14 +72,14 @@ public abstract class AbstractMain {
     private static final Logger log = LoggerFactory.getLogger(AbstractMain.class);
 
     // Launch banner
+    // http://patorjk.com/software/taag/#p=display&f=Standard&t=Apache%20Brooklyn
     public static final String DEFAULT_BANNER =
-        " _                     _    _             \n" +
-        "| |__  _ __ ___   ___ | | _| |_   _ _ __ (R)\n" +
-        "| '_ \\| '__/ _ \\ / _ \\| |/ / | | | | '_ \\ \n" +
-        "| |_) | | | (_) | (_) |   <| | |_| | | | |\n" +
-        "|_.__/|_|  \\___/ \\___/|_|\\_\\_|\\__, |_| |_|\n" +
-        "                              |___/             "+BrooklynVersion.get()+"\n";
-
+        "    _                     _            ____                  _    _             \n" +             
+        "   / \\   _ __   __ _  ___| |__   ___  | __ ) _ __ ___   ___ | | _| |_   _ _ __ (R)\n" +
+        "  / _ \\ | '_ \\ / _` |/ __| '_ \\ / _ \\ |  _ \\| '__/ _ \\ / _ \\| |/ / | | | | '_ \\ \n" +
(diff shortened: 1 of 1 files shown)


m4rkmckenna:
@aledsage I think we took our queue from Apache Karaf who only use [karaf ](https://github.com/apache/karaf/blob/master/shell/console/src/main/filtered-resources/org/apache/karaf/shell/console/branding.properties)

drigodwin:
We should not have an (R) as it's not a fully registered trademark.

ahgittin:
i prefer it as was; so long as we say `Apache Brooklyn` underneath we should be okay?

what you've got here is okay.  slight preference not to exceed 80 chars, but strong preference to avoid the compressed variant here which is hideous!  :)

drigodwin:
I'm not sure we should use the (R) symbol as "Apache Brooklyn" is not a registered trademark anywhere.

rdowner:
This is at risk of being a bike-shed discussion. Let's keep it simple.

* The ASCII art logo is not important. If it is causing debate or trouble, then just drop completely. 
* The name of our product is Apache Brooklyn, so that must appear somewhere.
* @drigodwin is correct to say we should not be using the `(R)` designation.
* It is acceptable to say "Brooklyn" sometimes, but only if the full "Apache Brooklyn" is mentioned nearby.
* Just because another Apache project have done "X", doesn't mean they are right and we can copy them. During incubation we were bitten by this several times.

As long as our product clearly identifies itself as "Apache Brooklyn" then I don't have an opinion on aesthetics.

aledsage:
@rdowner Closing this, replaced by https://github.com/apache/brooklyn-server/pull/670 (which ensures "Apache Brooklyn" is written out, and leaves the same banner but with "(R)" deleted.

PR #661, opened 2017-04-28
fix icon loading

ahgittin:
fixes all issues around icons revealed by @geomacy's testing in #645 

last commit is worth looking at in detail, pretty sure it's reasonable but appreciate feedback.  see the comment there.

also note https://github.com/ahgittin/brooklyn-server/commit/b4ecbbb7a751cf2a69155268ab7828cab94e8da8 -- this seems eminently sensible and there may be other places we want to do it, but this one was caught by a test that failed after the fix in the previous commit

cc @neykov @aledsage @geomacy


ahgittin:
unrelated failure https://builds.apache.org/job/brooklyn-server-pull-requests/2078/console --
 fixed in https://github.com/apache/brooklyn-server/pull/663

PR #663, opened 2017-05-02
expand test of future value

ahgittin:
fix logic where future no longer guaranteed to be available when non-blocking

### DIFF
--- a/core/src/test/java/org/apache/brooklyn/core/config/DeferredConfigTest.java
+++ b/core/src/test/java/org/apache/brooklyn/core/config/DeferredConfigTest.java
@@ -19,7 +19,6 @@
 package org.apache.brooklyn.core.config;
 
 import static org.testng.Assert.assertEquals;
-import static org.testng.Assert.assertTrue;
 
 import java.util.List;
 import java.util.concurrent.Callable;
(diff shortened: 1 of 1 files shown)


ahgittin:
tiny change, but fixes failures -- as NonBlocking no longer blocks on futures the test needed a slight change, and felt reasonable also to cover the "not available" case

core/src/test/java/org/apache/brooklyn/core/config/DeferredConfigTest.java
```diff
@@ -30,32 +29,51 @@
...
         Task<Sensor<?>> sensorFuture = app.getExecutionContext().submit(new Callable<Sensor<?>>() {
             @Override
             public Sensor<?> call() throws Exception {
+                if (delay) Time.sleep(Duration.FIVE_SECONDS);
                 return TestApplication.MY_ATTRIBUTE;
             }
         });
         app.config().set(SENSORS_UNTYPED, (Object)ImmutableList.of(sensorFuture));
 
+        if (!delay) sensorFuture.get(Duration.ONE_SECOND);
```
geomacy: Why so large a timeout? Would a millisecond or two not be sufficient? This might be a bit confusing for readers as it suggests that there is some need to wait for a long time, even though the test in the `!delay` case is for a case where the future will execute and complete very quickly.

ahgittin: it shouldn't take more than a millisecond on most machines but on a slow machine the 1s gives it plenty of time

PR #665, opened 2017-05-03
Delete catalog.xml support

aledsage:
Use of `~/.brooklyn/catalog.xml` was deprecated in 0.7.0. This PR finally deletes the code for it.

Once this is reviewed/merged, please also see https://github.com/apache/brooklyn-client/pull/50.

Note that it deletes the rest api endpoint `/v1/catalog/reset` (which relied on there being a `catalog.xml`).

It also changes the way `fragment` is handled in things like `/v1/catalog/applications?fragment=...`. Previously it was use the `toXml`, and then do a case-insensitive search for the fragment. To try to keep the functionality as close as possible (with code as simple as possible), I've reused the persisted state serialization mechanism to produce the memento, and then do the case-insensitive search for the fragment in that.


drigodwin:
The error in jenkins looks like some system problem:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:2.19.1:test (default-test) on project brooklyn-core: Execution default-test of goal org.apache.maven.plugins:maven-surefire-plugin:2.19.1:test failed: The forked VM terminated without properly saying goodbye. VM crash or System.exit called?
```
Could you re-push to run the tests again please @aledsage?

drigodwin:
Again the error in the jenkins appears to be unrelated to this, tests pass locally and Infra seem to have fixed the problem now. This PR looks good to me so I'm going to merge it now, thanks @aledsage!

PR #669, opened 2017-05-05
Delete hazelcast datagrid (and delete datagrid config)

aledsage:
This has not been documented, used in production, or recommended ever! It was more experimental, as an approach that might have replaced the existing persisted state approach but never did.

We can also delete the use of `BrooklynStorage`, but we can do that incrementally (i.e. in a subsequent PR) - that will be more intrusive (changing `AbstractEntity` etc). By first deleting/hiding the `Datagrid` code, it makes the next stage much simpler as we can be much more confident about what `BrooklynStorage` is doing.


andreaturli:
I agree with you, this should be removed as not used +1

fwiw, I think a datagrid was a good idea especially if we want to support "brooklyn of brooklyns" kind of approach, although maybe with the shift to karaf there is a better way.
> It was more experimental, as an approach that might have replaced the existing persisted state approach but never did.

I tried that as well but never finished the poc so either we need some experts or sounds unlikely to have it done correctly. We can always re-introduce datagrid in case needed.

drigodwin:
This looks good and agree about removal. The failure is an unrelated problem with the build server. I've run the tests manually and everything seems fine so I'm going to merge now. Thanks @aledsage.

PR #671, opened 2017-05-08
define canonical synch order to fix potential deadlock

ahgittin:
and now preserve happens-before for child add-remove

see: https://issues.apache.org/jira/browse/BROOKLYN-498

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/entity/AbstractEntity.java
+++ b/core/src/main/java/org/apache/brooklyn/core/entity/AbstractEntity.java
@@ -217,8 +217,15 @@ public abstract class AbstractEntity extends AbstractBrooklynObject implements E
     // values. They must be thread-safe, and where necessary (e.g. group) they should preserve order
     // if possible.
     private Reference<Entity> parent = new BasicReference<Entity>();
-    private Set<Group> groupsInternal = Collections.synchronizedSet(Sets.<Group>newLinkedHashSet());
+    /** Synchronize on this when updating to ensure addition/removals done in order, 
+     * and notifications done in order.
+     * If calling other code while holding this synch lock, any synch locks it might call should be called first
+     * (in particular the AttributeMap.values should be obtained before this if publishing.) */
     private Set<Entity> children = Collections.synchronizedSet(Sets.<Entity>newLinkedHashSet());
+    /** Synchronize on this to ensure group and members are updated at the same time.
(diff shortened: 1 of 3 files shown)


ahgittin:
Mark https://issues.apache.org/jira/browse/BROOKLYN-498 resolved when this is merged.

@aledsage you've looked more closely at the synch behaviour than anyone else so I'd value your thoughts.  As noted at the Jira issue we could drop the guarantee about add/remove event notification -- that would make for a simpler synch model -- but thinking about it having a primary synch object (the attributes map's values) feels like a good solution.  A tidy might be to introduce a synch object at the `AbstractEntity` level and have everything synch on it -- aspire never to have nested/multiple synchronization.  But that could be an iteration on top of this PR.

aledsage:
@ahgittin (cc @neykov) your fix looks good to me, with a few (longer term) caveats:

* Exposing `abstractEntity.getAttributesSynchObjectInternal()` and `attributeMap. getSynchObjectInternal()` feels scary, like it's going in the wrong direction - it's exposing the internal synchronization which could make it harder to reason about and subsequent simplify (e.g. what else will call it?!). However, I agree it's the right thing to do for now.

* I agree about having a "sync object" at the `AbstractEntity` level, used for everything that the entity is doing (including children, groups and sensors).

* Things like `LocalSubscriptionManager.submitPublishEvent` are scary. It calls `subscription.eventFilter.apply()` in the current thread (thus potentially holding locks). This is an example of calling "alien code". In your example, the filter comes from `subscribeToMembers`, so does `parent.getMembers().contains(input.getSource())`. This has led back to a different entity's synchronization block (for group membership) while holding the `AttributeMap`'s lock on the first entity.

* I worry there is still a serious risk of deadlock in more complex use-cases. For example, if involving multiple entities that interact via their group membership and their sensor notifications/subscriptions. We need a proper review (and simplification!) of this.

* Part of the complexity is in our parent/child relationship stuff. We can massively simplify that part once we delete the entity-constructor way of creating entities. We can assume that the entity's parent is set from the very start. That has been deprecated since 0.5.0, but is still used in too many tests for us to be able to actually delete it yet!

In conclusion, I'm happy to (tentatively!) merge this. I can see how it fixes this particular deadlock and makes other deadlocks less likely.

In future PRs, I'll delete the entity-construction, simplify the parent-child code, and I'll also think more about how to simplify (and document) our threading/synchronization model. It definitely deserves attention!

PR #672, opened 2017-05-08
Bundle uninstall and snapshot

ahgittin:
Supports bundle install and auto-reinstall for snapshot POSTs.

Doesn't do anything for auto-upgrade but if you POST a new snapshot ZIP then redeploy, the newer version gets used.  Also cleanly uninstalls catalog items.


ahgittin:
Failure due to deadlock described and fixed in #671 .  Tests pass locally.

geomacy:
@ahgittin have made a start to this but not got all the way through it yet; looks great; I like the tests `testClassAccessAfterUninstall` and friends

core/src/main/java/org/apache/brooklyn/core/typereg/BasicRegisteredType.java
```diff
@@ -39,6 +39,7 @@
     final RegisteredTypeKind kind;
     final String symbolicName;
     final String version;
+    String containingBundle;
```
geomacy: Would it not be better to make this a `VersionedName`?

api/src/main/java/org/apache/brooklyn/api/catalog/CatalogItem.java
```diff
@@ -123,6 +123,8 @@ public String toString() {
     @Nullable public String getIconUrl();
 
     public String getSymbolicName();
+    
+    public String getContainingBundle();
```
geomacy: Why not make this a `VersionedName`?

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/resources/CatalogResource.java
```diff
@@ -158,6 +149,39 @@ public Response createFromYaml(String yaml) {
...
+            result.metadata = in.getMetadata();
+            result.code = in.getCode();
+            if (in.getCatalogItemsInstalled()!=null) {
+                result.types = MutableMap.of();
+                for (String id: in.getCatalogItemsInstalled()) {
+                    // TODO prefer to use RegisteredType, but we need transformer for those in REST
+                    //RegisteredType ci = mgmt.getTypeRegistry().get(id);
+                    
+                    CatalogItem<?, ?> ci = CatalogUtils.getCatalogItemOptionalVersion(mgmt, id);
+                    CatalogTransformer.catalogItemSummary(brooklynU, ci, ui.getBaseUriBuilder());
```
geomacy: Shouldn't this be
```
CatalogItemSummary summary = CatalogTransformer.catalogItemSummary(brooklynU, ci, ui.getBaseUriBuilder());
result.types().put(id, summary);
```

core/src/main/java/org/apache/brooklyn/core/mgmt/ha/OsgiArchiveInstaller.java
```diff
@@ -0,0 +1,429 @@
...
+            discoveredManifest.getMainAttributes().putValue(Constants.BUNDLE_VERSION, inferredMetadata.getVersion());
+        }
+        if (Strings.isBlank(inferredMetadata.getSymbolicName())) {
+            throw new IllegalArgumentException("Missing bundle symbolic name in BOM or MANIFEST");
+        }
+        if (Strings.isBlank(inferredMetadata.getVersion())) {
+            throw new IllegalArgumentException("Missing bundle version in BOM or MANIFEST");
+        }
+        if (discoveredManifest.getMainAttributes().getValue(Attributes.Name.MANIFEST_VERSION)==null) {
+            discoveredManifest.getMainAttributes().putValue(Attributes.Name.MANIFEST_VERSION.toString(), "1.0");
```
aledsage: I think this is always "1.0" (I think it's about the version of the manifest format?), rather than being the version of the code within the jar.

geomacy:
I have **finally** got round to doing a test on this, sorry to have taken so long. 

I will attach a couple of zip files here for test purposes, each being a separate -SNAPSHOT revision of a simple blueprint. The second builds upon the first, getting a bit more complicated.

My first test zip file causes an error when I try to upload it; what you see on the command line is 

    br catalog add file://$(pwd)/test672a.zip
    json: cannot unmarshal string into Go value of type models.CatalogEntitySummary

and in the Karaf log you get 

    2017-05-26 15:46:42,844 WARN  121 o.a.b.u.c.j.ErrorAndToStringUnknownTypeSerializer [qtp941342355-186] Standard serialization not possible for class org.apache.brooklyn.rest.resources.CatalogResource$BundleInstallationRestResult (org.apache.brooklyn.rest.resources.CatalogResource$BundleInstallationRestResult@3f7a9dbf)
    ... lots of repetition deleted...
    com.fasterxml.jackson.databind.JsonMappingException: Infinite recursion (StackOverflowError) (through reference chain: org.apache.brooklyn.core.relations.ByObjectBasicRelationSupport["source"]->org.apache.brooklyn.core.typereg.BasicManagedBundle["relations"]->org.apache.brooklyn.core.relations.ByObjectBasicRelationSupport["source"]->org.apache.brooklyn.core.typereg.BasicManagedBundle["relations"]->org.apache.brookly
            at com.fasterxml.jackson.databind.ser.std.BeanSerializerBase.serializeFields(BeanSerializerBase.java:694)
            at com.fasterxml.jackson.databind.ser.BeanSerializer.serialize(BeanSerializer.java:157)
            at com.fasterxml.jackson.databind.ser.BeanPropertyWriter.serializeAsField(BeanPropertyWriter.java:672)
            at com.fasterxml.jackson.databind.ser.std.BeanSerializerBase.serializeFields(BeanSerializerBase.java:678)
            at com.fasterxml.jackson.databind.ser.BeanSerializer.serialize(BeanSerializer.java:157)
            at com.fasterxml.jackson.databind.ser.BeanPropertyWriter.serializeAsField(BeanPropertyWriter.java:672)
            ... many lines deleted...

I tested the same file against `master` and it loads ok there, so I think there is something in the PR to cause this. 

I will see if I can make a bit of time next week to look at this again and investigate what's going on.

geomacy:
Test files attached (both cause the problem described above.

[test672a.zip](https://github.com/apache/brooklyn-server/files/1032231/test672a.zip)
[test672b.zip](https://github.com/apache/brooklyn-server/files/1032232/test672b.zip)

core/src/main/java/org/apache/brooklyn/core/mgmt/ha/OsgiArchiveInstaller.java
geomacy: ah good point thanks @aledsage

geomacy:
Test files attached (both cause the problem described above.

[test672a.zip](https://github.com/apache/brooklyn-server/files/1033430/test672a.zip)
[test672b.zip](https://github.com/apache/brooklyn-server/files/1033429/test672b.zip)

core/src/main/java/org/apache/brooklyn/core/mgmt/ha/OsgiArchiveInstaller.java
```diff
@@ -0,0 +1,429 @@
...
+                if (!suppliedKnownBundleMetadata.isNameResolved()) {
+                    ManagedBundle mbFromUrl = osgiManager.getManagedBundleFromUrl(suppliedKnownBundleMetadata.getUrl());
+                    if (mbFromUrl!=null) {
+                        // user supplied just a URL (eg brooklyn.libraries), but we recognise it,
+                        // so don't try to reload it, just record the info we know about it to retrieve the bundle
+                        ((BasicManagedBundle)suppliedKnownBundleMetadata).setSymbolicName(mbFromUrl.getSymbolicName());
+                        ((BasicManagedBundle)suppliedKnownBundleMetadata).setVersion(mbFromUrl.getVersion());
+                    }
+                }
+                if (installedBundle.isAbsent() && suppliedKnownBundleMetadata.getOsgiUniqueUrl()!=null) {
```
ahgittin: there's an `isAbsent` check in each `if` test and the result of each body might be an absent so `else wouldn't be sufficient

core/src/main/java/org/apache/brooklyn/core/mgmt/ha/OsgiArchiveInstaller.java
ahgittin: have extracted `"1.0"` as a private static final with comment that this is what it has to be

core/src/main/java/org/apache/brooklyn/core/mgmt/ha/OsgiArchiveInstaller.java
```diff
@@ -0,0 +1,429 @@
...
+            if (!updating) { 
+                synchronized (osgiManager.managedBundles) {
+                    osgiManager.managedBundles.put(result.getMetadata().getId(), result.getMetadata());
+                    osgiManager.managedBundlesByNam.put(result.getMetadata().getVersionedName(), result.getMetadata().getId());
+                    if (Strings.isNonBlank(result.getMetadata().getUrl())) {
+                        osgiManager.managedBundlesByUrl.put(result.getMetadata().getUrl(), result.getMetadata().getId());
+                    }
+                }
+                result.code = OsgiBundleInstallationResult.ResultCode.INSTALLED_NEW_BUNDLE;
+                result.message = "Installed "+result.getMetadata().getVersionedName()+" with ID "+result.getMetadata().getId();
```
ahgittin: good idea

core/src/main/java/org/apache/brooklyn/core/mgmt/ha/OsgiArchiveInstaller.java
```diff
@@ -0,0 +1,429 @@
...
+                    result.bundle.update(fin);
+                }
+            }
+            
+            osgiManager.checkCorrectlyInstalled(result.getMetadata(), result.bundle);
+            ((BasicManagedBundle)result.getMetadata()).setTempLocalFileWhenJustUploaded(zipFile);
+            zipFile = null; // don't close/delete it here, we'll use it for uploading, then it will delete it
+            
+            if (!updating) { 
+                synchronized (osgiManager.managedBundles) {
```
ahgittin: yes, coming back after break, i agree this is sailing too close to the wind! :)

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/resources/CatalogResource.java
ahgittin: absolutely -- good spot!

core/src/main/java/org/apache/brooklyn/core/mgmt/ha/OsgiManager.java
```diff
@@ -209,54 +213,114 @@ public Boolean call() {
...
+     * removes all catalog items it defined,
+     * and then uninstalls the bundle from OSGi.
+     * <p>
+     * No checking is done whether anything is using the bundle;
+     * behaviour of such things is not guaranteed. They will work for many things
+     * but attempts to load new classes may fail.
+     * <p>
+     * Callers should typically fail if anything from this bundle is in use.
+     */
+    public void uninstallUploadedBundle(ManagedBundle bundleMetadata) {
```
ahgittin: +1 to needing `delete`

api/src/main/java/org/apache/brooklyn/api/catalog/CatalogItem.java
ahgittin: I've done this and a commit, but then reverted it.  I'm not sure everything comes from a bundle just yet.  When it does we can reinstate.  To me it is tied up with use of search paths etc.

ahgittin:
This should address everything, including backwards compatibility so the existing CLI works (though it doesn't show the detail).  New PR coming for the CLI to be able to show the detail.

All changes minor and obvious in light of the PR except for https://github.com/apache/brooklyn-server/pull/672/commits/33d9fd63aae2f7da3b2467bba3e47df8269d5acd#diff-3da051230841536d82a7bcceabdb280aR113 which has a detailed comment in the code (above the method).  @geomacy LMK if you think that is too bad or worth it as a temporary fix until we have a new `/bundles` endpoint.

geomacy:
@ahgittin changes look good.  I'm comfortable with the variant return type on `createFromArchive`, especially given the `@Beta` annotation on it.  I shall pull this and try to get my test going and hopefully merge.

geomacy:
Tests all look good, I added test672a.zip to catalog
```
[geoff:] master(+0/-889)* ± br catalog add test672a.zip
myTestEnt:0.12.0.SNAPSHOT
myOtherTest:0.12.0.SNAPSHOT
testent1:0.12.0.SNAPSHOT
```
then deployed a `testEnt1` from it, then added the revised snapshot version of the blueprint by adding  test672b.zip:
```
[geoff:] master(+0/-889)* ± br catalog add test672b.zip
myTestEnt:0.12.0.SNAPSHOT
myOtherTest:0.12.0.SNAPSHOT
myOtherTest2:0.12.0.SNAPSHOT
furtherTest:0.12.0.SNAPSHOT
testent1:0.12.0.SNAPSHOT
testent2:0.12.0.SNAPSHOT
```
and again deployed a `testEnt1`; on this second deployment the updated definition from the second zip was used and the entity created with two children.

LGTM, will merge.

ahgittin:
Thanks @geomacy.  Merged https://github.com/apache/brooklyn-client/pull/53 following @Graeme-Miller 's review there but as I'm a go newbie (g00b ?) appreciate any comments you have there.

ahgittin:
Thanks @geomacy for the super-quick follow-up fix in #719.

neykov:
I've been catching up on some of the merged PRs and noticed the following. When doing `br catalog add .` with the following `catalog.bom` in the folder:
```
brooklyn.catalog:
  bundle: org.apache.brooklyn.test
  version: "0.1.0-SNAPSHOT"
  itemType: entity
  item:
    type: org.apache.brooklyn.core.test.entity.TestEntity
```

I get the error:

```
Server error (400): invalid version "0.1.0-SNAPSHOT": non-numeric "0-SNAPSHOT": NumberFormatException: For input string: "0-SNAPSHOT"
```

In `catalog.bom` we are using maven style versioning and the OSGi style of versioning differs from it, mostly by enforcing a `.`delimiter for the extra tags. It's not restricted to `SNAPSHOT` versions (which is not a special version in OSGi). It applies to any tagged version.

The code must take this into account and:
  * Convert maven style versions to osgi style versions when initializing the bundle metadata. There's `OsgiUtils.toOsgiVersion(String)` to do the mapping which uses the same code as maven does to do the conversion.
  * Keep the duality hidden to the internals. The version from `catalog.bom` should be the one we see the catalog item created with, etc. 
  * For the case where there's no `catalog.bom` and we rely on the `MANIFEST.MF` - try to get the maven version and use that for the user-visible parts. For example `Implementation-Version` header  or `pom.properties`. Fallback to the `Version` header when no alternative source is available.
  * The conversion is one way - we can't recover the maven version from an OSGi version. This means we can't directly map from a bundle version to a catalog version without additional metadata.

geomacy:
Very good point Svet.  I got my test above working with snapshot versions exactly by supplying the OSGI version in the bom, without really thinking about whether that was the right thing to do or not:

```
brooklyn.catalog:
  bundle: io.cloudsoft.cli44
  version: 0.12.0.SNAPSHOT
```

but indeed it's not right that someone should have to do that.

+1 to your bullets above.

geomacy:
P.S. I can have a look at implementing those changes, unless anyone else is particularly keen to do so?

ahgittin:
@neykov @geomacy I've been working in this area for a different reason - if people use `v1` as a version a few minor things break.

wdyt about using OSGi versions everywhere, treating anything else including maven versions, as legacy/deprecated?  i'm hoping in future all catalog item versions inherit from the bundle version, IE we stop versioning items separately, but that needs to be discussed.  (if you installed a non-bundle BOM and bundle would be created.)

for convenience if uploading a bundle with a maven (non-osgi) version, we should convert it (one-way) to an osgi version, and also log a warning.  if we do that i guess there is a question should we use the converted-to-osgi version string for items that inherit (items that define their own version will use what they define but if that differs from bundle maybe we will warn?)

geomacy:
@ahgittin re. the version numbers;  suggest we take this to the mailing list to disuss, but my feeling is that people are used to using Maven versions and it's convenient to regard those as "_the_" version for a module ( = bundle). Could be annoying/error-prone to have to maintain both a version in a pom.xml and another (separate, but almost the same) version in a catalog.bom.

neykov:
I'm not sold on the idea, but that could be because I spend too much time in Java land.
OSGi style versioning feels niche - not widely known/used. It's more restrictive. We'll still need to special case `SNAPSHOT` even though it's not a thing in OSGi. `semver` which is much wider used, even outside of the Java community, better matches our current (maven) versioning style.

I like the vision of having bundle-first design, but feel it's hidden from the user so we are free not not leak the details to the UI.

In any case this deserves a mailing list discussion.

ahgittin:
Thanks @geomacy @neykov -- good points.  Will research and take to ML.

ahgittin:
Results of ML discussion summarised in https://github.com/apache/brooklyn-docs/pull/198 , with PR to follow on this project implementing that (which will be referenced there)

PR #674, opened 2017-05-09
[WIP] [JCLOUDS 2.1.0] Remove Azure workarounds already included in jclouds 2.1.0

neykov:
* The Azure ARM provider now supports the standard naming convention with dashes in the names.
* Retry handler already part of [azurecompute-arm](https://github.com/jclouds/jclouds-labs/blob/master/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/AzureManagementApiMetadata.java#L79)

Tested by deploying a three-tier app to azure using current jclouds master.

Merge only after switching to jclouds-2.1.0.

/cc @drigodwin


neykov:
:) oops

neykov:
Looks like this has been addressed as well so closing.

PR #677, opened 2017-05-10
Adds EBS Location Customizers

nakomis:
Allows user to specify an additional AWS hard drive to be added at provisioning time, e.g.:

```
provisioning.properties:
  customizers:
  - $brooklyn:object:
      type:                               org.apache.brooklyn.location.jclouds.aws.EbsVolumeCustomizers$WithNewVolume
      object.fields:
        availabilityZone:                 us-west-1a
        ec2DeviceName:                    /dev/sdf
        osDeviceName:                     /dev/xvdf
        mountPoint:                       /data
        permissions:                      700
        sizeInGib:                        20
        filesystemType:                   ext4
        deleteOnTermination:              false
```

### DIFF
--- a/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/aws/AbstractEbsVolumeCustomizer.java
+++ b/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/aws/AbstractEbsVolumeCustomizer.java
@@ -1,0 +1,96 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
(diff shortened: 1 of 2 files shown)


drigodwin:
Would you be able to put some more information on what this PR adds please @nakomis? Possibly with an example. Thankyou!

sjcorbett:
Is an AWS-specific feature something we want in Brooklyn or should it be provided in a self-contained repository? There's some precedent for the latter. 
- https://github.com/cloudsoft/brooklyn-aws-elb
- https://github.com/cloudsoft/brooklyn-aws-route53

sjcorbett:
Those projects are both licensed to Cloudsoft so are perhaps not the best examples but my question stands: should we encourage cloud-specific features in core Brooklyn?

nakomis:
@drigodwin description an example added
@sjcorbett the classes sit adjacent to existing cloud-specific classes such as `SoftLayerSameVlanLocationCustomizer` and `AwsAvailabilityZoneExtension` so it's not like this introduces the idea of cloud-specific classes. The alternative would be to move them out of Brooklyn and into a 3rd party (i.e. non-Apache) project

neykov:
@nakomis Do you think it's worth making the customizers skip unknown locations (i.e. do their work only for amazon)?
This would let blueprints customize their environment through `provisioning.properties` but still be cloud agnostic.

nakomis:
@neykov I did look at that, but the `availabilityZone: us-west-1a` field gets automagically pulled into the provisioning properties (grr!), so it's always going to be cloud-specific :-(

neykov:
I mean if you deploy to another cloud it just doesn't do anything, not even populating the template options.

nakomis:
@neykov Done

PR #680, opened 2017-05-15
Adds JcloudsLocationCustomizer hook to allow node / config configuration

nakomis:
Adds JcloudsLocationCustomizer hook to allow node / config configuration which can be used by a customizer to configure the ssh / winrm config tool with runtime data such as machineid

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/util/core/internal/ssh/SshTool.java
+++ b/core/src/main/java/org/apache/brooklyn/util/core/internal/ssh/SshTool.java
@@ -86,6 +86,10 @@ public interface SshTool extends ShellTool {
     public static final ConfigKey<Long> PROP_LAST_MODIFICATION_DATE = newConfigKey("lastModificationDate", "Last-modification-date to be set on files copied/created (should be UTC/1000, ie seconds since 1970; default 0 usually means current)", 0L);
     public static final ConfigKey<Long> PROP_LAST_ACCESS_DATE = newConfigKey("lastAccessDate", "Last-access-date to be set on files copied/created (should be UTC/1000, ie seconds since 1970; default 0 usually means lastModificationDate)", 0L);
     public static final ConfigKey<Integer> PROP_OWNER_UID = newConfigKey("ownerUid", "Default owner UID (not username) for files created on remote machine; default is unset", -1);
+
+    ConfigKey<String> ADDITIONAL_CONNECTION_METADATA = newStringConfigKey("additional.connection.metadata",
+            "Can be used to pass additional custom data to the SshTool, which is especially useful " +
+                    "if writing a bespoke tool implementation");
     
     // TODO remove unnecessary "public static final" modifiers
     
--- a/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
+++ b/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
@@ -756,6 +756,10 @@ public class JcloudsLocation extends AbstractCloudMachineProvisioningLocation im
             if (node == null)
                 throw new IllegalStateException("No nodes returned by jclouds create-nodes in " + getCreationString(setup));
 
+            for (JcloudsLocationCustomizer customizer : customizers) {
+                customizer.customize(this, node, setup);
+            }
+
             boolean windows = isWindows(node, setup);
 
             if (windows) {
--- a/locations/jclouds/src/test/java/org/apache/brooklyn/location/jclouds/JcloudsSshMachineLocationStubbedTest.java
+++ b/locations/jclouds/src/test/java/org/apache/brooklyn/location/jclouds/JcloudsSshMachineLocationStubbedTest.java
@@ -18,14 +18,20 @@
  */
 package org.apache.brooklyn.location.jclouds;
 
+import static org.apache.brooklyn.location.jclouds.JcloudsLocationConfig.JCLOUDS_LOCATION_CUSTOMIZERS;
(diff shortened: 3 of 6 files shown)


bostko:
You can assign any ssh config using 
```
provisioning.properties:
  brooklyn.ssh.config.additionalSshConfig: myValue
```

nakomis:
@bostko If declared in the YAML, then the property must be known *before* the machine is provisioned. The hook allows for a customizer to configure the ssh / winrm config tool with runtime data such as the machine id

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -751,6 +751,10 @@ protected MachineLocation obtainOnce(ConfigBag setup) throws NoMachinesAvailable
                 machineCreationSemaphore.release();
             }
 
+            for (JcloudsLocationCustomizer customizer : customizers) {
+                customizer.customize(this, node, setup);
```
aledsage: Looks like this should be moved down slightly - we don't set `node` until just after this, so we'll always be passing `null` to the customizer.

aledsage: I'd move this to just after the `if (null == null)` check.

locations/jclouds/src/test/java/org/apache/brooklyn/location/jclouds/JcloudsSshMachineLocationStubbedTest.java
```diff
@@ -118,4 +123,17 @@ public void testWinrmConfigPassedToMachine() throws Exception {
...
         assertEquals(machine.config().get(WinRmTool.PROP_EXEC_TRIES), Integer.valueOf(456));
     }
+
+    @Test
+    public void testNodeSetupCustomizer() throws Exception {
+        final String testMetadata = "test-metadata";
+        obtainMachine(ImmutableMap.of(JCLOUDS_LOCATION_CUSTOMIZERS, ImmutableList.of(new BasicJcloudsLocationCustomizer(){
+            @Override
+            public void customize(JcloudsLocation location, NodeMetadata node, ConfigBag setup) {
+                setup.configure(ADDITIONAL_CONNECTION_METADATA, testMetadata);
```
aledsage: Let's also assert that `node` is not null (given it will be, based on the bug I spotted (I believe)).
And might as well also assert that `location` is equal to the object we expect.

software/winrm/src/main/java/org/apache/brooklyn/util/core/internal/winrm/WinRmTool.java
```diff
@@ -86,6 +86,8 @@
             "Size of file chunks (in bytes) to be used when copying a file to the remote server", 
             1024);
 
+    ConfigKey<String> ADDITIONAL_CONNECTION_METADATA = newStringConfigKey("additional.connection.metadata", "Can be used by a location customizer during provisioning to provide additional information, which in turn can be consumed by a custom ssh tool");
```
aledsage: Personal preference for splitting onto multiple lines when it's this crazy-long a line. Otherwise, it's hard to tell if there's a default value or anything like that lurking at the very end. I'd go for one line per arg.

core/src/main/java/org/apache/brooklyn/util/core/internal/ssh/SshTool.java
```diff
@@ -86,6 +86,8 @@
     public static final ConfigKey<Long> PROP_LAST_MODIFICATION_DATE = newConfigKey("lastModificationDate", "Last-modification-date to be set on files copied/created (should be UTC/1000, ie seconds since 1970; default 0 usually means current)", 0L);
     public static final ConfigKey<Long> PROP_LAST_ACCESS_DATE = newConfigKey("lastAccessDate", "Last-access-date to be set on files copied/created (should be UTC/1000, ie seconds since 1970; default 0 usually means lastModificationDate)", 0L);
     public static final ConfigKey<Integer> PROP_OWNER_UID = newConfigKey("ownerUid", "Default owner UID (not username) for files created on remote machine; default is unset", -1);
+
+    ConfigKey<String> ADDITIONAL_CONNECTION_METADATA = newStringConfigKey("additional.connection.metadata", "Can be used by a location customizer during provisioning to provide additional information, which in turn can be consumed by a custom ssh tool");
```
aledsage: Split onto multiple lines.

software/winrm/src/main/java/org/apache/brooklyn/util/core/internal/winrm/WinRmTool.java
aledsage: Says "custom ssh tool".

aledsage: Perhaps word as something like "Can be used to pass additional custom data to the `WinrmTool`, which is especially useful if writing a bespoke tool implementation."

Your description is accurate for your use-case, but really it's not specific to the customizer - it could be passed in as normal location config or entity `provisioning.properties`.

locations/jclouds/src/test/java/org/apache/brooklyn/location/jclouds/JcloudsSshMachineLocationStubbedTest.java
nakomis: I've added a null check for `location`, but I'm not sure what else to assert on it

aledsage:
LGTM; merging

PR #681, opened 2017-05-15
Flag allowing a Winrm4jTool to use unique context

bostko:
Fixes https://github.com/cloudsoft/winrm4j/issues/63

To test add in location configuration:

    brooklyn.winrm.config.use.unique.context: true

### DIFF
--- a/software/winrm/src/main/java/org/apache/brooklyn/util/core/internal/winrm/winrm4j/Winrm4jTool.java
+++ b/software/winrm/src/main/java/org/apache/brooklyn/util/core/internal/winrm/winrm4j/Winrm4jTool.java
@@ -59,7 +59,7 @@ import io.cloudsoft.winrm4j.winrm.WinRmToolResponse;
 public class Winrm4jTool implements org.apache.brooklyn.util.core.internal.winrm.WinRmTool, ManagementContextInjectable {
 
     private static final Logger LOG = LoggerFactory.getLogger(Winrm4jTool.class);
-    
+
     private static final ConfigKey<WinRmClientContext> CONTEXT = ConfigKeys.newConfigKey(WinRmClientContext.class, "winrm.context");
 
     // TODO Should we move this up to the interface?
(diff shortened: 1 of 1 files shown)


aledsage:
LGTM (though I'v just read the code rather than testing it).

Note that in `SshMachineLocation`, it caches the sessions and reuses them only if a given set of config keys are identical (e.g. usename, password, etc).

I'm not suggesting you change this code to do that, but it's something worth bearing in mind.

@neykov any more comments/concerns before we merge this?

software/winrm/src/main/java/org/apache/brooklyn/util/core/internal/winrm/winrm4j/Winrm4jTool.java
```diff
@@ -237,7 +240,7 @@ private static synchronized WinRmClientContext createWinrmContext(ManagementCont
         // TODO Use getScratchpad()
         BrooklynProperties props = ((ManagementContextInternal)mgmt).getBrooklynProperties();
         WinRmClientContext instance = props.getConfig(CONTEXT);
-        if (instance == null) {
+        if (instance == null || Boolean.TRUE.equals(props.getConfig(USE_UNIQUE_CONTEXT))) {
```
neykov: This will cause buildup of connection pools & threads for them (pool per `executePs` call).
Instead I suggest that we don't pass the `context` at all to `WinRmTool.Builder`. When no `context` is passed externally, `WinRmClient` will create one and destroy it when done with the request.
The flag doesn't seem too useful since it causes incorrect behaviour when activated - suggest doing the changes unconditionally.

Longer term would be great if we implement caching like in `SshMachineLocation` as @aledsage suggests, but with the addition of a `release` step so we can terminate the `context` when no longer needed.

bostko:
Thank you @neykov. I changed PR as you suggested.

software/winrm/src/main/java/org/apache/brooklyn/util/core/internal/winrm/winrm4j/Winrm4jTool.java
```diff
@@ -59,8 +59,11 @@
 public class Winrm4jTool implements org.apache.brooklyn.util.core.internal.winrm.WinRmTool, ManagementContextInjectable {
 
     private static final Logger LOG = LoggerFactory.getLogger(Winrm4jTool.class);
-    
+
     private static final ConfigKey<WinRmClientContext> CONTEXT = ConfigKeys.newConfigKey(WinRmClientContext.class, "winrm.context");
+    private static final ConfigKey<Boolean> USE_UNIQUE_CONTEXT = ConfigKeys.newBooleanConfigKey("use.unique.context",
```
neykov: Not used, remove. otherwise LGTM.

neykov:
Thanks @bostko, merging.

PR #682, opened 2017-05-15
Delete support for `--app <groovy-file>`

aledsage:
I couldn't see this documented in http://brooklyn.apache.org/v/latest/ops/server-cli-reference.html, so I don't know of any docs that also need to be deleted.


bostko:
Looks good.

PR #684, opened 2017-05-18
Switch to Java 8

geomacy:
This updates the Java version to `1.8` and makes a couple of changes to get Brooklyn to build again.  

For demonstration purposes I've switched a `new Callable ...` in 
[BasicExecutionContext.java](https://github.com/apache/brooklyn-server/compare/master...geomacy:switch-to-java-8?expand=1#diff-1876d3c7b9d3cc5d485f7469b403e281R216) to use a Java 8 lambda expression, really just to prove that it is Java 8! (This was useful as it showed we needed to update the Checkstyle plugin to a level that works with 8).   

I've tested that Brooklyn runs in both classic and Karaf launchers, and can successfully deploy a basic server and a couple of apps from Brooklyn Central, (a three tier app and `brooklyn-etcd`). 

There's a matching update to `brooklyn-ui`, https://github.com/apache/brooklyn-ui/pull/44.

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/objs/BasicSpecParameter.java
+++ b/core/src/main/java/org/apache/brooklyn/core/objs/BasicSpecParameter.java
@@ -571,10 +571,11 @@ public class BasicSpecParameter<T> implements SpecParameter<T>{
         }
 
         /** used if yaml specifies a parameter, but possibly incomplete, and a spec supertype has a parameter */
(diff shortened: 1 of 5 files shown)


duncangrant:
That's a nice example of java 8ifyng code.  LGTM

m4rkmckenna:
Thanks @geomacy verified this locally 

merging now

PR #690, opened 2017-05-21
External ip config for winrm byon location

iyovcheva:


core/src/main/java/org/apache/brooklyn/location/byon/ByonLocationResolver.java
```diff
@@ -152,49 +153,54 @@ protected ConfigBag extractConfig(Map<?,?> locationFlags, String spec, LocationR
...
-        Map<Integer, String> tcpPortMappings = (Map<Integer, String>) machineConfig.get("tcpPortMappings");
-        
-        checkArgument(ssh != null ^ winrm != null, "Must specify exactly one of 'ssh' or 'winrm' for machine: %s", valSanitized);
-        
-        UserAndHostAndPort userAndHostAndPort;
-        String host;
-        int port;
-        if (ssh != null) {
-            userAndHostAndPort = parseUserAndHostAndPort(ssh, 22);
+        if (machineConfig.containsKey("winrm") && !(machineConfig.get("winrm") instanceof String)) {
```
bostko: Looks good.
@iyovcheva what do you think about supporting external provider for ssh address as well. It will be more consistent.

iyovcheva: @bostko I agree. Will need to test it.

iyovcheva: External config is not supported for ssh. If we implement it, it's good to go in separate PR.

core/src/main/java/org/apache/brooklyn/location/byon/ByonLocationResolver.java
```diff
@@ -152,49 +153,55 @@ protected ConfigBag extractConfig(Map<?,?> locationFlags, String spec, LocationR
...
```
neykov: @iyovcheva The changes here break the support for specifying `winrm` as `user@host:port`. Can you add it to the test cases.
On the other hand your approach is also interesting because it doesn't resolve the value here - at the time of deploying the blueprint. Obtaining the machine could happen much later in the lifecycle where the external values have already changed.

A proper solution would involve moving all of non-essential parsing to the machine location, leaving only minimal sanity checks here.

core/src/main/java/org/apache/brooklyn/location/byon/ByonLocationResolver.java
```diff
@@ -152,49 +153,55 @@ protected ConfigBag extractConfig(Map<?,?> locationFlags, String spec, LocationR
...
-        if (userAndHostAndPort.getUser() != null) {
-            checkArgument(!vals.containsKey("user"), "Must not specify user twice for machine: %s", valSanitized);
-            machineConfig.put("user", userAndHostAndPort.getUser());
-        }
-        if (userAndHostAndPort.getHostAndPort().hasPort()) {
-            checkArgument(!vals.containsKey("port"), "Must not specify port twice for machine: %s", valSanitized);
-            machineConfig.put("port", port);
+            // If there is a tcpPortMapping defined for the connection-port, then use that for ssh/winrm machine
+            port = userAndHostAndPort.getHostAndPort().getPort();
+            if (tcpPortMappings != null && tcpPortMappings.containsKey(port)) {
```
geomacy: I would do this `if` in a separate method, this one is long enough already!

core/src/main/java/org/apache/brooklyn/location/byon/ByonLocationResolver.java
```diff
@@ -152,49 +153,55 @@ protected ConfigBag extractConfig(Map<?,?> locationFlags, String spec, LocationR
...
+                port = hostAndPortOverride.getPort();
+                host = hostAndPortOverride.getHostText().trim();
+            } else {
+                host = userAndHostAndPort.getHostAndPort().getHostText().trim();
+            }
+
+            machineConfig.put("address", host);
+            try {
+                InetAddress.getByName(host);
+            } catch (Exception e) {
```
geomacy: `UnknownHostException`

core/src/main/java/org/apache/brooklyn/location/byon/ByonLocationResolver.java
```diff
@@ -152,49 +153,55 @@ protected ConfigBag extractConfig(Map<?,?> locationFlags, String spec, LocationR
...
+            checkArgument(ssh != null ^ winrm != null, "Must specify exactly one of 'ssh' or 'winrm' for machine: %s", valSanitized);
+
+            UserAndHostAndPort userAndHostAndPort;
+            String host;
+            int port;
+            if (ssh != null) {
+                userAndHostAndPort = parseUserAndHostAndPort(ssh, 22);
+            } else {
+                // TODO set to null and rely on the MachineLocation. If not then make a dependency to WinRmMachineLocation and use its config key name.
+                userAndHostAndPort = parseUserAndHostAndPort(winrm, vals.get("winrm.useHttps") != null && (Boolean)vals.get("winrm.useHttps") ? 5986 : 5985);
```
geomacy: Do the boolean expression in a separate variable for readability.

Second clause should be `Boolean.parseValue(vals.get("winrm.useHttps").toString())`.

core/src/main/java/org/apache/brooklyn/location/byon/ByonLocationResolver.java
geomacy: `!(machineConfig.get("winrm") instanceof String)` might be better as `machineConfig.get("winrm") instanceof DeferredSupplier`?

geomacy:
hi @iyovcheva just checking over brooklyn-server open PRs, what's the latest on this one?  Do you want to update it as @neykov mentioned [above](https://github.com/apache/brooklyn-server/pull/690#discussion_r118445857)?

iyovcheva:
@geomacy yep, I'm currently working on it

iyovcheva:
retest this please

iyovcheva:
Failing with "cannot create temporary folders in parent /tmp after 1000 attempts" which is not related to the PR

tbouron:
retest this please

nakomis:
retest this please

PR #691, opened 2017-05-22
Fix generics usage (for compilation in Eclipse)

aledsage:
Fixes generics compilation errors that the Eclipse compiler reports, with Java 8 (in Eclipse 4.6.3).


core/src/test/java/org/apache/brooklyn/core/network/OnPublicNetworkEnricherTest.java
```diff
@@ -372,7 +371,7 @@ public void testIgnoresWithNoPort() throws Exception {
         
         // Ugly casting in java, but easy to get passed this when constructed from YAML
```
geomacy: don't need this comment any more?

aledsage:
Thanks @geomacy - I've deleted that comment. Merging now.

PR #694, opened 2017-05-23
Fix non-deterministic DefaultCOnnectivityResolverTest

aledsage:
Was failing on apache Jenkins sometimes, due to a 1ms timeout!
Increased this to 100ms, which still feels low (but unfortunately
The test will take a minimum of `Math.min(timeout, gracePeriod)`.


sjcorbett:
See my comment on https://github.com/apache/brooklyn-server/pull/651#discussion_r114961921 for why these failures are likelier to occur with a short timeout.

PR #693, opened 2017-05-22
jclouds TemplateOptions type coercions

aledsage:

PR #696, opened 2017-05-24
Fix DslYamlTest

aledsage:
Test previously failed non-deterministically in jenkins (e.g. see https://github.com/apache/brooklyn-server/pull/693#issuecomment-303414269)


sjcorbett:
Looks good, thanks @aledsage.

PR #705, opened 2017-05-29
jclouds stub test: better allow config to be overridden

aledsage:
This allows sub-classes to better override the config values
(the `jcloudsLocationConfig` can contain keys of type ConfigKey or String).

Also fixes the log.debug, to say what the default value is overridden with.

The motivator for this change is my use of it in a downstream project, which failed because the overrides weren't being applied (when passed in as string keys).


aledsage:
jenkins build passed (see https://builds.apache.org/job/brooklyn-server-pull-requests/2173). Not sure why it's not updating the PR (to say that the PR was being built, or with the result).

Merging now.

PR #708, opened 2017-05-30
Getting the correct value for WinRmMachineLocation.USE_HTTPS_WINRM

ygy:
- the config still doesn't contain the correct value for
  USE_HTTPS_WINRM, when it is set via provisioning.properties


bostko:
LGTM.

aledsage:
This fixes https://issues.apache.org/jira/browse/BROOKLYN-509. Merging.

PR #712, opened 2017-05-30
Updated GCE features

drigodwin:
Added missing GCE dependency to features and moved to load order. Tested deployment on GCE (which was previously not working).


andreaturli:
thanks @drigodwin, I wonder if we shouldn't use this https://github.com/jclouds/jclouds-karaf/blob/master/feature/src/main/feature/feature.xml and import, for example, https://github.com/jclouds/jclouds-karaf/blob/master/feature/src/main/feature/feature.xml#L32 to avoid this issues

m4rkmckenna:
@andreaturli I agree ... has this been included in a JClouds release yet?

andreaturli:
not yet

drigodwin:
I agree @andreaturli, we should use this when it becomes available. I think we should also make this change in the meantime however as this fixes GCE in the short term.

andreaturli:
lgtm

andreaturli:
testing locally and merging shortly

andreaturli:
thanks @drigodwin

PR #713, opened 2017-05-30
Fix for BROOKLYN-510: failure with inboundPorts specified

drigodwin:
Tested on GCE / Centos 7

### DIFF
--- a/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
+++ b/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
@@ -47,6 +47,7 @@ import java.util.concurrent.atomic.AtomicReference;
 import javax.annotation.Nullable;
 import javax.xml.ws.WebServiceException;
 
+import com.google.common.primitives.Ints;
 import org.apache.brooklyn.api.entity.Entity;
 import org.apache.brooklyn.api.location.LocationSpec;
 import org.apache.brooklyn.api.location.MachineLocation;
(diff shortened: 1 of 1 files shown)


locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -936,8 +937,7 @@ protected MachineLocation obtainOnce(ConfigBag setup) throws NoMachinesAvailable
                     } else {
                         LOG.warn("Using DEPRECATED flag OPEN_IPTABLES (will not be supported in future versions) for {} at {}", machineLocation, this);
 
-                        @SuppressWarnings("unchecked")
-                        Iterable<Integer> inboundPorts = (Iterable<Integer>) setup.get(INBOUND_PORTS);
+                        Iterable<Integer> inboundPorts = Ints.asList(template.getOptions().getInboundPorts());
 
                         if (inboundPorts == null || Iterables.isEmpty(inboundPorts)) {
```
aledsage: Very minor: we could now get rid of the null check, given `Ints.asList()` (and `template.getOptions().getInboundPorts()`) will never return null.

Also, we don't need to convert it to an `Iterable<Integer>`. We could leave it as a `int[]`, and change the line below to `inboundPorts.length == 0`. No strong feelings.

But will merge as-is.

aledsage:
LGTM - merging now.

bostko:
To what it will transform for the blueprint in BROOKLYN-510?
Will the yaml bellow 
```
    inboundPorts:
    - 22
    - 9100-9200
```
mean `22, 9101, 9202,..., 9200`?

drigodwin:
@bostko, it should open `22, 9100, 9101... 9199, 9200`

PR #714, opened 2017-06-01
BROOKLYN-513: fix coercion for jclouds reflective builder calls

aledsage:


utils/common/src/main/java/org/apache/brooklyn/util/javalang/MethodAccessibleReflections.java
```diff
@@ -0,0 +1,142 @@
...
+            return false;
+        }
+    }
+
+    /**
+     * @see {@link Reflections#findAccessibleMethod(Method)}
+     */
+    static Method findAccessibleMethod(Method method) {
+        if (!Modifier.isPublic(method.getModifiers())) {
+            trySetAccessible(method);
```
neykov: Let's not do this and use only publicly accessible methods, just as one would do when creating the object from java.

utils/common/src/main/java/org/apache/brooklyn/util/javalang/MethodAccessibleReflections.java
```diff
@@ -0,0 +1,142 @@
...
+
+    private static Maybe<Method> tryFindPublicMethod(Class<?> clazz, String methodName, Class<?>... parameterTypes) {
+        if (!Modifier.isPublic(clazz.getModifiers())) {
+            return Maybe.absent();
+        }
+        
+        try {
+            Method altMethod = clazz.getMethod(methodName, parameterTypes);
+            if (Modifier.isPublic(altMethod.getModifiers()) && !Modifier.isStatic(altMethod.getModifiers())) {
+                return Maybe.of(altMethod);
```
neykov: Skip it if the class is not accessible. This will fail with same problem if you have:
public class 1 - public method
protected class 2 < class 1 - public method
protected class 3 < class 2 - public method

core/src/main/java/org/apache/brooklyn/util/core/flags/MethodCoercions.java
```diff
@@ -83,10 +86,11 @@ public boolean apply(@Nullable Method input) {
         Optional<Method> matchingMethod = Iterables.tryFind(methods, matchSingleParameterMethod(methodName, argument));
         if (matchingMethod.isPresent()) {
             Method method = matchingMethod.get();
+            Method accessibleMethod = Reflections.findAccessibleMethod(method);
```
neykov: The approach will fail for classes which have similar methods and one of them is not accessible.
For example the builder of `IpOptions` has
```
      abstract Builder address(Optional<String> address);
      public Builder address(String address) {
```

If we have a registered coercion for `T -> Optional<T>`  it will choose the first one, but there's no public alternative for it in the superclasses. A simpler case would be two methods to have `int`, `string` arguments where again the former can be coerced to the latter.

A more resilient approach would be to get a list of the public methods for all interfaces and superclasses (at line 76), filter them for public access and then do the name/parameter matching.

neykov:
Seems the method matching logic is already implemented in `org.apache.commons.lang3.reflect.MethodUtils` (`commons-lang3`) which we already have as a dependency, so might be better if we reuse that.

neykov:
Scratch last comment - doesn't take coercion into account.

ygy:
I've successfully tested it with a manual build. It covers the use case for specifying ipOptions in jclouds.

aledsage:
Thanks @neykov and @ygy - I'll incorporate some of those comments and then merge.

utils/common/src/main/java/org/apache/brooklyn/util/javalang/MethodAccessibleReflections.java
aledsage: Agreed - will remove the `setAccessible` calls from this method.

utils/common/src/main/java/org/apache/brooklyn/util/javalang/MethodAccessibleReflections.java
aledsage: Not sure I follow - at the top of this method it checks for `Modifier.isPublic(clazz.getModifiers())`. Does that cover what you're suggesting?

core/src/main/java/org/apache/brooklyn/util/core/flags/MethodCoercions.java
aledsage: Good point. I'll defer that, and not try to do it in this PR.

aledsage:
@neykov I've updated the PR. If you get a chance to review again, that would be great. Otherwise I'll wait for apache jenkins and then squash it into one commit and merge.

utils/common/src/main/java/org/apache/brooklyn/util/javalang/MethodAccessibleReflections.java
neykov: Somehow missed that, it's all good then.

utils/common/src/main/java/org/apache/brooklyn/util/javalang/MethodAccessibleReflections.java
```diff
@@ -0,0 +1,173 @@
...
+     */
+    private static final Set<String> SET_ACCESSIBLE_SUCCEEDED_LOGGED_METHODS = Sets.newConcurrentHashSet();
+    
+    /**
+     * Contains method.toString() representations for methods for which {@link #findAccessibleMethod(Method)} 
+     * failed to find a suitable publicly accessible method.
+     */
+    private static final Set<String> FIND_ACCESSIBLE_FAILED_LOGGED_METHODS = Sets.newConcurrentHashSet();
+
+    static boolean trySetAccessible(Method method) {
```
neykov: Should we merge this if it's not used?

neykov:
LGTM. Jenkins hasn't kicked in yet. Building locally and will merge if no errors.

PR #715, opened 2017-06-01
Feature/http commnand effector improvements

andreaturli:


core/src/main/java/org/apache/brooklyn/core/effector/http/HttpCommandEffector.java
```diff
@@ -71,9 +88,10 @@
     public static final ConfigKey<Map<String, String>> EFFECTOR_HTTP_HEADERS = new MapConfigKey(String.class, "headers");
     public static final ConfigKey<Object> EFFECTOR_HTTP_PAYLOAD = ConfigKeys.newConfigKey(Object.class, "httpPayload");
     public static final ConfigKey<String> JSON_PATH = ConfigKeys.newStringConfigKey("jsonPath", "JSON path to select in HTTP response");
-    public static final ConfigKey<String> PUBLISH_SENSOR = ConfigKeys.newStringConfigKey("publishSensor", "Sensor name where to store json path extracted value");
+    public static final ConfigKey<Map<String, String>> JSON_PATHS_AND_SENSORS = new MapConfigKey(String.class, "jsonPathAndSensors", "json path selector and correspondant sensor name tha will publish the json path extracted value");
```
drigodwin: tha -> that

core/src/test/java/org/apache/brooklyn/core/effector/http/HttpCommandEffectorHttpBinTest.java
```diff
@@ -147,8 +147,10 @@ public void testHttpEffectorWithPayload() throws Exception {
                         "public", "false",
                         "files", ImmutableMap.of("demo.txt", ImmutableMap.of("content","Demo"))))
                 .configure(HttpCommandEffector.EFFECTOR_HTTP_HEADERS, ImmutableMap.of("Content-Type", "application/json"))
-                .configure(HttpCommandEffector.JSON_PATH, "$.url")
-                .configure(HttpCommandEffector.PUBLISH_SENSOR, "result")
+//                .configure(HttpCommandEffector.JSON_PATH, "$.url")
+//                .configure(HttpCommandEffector.PUBLISH_SENSOR, "result")
```
drigodwin: Why not remove these if they're no longer needed?

core/src/main/java/org/apache/brooklyn/core/effector/http/HttpCommandEffector.java
geomacy: should `PUBLISH_SENSOR` not be deprecated rather than removed?

core/src/main/java/org/apache/brooklyn/core/effector/http/HttpCommandEffector.java
```diff
@@ -181,7 +204,19 @@ private HttpRequest buildHttpRequest(String httpVerb, URI uri, Map<String, Strin
                 if (contentType == null || contentType.equalsIgnoreCase(APPLICATION_JSON)) {
                     LOG.warn("Content-Type not specified. Using {}, as default (continuing)", APPLICATION_JSON);
                     body = toJsonString(payload);
-                } else if (!contentType.equalsIgnoreCase(APPLICATION_JSON)) {
+                } else if (contentType.equalsIgnoreCase(APPLICATION_X_WWW_FORM_URLENCODE)) {
+                    if (payload instanceof Map) {
+                        for (Map.Entry<String, String> entry : ((Map<String, String>) payload).entrySet()) {
+                            try {
+                                if (!body.equals("")) body += "&";
+                                body += URLEncoder.encode(entry.getKey(), StandardCharsets.UTF_8.toString()) + "=" + URLEncoder.encode(entry.getValue(), StandardCharsets.UTF_8.toString());
```
geomacy: `URLEncoder`, despite its name, is for HTML form encoding, so it's probably not exactly what you want here, e.g. a space character will become a `+`.   I would probably use `java.net.URI`, e.g.
```
new URI(null,  "<img>",  null).toASCIIString()
```
returns `%3Cimg%3E`.

core/src/main/java/org/apache/brooklyn/core/effector/http/HttpCommandEffector.java
andreaturli: good question @geomacy, I think it is safer to deprecate it before removing it completely although it is a @Beta feature

core/src/main/java/org/apache/brooklyn/core/effector/http/HttpCommandEffector.java
andreaturli: thanks, @geomacy but I think this is the intended behavior.
`URLEncoder.encode` javadoc explicitly says 
```
Translates a string into {@code application/x-www-form-urlencoded}
     * format using a specific encoding scheme. 
```
so I think it is ok, no?

fyi, the `UrlEncoder.encode("")` is `%3Cimg%3E`. I'll add more unit tests for the encoding anyway.

Am I missing something? thanks

geomacy: Yes, but 
```java
    public static final Logger LOG = LoggerFactory.getLogger(TempDeleteMe.class);
    @Test
    public void testit() throws Exception {
        String uri = new URI(null,  "<b>hello world</b> ",  null).toASCIIString();
        String urlEncoder = URLEncoder.encode("<b>hello world</b> ", "UTF-8");
        System.out.println(uri);
        System.out.println(urlEncoder);
        Assert.assertEquals(uri, urlEncoder);
    }
```
fails and prints
```
%3Cb%3Ehello%20world%3C/b%3E%20
%3Cb%3Ehello+world%3C%2Fb%3E+
```
You probably just want your values to be escaped according to RFC 2396, which is what URI does (almost).  URLEncoder turns your value into `application/x-www-form-urlencoded`, which is not the same, and is especially unhelpful in how it treats space characters (`+`), which you probably want as `%20`, right?

andreaturli: Thanks @geomacy for helping me here. To be honest I don't know exactly what to do, probably best to implement the less surprising behavior. I'll try to clarify with more tests and docs

geomacy: ah I take it back - of course it's not the URL you're processing here, you're doing a POST on a form body; so in this case URLEncoder is what you want!

geomacy: I managed to get myself misled by the `convertToUri` at the start of the diff above, and missed the fact that this bit of code is down inside `buildHttpRequest`. Sorry about that!

andreaturli: ok cool, no problem. I will add more tests and submit a new commit. Thanks

andreaturli:
@geomacy can you have another look at this, please?

core/src/main/java/org/apache/brooklyn/core/effector/http/HttpCommandEffector.java
```diff
@@ -88,7 +91,9 @@
     public static final ConfigKey<Map<String, String>> EFFECTOR_HTTP_HEADERS = new MapConfigKey(String.class, "headers");
     public static final ConfigKey<Object> EFFECTOR_HTTP_PAYLOAD = ConfigKeys.newConfigKey(Object.class, "httpPayload");
     public static final ConfigKey<String> JSON_PATH = ConfigKeys.newStringConfigKey("jsonPath", "JSON path to select in HTTP response");
-    public static final ConfigKey<Map<String, String>> JSON_PATHS_AND_SENSORS = new MapConfigKey(String.class, "jsonPathAndSensors", "json path selector and correspondant sensor name tha will publish the json path extracted value");
+    @Deprecated
```
geomacy: Needs a comment with "Deprecated since=0.12.0"

core/src/main/java/org/apache/brooklyn/core/effector/http/HttpCommandEffector.java
```diff
@@ -88,7 +91,9 @@
     public static final ConfigKey<Map<String, String>> EFFECTOR_HTTP_HEADERS = new MapConfigKey(String.class, "headers");
     public static final ConfigKey<Object> EFFECTOR_HTTP_PAYLOAD = ConfigKeys.newConfigKey(Object.class, "httpPayload");
     public static final ConfigKey<String> JSON_PATH = ConfigKeys.newStringConfigKey("jsonPath", "JSON path to select in HTTP response");
-    public static final ConfigKey<Map<String, String>> JSON_PATHS_AND_SENSORS = new MapConfigKey(String.class, "jsonPathAndSensors", "json path selector and correspondant sensor name tha will publish the json path extracted value");
+    @Deprecated 
+    public static final ConfigKey<String> PUBLISH_SENSOR = ConfigKeys.newStringConfigKey("publishSensor", "Sensor name where to store json path extracted value");
+    public static final ConfigKey<Map<String, String>> JSON_PATHS_AND_SENSORS = new MapConfigKey(String.class, "jsonPathAndSensors", "json path selector and correspondant sensor name that will publish the json path extracted value");
```
geomacy: maybe better "corresponding" rather than "correspondent"

core/src/main/java/org/apache/brooklyn/core/effector/http/HttpCommandEffector.java
```diff
@@ -72,7 +74,8 @@
  * It deals with some {@link HttpHeaders.CONTENT_TYPE} namely 'application/json' (as default) and 'application/x-www-form-urlencoded'. 
  * In the latter case, a map payload will be URLEncoded in a single string
  * 
- * With optional JSON_PATH config key, the effector will extract a section of the json response.
+ * With optional JSON_PATH config key, the effector will extract a section of the json response. 
+ * Notice if both JSON_PATH and JSON_PATHS_AND_SENSORS are both defined, JSON_PATH will take precedence.
```
geomacy: This is different from the behaviour implemented at lines 139-141

core/src/main/java/org/apache/brooklyn/core/effector/http/HttpCommandEffector.java
```diff
@@ -211,7 +226,7 @@ private HttpRequest buildHttpRequest(String httpVerb, URI uri, Map<String, Strin
                                 if (!body.equals("")) body += "&";
                                 body += URLEncoder.encode(entry.getKey(), StandardCharsets.UTF_8.toString()) + "=" + URLEncoder.encode(entry.getValue(), StandardCharsets.UTF_8.toString());
                             } catch (UnsupportedEncodingException e) {
-                                e.printStackTrace();
+                                throw Throwables.propagate(e);
```
geomacy: why not `Exceptions.propagate(e);`?

core/src/test/java/org/apache/brooklyn/core/effector/http/HttpCommandEffectorTest.java
```diff
@@ -248,6 +248,27 @@ public void testPayloadWithContentTypeHeaderXml() throws InterruptedException {
...
+              .configure(HttpCommandEffector.EFFECTOR_HTTP_HEADERS, ImmutableMap.of(HttpHeaders.CONTENT_TYPE, HttpCommandEffector.APPLICATION_X_WWW_FORM_URLENCODE))
+              .configure(HttpCommandEffector.JSON_PATH, "$.data")
+      );
+      assertNotNull(httpCommandEffector);
+      TestEntity testEntity = app.createAndManageChild(buildEntitySpec(httpCommandEffector));
+      Object output = testEntity.invoke(EFFECTOR_HTTP_COMMAND, ImmutableMap.of()).getUnchecked(Duration.seconds(1));
+      assertEquals(output, "{\"key\", \"%3Cimg%3E\"}");
+      
+      assertEquals(server.getRequestCount(), 1);
+      assertSent(server, "POST", "/post");
```
geomacy: you could add a check on the result of `assertSent` in here to verify that the payload key got successfully URL encoded

core/src/main/java/org/apache/brooklyn/core/effector/http/HttpCommandEffector.java
andreaturli: thanks! :)

core/src/main/java/org/apache/brooklyn/core/effector/http/HttpCommandEffector.java
andreaturli: thanks!

core/src/main/java/org/apache/brooklyn/core/effector/http/HttpCommandEffector.java
andreaturli: forgot to remove this comment after changing the impl. thx

core/src/test/java/org/apache/brooklyn/core/effector/http/HttpCommandEffectorTest.java
andreaturli: good idea, added it to the test

andreaturli:
@geomacy I think I've addressed all of your comments. please let me know if it is good to merge

andreaturli:
thanks!

PR #721, opened 2017-06-07
DynamicFabric: add support for firstMemberSpec

rdowner:
Just like DynamicCluster does.


aledsage:
@rdowner agree with @drigodwin - code looks good, but would be good to have a test similar to `DynamicClusterTest.testDifferentFirstMemberSpec` (see
https://github.com/apache/brooklyn-server/blob/master/core/src/test/java/org/apache/brooklyn/entity/group/DynamicClusterTest.java#L975)

rdowner:
Thanks for the comments. Sorry for the delay so far, mainly due to vacation. Having a few problems with writing a good test but will get this PR updated as soon as I can.

rdowner:
Ready for re-review @drigodwin @aledsage

PR #722, opened 2017-06-07
[BROOKLYN-515] WindowsPerformanceCounterSensors does not show values

bostko:
WindowsPerformanceCounterSensors does not show values on some deployments.


bostko:
retest this please

PR #731, opened 2017-06-12
BROOKLYN-440: ssh not use StreamGobbler for logging stdout

aledsage:
Improves thread usage when executing ssh remote commands. Simplifies the code to use a new `LoggingOutputStream`, rather than using the `StreamGobbler` thread.

@neykov I'd appreciate your thoughts. It only does the first part of https://issues.apache.org/jira/browse/BROOKLYN-440; it doesn't do your suggestion of having the single thread polling to see if the streams have any bytes available.

### DIFF
--- a/utils/common/src/main/java/org/apache/brooklyn/util/stream/LoggingOutputStream.java
+++ b/utils/common/src/main/java/org/apache/brooklyn/util/stream/LoggingOutputStream.java
@@ -21,12 +21,17 @@ package org.apache.brooklyn.util.stream;
 import java.io.FilterOutputStream;
 import java.io.IOException;
 import java.io.OutputStream;
+import java.nio.charset.StandardCharsets;
 import java.util.concurrent.atomic.AtomicBoolean;
 
+import org.apache.http.util.ByteArrayBuffer;
 import org.slf4j.Logger;
 
 /**
  * Wraps another output stream, intercepting the writes to log it.
+ * 
+ * This is <em>not</em> thread safe. It assumes that calls like write, flush and close 
+ * will be done by the same thread.
  */
 public class LoggingOutputStream extends FilterOutputStream {
 
@@ -68,7 +73,9 @@ public class LoggingOutputStream extends FilterOutputStream {
     protected final Logger log;
     protected final String logPrefix;
     private final AtomicBoolean running = new AtomicBoolean(true);
-    private final StringBuilder lineSoFar = new StringBuilder(16);
+    
+    // Uses byte array, rather than StringBuilder, to handle Unicode chars longer than one byte
+    private ByteArrayBuffer lineSoFar = new ByteArrayBuffer(16);
 
     private LoggingOutputStream(Builder builder) {
         super(builder.out != null ? builder.out : NOOP_OUTPUT_STREAM);
@@ -78,7 +85,7 @@ public class LoggingOutputStream extends FilterOutputStream {
 
     @Override
     public void write(int b) throws IOException {
-        if (running.get() && b >= 0) onChar(b);
+        if (running.get()) onChar(b);
         out.write(b);
     }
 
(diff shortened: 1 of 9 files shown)


neykov:
retest this please

utils/common/src/main/java/org/apache/brooklyn/util/stream/LoggingOutputStream.java
```diff
@@ -0,0 +1,134 @@
...
+
+    private LoggingOutputStream(Builder builder) {
+        super(builder.out != null ? builder.out : NOOP_OUTPUT_STREAM);
+        log = builder.log;
+        logPrefix = (builder.logPrefix != null) ? builder.logPrefix : "";
+      }
+
+    @Override
+    public void write(int b) throws IOException {
+        if (running.get() && b >= 0) onChar(b);
```
geomacy: the `b >= 0` here, plus the use of `StringBuilder` and `lineSoFar.append((char)c);` in `onChar`, will make things work wrongly for unicode, e.g. 

```
    @Test
    public void testLogsUnicode() throws Exception {
        LoggingOutputStream out = LoggingOutputStream.builder().logger(mockLogger).build();
        String test = "Лорем.";
        out.write("Лорем.\n".getBytes(StandardCharsets.UTF_8));
        out.flush();

        assertEquals(logs, ImmutableList.of(test));
    }
```
fails with `java.lang.AssertionError: Lists differ at element [0]: Лорем. != . expected [Лорем.] but found [.]
`.

We could drop the `b >= 0` and use a list of bytes for `lineSoFar` in order to support Unicode if you felt that was valuable.

neykov: Nice catch

utils/common/src/main/java/org/apache/brooklyn/util/stream/LoggingOutputStream.java
```diff
@@ -0,0 +1,134 @@
...
+        out.write(b);
+    }
+
+    @Override
+    public void flush() throws IOException {
+        try {
+            if (lineSoFar.length() > 0) {
+                onLine(lineSoFar.toString());
+                lineSoFar.setLength(0);
```
neykov: This could potentially leak a big buffer - when logging a very long line once, `StringBuffer` will keep the allocated memory.

aledsage: Is your concern about leaking that the `lineSoFar` might now have a very large internal array in `StringBuilder.value`, which won't be GC'ed until the `LoggingOutputStream` is GC'ed?

I could add a call to `StringBuilder.trimToSize()` as well (which would set it to size 0, rather than our initial size of 16). That would be less efficient (causing array-copy to grow again), but probably not noticeably so. Maybe I'll compromise and say that if the size was `>= 1024` then I'll `trimToSize`.

Note that this code was copy-pasted from `StreamGobbler`.

aledsage:
@neykov @geomacy Do you think I need to worry about multi-threaded calls to this new `LoggingOutputStream` stream's `write`, `flush` and `close` methods?!

utils/common/src/main/java/org/apache/brooklyn/util/stream/LoggingOutputStream.java
neykov: > Is your concern about leaking that the lineSoFar might now have a very large internal array in StringBuilder.value, which won't be GC'ed until the LoggingOutputStream is GC'ed?

Yes, that's what I meant.

> Maybe I'll compromise and say that if the size was >= 1024 then I'll trimToSize.

Sounds like a good balance. Alternatively could re-create the `StringBuilder` object with size 16 if you feel that's a better metric.

neykov:
@aledsage I'm fine with not updating the code to handle multi-threaded calls. Would be nice if you mark it explicitly as not supported it in the docs. It's been used just fine as is and we'll keep using it in the same context as before where it's single threaded.

aledsage:
@neykov @geomacy thanks for comments - I've added another commit to address those. Could you review again please?

geomacy:
Failure unrelated?

Retest this please.

PR #732, opened 2017-06-12
Adds brooklyn-container-service code

grkvlt:
See [IP clearance details](http://incubator.apache.org/ip-clearance/brooklyn-container-service.html) for further details.


neykov:
@grkvlt Instead of moving the resources in a separate commit shouldn't we do some history surgery, similar to the split-repo work, where the files will appear as always been at the current location?

grkvlt:
@neykov I didn't move things like that, but remember there is no shared history for the commits, which is why things look like this. It's the best we can do...

neykov:
Makes sense @grkvlt.

Tried to deploy to kubernetes, here are my findings so far:
  * classic Brooklyn - need to add the new jar as a dependency to the `brooklyn-all` module so it's included in the distribution
  * Karaf - need to add the new bundle plus its dependencies (kubernetes-client, openshift-client, okhttp,  etc) to `feature.xml`.

drigodwin:
I've tested this connecting to a docker swarm, k8s and openshift endpoint. I was able to connect and deploy things on each so this all looks good to me. I'm going to merge now. Thankyou very much @grkvlt and all the other contributors.

PR #733, opened 2017-06-13
Makes core-tests.jar installable in the catalog, add metadata

neykov:
Fixes CatalogDtoTest.testCatalogLookup which installs the jar even though it's not a proper osgi bundle.

https://builds.apache.org/view/B/view/Brooklyn/job/brooklyn-integration-tests/73/testngreports/org.apache.brooklyn.core.catalog.internal/CatalogDtoTest/testCatalogLookup/


drigodwin:
When I run this I get `Missing bundle symbolic name in BOM or MANIFEST`, am I doing something wrong @neykov?

neykov:
@drigodwin I think the tests relies on a previously build `brooklyn-core-tests.jar`. Try building once with tests disabled and do a separate run of the test.

PR #734, opened 2017-06-16
Support ConfigKey deprecated names

aledsage:
As discussed on the dev@brooklyn mailing list in the thread "[PROPOSAL] Deprecate @SetFromFlag".

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/config/BasicConfigKey.java
+++ b/core/src/main/java/org/apache/brooklyn/core/config/BasicConfigKey.java
@@ -47,6 +47,7 @@ import com.google.common.base.Preconditions;
 import com.google.common.base.Predicate;
 import com.google.common.base.Predicates;
 import com.google.common.base.Splitter;
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Lists;
 import com.google.common.reflect.TypeToken;
 
@@ -100,6 +101,7 @@ public class BasicConfigKey<T> implements ConfigKeySelfExtracting<T>, Serializab
 
     public abstract static class Builder<T, B extends Builder<T,B>> {
         protected String name;
+        protected Collection<String> deprecatedNames = ImmutableList.of();
         protected TypeToken<T> type;
         protected String description;
         protected T defaultValue;
@@ -125,6 +127,7 @@ public class BasicConfigKey<T> implements ConfigKeySelfExtracting<T>, Serializab
         public Builder(String newName, ConfigKey<T> key) {
             this.type = checkNotNull(key.getTypeToken(), "type");
             this.name = checkNotNull(newName, "name");
+            this.deprecatedNames = checkNotNull(key.getDeprecatedNames(), "deprecatedNames");
             description(key.getDescription());
             defaultValue(key.getDefaultValue());
             reconfigurable(key.isReconfigurable());
@@ -134,6 +137,12 @@ public class BasicConfigKey<T> implements ConfigKeySelfExtracting<T>, Serializab
         }
         public B name(String val) {
             this.name = val; return self();
+        }
+        public B deprecatedNames(Collection<String> val) {
+            this.deprecatedNames = val; return self();
+        }
+        public B deprecatedNames(String... val) {
+            return deprecatedNames(val == null ? ImmutableList.of() : ImmutableList.copyOf(val));
         }
         public B type(Class<T> val) {
             this.type = TypeToken.of(val); return self();
@@ -187,6 +196,7 @@ public class BasicConfigKey<T> implements ConfigKeySelfExtracting<T>, Serializab
     }
     
     protected String name;
+    protected Collection<String> deprecatedNames;
     protected TypeToken<T> typeToken;
     protected Class<? super T> type;
     protected String description;
@@ -237,6 +247,7 @@ public class BasicConfigKey<T> implements ConfigKeySelfExtracting<T>, Serializab
     public BasicConfigKey(TypeToken<T> type, String name, String description, T defaultValue) {
         this.description = description;
         this.name = checkNotNull(name, "name");
+        this.deprecatedNames = ImmutableList.of();
         
         this.type = TypeTokens.getRawTypeIfRaw(checkNotNull(type, "type"));
         this.typeToken = TypeTokens.getTypeTokenIfNotRaw(type);
@@ -248,6 +259,7 @@ public class BasicConfigKey<T> implements ConfigKeySelfExtracting<T>, Serializab
 
     public BasicConfigKey(Builder<T,?> builder) {
         this.name = checkNotNull(builder.name, "name");
+        this.deprecatedNames = checkNotNull(builder.deprecatedNames, "deprecatedNames");
         this.type = TypeTokens.getRawTypeIfRaw(checkNotNull(builder.type, "type"));
         this.typeToken = TypeTokens.getTypeTokenIfNotRaw(builder.type);
         this.description = builder.description;
@@ -263,6 +275,13 @@ public class BasicConfigKey<T> implements ConfigKeySelfExtracting<T>, Serializab
 
     /** @see ConfigKey#getName() */
     @Override public String getName() { return name; }
+
+    /** @see ConfigKey#getDeprecatedNames() */
+    @Override public Collection<String> getDeprecatedNames() {
+        // check for null, for backwards compatibility of serialized state
+        if (deprecatedNames == null) deprecatedNames = ImmutableList.of();
+        return deprecatedNames;
+    }
 
     /** @see ConfigKey#getTypeName() */
     @Override public String getTypeName() { return getType().getName(); }
@@ -365,6 +384,7 @@ public class BasicConfigKey<T> implements ConfigKeySelfExtracting<T>, Serializab
     }
 
     /** @see ConfigKey#getNameParts() */
+    @Deprecated
     @Override public Collection<String> getNameParts() {
         return Lists.newArrayList(dots.split(name));
     }
--- a/core/src/main/java/org/apache/brooklyn/core/entity/AbstractEntity.java
+++ b/core/src/main/java/org/apache/brooklyn/core/entity/AbstractEntity.java
@@ -64,6 +64,7 @@ import org.apache.brooklyn.core.config.ConfigConstraints;
 import org.apache.brooklyn.core.config.internal.AbstractConfigMapImpl;
 import org.apache.brooklyn.core.config.render.RendererHints;
 import org.apache.brooklyn.core.enricher.AbstractEnricher;
+import org.apache.brooklyn.core.entity.internal.ConfigUtilsInternal;
 import org.apache.brooklyn.core.entity.internal.EntityConfigMap;
 import org.apache.brooklyn.core.entity.lifecycle.PolicyDescriptor;
 import org.apache.brooklyn.core.entity.lifecycle.ServiceStateLogic;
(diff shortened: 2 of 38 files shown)


core/src/main/java/org/apache/brooklyn/core/objs/AbstractEntityAdjunct.java
```diff
@@ -151,6 +153,19 @@ public AbstractEntityAdjunct configure(Map flags) {
...
+        // Aled thinks it would be sensible to remove the consumed flags below (i.e. flags = ...).
+        // However, that causes PolicyConfigTest.testConfigFlagsPassedInAtConstructionIsAvailable to fail.
+        // However, tht looks mad - do we really need to support it?!
+        // The policy is defined with one key using a name in SetFromFlag, and another key using the same name.
+        // It expects both of the config keys to have been set.
+        //     @SetFromFlag("strKey")
+        //     public static final ConfigKey<String> STR_KEY = new BasicConfigKey<String>(String.class, "akey", "a key");
+        //     public static final ConfigKey<String> STR_KEY_WITH_DEFAULT = new BasicConfigKey<String>(String.class, "strKey", "str key", "str key default");
+        // I've preserved that behaviour (for now).
+        ConfigUtilsInternal.setAllConfigKeys(flags, getAdjunctType().getConfigKeys(), this);
```
ahgittin: agree that behaviour is way overkill.  if it's not too hard to log a warning wherever we rely on it, however, that would be ideal.  if that's not easy then fine to abandon.

core/src/test/java/org/apache/brooklyn/core/config/ConfigKeyDeprecationTest.java
```diff
@@ -0,0 +1,328 @@
...
+        EntityInternal entity = app.addChild(EntitySpec.create(MyBaseEntity.class)
+                .configure("superKey1", "myval")
+                .configure("oldSuperKey1", "mywrongval"));
+        assertEquals(entity.config().get(MyBaseEntity.SUPER_KEY_1), "myval");
+    }
+    
+    @Test
+    public void testPrefersFirstDeprecatedNameIfMultiple() throws Exception {
+        EntityInternal entity = app.addChild(EntitySpec.create(MyBaseEntity.class)
+                .configure("oldSuperKey1", "myval1")
```
geomacy: Maybe nicer to order these as 
```
.configure("oldSuperKey1b", "myval2")
.configure("oldSuperKey1", "myval1")
```
to demonstrate that the order in which they are configured here is not the determining factor; maybe even call `oldSuperKey1` something instead that is lexically greater than `oldSuperKey2` to demonstrate that it's not alphabetic either!

core/src/test/java/org/apache/brooklyn/core/config/ConfigKeyDeprecationTest.java
```diff
@@ -0,0 +1,328 @@
...
+    // any existing value that was set using the real config key name. I think this is because 
+    // EntityDynamicType.addConfigKey() is not called when KEY_1 is first set, so it doesn't have
+    // access to the deprecated names on subsequent calls to set(String, val). It therefore stores
+    // in EntityConfigMap (backed by a ConfigBag) the original key-value and the deprecated key-value.
+    // When we subsequently call get(key), it retrieved the original key-value.
+    //
+    // Contrast this with EntityDynamicType.addSensorIfAbsent().
+    //
+     * However, it's (probably) not straight forward to just add and call a addConfigKeyIfAbsent. This
+     * is because config().set() is used for dynamic config declard in yaml, which the entity doesn't
```
geomacy: typo `declard`

core/src/main/java/org/apache/brooklyn/core/config/BasicConfigKey.java
```diff
@@ -365,6 +384,7 @@ public boolean isValueValid(T value) {
     }
 
     /** @see ConfigKey#getNameParts() */
+    @Deprecated
```
aledsage: That's included in the super-type `ConfigKey.getNameParts()`, which says `@deprecated since 0.12.0; use {@link #getName()}`

core/src/main/java/org/apache/brooklyn/util/core/config/ConfigBag.java
```diff
@@ -507,6 +525,44 @@ protected Object getStringKey(String key, boolean markUsed) {
...
+            if (deprecatedVal.isPresent()) {
+                if (val.isPresent()) {
+                    if (!Objects.equal(val.get(), deprecatedVal.get())) {
+                        log.warn("Conflicting value for key "+key+" from deprecated name '"+deprecatedName+"'; "
+                                + "using value from preferred name "+key.getName());
+                    } else {
+                        log.warn("Duplicate value for key "+key+" from deprecated name '"+deprecatedName+"'; "
+                                + "using same value from preferred name "+key.getName());
+                    }
+                } else if (firstDeprecatedVal.isPresent()) {
```
aledsage: Good spot! I've added a failing test, and then fixed the NPE.

core/src/main/java/org/apache/brooklyn/util/core/config/ConfigBag.java
```diff
@@ -518,6 +574,46 @@ protected Object getStringKey(String key, boolean markUsed) {
     /**
      * @return Unresolved configuration value. May be overridden to provide resolution - @see {@link ResolvingConfigBag#getStringKeyMaybe(String, boolean)}
      */
+    protected synchronized Maybe<Object> getKeyMaybe(ConfigKey<?> key, boolean markUsed) {
```
aledsage: Yeah, not sure how to remove that without an almost-equal amount of code! The difference is it needs to call a `getRawStringKeyMaybe(name, markUsed)` or `getStringKeyMaybe(name, markUsed)`. Thinking of how we could achieve that:
* We could pass another boolean to say whether raw is desired. But I don't really like that pattern.
* There's not a nice java interface for passing two parameters, so we'd end up declaring our own interface to pass the different getter impl.

Do you think either of those is worth it @geomacy?

core/src/main/java/org/apache/brooklyn/core/location/AbstractLocation.java
```diff
@@ -215,6 +216,16 @@ public AbstractLocation configure(Map<?,?> properties) {
             config().removeKey(PARENT_LOCATION);
         }
 
+        // allow config keys to be set by name (or deprecated name)
+        //
+        // Aled thinks it would be sensible to remove the consumed flags below (i.e. properties = ...).
+        // However, that caused ClockerDynamicLocationPatternTest to fail because there is a field of 
+        // StubContainerLocation annotated with `@SetFromFlag("owner")`, as well as a config key with
```
aledsage: Not certain - I deleted the test that used it in #728. My concern is whether that test highlighted that we're supposed to be supporting such madness!

@ahgittin's comment was: "if it's not too hard to log a warning wherever we rely on it, however, that would be ideal. if that's not easy then fine to abandon.".

It's probably tricky to log.warn on it reliably (because `FlagUtils.setFieldsFromFlagsWithBag` and `FlagUtils.setAllConfigKeys` also does logic for setting config, so we can't easily tell how it's using the key).

We could try adding some new code, e.g. in `EntityDynamicType`, to detect it, and warn (or even fail) if we see the name being used in multiple places. I suggest we do that separately from this PR, and I suggest that we fail-fast.

I'd be happy with abandoning - i.e. change it to what I proposed in the comment, to not pass the used properties to `FlagUtils.setFieldsFromFlagsWithBag` or `FlagUtils.setAllConfigKeys`.

I'll go with abandoning, unless anyone feels strongly to the contrary?

core/src/test/java/org/apache/brooklyn/core/config/ConfigKeyDeprecationRebindTest.java
```diff
@@ -0,0 +1,373 @@
...
+        
+        Entity newApp = mgmt().getEntityManager().getEntity(appId);
+        assertEquals(newApp.config().get(MyApp.KEY_2), "myval");
+        
+        // Expect the persisted state to have been re-written with the new key value.
+        switchOriginalToNewManagementContext();
+        rebind();
+        
+        String allLines = getPersistanceFileContents(BrooklynObjectType.ENTITY, appId);
+        assertFalse(allLines.contains("oldFlagKey2"), "should not contain 'oldFlagKey2', allLines="+allLines);
```
aledsage: Yeah, it's a bit of a weird test. It was sensible to decide to write it - I generated the persisted state (as per the javadoc) with:
```
Entity app = mgmt().getEntityManager().createEntity(EntitySpec.create(MyApp.class)
        .configure("oldFlagKey2", "myval"));
```

That turns out not to contain `oldFlagKey2` because it was transformed at the time of writing the persisted state (even before any changes in this PR).

I think it's important that we record that this upgrade-path works. I could delete the test, and replace it with just a comment in the test class to say why we're not explicitly testing this scenario.

Or I could leave in the test, and add to the javadoc an explanation of why it's weird.

Thoughts @geomacy?

aledsage:
@geomacy comments addressed (except where I responded explicitly to your comment). I've added two commits on top of the original - could you review those please?

geomacy:
hi @aledsage sorry missed the update above last week, will look at that presto.

core/src/main/java/org/apache/brooklyn/core/entity/AbstractEntity.java
```diff
@@ -396,6 +396,9 @@ public AbstractEntity configure(Map flags) {
         }
 
         // allow config keys to be set by name (or deprecated name)
+        //
+        // The resulting `flags` will no longer contain the keys that we matched;
+        // we will not also use them to for `SetFromFlag` etc.
```
geomacy: typo "to for"

core/src/main/java/org/apache/brooklyn/core/location/AbstractLocation.java
```diff
@@ -216,17 +216,14 @@ public AbstractLocation configure(Map<?,?> properties) {
...
-        // name "owner" (and with `@SetFromFlag("owner")`) in the super-type (DynamicLocation).
-        // However, that looks mad - do we really need to support it?!
-        // I've preserved that behaviour (for now).
-        ConfigUtilsInternal.setAllConfigKeys(properties, getLocationTypeInternal().getConfigKeys().values(), this);
+        // The resulting `properties` will no longer contain the keys that we matched;
+        // we will not also use them to for `SetFromFlag` etc.
+        properties = ConfigUtilsInternal.setAllConfigKeys(properties, getLocationTypeInternal().getConfigKeys().values(), this);
 
         // NB: flag-setting done here must also be done in BasicLocationRebindSupport
+        // Must call setFieldsFromFlagsWithBag, even if properites is empty, so defaults are extracted from annotations
```
geomacy: typo `properites`

core/src/main/java/org/apache/brooklyn/core/objs/AbstractEntityAdjunct.java
```diff
@@ -153,19 +153,13 @@ public AbstractEntityAdjunct configure(Map flags) {
...
-        //     @SetFromFlag("strKey")
-        //     public static final ConfigKey<String> STR_KEY = new BasicConfigKey<String>(String.class, "akey", "a key");
-        //     public static final ConfigKey<String> STR_KEY_WITH_DEFAULT = new BasicConfigKey<String>(String.class, "strKey", "str key", "str key default");
-        // I've preserved that behaviour (for now).
-        ConfigUtilsInternal.setAllConfigKeys(flags, getAdjunctType().getConfigKeys(), this);
+        // The resulting `flags` will no longer contain the keys that we matched;
+        // we will not also use them to for `SetFromFlag` etc.
+        flags = ConfigUtilsInternal.setAllConfigKeys(flags, getAdjunctType().getConfigKeys(), this);
 
+        // Must call setFieldsFromFlagsWithBag, even if properites is empty, so defaults are extracted from annotations
```
geomacy: shouldn't this read `// Must call setFieldsFromFlags,` as per 164 below?

geomacy:
Going to merge this, should have done so before now, sorry! Any further changes can be done as separate PRs.

PR #737, opened 2017-06-16
Clean up inference around parsing `name:version` strings

ahgittin:
We've had some rough heuristics how to find name and version but not a strict definition.

This adds strict definitions in `RegisteredTypeNaming` (first commit).

It then changes core to prefer these methods.  It doesn't change behaviour (except in a few fringe cases noted), but it does add log warnings where we were relying on hokey old strategies (previously assumed something was a version if it started with a number, that's it).

Then clean up a bunch of deprecated stuff where the above was used.

The methods in `RegisteredTypeNaming` let us hedge a little bit whether we want to be strict with OSGi versions or not.  That's a subject for another PR (but an important one as discussed at #672).


ahgittin:
(There is a failing test, I'm working on that, otherwise this is good to review.)

ahgittin:
@geomacy @neykov ready for review.  does not do the version changes discussed on ML but prepares us for them (and helped me understand them).  this is mainly to fix issues where people have used funny `v1` version strings.

ahgittin:
(the test failure is a test env config issue in hudson remoting)

ahgittin:
re tests - see #740 which merges this and test pass

geomacy:
@ahgittin haven't ignored this but haven't been able to find the bandwidth to look at it - will be off tomorrow, but will look at this and related PRs on Monday!

core/src/main/java/org/apache/brooklyn/core/catalog/internal/BasicBrooklynCatalog.java
```diff
@@ -855,24 +893,50 @@ private boolean attemptType(String key, CatalogItemType candidateCiType) {
...
+                        catalogItemType = candidateCiType;
+                        planYaml = candidateYaml;
+                        resolved = true;
+                        return true;
+                    }
+                }
+            }
+            {
+                // legacy routine; should be the same as above code added in 0.12 because:
+                // if type is symbolic_name, the type will match below, and version will be null so any version allowed to match
```
geomacy: I'm afraid I don't follow this - is it the case that a type which is just a `symbolic_name` will be taken care of by line 900 above (so you actually mean "the type will match above" on this line i.e. at line 900); but how is the next case handled, where type is of form `symbolic_name:version`?   When you say `the id will match`, where does that happen?

geomacy:
Would also be interested in @neykov's comments on this!

neykov:
@geomacy left my comments in https://github.com/apache/brooklyn-server/pull/740 which includes this PR.

geomacy:
👍  I was going to look at that next :-)

core/src/main/java/org/apache/brooklyn/core/catalog/internal/BasicBrooklynCatalog.java
```diff
@@ -601,14 +602,34 @@ private void collectCatalogItems(String sourceYaml, Map<?,?> itemMetadata, List<
         // if symname not set, infer from: id, then name, then item id, then item name
         if (Strings.isBlank(symbolicName)) {
```
ahgittin: want to completely gut this class as soon as #363 YOML is available, and remove much of this once 0.12.0 is out -- refactoring i think would make that harder as well as of course not being worth the effort if we're gutting it

core/src/main/java/org/apache/brooklyn/core/catalog/internal/BasicBrooklynCatalog.java
ahgittin: yes, should be `above`

ahgittin:
great comments here @geomacy , addressed all except where commented above; moving on to #740

core/src/main/java/org/apache/brooklyn/core/catalog/internal/BasicBrooklynCatalog.java
geomacy: grand, but how about the question above where type is of form symbolic_name:version? When you say "the id will match", where does that happen?

PR #738, opened 2017-06-19
Delete deprecated ApplicationBuilder

aledsage:
Required updating lots of tests, often to extend `BrooklynAppLiveTestSupport` or `BrooklynAppUnitTestSupport`.

Please first review/merge (to avoid those other repos failing when we delete `ApplicationBuilder`):
* https://github.com/apache/brooklyn-library/pull/118
* https://github.com/apache/brooklyn-dist/pull/96


ahgittin:
big job, very nice to have.  gotta love this:

> +299 −1,329

looks good and downstream changes are merged, so merging

ahgittin:
Something seems to have sped up tests significantly (1 to 2 mins) in the past week or two.  I'm blaming this PR!  In any case -- I'm happy (though slightly curious @aledsage if we can't explain the speedup there is a chance it's a sign we've disabled things we want!)

aledsage:
@ahgittin changing a bunch of tests to extend `BrooklynAppUnitTestSupport` might well have sped things up (before the change, some of their `setUp` method created a normal `LocalManagementContext` rather than the minimalist `LocalManagementContextForTests.newInstance()`; the former would include the embedded Felix etc).

PR #739, opened 2017-06-20
A default network is now created when deploying to Azure ARM

Graeme-Miller:
In an upcoming release of JClouds, support for Azure ARM will be added. The JClouds code for Azure ARM will place each created VM into a network based on it's group ID. As we do not use the concept of a group ID in Brooklyn, this means that each VM deployed by Brooklyn is deployed in it's own subnet. 

For most other clouds, a sensible default network is provided by the cloud. As such a Brooklyn user can be assured that if he deploys a blueprint with multiple VMs, they can reach each other on the private IPs that are provided. Unless we create a default network in Azure, those blueprints will not be portable and a user will have to manually create a network first. I believe that it is confusing for a user to have the clouds act differently on networking, which is an essential part of Brooklyn.

A few things to mention:
*) The default network/subnet is only created when needed
*) The default network/subnet is created per region
*) It is never cleaned up or deleted

--------------------------------------------------------------------
As a side note, I discovered this as AMP, which is a downstream project of Brooklyn, pulled in and was using Azure ARM support from JClouds. An older version of the JClouds code created the default network, but was recently removed meaning a regression in AMP. The code which I have added to create a default network is based on the JClouds code.

### DIFF
--- a/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/creator/DefaultAzureArmNetworkCreator.java
+++ b/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/creator/DefaultAzureArmNetworkCreator.java
@@ -37,6 +37,9 @@ import org.jclouds.azurecompute.arm.domain.Subnet;
 import org.jclouds.azurecompute.arm.domain.VirtualNetwork;
 import org.jclouds.compute.ComputeService;
 
+import org.apache.brooklyn.config.ConfigKey;
+import org.apache.brooklyn.core.config.ConfigKeys;
+import org.apache.brooklyn.util.collections.MutableMap;
 import org.apache.brooklyn.util.core.config.ConfigBag;
 
 public class DefaultAzureArmNetworkCreator {
@@ -47,23 +50,35 @@ public class DefaultAzureArmNetworkCreator {
     private static final String DEFAULT_NETWORK_NAME = "brooklyn-default-network";
     private static final String DEFAULT_SUBNET_NAME = "brooklyn-default-subnet";
 
-    private static final String defaultVnetAddressPrefix = "10.1.0.0/16";
-    private static final String defaultSubnetAddressPrefix = "10.1.0.0/24";
+    private static final String DEFAULT_VNET_ADDRESS_PREFIX = "10.1.0.0/16";
+    private static final String DEFAULT_SUBNET_ADDRESS_PREFIX = "10.1.0.0/24";
 
-    public static void createDefaultNetworkAndAddToTemplateOptions(ComputeService computeService, ConfigBag config) {
+    public static ConfigKey<Boolean> AZURE_ARM_DEFAULT_NETWORK_ENABLED = ConfigKeys.newBooleanConfigKey(
+            "azure.arm.default.network.enabled",
+            "When set to true, AMP will create a default network and subnet per Azure region and " +
+                    "deploy applications there (if no network configuration has been set for the application).",
+            true);
+
+    public static void createDefaultNetworkAndAddToTemplateOptionsIfRequired(ComputeService computeService, ConfigBag config) {
+        if (!config.get(AZURE_ARM_DEFAULT_NETWORK_ENABLED)) {
+            LOG.info("azure.arm.default.network.enabled is disabled, not creating default network");
+            return;
+        }
+
+
         Map<String, Object> templateOptions = config.get(TEMPLATE_OPTIONS);
 
         //Only create a default network if we haven't specified a network name (in template options or config) or ip options
         if (config.containsKey(NETWORK_NAME)) {
-            LOG.info("Network config specified when creating Azure location. Not creating default network");
+            LOG.info("Network config specified when provisioning Azure machine. Not creating default network");
             return;
         }
         if (templateOptions != null && (templateOptions.containsKey(NETWORK_NAME.getName()) || templateOptions.containsKey("ipOptions"))) {
-            LOG.info("Network config specified when creating Azure location. Not creating default network");
+            LOG.info("Network config specified when provisioning Azure machine. Not creating default network");
             return;
         }
 
-        LOG.info("Network config not specified when creating Azure location. Creating default network if doesn't exist");
+        LOG.info("Network config not specified when provisioning Azure machine. Creating default network if doesn't exist");
 
         AzureComputeApi api = computeService.getContext().unwrapApi(AzureComputeApi.class);
         String location = config.get(CLOUD_REGION_ID);
@@ -87,10 +102,10 @@ public class DefaultAzureArmNetworkCreator {
 
         //Setup properties for creating subnet/network
         Subnet subnet = Subnet.create(subnetName, null, null,
-                Subnet.SubnetProperties.builder().addressPrefix(defaultSubnetAddressPrefix).build());
+                Subnet.SubnetProperties.builder().addressPrefix(DEFAULT_SUBNET_ADDRESS_PREFIX).build());
 
         VirtualNetwork.VirtualNetworkProperties virtualNetworkProperties = VirtualNetwork.VirtualNetworkProperties
-                .builder().addressSpace(VirtualNetwork.AddressSpace.create(Arrays.asList(defaultVnetAddressPrefix)))
+                .builder().addressSpace(VirtualNetwork.AddressSpace.create(Arrays.asList(DEFAULT_VNET_ADDRESS_PREFIX)))
                 .subnets(Arrays.asList(subnet)).build();
 
         //Create network
@@ -103,19 +118,22 @@ public class DefaultAzureArmNetworkCreator {
     }
 
     private static void updateTemplateOptions(ConfigBag config, Subnet createdSubnet){
-        Map<String, Object> templateOptions = config.get(TEMPLATE_OPTIONS);
+        Map<String, Object> templateOptions;
 
-        if(templateOptions == null) {
+        if(config.containsKey(TEMPLATE_OPTIONS)) {
+            templateOptions = MutableMap.copyOf(config.get(TEMPLATE_OPTIONS));
+        } else {
             templateOptions = new HashMap<>();
-            config.put(TEMPLATE_OPTIONS, templateOptions);
         }
 
-        Map<String, Object> ipOptions = new HashMap<>();
-        ipOptions.put("allocateNewPublicIp", true); //JClouds will not provide a public IP unless we set this
-        ipOptions.put("subnet", createdSubnet.id());
-        templateOptions.put("ipOptions", ipOptions);
-    }
+        templateOptions.put("ipOptions", ImmutableMap.of(
+                "allocateNewPublicIp", true, //jclouds will not provide a public IP unless we set this
+                "subnet", createdSubnet.id()
+        ));
 
+        config.put(TEMPLATE_OPTIONS, templateOptions);
+
+    }
 
     private static void createResourceGroupIfNeeded(AzureComputeApi api, String resourceGroup, String location) {
         LOG.debug("using resource group [%s]", resourceGroup);
--- a/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/creator/DefaultAzureArmNetworkCreator.java
+++ b/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/creator/DefaultAzureArmNetworkCreator.java
@@ -1,0 +1,129 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.brooklyn.location.jclouds.networking.creator;
+
+import static org.apache.brooklyn.core.location.cloud.CloudLocationConfig.CLOUD_REGION_ID;
+import static org.apache.brooklyn.location.jclouds.api.JcloudsLocationConfigPublic.NETWORK_NAME;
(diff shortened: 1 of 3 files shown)


drigodwin:
retest this please

Graeme-Miller:
The failure was due to my test class having an un-approve licence. I have updated the class and pushed- this should force a retest

locations/jclouds/src/test/java/org/apache/brooklyn/location/jclouds/networking/creator/DefaultAzureArmNetworkCreatorTest.java
```diff
@@ -0,0 +1,184 @@
...
+    final String TEST_SUBNET_NAME = "brooklyn-default-subnet-test-loc";
+    final String TEST_SUBNET_ID = "/test/resource/id";
+    final String TEST_LOCATION = "test-loc";
+
+    @BeforeMethod
+    public void setUp() {
+        MockitoAnnotations.initMocks(this);
+
+//
+//        when(azureComputeApi.getVirtualNetworkApi(TEST_RESOURCE_GROUP)).thenReturn(virtualNetworkApi);
```
drigodwin: Should this be removed if it's commented?

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/creator/DefaultAzureArmNetworkCreator.java
```diff
@@ -0,0 +1,129 @@
...
+            updateTemplateOptions(config, preexistingSubnet);
+            return;
+        }
+
+
+        LOG.info("Network config not specified when creating Azure location and default network/subnet does not exists. Creating");
+
+        createResourceGroupIfNeeded(api, resourceGroupName, location);
+
+        //Setup properties for creating subnet/networl
```
drigodwin: subnet/networl -> subnet/network

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/creator/DefaultAzureArmNetworkCreator.java
```diff
@@ -0,0 +1,129 @@
...
+    public static final Logger LOG = LoggerFactory.getLogger(DefaultAzureArmNetworkCreator.class);
+
+    private static final String DEFAULT_RESOURCE_GROUP = "brooklyn-default-resource-group";
+    private static final String DEFAULT_NETWORK_NAME = "brooklyn-default-network";
+    private static final String DEFAULT_SUBNET_NAME = "brooklyn-default-subnet";
+
+    private static final String defaultVnetAddressPrefix = "10.1.0.0/16";
+    private static final String defaultSubnetAddressPrefix = "10.1.0.0/24";
+
+    public static void createDefaultNetworkAndAddToTemplateOptions(ComputeService computeService, ConfigBag config) {
```
aledsage: I'd split this into two separate methods: `isDefaultNetworkRequired(ConfigBag)` and `createDefaultNetworkAndAddToTemplateOptions(ComputeService, ConfigBag)`.

I think it's a little confusing (given the method name) that it is a no-op if certain config options are set explicitly.

First thing in the `createDefaultNetworkAndAddToTemplateOptions`, I'd then do if (!isDefaultNetworkRequired(config)) throw new IllegalArgumentException(...)`.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/creator/DefaultAzureArmNetworkCreator.java
```diff
@@ -0,0 +1,129 @@
...
+        Map<String, Object> templateOptions = config.get(TEMPLATE_OPTIONS);
+
+        //Only create a default network if we haven't specified a network name (in template options or config) or ip options
+        if (config.containsKey(NETWORK_NAME)) {
+            LOG.info("Network config specified when creating Azure location. Not creating default network");
```
aledsage: Instead of "creating Azure location", I'd say "provisioning Azure machine".

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/creator/DefaultAzureArmNetworkCreator.java
```diff
@@ -0,0 +1,129 @@
...
+
+public class DefaultAzureArmNetworkCreator {
+
+    public static final Logger LOG = LoggerFactory.getLogger(DefaultAzureArmNetworkCreator.class);
+
+    private static final String DEFAULT_RESOURCE_GROUP = "brooklyn-default-resource-group";
+    private static final String DEFAULT_NETWORK_NAME = "brooklyn-default-network";
+    private static final String DEFAULT_SUBNET_NAME = "brooklyn-default-subnet";
+
+    private static final String defaultVnetAddressPrefix = "10.1.0.0/16";
```
aledsage: I'd prefer all caps for constants.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/creator/DefaultAzureArmNetworkCreator.java
```diff
@@ -0,0 +1,129 @@
...
+            return;
+        }
+
+
+        LOG.info("Network config not specified when creating Azure location and default network/subnet does not exists. Creating");
+
+        createResourceGroupIfNeeded(api, resourceGroupName, location);
+
+        //Setup properties for creating subnet/network
+        Subnet subnet = Subnet.create(subnetName, null, null,
```
aledsage: I wondered whether we could move this chunk into a method like `Subnet createSubnet(api, resourceGroupName, vnetName, subnetName)`. However, having a method taking three strings in a row might well not make things clearer :-(

Reasoning is that decreasing the size of this method would make it a bit clearer.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/creator/DefaultAzureArmNetworkCreator.java
```diff
@@ -0,0 +1,129 @@
...
+        api.getVirtualNetworkApi(resourceGroupName).createOrUpdate(vnetName, location, virtualNetworkProperties);
+        Subnet createdSubnet = api.getSubnetApi(resourceGroupName, vnetName).get(subnetName);
+
+        //Add config
+        updateTemplateOptions(config, createdSubnet);
+
+    }
+
+    private static void updateTemplateOptions(ConfigBag config, Subnet createdSubnet){
+        Map<String, Object> templateOptions = config.get(TEMPLATE_OPTIONS);
```
aledsage: Don't rely on `templateOptions` from the config map being mutable. Instead, make a copy.

I'd write something like:
```
Map<String, Object> templateOptions = MutableMap.copyOf(config.get(TEMPLATE_OPTIONS))
        .put("ipOptions", ImmutableMap.of(
                "subnet", createdSubnet.id(),
                "allocateNewPublicIp", true)) // jclouds will not provide a public IP unless we set this
        .build();
config.put(TEMPLATE_OPTIONS, templateOptions);
```

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/creator/DefaultAzureArmNetworkCreator.java
```diff
@@ -0,0 +1,129 @@
...
+
+        if(templateOptions == null) {
+            templateOptions = new HashMap<>();
+            config.put(TEMPLATE_OPTIONS, templateOptions);
+        }
+
+        Map<String, Object> ipOptions = new HashMap<>();
+        ipOptions.put("allocateNewPublicIp", true); //JClouds will not provide a public IP unless we set this
```
aledsage: Why set `allocateNewPublicIp` if the user hasn't asked for it? Is that definitely the default that we want? I probably agree, as it's more like other clouds and this is just the defaults in brooklyn (when a user isn't configuring the security options they really desire). But I worry that injecting defaults like this will be confusing - e.g. if the user supplies their own network or subnet name, then this value doesn't get set (I don't think that difference is intuitive to the user).

aledsage: Note that jclouds uses all lower-case in their branding

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/creator/DefaultAzureArmNetworkCreator.java
aledsage: I'd be tempted to add another config key that says whether to enable the automatic-default-network stuff or to skip it (where that config key defaults to true). You could declare the config key in this class, and look it up in the `ConfigBag` passed in. For example, `azureArmDefaultNetworkCreationEnabled`. Hopefully use of the word "enabled" makes it clear that setting this config key doesn't mean you'll definitely get a default network/subnet (it still depends on the other config for that vm's provisioning), but that setting it to false will always turn it off.

Graeme-Miller: re separate methods. I would prefer not to, I think it is easier to read if the logic is in one place. How about I change the name to createDefaultNetworkAndAddToTemplateOptionsIfRequired. A bit wordy, but it is clear what it is doing.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/creator/DefaultAzureArmNetworkCreator.java
Graeme-Miller: yeah, I'm not convinced it would make things clearer. Will change though if there are any strong opinions

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/creator/DefaultAzureArmNetworkCreator.java
Graeme-Miller: setting allocateNewPublicIp makes sense to me. When a user doesn't setup any network options in brooklyn, I think they can expect:
a public IP associated with the VM
a private IP on a default network.

Setting this config gives them that. If we don't set this config, they won't be able to access their nodes (in fact, I am not sure brooklyn will be able to access the vm- i'd have to check).

Currently, without these changes, it is already confusing for a user to specify there own network as they need to know to set this option. I don't think the answer is to not set this by default, but rather to make it clearer that this needs to be set.

drigodwin:
I'm going to merge this now, thankyou @Graeme-Miller!

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/creator/DefaultAzureArmNetworkCreator.java
```diff
@@ -0,0 +1,147 @@
...
+
+
+        Map<String, Object> templateOptions = config.get(TEMPLATE_OPTIONS);
+
+        //Only create a default network if we haven't specified a network name (in template options or config) or ip options
+        if (config.containsKey(NETWORK_NAME)) {
+            LOG.info("Network config specified when provisioning Azure machine. Not creating default network");
+            return;
+        }
+        if (templateOptions != null && (templateOptions.containsKey(NETWORK_NAME.getName()) || templateOptions.containsKey("ipOptions"))) {
```
neykov: The `TemplateOptions` property is named `networks`, whereas `NETWORK_NAME.getName()` is `networkName`, so it will miss user-supplied options.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/creator/DefaultAzureArmNetworkCreator.java
```diff
@@ -0,0 +1,147 @@
...
+            LOG.info("Network config specified when provisioning Azure machine. Not creating default network");
+            return;
+        }
+
+        LOG.info("Network config not specified when provisioning Azure machine. Creating default network if doesn't exist");
+
+        AzureComputeApi api = computeService.getContext().unwrapApi(AzureComputeApi.class);
+        String location = config.get(CLOUD_REGION_ID);
```
neykov: Might not be set. When using default this will be `null`.

neykov: Can only get it from  `template.getLocation()` reliably. If you move the default network creation after `template` is created you'll be able to get it. This will also make the above checks easier - all you need to do is look in the `templateOptions` object.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/creator/DefaultAzureArmNetworkCreator.java
```diff
@@ -0,0 +1,147 @@
...
+
+        String resourceGroupName = DEFAULT_RESOURCE_GROUP  + "-" + location;
+        String vnetName = DEFAULT_NETWORK_NAME + "-" + location;
+        String subnetName = DEFAULT_SUBNET_NAME + "-" + location;
+
+        //Check if default already exists
+        Subnet preexistingSubnet = api.getSubnetApi(resourceGroupName, vnetName).get(subnetName);
+        if(preexistingSubnet != null){
+            LOG.info("Default Azure network and subnet already created, "+vnetName);
```
neykov: Could log that are part of the normal workflow at `debug` level.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/creator/DefaultAzureArmNetworkCreator.java
```diff
@@ -0,0 +1,147 @@
...
+
+        LOG.info("Network config not specified when creating Azure location and default network/subnet does not exists. Creating");
+
+        createResourceGroupIfNeeded(api, resourceGroupName, location);
+
+        //Setup properties for creating subnet/network
+        Subnet subnet = Subnet.create(subnetName, null, null,
+                Subnet.SubnetProperties.builder().addressPrefix(DEFAULT_SUBNET_ADDRESS_PREFIX).build());
+
+        VirtualNetwork.VirtualNetworkProperties virtualNetworkProperties = VirtualNetwork.VirtualNetworkProperties
```
neykov: The vnet might exist, but subnet could be deleted for whatever reason. I'd split this in two steps, doing the exists check for each one.

neykov:
The most important from above comments is that `location` (which is region in brooklyn speak) could not be set, using the jclouds' default. Would be useful to have that improved.
Another thing that I think is important is locking. When deploying a large application it's almost certain that the logic will get confused because everyone is trying to create the default network in parallel.

PR #740, opened 2017-06-21
New versioning rules prep

ahgittin:
Creates `BrooklynVersionSyntax` to assist with brooklyn/osgi versioning.  Includes conversion routines, many tests, and many usages updated.

Versions are handled according to https://github.com/apache/brooklyn-docs/pull/198 -- but we're not yet applying them in the code -- so the problem discussed at https://github.com/apache/brooklyn-server/pull/672 still stands -- but should be easy to fix using the groundwork here and in #737.

Builds on #737 which laid the foundation but didn't recognise the need for a specific recommended brooklyn version syntax (this includes and closes that in case it's easier to review and merge just this, ignoring that).


neykov:
LGTM, but a bit nervous about not relying on `DefaultMaven2OsgiConverter` for the version conversion. The same version string will create a maven bundle with one OSGi version and a catalog item with a different Brooklyn version. This then could lead to failure to load a bundle because the catalog item version is different from the wrapping bundle. Or failure to find `bundle:version` reference.

neykov:
Couldn't find where we sanitise the catalog items ID to be used for persistence file names. If changing the naming rules for the IDs then perhaps changing the sanitising rules would be needed as well?

utils/common/src/test/java/org/apache/brooklyn/util/text/VersionComparatorTest.java
```diff
@@ -56,40 +35,51 @@ public void testSnapshotSuffixComparison() {
 
     @Test
     public void testComparison() {
```
geomacy: BTW the tests would need to be careful to supply actually different String objects, to avoid hitting the `if (Objects.equal(v1, v2)) return 0;` [here](https://github.com/apache/brooklyn-server/pull/743/files#diff-a66acd1b299b8d264a4c2c6f052b27c5R77). We'd want to be sure to exercise the code below that test.

utils/common/src/main/java/org/apache/brooklyn/util/text/NaturalOrderComparator.java
```diff
@@ -140,6 +147,32 @@ public int compare(String a, String b) {
...
+                        return result;
+                    }
+                }
+                // numbers are equal value, remaining string compares identical; 
+                // comes down to whether there are any leading zeroes here
+                return nzb-nza;
+                
             } else if ((Character.isDigit(ca) || nza>0) && (Character.isDigit(cb) || nzb>0)) {
                 // both sides are numbers, but at least one is a sequence of zeros
                 if (nza==0) {
```
ahgittin: his `charAt` returns 0 if we've gone beyond the string length, i think it might rely on that

utils/common/src/main/java/org/apache/brooklyn/util/text/NaturalOrderComparator.java
```diff
@@ -140,6 +147,32 @@ public int compare(String a, String b) {
                 if ((result = compareRight(a.substring(ia), b.substring(ib))) != 0) {
                     return result;
                 }
+                // numeric portion is the same; previously we incremented and checked again
```
ahgittin: i noticed some behaviour which i thought was undesired behaviour of `NaturalOrderComparator`, as indicated by the new tests

utils/common/src/main/java/org/apache/brooklyn/util/text/VersionComparator.java
```diff
@@ -56,149 +56,69 @@ public static boolean isSnapshot(String version) {
...
-    
-    private int compareDotSplitPart(String v1, String v2) {
-        String[] v1Parts = splitOnNonWordChar(v1);
-        String[] v2Parts = splitOnNonWordChar(v2);
+        TwoBooleans snapshots = TwoBooleans.of(isSnapshot(v1), isSnapshot(v2));
+        if (snapshots.different()) return snapshots.compare(true);
+
+        String u1 = versionWithQualifier(v1);
+        String u2 = versionWithQualifier(v2);
+        int uq = NaturalOrderComparator.INSTANCE.compare(u1, u2);
```
ahgittin: i think it's important for this comparator not to return `0` unless strings are truly equal; otherwise we risk indeterminate ordering for things which other routines consider different.

therefore in perverse cases we return a consistent ordering, even if it seems a little arbitrary.

in this case, if someone did perversely use `1.01`, `1.1`, `1.09`, `1.9`, and `1.10` as versions, i think that order -- which is what we return -- is the least surprising.

utils/common/src/main/java/org/apache/brooklyn/util/text/VersionComparator.java
```diff
@@ -56,149 +56,69 @@ public static boolean isSnapshot(String version) {
...
+        String q1 = v1.substring(u1.length());
+        String q2 = v2.substring(u2.length());
+        
+        String qq1 = q1, qq2 = q2;
+        if (q1.startsWith("-") || q1.startsWith(".") || q1.startsWith("_")) q1 = q1.substring(1);
+        if (q2.startsWith("-") || q2.startsWith(".") || q2.startsWith("_")) q2 = q2.substring(1);
+        uq = NaturalOrderComparator.INSTANCE.compare(q1, q2);
+        if (uq!=0) return uq;
+        
+        // if qualifiers the same, look at qualifier separator char
```
ahgittin: as above if someone requests a comparison between different strings we should give one.  it's a perverse case but equality determination should be done by someone else.

utils/common/src/main/java/org/apache/brooklyn/util/text/VersionComparator.java
```diff
@@ -56,149 +56,69 @@ public static boolean isSnapshot(String version) {
...
+        int sep1 = qq1.startsWith("-") ? 3 : qq1.startsWith(".") ? 2 : qq1.startsWith("_") ? 1 : 0;
+        int sep2 = qq2.startsWith("-") ? 3 : qq2.startsWith(".") ? 2 : qq2.startsWith("_") ? 1 : 0;
+        uq = Integer.compare(sep1, sep2);
+        if (uq!=0) return uq;
         
-        // use lookahead to split on all non-letter non-numbers, putting them into their own buckets 
-        parts.addAll(Arrays.asList(remaining.split("(?<=[^0-9\\p{L}])|(?=[^0-9\\p{L}])")));
-        return parts.toArray(new String[parts.size()]);
+        // finally, do normal comparison if both have qualifier or both unqualified (in case leading 0's are used)
+        return NaturalOrderComparator.INSTANCE.compare(v1, v2);
```
ahgittin: i think this block handles the case of `01` v `0` and possibly separators -- in any case useful to have a fallback (not return 0 unless it's the same string)

utils/common/src/main/java/org/apache/brooklyn/util/text/VersionComparator.java
geomacy: But it's surely not correct for Versions - I agree `NaturalOrderComparator` should return 0 only if the Strings are truly equal; I don't think it's the right thing to use for the major/minor/patch - these are meant to be compared numerically.

utils/common/src/main/java/org/apache/brooklyn/util/text/VersionComparator.java
```diff
@@ -56,149 +56,69 @@ public static boolean isSnapshot(String version) {
...
-        if (parts==null || parts.length<=i) return false;
-        if (isNumberInFirstChar(parts[i])) return true;
-        if (".".equals(parts[i])) {
-            if (parts.length>i+1)
-                if (isNumberInFirstChar(parts[i+1])) 
-                    return true;
-        }
-        return false;
+    private String versionWithQualifier(String v1) {
+        Matcher q = Pattern.compile("([0-9]+(\\.[0-9]+(\\.[0-9])?)?)(.*)").matcher(v1);
```
ahgittin: good spot!

utils/common/src/test/java/org/apache/brooklyn/util/text/VersionComparatorTest.java
ahgittin: as above, i think the comparators should never return 0 unless the strings are actually equal; someone else has determined equality beforehand, and wants this to give a canonical order on the things it thinks are unequal

utils/common/src/main/java/org/apache/brooklyn/util/text/BrooklynVersionSyntax.java
```diff
@@ -84,4 +89,100 @@ public static boolean isValidOsgiVersion(String candidate) {
...
+     * and {@link #toValidOsgiVersion(String)} can be used to reverse the input (modulo insertion of "0"'s for minor/patch numbers if missing).
+     * Precise behaviour for other input is not guaranteed but callers can expect output which resembles the input,
+     * with any major/minor/patch string at the front preserved and internal contiguous alphanumeric sequences preserved. */
+    public static String toGoodBrooklynVersion(String input) {
+        return toGoodVersion(input, "-", false);
+    }
+    
+    private static String toGoodVersion(String input, String qualifierSeparator, boolean requireMinorAndPatch) {
+        Preconditions.checkNotNull(input);
+        final String FUZZY_REGEX =
```
ahgittin: most are; the few that aren't i think either aren't worth it or are special to how they are used; moving them elsewhere risks things breaking if e.g. the grouping rules are changed

ahgittin:
i've cleaned up several things, but multiple discussion threads remain:

>  DefaultMaven2OsgiConverter different

for recommended versions it is the same, isn't it?  that's the important thing.  for other things yes it will be surprising but someone making an OSGi bundle relying on that converter's obscure behaviour is asking for trouble.  i'd rather optimize to be sensible for people doing foolish non-recommended things in non-maven / non-OSGi cases, where version `1.2_SNAPSHOT` is probably `1.2.0.SNAPSHOT` not `1.0.0.2_SNAPSHOT`.

> sanitize catalog item IDs filenames

again only an issue if using non-recommended IDs, and we don't rely on those files very much now (esp once YAML BOMs are added as bundles), so content to leave this as a bug.

> different ordering to OSGi

semver says qualifer makes something less than unqualified, e.g. `1.0.0-rc1` < `1.0.0`, and i think that's what people expect now.  also i think they are surprising i the case of `-rc9` v `-rc10`.

our ordering is different to both; we should make this clear in the docs but i think that's better than forcing people to understand the surprising ordering in osgi.

the one place this could be an issue is if we've relied on OSGi to resolve version ranges.  but i think let's cross that bridge when we come to it, rather than have `1.0.0-rc1` > `1.0.0` -- no?

> comparators shouldn't return 0

discussed above, but captured here for completeness

neykov:
> for other things yes it will be surprising but someone making an OSGi bundle relying on that converter's obscure behaviour is asking for trouble

That's everyone using maven. When the bundle's version is different from the catalog.bom's derived OSGi version Brooklyn will deny loading the file, right? If so we should just say that we are not supporting non-standard versions (and not accept them from the start) instead of documenting the format as deprecated, but working and letting the user find out if it works or not for them. Wdyt?

geomacy:
I don't think there's any harm in having a slightly stricter validation in place, in fact as a user I would find that more helpful than having Brooklyn adapt to me doing things badly.

Should we take the discussion of the above points back on to the mailing list or just deal with them here?

ahgittin:
@neykov do we know anyone actually using maven to make brooklyn bundles with a version syntax that will break it?  i think it's far enough below the radar i'm not concerned about it.  we should log warns on non-recommended versions (which i think we do, probably too much) but disallowing them now is going to break backwards compatibility.  also it's not uncommon that people use OSGi versions directly in maven which will work fine in Brooklyn even though it's not the recommended syntax.  feels like it might be antisocial to prevent that.

in short current approach -- have a recommended syntax, and warn but make best effort if people not using it -- is best for now, and we can become stricter in the next version.

PR #741, opened 2017-06-22
Update maven-compiler-plugin to 3.6.1

bostko:
I think since we upgrade java it will be useful to have later maven-compiler-plugin.
If you approve this change, I will do the same PR for brooklyn-ui and brooklyn-dist.


neykov:
+1

neykov:
Builds fine for me, merging.

PR #743, opened 2017-06-22
Version new rules

ahgittin:
Implements the behaviour discussed on ML re version syntax.

Places where OSGi syntax is required (ie registering/searching OSGi bundles) convert to the OSGi syntax.  Everywhere else uses the version string exactly as supplied by the caller.  Places where mappings to OSGi are not unique result in warnings.

The fixes the issue described in #672, and `catalog.bom` files can (should) now use familiar maven/semver versions (`#.#.#-qual`) which most people seem to use everywhere else.  This is converted to OSGi syntax (`#.#.#.qual`) when creating the bundle, or comparing with installed bundles, but that's all hidden from users (unless they go deep into OSGi).

You _can_ supply OSGi syntax (or even crazy things like `v1`) in the `catalog.bom`, and the syntax supplied is exactly what will be used in the Brooklyn catalog.  There will be warnings if it does not follow the recommended syntax but they can be ignored.

This builds on (and closes) #737 and #740, depending whether people want to review those independently or in the context of this.

Merge https://github.com/apache/brooklyn-docs/pull/198 once this is merged.


ahgittin:
unrelated non-det test failure `org.apache.brooklyn.entity.group.SequenceGroupTest.testGroupWithMatchingFilterReturnsEverythingThatMatches`

neykov:
LGTM
retest this please.

geomacy:
hi Alex, looks good but I did have some comments and concerns especially around the comparison of versions. I recorded the comments against  #737 and #740 (and in https://github.com/apache/brooklyn-docs/pull/198) rather than here, probably should just have put them here but was going in order!...  let me know what you think.

ahgittin:
Updated with fixes to upstream PRs.

geomacy:
I've had a further think about the versions as discussed in [#740](https://github.com/apache/brooklyn-server/pull/740#issuecomment-311631969).  In a (long) nutshell:

* I think we agree that there _could_ be problems with the approach here _in edge cases_ where people are using version formats that we don't recommend; in particular, as you note "one place this could be an issue is if we've relied on OSGi to resolve version ranges."  

Re. the latter I'm imagining a case like your 'acme-cluster' ([mail link](http://mail-archives.apache.org/mod_mbox/brooklyn-dev/201706.mbox/%3C2c13060e-58d9-1098-ddd0-e713575b80e7%40cloudsoft.io%3E)) where someone has released a series of drafts, say, `acme:1.0.0-v1` , ..  `acme:1.0.0-v9` and `acme:1.0.0-v10`.  Things started settling down around `v5`, so if I have a bundle that depends on `acme`, and I specify an OSGI dependency on it, I would like to be able to say 

    Import-Package: acme;version="[1.0.0.v5, 2)"

but I will have to understand that this will exclude  `1.0.0.v10` when it becomes available and write my poms/boms appropriately.  That's probably something you would expect me to understand if I am writing `Import-Package` declarations and have read the new Brooklyn docs on versions.  However it's something we should bear in mind if we introduce version ranges in Brooklyn catalogs, where if I write the following I think I should be entitled to expect it to work with `v10`.

    classpath://acme:[1.0.0-v5,2):/catalog.bom 

*  I guess I agree that this is too far below the radar to be a deal breaker for this PR, and we can go with your [suggestion](https://github.com/apache/brooklyn-server/pull/740#issuecomment-311903131) of "a recommended syntax, and warn but make best effort if people not using it -- is best for now, and we can become stricter in the next version."

* However, one final caveat, worth considering before merging: in your updates to `Osgis.BundleFinder` you haven't changed the [comparison](https://github.com/ahgittin/brooklyn-server/blob/e742da7bdf22ac41aaa68002f6b05058a65f8442/core/src/main/java/org/apache/brooklyn/util/core/osgi/Osgis.java#L215) in `findAll` where the bundles are sorted in order to be able to return the latest in `findOne`.  In the `acme` example above this will mean `findOne` will give you `v9` as the latest even when `v10` is out; I would argue that this should return **Brooklyn's** notion of what is "latest", so we should be re-mapping from OSGI bundle version to `BrooklynVersion` before doing the comparison.  Otherwise I think it really will give us problems where the wrong bundle is picked.

ahgittin:
very good spot on `BundleFinder` @geomacy - have cleaned it up so instead of sorting by version (and _ignoring_ bundle _names_ altogether!) it now sorts alphanumerically on name, then on version, and in the code where we `findOne` we take the highest version of the first entry.

as for OSGi ranges that may be a bit of a headache, or at least potentially a surprise to OSGi users.  _however_ so long as we can make it follow our scheme it should be fine:  i would think anyone using a combination of qualified and unqualified versions would expect semver semantics where unqualified is last.  in osgi people using qualifiers presumably _always_ do so, and are used to using a datestamp or something similar to ensure ordering is correct and it will almost certainly still be correct.  (IE I don't imagine anyone would intentionally use `1.0.0.M9` then `1.0.0.M10`, or `1.0.0.M1` then `1.0.0` and want OSGi semantics to apply!)

PR #744, opened 2017-06-23
Allow catalog item to be force-updated via query string flag on the REST API

tbouron:
# Issue

When `POST`ing a catalog item (YAML file) with an ID and version that already exists, the REST API returns a `400` with a message inviting the user to use a `forceUpdate` flag if one wishes to. However, this flag is only available internally and not exposed to the REST API.

Another issue is that under the same condition for ZIP/JAR archives, the `OsgiArchiveInstaller` silently ignores the update and therefore, the REST API returns a `201` even though it has done anything.

# Fix

This fixes both above issues by:
- exposing a `forceUpdate` flag on the REST API for `/v1/catalog`
- checking the result code from `OsgiArchiveInstaller.install()` method and throws the same error message inviting the user to use the `forceUpdate` flag

### DIFF
--- a/rest/rest-api/src/main/java/org/apache/brooklyn/rest/api/CatalogApi.java
+++ b/rest/rest-api/src/main/java/org/apache/brooklyn/rest/api/CatalogApi.java
@@ -67,7 +67,7 @@ public interface CatalogApi {
             @ApiResponse(code = 400, message = "Error processing the given YAML"),
             @ApiResponse(code = 201, message = "Catalog items added successfully")
     })
-    public Response create(String yaml);
+    public Response create(String yaml, @QueryParam("forceUpdate") @DefaultValue("false") boolean forceUpdate);
 
     @POST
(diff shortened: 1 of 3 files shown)


rest/rest-api/src/main/java/org/apache/brooklyn/rest/api/CatalogApi.java
```diff
@@ -67,7 +67,7 @@
             @ApiResponse(code = 400, message = "Error processing the given YAML"),
             @ApiResponse(code = 201, message = "Catalog items added successfully")
     })
-    public Response create(String yaml);
+    public Response create(String yaml, @QueryParam("forceUpdate") @DefaultValue("false") boolean forceUpdate);
```
m4rkmckenna: Can you add `@ApiParam(***)` so that the swagger doc has some detail

tbouron: Good call @m4rkmckenna, I just pushed the changes

PR #745, opened 2017-06-26
Azure ARM default network fixes from PR suggestions

Graeme-Miller:
Important Note: Azure ARM doesn't currently work with the master of Brooklyn. Currently, after a Azure ARM VM is created, JClouds attempts to list all of the nodes in the resource group but cannot parse the result due to a missing property (vhd for DataDisk). I have tested with the latest JClouds code in master and we do not have this issue. 

However, we are currently on the latest released version of JClouds (2.0.1) and we cannot pull in the code in mater until there is another JClouds release. As such, the code that has been written for default network has been written against the code in JClouds master and will work when we upgrade to the next version of JClouds. If you need working Azure ARM support, and default networking build JClouds master locally, and build brooklyn against it.

---------------------

Various fixes:
- IpOptions were being added incorrectly to templateOptions. They were being set as a map, when they should have been a list of maps
- When checking templateOptions for existing config, look for 'networks' rather than NETWORK_NAME config (which resolves to 'networkName'). Still use NETWORK_NAME config when checking in brooklyn config
- Check that a region has been set in brooklyn config (config.get(CLOUD_REGION_ID)) if not, don't try and create a default network. It was suggested that I could move the default network creation to later in JCloudsLocation after the templateOptions have been built. This would mean that a default region would have been added to the templateOptions if none was specified in config. I attempted to implement this, but ran into an issue. The AzureTemplateOptions in JClouds 2.0.1 is different to the AzureTemplateOptions in JClouds master. As such, it is not possible to amend the AzureTemplateOptions from Brooklyn code, as we would not be able to modify the ipOptions (a concept that only exists in JClouds master). As such, I have put the null check in for the time being. After the next JClouds release, I will come back and fix this
- Check for the situation where a default network exists, but a default subnet does not. In this situation create a default subnet only

------------------------------

@neykov pointed out there could be issues if we are trying to create multiple VM's at the same time due to the these VMs all trying to create the default network/subnet. I have investigated and I don't think this will be an issue. The default network creation is done in the obtainOnce method of JCloudsLocation. If there were a situation where multiple VMs tried to create the network at the same time, then all but one would fail. Brooklyn will by default will then attempt to get a VM again, and this time it will succeed. To test this theory I spooled up three AMPs, and at the same time kicked off a deployment of 5 VMs on each AMP (15 total) to try and force this situation. Only 3/15 came up correctly, but the failures were quota issues which means they got passed network creation.

### DIFF
--- a/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/creator/DefaultAzureArmNetworkCreator.java
+++ b/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/creator/DefaultAzureArmNetworkCreator.java
@@ -45,6 +45,8 @@ import org.apache.brooklyn.config.ConfigKey;
 import org.apache.brooklyn.core.config.ConfigKeys;
 import org.apache.brooklyn.util.collections.MutableMap;
 import org.apache.brooklyn.util.core.config.ConfigBag;
+import org.apache.brooklyn.util.time.CountdownTimer;
+import org.apache.brooklyn.util.time.Duration;
 
 public class DefaultAzureArmNetworkCreator {
 
@@ -53,6 +55,9 @@ public class DefaultAzureArmNetworkCreator {
     private static final String DEFAULT_RESOURCE_GROUP_PREFIX = "brooklyn-default-resource-group";
     private static final String DEFAULT_NETWORK_NAME_PREFIX = "brooklyn-default-network";
     private static final String DEFAULT_SUBNET_NAME_PREFIX = "brooklyn-default-subnet";
+
+    private static final String PROVISIONING_STATE_UPDATING = "Updating";
+    private static final String PROVISIONING_STATE_SUCCEEDED = "Succeeded";
 
     private static final String DEFAULT_VNET_ADDRESS_PREFIX = "10.1.0.0/16";
     private static final String DEFAULT_SUBNET_ADDRESS_PREFIX = "10.1.0.0/24";
@@ -130,6 +135,24 @@ public class DefaultAzureArmNetworkCreator {
 
         //Get created subnet
         Subnet createdSubnet = api.getSubnetApi(resourceGroupName, vnetName).get(subnetName);
+
+        //Wait until subnet is created
+        CountdownTimer timeout = CountdownTimer.newInstanceStarted(Duration.minutes(new Integer(20)));
+        while (PROVISIONING_STATE_UPDATING.equals(createdSubnet.properties().provisioningState())) {
+            if (timeout.isExpired()) {
+                throw new IllegalStateException("Creating subnet " + subnetName + " stuck in the updating state, aborting.");
+            }
+            LOG.debug("Created subnet {} is still in updating state, waiting for it to complete", createdSubnet);
+            Duration.sleep(Duration.ONE_SECOND);
+            createdSubnet = api.getSubnetApi(resourceGroupName, vnetName).get(subnetName);
+        }
+
+        String lastProvisioningState = createdSubnet.properties().provisioningState();
+        if (!lastProvisioningState.equals(PROVISIONING_STATE_SUCCEEDED)) {
+            LOG.debug("Created subnet {} in wrong state, expected state {} but found {}", new Object[] {subnetName, PROVISIONING_STATE_SUCCEEDED, lastProvisioningState});
+            throw new IllegalStateException("Created subnet " + subnetName + " in wrong state, expected state " + PROVISIONING_STATE_SUCCEEDED +
+                    " but found " + lastProvisioningState );
+        }
 
         //Add config
         updateTemplateOptions(config, createdSubnet);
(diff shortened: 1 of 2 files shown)


duncangrant:
Tested with a location with no virtual network set but with overrideAuthenticateSudo set to true.
Tested with a specified virtual network.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/creator/DefaultAzureArmNetworkCreator.java
```diff
@@ -64,57 +68,71 @@ public static void createDefaultNetworkAndAddToTemplateOptionsIfRequired(Compute
...
+                    + "Creating subnet [{}] on network [{}], and updating template options", subnetName, vnetName);
 
-        //Create network
-        api.getVirtualNetworkApi(resourceGroupName).createOrUpdate(vnetName, location, virtualNetworkProperties);
+            //Just creating the subnet
+            subnetApi.createOrUpdate(subnetName, subnetProperties);
+        }
+
+        //Get created subnet
         Subnet createdSubnet = api.getSubnetApi(resourceGroupName, vnetName).get(subnetName);
```
neykov: You already have `subnetApi` available here, could reuse it.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/networking/creator/DefaultAzureArmNetworkCreator.java
```diff
@@ -64,57 +68,71 @@ public static void createDefaultNetworkAndAddToTemplateOptionsIfRequired(Compute
...
+                    .subnets(Arrays.asList(subnet)).build();
+            virtualNetworkApi.createOrUpdate(vnetName, location, virtualNetworkProperties);
+        } else {
+            LOG.info("Network config not specified when provisioning Azure machine, and default subnet does not exists. "
+                    + "Creating subnet [{}] on network [{}], and updating template options", subnetName, vnetName);
 
-        //Create network
-        api.getVirtualNetworkApi(resourceGroupName).createOrUpdate(vnetName, location, virtualNetworkProperties);
+            //Just creating the subnet
+            subnetApi.createOrUpdate(subnetName, subnetProperties);
```
neykov: Need to wait for the `createOrUpdate` operation to complete (here and above). The call is async. Can be done by polling the `properties.provisioningState()` returned by the `get` call on the resource.
If not done, any operations using the resource will fail.

neykov:
* re azure jclouds 2.0.1 not having enough functionality - the changes here (and in previous PR) will create the default network, but will not assign it to the nodes because `ipOptions` doesn't exist in `templateOptions`. That's actually fine because 2.0.1 will create a single network for all nodes. The behaviour this PR fixes is introduced post `2.0.1`. That's why I suggest to do the changes against jclouds master and mark the PR as `WIP` until Brooklyn upgrades to a new jclouds release.

* re quota failures - not waiting for the `createOrUpdate` operation to complete will result in failures which look much like quota exceeded failures you are seeing. The API calls are async and can return before the resource is fully created. Attempts to use the resource in other operations will fail. That's why its important to:
  1. block for the operation to complete
  2. do an effort to `createOrUpdate` the resource only once (or as little as possible) to minimize the risk of unblocking too early while a subsequent call re-creates the resource, putting it back in the `UPDATING` state.

Depending on the region and instance type the quota is between 10-30 instances. So I'd expect at least 10 instances to come up fine, ideally all of them on regions that have the limit increased.

Graeme-Miller:
@neykov that's a good point about the default networking functionality being different in 2.0.1. I'd still recommend merging this PR as it is useful for downstream projects that use the latest JClouds.

Not sure I 100% follow your comment about quota failures. However, if it is acceptable I will research and fix in a separate PR?

neykov:
Fine by me to merge the PR as it already improves on the existing functionality. Can you create another PR which targets post 2.0.1 jclouds, handling the default region case and mark it `WIP` so we can merge it once we have a new jclouds release.
Note that because of my second point in above comment this will actually break deployments for anyone relying on the functionality. Happy to pair to discuss how it can be improved on.

neykov:
LGTM with the "waiting to complete" changes.
@Graeme-Miller have you retried your test, has the failure rate improved? If no failure this is good to merge. Otherwise we need to look for further improvements to prevent deployment errors.
Prefer rebasing of PR instead merging from master, especially for small ones, but not a blocker for merging.

Graeme-Miller:
@neykov my test failed again, but with a better rate. 8/15 came up.
This doesn't surprise me as the failures are all for exceeding the cpu quota limit.

Graeme-Miller:
@neykov I have rebased

neykov:
Thanks @Graeme-Miller, merging.

neykov:
@Graeme-Miller Brooklyn has upgraded to jclouds 2.0.2, which opens the door to supporting default region as well.

PR #746, opened 2017-06-28
Auto wrap YAML BOMs as catalog bundles

ahgittin:
Wraps pure-YAML BOMs uploaded to catalog in a bundle, creating a name and version if needed.

This allows us to see all items added as a unit.  This will allow us to prefer loading co-defined items (though that's not fully done, still relies on preferring search path when doing item lookup), and to edit or remove that item as a set (crucially we now keep the BOM, previously we discarded it!).

Docs changed to reflect this and explain differences in another PR.

The basic wrapping was pretty simple but a surprising number of subtleties emerged which this fixes:

* java scanning doesn't work in OSGi; docs and errors make this clear
* impacts replacement semantics (now it does a hash on the bundle to make most things the same, though a re-post of a non-SNAPSHOT BOM that doesn't declare a bundle will now fail because the random bundle name is stored on the item as containingBundle and is different on a re-post; and needed extra work to remove leaked anonymous bundles)
* had to change the way we set up the libraries / search path, to exclude this "bundle" (mainly so tests are unchanged, but also improves efficiency for simple BOMs; also the code is cleaner now)

And there is one thing that emerged that I *haven't* fixed:  on rebind it became vital to handle the situation where dependent bundles might not be loaded before the depender bundle; see failing test`testLongReferenceSequence`.  Not entirely sure how to do this /cc @neykov @geomacy @aledsage.

Depends on #743.


ahgittin:
have done a big job to make things use the TypeRegistry instead of the old catalog, and to add to type registry in an `UNRESOLVED` state first.  this fixes the big problem by adding types from all bundles into an unresolved state, then resolving them.

it has ended up much bigger than i meant BUT it has several main advantages:

* we no longer persist catalog item XML for things where we persist the OSGi bundle
* we're very close to being able to remove nearly all the clunky "catalog" code; apart from items that were persisted as XML everything uses the new type registry now

the logic for guessing the supertypes is messy, hooking in to the old `Guessing` interpreter in a brittle way, but if we plan to remove old catalog code (and annotations scanning) we can rewrite it all to be much simpler and nicer.

note this PR is still WIP, not working 100%; we don't support templates in the registry yet, locations need attention, and there are quite a few TODO items to look at.  but wanted to give people sight of it.

ahgittin:
This now addresses templates in TypeRegistry -- simply using a "catalog_template" tag.  There was no immediate need for "deployable" or "final" semantics so those have not been added.

This also addresses a newly-found issue to do with locations:  the location registry had a hokey relationship with the old catalog; it's now a cleaner facade.  See comments in BasicLocationRegistry for this.

Finally I found a way to preserve the capability where the same BOM YAML is uploaded multiple times:  if the bundle of the currently installed type was a wrapper, we now allow to replace it if the plan itself is unchanged.  This removes the backwards compatibility issue introduced around uploading a BOM multiple times and it disallowing because the containing bundle was changed.

The REST API however still points at the old catalog routines, so it won't yet show items added through OSGi.  This is a showstopper of course because all BOMs will be added through OSGi.  Fix for that coming soon.

ahgittin:
REST API is now compatible.  So this is ready for review @geomacy @neykov @aledsage .

Apologies for the size -- hopefully between this and the ML it is clear why there was so much.  Perhaps I should have disabled it so it could have been merged incrementally but I was not expecting so much more to be needed.

Still it is all good stuff moving forwards.  We are now very close to being able to remove _all_ the catalog code, in favour of the very simple `TypeRegistry`.

ahgittin:
@geomacy This should solve load order problems and any problems around BOM scanner.  (I think we want to disable the auto-scan osgi bundles BTW.)  Let me know if not.  And of course please don't go fixing other things -- ask me if in any doubt.

geomacy:
hi @ahgittin you've accidentally committed an unresolved conflict in [CatalogUtils](https://github.com/ahgittin/brooklyn-server/commit/47b54d46750b03bb10bcfed3e851f827fa046eb1#diff-61553f1ab67d6bd6673b69a1bc99d136R394) in 47b54d4

ahgittin:
oops - thanks @geomacy; fixed that

camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/catalog/CatalogYamlEntityOsgiTypeRegistryTest.java
```diff
@@ -0,0 +1,66 @@
...
+import com.google.common.collect.Iterables;
+
+/** Variant of parent tests using OSGi, bundles, and type registry, instead of lightweight non-osgi catalog */
+@Test
+public class CatalogYamlEntityOsgiTypeRegistryTest extends CatalogYamlEntityTest {
+
+    // use OSGi here
+    @Override protected boolean disableOsgi() { return false; }
+    
+    // use type registry appraoch
```
geomacy: typo "appraoch" here and a couple of other places

geomacy:
Attached file contains logs for the above test results, and the persisted bundle data for the second.
[test-notes.zip](https://github.com/apache/brooklyn-server/files/1128546/test-notes.zip)

ahgittin:
Thanks @geomacy.  Glad you tried Karaf mode -- didn't think to test that.  Will repair.

BTW wanted to note, this fixes https://issues.apache.org/jira/browse/BROOKLYN-343 .

ahgittin:
right, see note on last commit which should solve the karaf mode problem you spotted @geomacy.  also does a little something to solve the situation you encountered @richardcloudsoft .

core/src/main/java/org/apache/brooklyn/core/typereg/RegisteredTypes.java
```diff
@@ -125,21 +178,52 @@ public static RegisteredType bean(@Nonnull String symbolicName, @Nonnull String
...
         return new BasicRegisteredType(RegisteredTypeKind.SPEC, symbolicName, version, plan);
     }
-    /** Convenience for {@link #cpec(String, String, TypeImplementationPlan)} when there is a single known java signature/super type */
+    /** Convenience for {@link #spec(String, String, TypeImplementationPlan)} when there is a single known java signature/super type */
     public static RegisteredType spec(@Nonnull String symbolicName, @Nonnull String version, @Nonnull TypeImplementationPlan plan, @Nonnull Class<?> superType) {
-        if (superType==null) log.warn("Deprecated use of RegisteredTypes API passing null name/version", new Exception("Location of deprecated use, wrt "+symbolicName+":"+version+" "+plan));
+        if (superType==null) log.warn("Deprecated use of RegisteredTypes API passing null supertype", new Exception("Location of deprecated use, wrt "+symbolicName+":"+version+" "+plan));
         return addSuperType(spec(symbolicName, version, plan), superType);
     }
+    public static RegisteredType newInstance(@Nonnull RegisteredTypeKind kind, @Nonnull String symbolicName, @Nonnull String version,
```
geomacy: Maybe worth introducing a Builder, for readability at the call site.

api/src/main/java/org/apache/brooklyn/api/typereg/ManagedBundle.java
```diff
@@ -27,9 +27,13 @@
...
      * to uniquely identify this bundle-instance.
-     * This typically includes the unique {@link #getId()} of this item. */
+     * This typically includes the unique {@link #getId()} of this item. 
+     * This will not normally be a URL that can be loaded. */
     String getOsgiUniqueUrl();
 
     VersionedName getVersionedName();
 
+    /** MD5 checksum of the contents of bundle as installed to OSGi */
+    String getChecksum();
```
ahgittin: it uses `md5` but it has some opinions about what it uses for checksum so better to consider it its own checksum

core/src/main/java/org/apache/brooklyn/core/catalog/internal/BasicBrooklynCatalog.java
```diff
@@ -449,21 +494,28 @@ public static VersionedName getVersionedName(Map<?,?> catalogMetadata, boolean r
...
+            containingBundle, catalogMetadata, result, requireValidation, parentMeta, 0, force);
         
         itemDef.remove("brooklyn.catalog");
         catalogMetadata.remove("item");
         catalogMetadata.remove("items");
         if (!itemDef.isEmpty()) {
-            log.debug("Reading brooklyn.catalog peer keys as item ('top-level syntax')");
+            // AH - i forgot we even supported this. probably no point anymore,
+            // now that catalog defs can reference an item yaml and things can be bundled together?
+            log.warn("Reading catalog item from sibling keys of `brooklyn.catalog` section, "
```
ahgittin: done

core/src/main/java/org/apache/brooklyn/core/catalog/internal/BasicBrooklynCatalog.java
```diff
@@ -472,12 +524,46 @@ private void collectCatalogItems(String yaml, List<CatalogItemDtoAbstract<?, ?>>
...
-            collectCatalogItems("item:\n"+makeAsIndentedObject(rootItemYaml), rootItem, result, catalogMetadata);
+            collectCatalogItemsFromItemMetadataBlock("item:\n"+makeAsIndentedObject(rootItemYaml), containingBundle, rootItem, result, requireValidation, catalogMetadata, 1, force);
         }
     }
 
+    /**
+     * Expects item metadata, containing an `item` containing the definition,
+     * and/or `items` containing a list of item metadata (recursing with depth).
+     * 
+     * Supports two modes:
```
ahgittin: added comment (it's actually `result` that controls the behaviour)

core/src/main/java/org/apache/brooklyn/core/catalog/internal/BasicBrooklynCatalog.java
```diff
@@ -511,30 +597,76 @@ private void collectCatalogItems(String sourceYaml, Map<?,?> itemMetadata, List<
...
         // TODO in 0.8.0 require brooklyn.libraries, don't allow "libraries" on its own
-        List<?> librariesNew = MutableList.copyOf(getFirstAs(itemMetadataWithoutItemDef, List.class, "brooklyn.libraries", "libraries").orNull());
-        Collection<CatalogBundle> libraryBundlesNew = CatalogItemDtoAbstract.parseLibraries(librariesNew);
-
-        List<?> librariesCombined = MutableList.copyOf(librariesNew)
-            .appendAll(getFirstAs(parentMetadata, List.class, "brooklyn.libraries", "libraries").orNull());
-        if (!librariesCombined.isEmpty())
-            catalogMetadata.put("brooklyn.libraries", librariesCombined);
-        Collection<CatalogBundle> libraryBundles = CatalogItemDtoAbstract.parseLibraries(librariesCombined);
+        List<?> librariesAddedHereNames = MutableList.copyOf(getFirstAs(itemMetadataWithoutItemDef, List.class, "brooklyn.libraries", "libraries").orNull());
```
ahgittin: now is a good time but not as part of this PR :)

lots of tests use the old syntax still :(

core/src/main/java/org/apache/brooklyn/core/catalog/internal/BasicBrooklynCatalog.java
```diff
@@ -1060,28 +1337,87 @@ private String makeAsIndentedObject(String yaml) {
...
+        return addItems(yaml, bundle, false);
+    }
+    
+    @Override
+    public List<? extends CatalogItem<?,?>> addItems(String yaml, ManagedBundle bundle, boolean forceUpdate) {
+        log.debug("Adding catalog item to "+mgmt+": "+yaml);
         checkNotNull(yaml, "yaml");
-        List<CatalogItemDtoAbstract<?, ?>> result = collectCatalogItems(yaml);
+        List<CatalogItemDtoAbstract<?, ?>> result = MutableList.of();
+        collectCatalogItemsFromCatalogBomRoot(yaml, bundle, result, true, ImmutableMap.of(), 0, false);
```
ahgittin: should be `forceUpdate`, good spot

core/src/main/java/org/apache/brooklyn/core/typereg/RegisteredTypes.java
ahgittin: would be a good addition in future.  for now we only call it in a few places.

core/src/main/java/org/apache/brooklyn/core/BrooklynFeatureEnablement.java
```diff
@@ -121,6 +120,8 @@
      * The functionality loads catalog items regardless of the persistence state so best used with persistence disabled.
      * If a bundle is uploaded its BOM is scanned regardless of this property (this only applies to bundles
      * installed through a non-brooklyn method, eg karaf.)
+     * 
+     * This installs legacy items and so should be deprecated in favour of uploading BOMs which Brooklyn manages.
```
ahgittin: okay

ahgittin:
@geomacy comments thus far addressed, including `CatalogBomScanner` removed

core/src/main/java/org/apache/brooklyn/core/catalog/internal/CatalogBundleLoader.java
```diff
@@ -64,14 +62,26 @@ public CatalogBundleLoader(Predicate<Bundle> applicationsPermitted, ManagementCo
...
+    
     /**
      * Scan the given bundle for a catalog.bom and adds it to the catalog.
      *
      * @param bundle The bundle to add
      * @return A list of items added to the catalog
      * @throws RuntimeException if the catalog items failed to be added to the catalog
      */
-    public Iterable<? extends CatalogItem<?, ?>> scanForCatalog(Bundle bundle) {
+    public Iterable<? extends CatalogItem<?, ?>> scanForCatalogLegacy(Bundle bundle) {
```
geomacy: Now that you've removed `CatalogBomScanner` this method isn't used any more, I think we can get rid of it and the one below.  See also comments on `OsgiManager.java`.  I think we can also get rid of `applicationsPermitted` in this class, in the one place that its constructor is called the predicate is hardcoded to `true`. The interesting case is on line 98 below.

core/src/main/java/org/apache/brooklyn/core/catalog/internal/CatalogBundleLoader.java
```diff
@@ -81,22 +91,54 @@ public CatalogBundleLoader(Predicate<Bundle> applicationsPermitted, ManagementCo
...
             String bomText = readBom(bom);
-            String bomWithLibraryPath = addLibraryDetails(bundle, bomText);
-            catalogItems = this.managementContext.getCatalog().addItems(bomWithLibraryPath, mb);
-            for (CatalogItem<?, ?> item : catalogItems) {
-                LOG.debug("Added to catalog: {}, {}", item.getSymbolicName(), item.getVersion());
+            if (mb==null) {
+                LOG.warn("Bundle "+bundle+" containing BOM is not managed by Brooklyn; using legacy item installation");
+                legacy = true;
+            }
+            if (legacy) {
```
geomacy: With the removal of the `CatalogBomScanner` I think the `legacy` parameter in this method's signature can be removed.  This is the interesting case where the code still calls the traditional catalog `addItems()` in the case where there is no managed bundle known in Brooklyn's management context. I'm wondering if this is still a path we need to support, or whether in fact now that the scanner is removed we can just call `addTypesFromBundleBom` in all cases within this method.

core/src/main/java/org/apache/brooklyn/core/catalog/internal/CatalogBomScanner.java
```diff
@@ -1,130 +0,0 @@
-/*
```
geomacy: Need to remove it from [blueprint.xml](https://github.com/apache/brooklyn-server/blob/3d8675e248cb229e6492d938655cb26222504fbc/karaf/init/src/main/resources/OSGI-INF/blueprint/blueprint.xml#L122) as well.

core/src/main/java/org/apache/brooklyn/core/mgmt/ha/OsgiArchiveInstaller.java
```diff
@@ -336,14 +358,14 @@ private synchronized void close() {
             if (!updating) { 
                 osgiManager.managedBundlesRecord.addManagedBundle(result);
                 result.code = OsgiBundleInstallationResult.ResultCode.INSTALLED_NEW_BUNDLE;
-                result.message = "Installed "+result.getMetadata().getVersionedName()+" with ID "+result.getMetadata().getId();
+                result.message = "Installed Brooklyn catalog bundle "+result.getMetadata().getVersionedName()+" with ID "+result.getMetadata().getId()+" ["+result.bundle.getBundleId()+"]";
```
geomacy: I'm getting this line being hit every time I stop and restart Karaf; the effect is that the default catalog bom gets added as a new bundle every time,

    karaf@brooklyn()> list -s | grep brooklyn-catalog-bom
    284 | Active   |  80 | 0.12.0.SNAPSHOT          | brooklyn-catalog-bom-kmJluM62                                                                                                                                                                                                                                                                   | brooklyn-catalog-bom-kmJluM62
    286 | Active   |  80 | 0.12.0.SNAPSHOT          | brooklyn-catalog-bom-Rx1SYA2u                                                                                                                                                                                                                                                                   | brooklyn-catalog-bom-Rx1SYA2u
    287 | Active   |  80 | 0.12.0.SNAPSHOT          | brooklyn-catalog-bom-WntzbKnZ                                                                                                                                                                                                                                                                   | brooklyn-catalog-bom-WntzbKnZ

 and you get another copy of its jar in the persisted `bundles` directory.

core/src/main/java/org/apache/brooklyn/core/mgmt/ha/OsgiArchiveInstaller.java
```diff
@@ -308,11 +326,16 @@ private synchronized void close() {
...
+                    // bundle already installed to OSGi subsystem but brooklyn not aware of it;
+                    // this will often happen on a karaf restart so don't be too strict!
+                    // in this case let's uninstall it to make sure we have the right bundle and checksum
+                    // (in case where user has replaced a JAR file in persisted state,
+                    // or where they osgi installed something and are now uploading it or something else) 
+                    // but let's just assume it's the same; worst case if not user will
+                    // have to uninstall it then reinstall it to do the replacement
+                    // (means you can't just replace a JAR in persisted state however)
+                    log.debug("Brooklyn install of "+result.getMetadata().getVersionedName()+" detected already loaded in OSGi; uninstalling that to reinstall as Brooklyn-managed");
+                    b.get().uninstall();
```
geomacy: We'll need some more checks around this, or maybe a rethink of how it's done altogether - try deploying this bom file via the REST API:

    brooklyn.catalog:
      id: hack
      bundle: org.apache.brooklyn.karaf-init
      version: 0.12.0.SNAPSHOT
      name: Hack
      itemType: template
  
      item:
        services:

and it will happily uninstall our karaf-init bundle; killing Brooklyn altogether (not the Karaf process).  You can even do 

    brooklyn.catalog:
      id: hack
      bundle: org.apache.karaf.bundle.core
      version: 4.0.8
      name: Hack
      itemType: template
    
      item:
        services:


which Karaf doesn't react well to.

core/src/main/java/org/apache/brooklyn/core/catalog/internal/CatalogBundleLoader.java
ahgittin: good idea, done

core/src/main/java/org/apache/brooklyn/core/catalog/internal/CatalogBundleLoader.java
ahgittin: yep

core/src/main/java/org/apache/brooklyn/core/catalog/internal/CatalogBundleLoader.java
```diff
@@ -81,22 +91,54 @@ public CatalogBundleLoader(Predicate<Bundle> applicationsPermitted, ManagementCo
...
+            }
+            
+            if (!legacy && BasicBrooklynCatalog.isNoBundleOrSimpleWrappingBundle(managementContext, mb)) {
+                ((ManagementContextInternal)managementContext).getOsgiManager().get().addInstalledWrapperBundle(mb);
             }
         } else {
             LOG.debug("No BOM found in {} {} {}", CatalogUtils.bundleIds(bundle));
         }
 
         if (!applicationsPermitted.apply(bundle)) {
```
ahgittin: agree re whtielist, removed that.

am hesitant about removing some of these methods as i find when i do that there's always something using it.  have added a warning and marked the legacy one deprecated but left enough that we can use it if we really have to, for one version,.

core/src/main/java/org/apache/brooklyn/core/catalog/internal/CatalogBomScanner.java
ahgittin: good spot; done

core/src/main/java/org/apache/brooklyn/core/mgmt/ha/OsgiManager.java
```diff
@@ -372,33 +376,46 @@ public synchronized Bundle registerBundle(CatalogBundle bundleMetadata) {
     // it probably works even if that is true, but we should consider what to do;
     // possibly remove that other capability, so that bundles with BOMs _have_ to be installed via this method.
     // (load order gets confusing with auto-scanning...)
-    public List<? extends CatalogItem<?,?>> loadCatalogBom(Bundle bundle) {
-        return MutableList.copyOf(loadCatalogBom(mgmt, bundle));
+    public List<? extends CatalogItem<?,?>> loadCatalogBomLegacy(Bundle bundle) {
```
ahgittin: removed

core/src/main/java/org/apache/brooklyn/core/location/BasicLocationRegistry.java
```diff
@@ -194,29 +168,71 @@ public boolean registerResolver(LocationResolver r) {
...
             }
-            return null;
+            return newDefinedLocation(lt);
         }
+        
+        return null;
     }
 
-    @Override
+    @Override @Deprecated
```
ahgittin: i omit them if it's covered by parent's javadoc

core/src/main/java/org/apache/brooklyn/core/catalog/internal/CatalogItemDtoAbstract.java
```diff
@@ -386,7 +387,7 @@ protected void setTags(Set<Object> tags) {
 
     /**
      * Parses an instance of CatalogLibrariesDto from the given List. Expects the list entries
-     * to be either Strings or Maps of String -> String. Will skip items that are not.
+     * to be either Strings or Maps of String -> String or bundles. Will skip items that are not.
      * <p>
      * If a string is supplied, this tries heuristically to identify whether a reference is a bundle or a URL, as follows:
      * - if the string contains a slash, it is treated as a URL (or classpath reference), e.g. <code>/file.txt</code>;
```
ahgittin: done

core/src/main/java/org/apache/brooklyn/core/mgmt/ha/OsgiArchiveInstaller.java
```diff
@@ -371,23 +394,36 @@ public void run() {
...
                 result.deferredStart = startRunnable;
+                log.debug(result.message+" (Brooklyn load deferred)");
             } else {
                 startRunnable.run();
+                if (!result.catalogItemsInstalled.isEmpty()) {
+                    // show fewer info messages, only for 'interesting' and non-deferred installations
+                    // (rebind is deferred, as are tests, but REST is not)
+                    MutableList<String> firstFive = MutableList.copyOf(Iterables.limit(result.catalogItemsInstalled, 5));
+                    log.info(result.message+", items: "+firstFive+
+                        (result.catalogItemsInstalled.size() > 5 ? " (and others, "+result.catalogItemsInstalled.size()+" total)" : "") );
```
ahgittin: done

core/src/main/java/org/apache/brooklyn/core/mgmt/ha/OsgiArchiveInstaller.java
ahgittin: GIGO -- adding a bundle should be (and is, at "add catalog" level, though we probably want finer grained) blockable by an RBAC scheme.  but i don't think we need to prevent it at a code level, and someone who has permission to install bundles can break things in many ways, that's not something we need to (or can) prevent.

core/src/main/java/org/apache/brooklyn/core/mgmt/ha/OsgiArchiveInstaller.java
ahgittin: this will happen if you install a `catalog.bom` which doesn't come from OSGi and which declare a bundle.  we should be declaring a `bundle` in all BOMs.  i've added this (some cheekily pushed to other projects).

the odd thing though is that the "delete empty BOMs" logic should be cleaning this up.  i'm looking at that.  (i think all other comments are addressed.)

ahgittin: @geomacy have pushed tests which confirm the logic to delete now-empty bundles is working correctly when it runs for items like this (the idea is if you add a second catalog with the same items as the first, the items become attached to the second so the first one is empty and we delete it).

i suspect in the problem you describe that for some reason it wasn't being triggered, either on startup or on rebind.  if you added another BOM it would probably clean out the old items.

we could run the "delete empties" more often but i'm inclined not to given it's already a warning to add something without a bundle declared and there's a workaround.

I've added the `bundle` declaration at:  https://github.com/apache/brooklyn-dist/pull/99

A related recommended improvement would be to put `catalog.bom` into each of the `brooklyn-library` sub-projects, and ensure that those bundles are referenced and installed as brooklyn-managed bundles in the `default.catalog.bom` (or the `library-*.bom` it calls).  Not that important however as we want to move away from those java library items in any case.  More important for yaml bundles in blueprint repo to follow the new pattern.

geomacy:
The build break on 98d9b0a is genuine, I get the same thing:

```
Failed tests: 
  ApplicationsYamlOsgiTest>ApplicationsYamlTest.testDoesNotWrapCatalogItemAndUsesCatalogName:216->ApplicationsYamlTest.assertDoesNotWrap:436 expected [catalogLevel] but found [Application (wiifyf67tn)]
  ApplicationsYamlOsgiTest>ApplicationsYamlTest.testNameInCatalogMetadata:364->ApplicationsYamlTest.assertDoesNotWrap:436 expected [My name in top-level] but found [Application (co8ot9tzt2)]
  ApplicationsYamlOsgiTest>ApplicationsYamlTest.testNameInItemMetadata:382->ApplicationsYamlTest.assertDoesNotWrap:436 expected [My name in item metadata] but found [Application (u1xjn8fsai)]
  ApplicationsYamlTest.testDoesNotWrapCatalogItemAndUsesCatalogName:216->assertDoesNotWrap:436 expected [catalogLevel] but found [Application (s5iotaeji9)]
  ApplicationsYamlTest.testNameInCatalogMetadata:364->assertDoesNotWrap:436 expected [My name in top-level] but found [Application (ojj1x9bq1v)]
  ApplicationsYamlTest.testNameInItemMetadata:382->assertDoesNotWrap:436 expected [My name in item metadata] but found [Application (hdis9rqcee)]
  CatalogOsgiYamlEntityTest.testOsgiNotLeakingToParent:539 Unexpected error message: java.lang.IllegalStateException: Bundle BasicManagedBundle{symbolicName=org.apache.brooklyn.test.osgi.entities.SimpleEntity, version=0.1.2.update, url=null} failed installation: Error installing catalog items: Failed to install org.apache.brooklyn.test.osgi.entities.SimpleEntity:0.1.2.update, types [BasicRegisteredType[org.apache.brooklyn.test.osgi.entities.SimpleEntity:0.1.2-update;null]] gave errors; 2 errors including: UnsupportedTypePlanException: Invalid plan; format could not be recognized, none of the available transformers [brooklyn-camp:CampTypePlanTransformer, java-type-name:JavaClassNameTypePlanTransformer] support BasicRegisteredType[org.apache.brooklyn.test.osgi.entities.SimpleEntity:0.1.2-update;null] expected [true] but found [false]
  CatalogYamlEntityNameTest.testUsesNameInCatalogItemMetadataIfNonSupplied:50 expected [nameInItemMetadata] but found [BasicEntity:faw9]
  CatalogYamlEntityOsgiTypeRegistryTest>CatalogYamlEntityTest.testAddCatalogItemVerySimple:60 expected 'services:' block: BasicRegisteredType[my.catalog.app.id.load:0.1.2;null]
type: org.apache.brooklyn.entity.stock.BasicEntity expected [true] but found [false]
  CatalogYamlEntityOsgiTypeRegistryTest>CatalogYamlEntityTest.testCreateSpecFromCatalogItem:371 » IllegalState
  CatalogYamlEntityTest.testAddCatalogItemVerySimple:60 expected 'services:' block: BasicRegisteredType[my.catalog.app.id.load:0.1.2;null]
type: org.apache.brooklyn.entity.stock.BasicEntity expected [true] but found [false]
```

ahgittin:
thanks @geomacy - let's watch this run, hopefully tests now all good!

ahgittin:
tests good now @geomacy

core/src/main/java/org/apache/brooklyn/core/catalog/internal/BasicBrooklynCatalog.java
```diff
@@ -792,28 +1026,58 @@ private String setFromItemIfUnset(String oldValue, Map<?,?> item, String fieldAt
...
+        // org.reflections requires the URL to be "file:" containg ".jar"
+        File fJar = Os.newTempFile(containingBundle.getVersionedName().toOsgiString(), ".jar");
+        try {
+            Streams.copy(ResourceUtils.create().getResourceFromUrl(url), new FileOutputStream(fJar));
+        } catch (FileNotFoundException e) {
+            throw Exceptions.propagate("Error extracting "+url+" to scan "+containingBundle.getVersionedName(), e);
+        }
+        subCatalog.addToClasspath(new String[] { "file:"+fJar.getAbsolutePath() });
+        Collection<CatalogItemDtoAbstract<?, ?>> result = scanAnnotationsInternal(mgmt, subCatalog, MutableMap.of("version", containingBundle.getSuppliedVersionString()), containingBundle);
+        fJar.delete();
```
geomacy: Could be put into a `finally` for the `try` at 1056.

core/src/main/java/org/apache/brooklyn/core/catalog/internal/BasicBrooklynCatalog.java
```diff
@@ -1060,28 +1336,87 @@ private String makeAsIndentedObject(String yaml) {
...
+            File bf = bm.createTempBundle(vn.getSymbolicName(), mf, MutableMap.of(
+                new ZipEntry(CATALOG_BOM), (InputStream) new ByteArrayInputStream(yaml.getBytes())) );
+
+            OsgiBundleInstallationResult result = null;
+            try {
+                result = osgiManager.get().install(null, new FileInputStream(bf), true, true, forceUpdate).get();
+            } catch (FileNotFoundException e) {
+                throw Exceptions.propagate(e);
+            }
+            bf.delete();
```
geomacy: put into a `finally`

core/src/main/java/org/apache/brooklyn/core/typereg/BasicManagedBundle.java
```diff
@@ -131,6 +129,7 @@ public String toString() {
 
     @Override
     public int hashCode() {
+        // checksum deliberately omitted here to match with OsgiBundleWithUrl
         return Objects.hashCode(symbolicName, version, url);
```
ahgittin: good spot

ahgittin:
thx @geomacy -- great comments.  all addressed.

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/transform/CatalogTransformer.java
```diff
@@ -67,11 +71,177 @@
...
+        }
+    }
+    private static String tidyIconLink(BrooklynRestResourceUtils b, RegisteredType item, String iconUrl, UriBuilder ub) {
+        if (b.isUrlServerSideAndSafe(iconUrl)) {
+            return serviceUriBuilder(ub, CatalogApi.class, "getIcon").build(item.getSymbolicName(), item.getVersion()).toString();
+        }
+        return iconUrl;
+    }
+
+    private static Set<Object> makeTags(EntitySpec<?> spec, RegisteredType item) {
```
ahgittin: this is done, just GH isn't noticing

core/src/main/java/org/apache/brooklyn/core/typereg/BasicRegisteredType.java
```diff
@@ -153,4 +160,49 @@ public String toString() {
...
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj) return true;
+        if (obj == null) return false;
+        if (getClass() != obj.getClass()) return false;
+        BasicRegisteredType other = (BasicRegisteredType) obj;
+        if (!Objects.equal(aliases, other.aliases)) return false;
+        if (!Objects.equal(bundles, other.bundles)) return false;
+        if (!Objects.equal(containingBundle, other.containingBundle)) return false;
+        if (!Objects.equal(deprecated, other.deprecated)) return false;
```
ahgittin: this is done, just GH isn't noticing

ahgittin:
@geomacy yes, there's an open question how to treat unresolved items, whether to uninstall the bundle (so it could be fixed) or leave them in the unresolved state; i think the latter (current beahviour) is pretty good, but we'll see what we find

as for deletion you're right there's no easy way to delete a bundle, that's a known issue, to be fixed soon.  (you have to stop brooklyn and go into persisted state, or run some hairy groovy.)

so i think merging.

geomacy:
👍  merging

PR #750, opened 2017-06-29
Adds a policy to create locations from an entity

grkvlt:
Adds `CreateLocationPolicy`which configures a location using sensor data from an entity, once the entity is running. The location is added to the catalog for use by other blueprints.

### DIFF
--- a/policy/src/main/java/org/apache/brooklyn/policy/location/CreateLocationPolicy.java
+++ b/policy/src/main/java/org/apache/brooklyn/policy/location/CreateLocationPolicy.java
@@ -225,6 +225,7 @@ public class CreateLocationPolicy extends AbstractPolicy {
 
     @Override
     public void setEntity(EntityLocal entity) {
+        super.setEntity(entity);
         subscriptions().subscribe(entity, getStatusSensor(), lifecycleEventHandler);
     }
 }
--- a/policy/src/main/java/org/apache/brooklyn/policy/location/CreateLocationPolicy.java
+++ b/policy/src/main/java/org/apache/brooklyn/policy/location/CreateLocationPolicy.java
@@ -22,9 +22,9 @@ import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicBoolean;
 
+import org.apache.brooklyn.api.catalog.CatalogItem;
 import org.apache.brooklyn.api.entity.EntityLocal;
 import org.apache.brooklyn.api.location.Location;
-import org.apache.brooklyn.api.location.LocationSpec;
 import org.apache.brooklyn.api.policy.PolicySpec;
 import org.apache.brooklyn.api.sensor.AttributeSensor;
 import org.apache.brooklyn.api.sensor.SensorEvent;
@@ -37,15 +37,31 @@ import org.apache.brooklyn.util.collections.MutableMap;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.google.common.base.Joiner;
 import com.google.common.base.Predicates;
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Iterables;
 import com.google.common.reflect.TypeToken;
 
 /**
  * Policy that is attached to an entity to create a location configured
  * using its sensor data. The created lifecycle is added to the catalog
  * and will be removed when the entity is no longer running.
+ * 
+ *
+ * in YAML blueprints:
+ * <pre>{@code
+ * name: My App
+ * brooklyn.policies:
+ * - type: org.apache.brooklyn.policy.location.CreateLocationPolicy
+ *   brooklyn.config:
+ *     location.catalogId: my-cloud
+ *     location.displayName: My Cloud
+ *     location.type: jclouds:aws-ec2
+ * }</pre>
+ *
  */
 public class CreateLocationPolicy extends AbstractPolicy {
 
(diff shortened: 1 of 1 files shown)


drigodwin:
This looks really useful but it could probably do with some tests if that's possible @grkvlt?

grkvlt:
Yes, that's probably useful - I'll put something together

policy/src/main/java/org/apache/brooklyn/policy/location/CreateLocationPolicy.java
```diff
@@ -0,0 +1,288 @@
...
+                                "  version: " + version.incrementAndGet(),
+                                "  itemType: location",
+                                "  item:",
+                                "    type: " + getLocationType(),
+                                "    brooklyn.config:",
+                                "      displayName: " + getLocationDisplayName());
+
+                        for (Map.Entry<String, AttributeSensor<?>> entry : getLocationConfiguration().entrySet()) {
+                            AttributeSensor<?> sensor = getLocationConfiguration().get(entry.getKey());
+                            Object value = entity.sensors().get(sensor);
```
andreaturli: what happens if the sensor is not yet populated, wouldn't be safer to require directly value so we can use `attributeWhenReady` in the blueprint? or maybe we can always assume that once `LOCATION_STATUS` is true, brooklyn would have sensor data ready?

andreaturli:
I agree with @drigodwin about the tests :)

FYI I've tested the `CreateLocationPolicy` and it works fine in some scenarios, only a minor comment.

andreaturli:
@grkvlt any updates on this?

policy/src/main/java/org/apache/brooklyn/policy/location/CreateLocationPolicy.java
```diff
@@ -0,0 +1,288 @@
...
+            this.displayName = val; return this;
+        }
+        public Builder type(String val) {
+            this.type = val; return this;
+        }
+        public Builder tags(Set<String> val) {
+            this.tags = val; return this;
+        }
+        public CreateLocationPolicy build() {
+            return new CreateLocationPolicy(toFlags());
```
aledsage: Delete this `build` method - folk should always use the `PolicySpec`. I'd probably go further and delete the builder entirely, relying on using the standard `PolicySpec` as the way to build it.

policy/src/main/java/org/apache/brooklyn/policy/location/CreateLocationPolicy.java
```diff
@@ -0,0 +1,288 @@
...
+
+    public static final ConfigKey<String> LOCATION_DISPLAY_NAME = ConfigKeys.builder(String.class)
+            .name("location.displayName")
+            .description("The display name for the created location")
+            .constraint(Predicates.notNull())
+            .build();
+
+    public static final ConfigKey<String> LOCATION_TYPE = ConfigKeys.builder(String.class)
+            .name("location.type")
+            .description("The type of location to create, i.e. 'jclouds:aws-ec2'")
```
aledsage: "e.g." rather than "i.e."

policy/src/main/java/org/apache/brooklyn/policy/location/CreateLocationPolicy.java
```diff
@@ -0,0 +1,288 @@
...
+    private AtomicBoolean created = new AtomicBoolean(false);
+    private Object mutex = new Object[0];
+    private CatalogItem<?,?> catalogItem;
+    private AtomicInteger version = new AtomicInteger(0);
+
+    public CreateLocationPolicy() {
+        this(MutableMap.<String,Object>of());
+    }
+
+    public CreateLocationPolicy(Map<String,?> props) {
```
aledsage: Remove this constructor - rely on `PolicySpec` for everything.

policy/src/main/java/org/apache/brooklyn/policy/location/CreateLocationPolicy.java
```diff
@@ -0,0 +1,288 @@
...
+    protected <T> void doReconfigureConfig(ConfigKey<T> key, T val) {
+        throw new UnsupportedOperationException("reconfiguring "+key+" unsupported for "+this);
+    }
+
+    private final SensorEventListener<Boolean> lifecycleEventHandler = new SensorEventListener<Boolean>() {
+        @Override
+        public void onEvent(SensorEvent<Boolean> event) {
+            synchronized (mutex) {
+                Boolean status = event.getValue();
+                if (status) {
```
aledsage: `if (Boolean.TRUE.equals(status))`

policy/src/main/java/org/apache/brooklyn/policy/location/CreateLocationPolicy.java
```diff
@@ -0,0 +1,288 @@
...
+                        if (items > 1) {
+                            LOG.warn("Got {} catalog items for location {}, using first: {}",
+                                    new Object[] { items, getLocationCatalogItemId(), Iterables.transform(catalogItems, item -> item.getCatalogItemId()) });
+                        } else if (items == 0) {
+                            throw new IllegalStateException("No catalog items returned for location: " + getLocationCatalogItemId());
+                        }
+                        catalogItem = Iterables.get(catalogItems, 0);
+                    }
+                } else {
+                    if (created.compareAndSet(true, false) && catalogItem != null) {
```
aledsage: `created` and `catalogItem` fields won't survive rebind. Neither will the auto-incrementing version. Therefore this will likely misbehave on rebind.

grkvlt:
Thanks for the review @aledsage, I will update appropriately. The use case is for blueprints like Kubernetes, which would be able to add a new `KubernetesLocation` or similar.

tbouron:
@grkvlt Did you had the time to looks at @aledsage's comments for this PR?

nakomis:
@grkvlt Could you take a look at this to see if it's still relevant, and address the comments please

Thanks

PR #751, opened 2017-06-30
Adds method to reset the shutdown flag within BrooklynShutdownHooks

m4rkmckenna:
Under karaf if you restarted the brooklyn-init bundle
either manually or due to a re-wire the management context would not
restart (requiring you to resart karaf). This change allows the OSGi
launcher to reset the shutdown flag enabling bundle / management context restart

See [BROOKLYN-522](https://issues.apache.org/jira/browse/BROOKLYN-522)

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/mgmt/internal/BrooklynShutdownHooks.java
+++ b/core/src/main/java/org/apache/brooklyn/core/mgmt/internal/BrooklynShutdownHooks.java
@@ -18,11 +18,8 @@
  */
 package org.apache.brooklyn.core.mgmt.internal;
 
-import java.util.List;
-import java.util.Set;
-import java.util.concurrent.Semaphore;
-import java.util.concurrent.atomic.AtomicBoolean;
-
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.Lists;
 import org.apache.brooklyn.api.entity.Entity;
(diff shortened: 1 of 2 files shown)


aledsage:
LGTM. Once jenkins has finished its checks, I'm happy for this to be merged.

core/src/main/java/org/apache/brooklyn/core/mgmt/internal/BrooklynShutdownHooks.java
```diff
@@ -100,6 +99,17 @@ public static void invokeStopOnShutdown(Entity entity) {
         }
     }
 
+    public static void resetShutdownFlag() {
```
geomacy: might be nice to make this `resetShutdownFlag(boolean newValue)` and then also use it in [run](https://github.com/m4rkmckenna/brooklyn-server/blob/7068308917b7424a4ff3cdab1f4416e45508c399/core/src/main/java/org/apache/brooklyn/core/mgmt/internal/BrooklynShutdownHooks.java#L190-L197).

geomacy: I guess you'd need both boolean parameters for the `compareAndSet`.

It's just a suggestion, matter of taste really, certainly not needed as such.

geomacy:
I think this is fine the way it is- merging.

PR #752, opened 2017-07-03
Cache the default region in location config

neykov:
The default region is expensive to retrieve when one is not set by the user. Very often it needs expensive clouds calls (e.x. retrieve list of images).
Cache the value so it can easily be used by code that needs it. It can be used by customizers in starting from `customizer(template)` step or any other code after the first `obtain` call.


neykov:
Closing because will not consistently set the property. When the location doesn't have a region set, but one is passed through `flags` (for example in `provisioning.properties`) we will not be able to obtain jclouds' default (as a non-default value is already passed in from `flags`).

PR #756, opened 2017-07-05
Support recursive option and limits when getting activities

ahgittin:
See `ActivityApi` and `EntityApi` where this adds better support for recursive task fetching, and an optional `limit` query parameter with default of 200 tasks; long lists of tasks are painful in the UI and on the server.  Recursive task call also has optional `maxDepth` (default unlimited).

This affects backwards compatibility if clients expect all tasks to be returned from these expensive calls, but they shouldn't.

## Pagination

This does not go so far as to support pagination but that would not be hard and we should start thining about it.  One strategy I like is to have _two_ optional parameters, `itemToContinueAfter` and `countToContinueAfter`, as follows, assuming the data is `A`, `B`, `C`.

Obvious things:

* request `limit=1` and `itemToContinueAfter=B`: gives `[C]`
* request `limit=1` and `countToContinueAfter=2`: gives `[C]`

If `A` is deleted:

* request `limit=1` and `itemToContinueAfter=B`: gives `[C]`
* request `limit=1` and `countToContinueAfter=2`: gives `[]`

If `A` is present but `B` is deleted:

* request `limit=1` and `itemToContinueAfter=B`: gives `[A]` (restarts if item is missing)
* request `limit=1` and `countToContinueAfter=2`: gives `[]`

If _both_ `item` and `count` are supplied we use `item` if it is present, otherwise we use `count`.  This lets clients do well when working with a changing list, paging forwards will continue from the previous item, never skipping anything unless that previous item was deleted in which rare case it skips some some but it doesn't completely restart.

(Are there better strategies?  The benefit here is that it is simple for both client and server.)


ahgittin:
retest please

ahgittin:
retest this please

ahgittin:
weird unrelated build failure at https://builds.apache.org/job/brooklyn-server-pull-requests/2396/console, new `container` location project had a dependency on jclouds `2.0.2` but other things were using `2.0.1` cc @grkvlt -- can't see why, poms look okay

retest this please

ahgittin:
another unrelated failure `org.apache.brooklyn.util.pool.BasicPoolTest.testConcurrentCallsNeverHaveSameVal`

retest this please

core/src/main/java/org/apache/brooklyn/util/core/task/Tasks.java
```diff
@@ -492,4 +494,29 @@ private static String getTimeoutString(Repeater repeater) {
         return "in "+timeout;
     }
 
+
+    public static void dumpInfo(Task<?> t) {
```
aledsage: Minor (and not for this PR): we should probably move this stuff into a separate `Dumper` utility class, rather than `Entities` and `Tasks` being a dumping ground for lots of stuff.

PR #757, opened 2017-07-07
[WIP] BROOKLYN-519 no BasicSpecParameter serialization

bostko:
BasicSpecParameters are only consumed in API
 and no need to keep them or keep their state.

- TODO rebind tests

- TODO test deserialization of a catalog item with
  brooklyn.parameters in it.

- Long term TODO - not serializing SpecParameters.
  (Will need bigger discussion in Apache Brooklyn mailing list)
  Two ways to do that is either by using @XStreamOmitField or
  by setting AbstractBrooklynObjectSpec#parameters transient.
  However `transient' change besides not serializing parameters
  it also do not deserialize the parameters field which cause a backward compatibility problem.

  Backward compatibility with transient `parameters`,
  reproduceable in test `org.apache.brooklyn.entity.group.EntitySpecRebindTest`.
  Causing problem with class object references deserialization, see https://x-stream.github.io/graphs.html

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/mgmt/persist/XmlMementoSerializer.java
+++ b/core/src/main/java/org/apache/brooklyn/core/mgmt/persist/XmlMementoSerializer.java
@@ -26,6 +26,7 @@ import java.util.Map;
 import java.util.NoSuchElementException;
 import java.util.concurrent.ExecutionException;
 
+import com.google.common.collect.ImmutableList;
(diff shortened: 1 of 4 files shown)


bostko:
Approach suggested on 07/07/2017 to dev@brooklyn.apache.org mailing list with subject "BROOKLYN-519 suggestion to remove BasicSpecParameter from serializing" is more suitable for solving BROOKLYN-519.
Closing PR.

PR #758, opened 2017-07-07
Adds patent pom relative path to `brooklyn-locations-container`

m4rkmckenna:
Without this the build will fail when preforming a release


geomacy:
merging

PR #760, opened 2017-07-11
Add DigitalOcean2 and ProfitBricks features to Karaf build

sjcorbett:


andreaturli:
thanks @sjcorbett -- would it be useful to add a comment in the `feature.xml` to remember to use `jclouds-all-compute` from https://github.com/jclouds/jclouds-karaf/blob/master/feature/src/main/feature/feature.xml and `jclouds-labs-all-compute` from https://github.com/jclouds/jclouds-karaf/blob/master/feature-labs/src/main/feature/feature.xml when we bump to `jclouds 2.1.x` ?

drigodwin:
I think that is a good point but probably outside the scope of this PR @andreaturli so it shouldn't stop this being merged. This looks fine to me, so I'm going to merge now, thanks @sjcorbett.

PR #761, opened 2017-07-13
Async entities

aledsage:
Adds `AsyncApplication`: An app that starts up asynchronously. Calling start will call start on its children, but it does not expect the children to have started by the time the start() effector has returned. Instead, it infers from the children's state whether they are up or not.

An example use-case is where the app kicks off some async children. These children call out to whichever system they integrate with (e.g. provisioning resources), and then several hours later get callbacks to say whether it started/succeeded. Instead of leaving the `start()` method running for several hours, we let the start method return and then we transition automatically to the right state. I didn't want to leave the `start()` method executing for a long time, because then on Brooklyn-restart the effector would have been interrupted and it would not handled it well!

---
The computation of service-up and service-state (i.e. the enrichers) is very different for the async app. As the javadoc says:
```
Calculates the "service.state" and "service.isUp", based on the state of the children. It 
also transitions from expected=starting to expected=running once all of the children have
finished their starting.

This does <em>not</em> just rely on the "service.problems" and "service.notUp.indicators"
because those make different assumptions about the expected state. Instead it seems much
easier to implement the specific logic for async startup here.
```
Before resorting to this approach, I tried to reuse/adapt the standard enrichers that are in `AbstractApplication` and `ServiceStateLogic`. However, I would either have made that complicated code much more complicated, or reimplemented a huge chunk as an alternative enricher. The big difference is that we want to automatically transition to `expected=running`.

Overall, it is therefore simpler to implement the specific enricher for AsyncApplication.

I also did *not* use the same pattern for `service.notUp.indicators` and `service.problems`. Instead of using these to infer whether we are up/healthy, it sets `service.isUp` and `service.state` directly. It also sets the indicators to explain why we are in the given state. This is a far simpler pattern when we can compute our state directly, and where we don't expect users to customize it with completely different ways of telling if we're up. In the past, the race conditions and complexity of entities using separate enrichers for indicators caused us a lot of pain in non-deterministic tests etc (which we ironed out over a long period of time).

---
The other commit (`AsyncStartable marker interface`) is important for if we restart Brooklyn or if we failover in HA. On rebind, a normal entity that is "starting" will be set as "on-fire" because it's effector will not automatically resume. That behaviour is not applicable for `AsyncStartable`.

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/entity/stock/AsyncApplicationImpl.java
+++ b/core/src/main/java/org/apache/brooklyn/entity/stock/AsyncApplicationImpl.java
@@ -1,0 +1,479 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.brooklyn.entity.stock;
+
+import static org.apache.brooklyn.core.entity.Attributes.SERVICE_STATE_ACTUAL;
+import static org.apache.brooklyn.core.entity.lifecycle.ServiceStateLogic.clearMapSensorEntry;
+import static org.apache.brooklyn.core.entity.lifecycle.ServiceStateLogic.updateMapSensorEntry;
+
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+import org.apache.brooklyn.api.entity.Entity;
+import org.apache.brooklyn.api.entity.EntityLocal;
+import org.apache.brooklyn.api.location.Location;
+import org.apache.brooklyn.api.sensor.EnricherSpec;
+import org.apache.brooklyn.api.sensor.Sensor;
+import org.apache.brooklyn.api.sensor.SensorEvent;
+import org.apache.brooklyn.api.sensor.SensorEventListener;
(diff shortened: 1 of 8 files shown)


core/src/main/java/org/apache/brooklyn/entity/stock/AsyncApplicationImpl.java
```diff
@@ -0,0 +1,473 @@
...
+     * This class is not to be instantiated directly. Instead, if cusotmization is desired then override 
+     * {@link AsyncApplicationImpl#initEnrichers()} to create and add this enricher (with the same unique 
+     * tag, to replace the default).
+     */
+    public static class ServiceStateComputer extends AbstractMultipleSensorAggregator<Void> implements SensorEventListener<Object> {
+        /** standard unique tag identifying instances of this enricher at runtime */
+        public final static String DEFAULT_UNIQUE_TAG = "async-service-state-computer";
+
+        public static final ConfigKey<QuorumCheck> RUNNING_QUORUM_CHECK = ConfigKeys.builder(QuorumCheck.class, "runningQuorumCheck")
+//              .description("Logic for checking whether this service is healthy, based on children and/or members running, defaulting to requiring none to be ON-FIRE")
```
drigodwin: Why are these commented and should they be here if they are not required?

aledsage: oops, good spot - will fix those! That's the previous copy-pasted description. I meant to update those comments!

core/src/main/java/org/apache/brooklyn/entity/stock/AsyncApplicationImpl.java
```diff
@@ -0,0 +1,473 @@
...
+                .configure(ServiceStateComputer.UP_QUORUM_CHECK, config().get(UP_QUORUM_CHECK))
+                .configure(ServiceStateComputer.RUNNING_QUORUM_CHECK, config().get(RUNNING_QUORUM_CHECK)));
+
+    }
+
+    // Only overrides AbstractApplication.start so as to disable the publishing of expected=running
+    @Override
+    public void start(Collection<? extends Location> locations) {
+        this.addLocations(locations);
+        // 2016-01: only pass locations passed to us, as per ML discussion
```
aledsage: Copy-pasted from AbstractApplication.start. I'll add to the method's comment:
`Code is copy-pasted verbatim from AbstractAppliation, except for not setting Lifecycle.RUNNING.`

aledsage:
Thanks @drigodwin - comments addressed (adding more explanation to `AsyncApplicationImpl.start`, and fixing the config key descriptions). Squashed and pushed here as a single commit.

PR #763, opened 2017-07-14
Add 'first' to aggregator

aledsage:
Use case is that we have a cluster - we want to propagate the sensor form any of the cluster's members, where that value is not null (in the same format as the child's value).

This is important for the cluster to be used in the same way as a simple child might be used (for "substitutability" of entity types, in object-oriented terms).


drigodwin:
I've added a PR to change the docs to match this [here](https://github.com/apache/brooklyn-docs/pull/203).

PR #765, opened 2017-07-17
REST API returns plane-id

aledsage:
This is useful for systems that use/manage Brooklyn - they want to know the stable id of that Brooklyn server, which will be the same after restarts or after HA failovers.


neykov:
Is it worth adding it to `/ha/states` as well, along `nodeId`? LGTM.

aledsage:
@neykov not sure what you mean. The second commit adds it to `/ha/states`, at the top-level as `planeId`. In that response, it doesn't repeat the planeId in the `nodes` map (alongside `nodeId`).

neykov:
:+1: somehow I've been viewing only the first commit. All good!

PR #767, opened 2017-07-19
Adds BrooklynViewerLauncher (for tests)

aledsage:
As per the javadoc:
```
/**
 * A convenience for started the Brooklyn REST api and web-app in a test, so that one can visually
 * inspect the app that the test creates. This is intended as a read-only view (but it has the real
 * management context so one can perform actions through this UI).
 * 
 * It expects to be configured with an existing {@link org.apache.brooklyn.api.mgmt.ManagementContext}.
 * 
 * Various other configuration options (e.g. {@link #application(String)) will be ignored.
 * 
 * An example of where this is very useful is when testing a blueprint where an effector hangs - 
 * one can visually inspect the app, drilling into the activities view to see what it is doing
 * and why it is blocked.
```

Note that `AbstractBlueprintTest` used to do this with `BrooklynLauncher`, but that is broken - that launcher also tries to set the `managementPlaneId` on the management context, but it's already set so it throws an exception.

My motivation for fixing + adding this is mostly for use in downstream projects. I want to run unit/integration tests that deploy + test some complicated blueprints. It's very useful to have this visualisation in such tests (particularly if it fails, and one places a breakpoint to see what activities have executed by that point).

### DIFF
--- a/launcher-common/src/main/java/org/apache/brooklyn/launcher/common/BasicLauncher.java
+++ b/launcher-common/src/main/java/org/apache/brooklyn/launcher/common/BasicLauncher.java
@@ -125,7 +125,7 @@ public class BasicLauncher<T extends BasicLauncher<T>> {
     private Duration haHeartbeatTimeoutOverride = null;
     private Duration haHeartbeatPeriodOverride = null;
     
-    private boolean started;
+    protected boolean started;
     
     private BrooklynProperties.Factory.Builder brooklynPropertiesBuilder;
 
--- a/launcher/src/main/java/org/apache/brooklyn/launcher/BrooklynLauncher.java
+++ b/launcher/src/main/java/org/apache/brooklyn/launcher/BrooklynLauncher.java
@@ -74,6 +74,7 @@ import com.google.common.collect.Maps;
  *     .start();
  * 
  * Entities.dumpInfo(launcher.getApplications());
+ * }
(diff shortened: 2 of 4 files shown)


neykov:
Looks a lot like `BrooklynJavascriptGuiLauncher`, but otherwise LGTM.

aledsage:
@neykov good point about `BrooklynJavascriptGuiLauncher` (and the `BrooklynRestApiLauncher` that it delegates to). I'd forgotten about those.

The existing `BrooklynJavascriptGuiLauncher` is only used in a couple of simple tests (in `BrooklynJavascriptGuiLauncherTest`). It is not configurable to inject a management context, which is what I want to do (obviously we could add that if we chose).

The `BrooklynRestApiLauncher` can be configured with a management context but does not include the UI. There is a surprisingly large amount of code in that class! I presume it's duplicating what is in `BrooklynLauncher`. I therefore much prefer my approach of just extending `BrooklynLauncher`, overriding `start()`.

At some point we could try re-writing `BrooklynRestApiLauncher` to extend `BrooklynLauncher`, but just now I don't want to risk accidentally breaking all the rest-api tests that use `BrooklynRestApiLauncher`. I suggest we come back to that in another PR in the future.

@neykov are you ok with merging this as-is?

neykov:
+1, merging.

PR #768, opened 2017-07-19
Improve localhost detection and webserver binding

ahgittin:
Fixes two annoying network issues:

* Sometimes when disconnected Brooklyn detects a non-usable NIC address as localhost and things like SSH fail subsequently
* If start-use-stop-immediately_restart brooklyn, it will normally take a different port eg 8082 because the first one eg 8081 will be in `TIME_WAIT` state.  Allow webserver to `REUSE_ADDR` so it takes the same port in this case.  Minor but hits me a lot when I'm testing!


ahgittin:
retest this please

failure:

```
java.lang.AssertionError: attribute=Sensor: a.sensor (java.lang.String); val=null
	at org.apache.brooklyn.camp.brooklyn.catalog.CatalogMakeOsgiBundleTest.assertBasic1DeploysAndHasSensor(CatalogMakeOsgiBundleTest.java:123)
	at org.apache.brooklyn.camp.brooklyn.catalog.CatalogMakeOsgiBundleTest.testCatalogBomFromBundleWithManifest(CatalogMakeOsgiBundleTest.java:114)
```

unrelated, maybe needs an `Eventually` ?

PR #772, opened 2017-07-21
Allow equivalent plans when replacing

ahgittin:
Needed because our routines sometimes rewrite the plan


ahgittin:
Follow-up to #746 making it slightly more forgiving when replacing, should fix issues when dealing with complex catalogs/blueprints

geomacy:
test failure is real, in `CatalogResourceTest`

geomacy:
For `testOsgiBundleWithBom` it's just an update needed to the test expectation - there are now more tags than before on the result:
```
    "equivalent-plan(A8FA402166F072F19CFE88AEA200B001)",
    "equivalent-plan(A74550A5F94D4AD144AC5EB816DE6C45)",
    {
      "traits": [
        ...
      ]
    }
```

ahgittin:
thanks @geomacy - fixed and merged

PR #773, opened 2017-07-22
prevent benign error in file deletion shutdown-hook

ahgittin:
- ensure required bundle has been loaded before shutdown
- fail gracefully if the deletion hook fails

### DIFF
--- a/utils/common/src/main/java/org/apache/brooklyn/util/os/Os.java
+++ b/utils/common/src/main/java/org/apache/brooklyn/util/os/Os.java
@@ -49,6 +49,7 @@ import com.google.common.base.Splitter;
 import com.google.common.base.Throwables;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Iterables;
+import com.google.common.collect.Ordering;
 import com.google.common.io.ByteStreams;
 import com.google.common.io.Files;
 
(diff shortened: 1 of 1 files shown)


utils/common/src/main/java/org/apache/brooklyn/util/os/Os.java
```diff
@@ -317,21 +318,28 @@ public void run() throws IOException {
...
+                                for (String path: pathsToDelete) {
+                                    try {
+                                        deletions.remove(path).run();
+                                    } catch (Exception e) {
+                                        log.warn("Unable to delete '"+path+"' on shutdown: "+e);
+                                    }
                                 }
                             }
+                        } catch (Exception e) {
+                            log.warn("Unable to delete one or mort paths ("+deletions+") on shutdown: "+e);
```
drigodwin: mort -> more

utils/common/src/main/java/org/apache/brooklyn/util/os/Os.java
```diff
@@ -317,21 +318,28 @@ public void run() throws IOException {
...
+                                Collections.sort(pathsToDelete, Strings.lengthComparator().reverse());
+                                for (String path: pathsToDelete) {
+                                    try {
+                                        deletions.remove(path).run();
+                                    } catch (Exception e) {
+                                        log.warn("Unable to delete '"+path+"' on shutdown: "+e);
+                                    }
                                 }
                             }
+                        } catch (Exception e) {
```
aledsage: I'd prefer to include `, e`, rather than just its toString. For example, if there were an NPE somewhere in the method then we'd only be told that there was an NPE rather than being told its line number etc.

ahgittin:
both done, thanks guys.  aled, i added the `, e` as a debug message.

merging.

PR #775, opened 2017-07-22
Useful getDataAsStringMap for VaultExternalConfigSupplier

bostko:
`getDataAsStringMap` allows obtaining vault data with single call.


drigodwin:
Hi @bostko, I'm not clear what this is for. Could you give a description of why you want this and perhaps a test? Thanks.

bostko:
Useful when consumer of the VaultConfigSuppliler wants to have finer control over number of calls made to vault.
Also this method returns Map with parameters on which a consumer can use `containsKey` to check whether key is available.

PR #777, opened 2017-07-24
Removes `FIRST` sensors on children

ahgittin:
They were buggy and could deadlock -- and I think fundamentally flawed (as comments there suggested).

@grkvlt I think you added these; can you check that the workarounds suggested in the last commit is sufficient?

There are a few other minor tidies also.

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/entity/lifecycle/ServiceStateLogic.java
+++ b/core/src/main/java/org/apache/brooklyn/core/entity/lifecycle/ServiceStateLogic.java
@@ -611,8 +611,20 @@ public class ServiceStateLogic {
             }
 
             return "Required entit"+Strings.ies(onesNotHealthy.size())+" not healthy: "+
-                (onesNotHealthy.size()>3 ? onesNotHealthy.get(0)+" and "+(onesNotHealthy.size()-1)+" others"
-                    : Strings.join(onesNotHealthy, ", "));
+                (onesNotHealthy.size()>3 ? nameOfEntity(onesNotHealthy.get(0))+" and "+(onesNotHealthy.size()-1)+" others"
+                    : Strings.join(nameOfEntity(onesNotHealthy), ", "));
+        }
+
+        private List<String> nameOfEntity(List<Entity> entities) {
+            List<String> result = MutableList.of();
+            for (Entity e: entities) result.add(nameOfEntity(e));
+            return result;
+        }
+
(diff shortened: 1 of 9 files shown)


ahgittin:
The deadlock causes important blueprints to fail so I am going to commit this now.

/cc @grkvlt @aledsage if one of you could do commit-then-review I'd appreciate it!

I'll be posting a follow-up which adds policies around managing primary elements.  That should provide an even nicer way to do some of the things the removed code was trying to do.

rdowner:
@ahgittin this appears to go against our published deprecation policy and is known to break some blueprints.

Your PR comments are making references to previous discussions, but it's not clear where those are, and I haven't been able to find them. Example: `as comments there suggested` in the first comment; `the new "primary" model` is mentioned in your last commit but no explanation of this model; `causes important blueprints to fail` but does not say which ones. So unfortunately I don't yet understand why this breaking change is needed.

I think we should prepare a response (explanation, workaround, new best practice recommendation) to the fact that we've broken some of our user's blueprints, with a clear justification of why it had to be done. I'd like to see this discussed on the mailing list, and noted in [the release notes](https://github.com/apache/brooklyn-docs/blob/master/guide/misc/release-notes.md), and documentation too.

Thanks

ahgittin:
@rdowner Release notes updated now in https://github.com/apache/brooklyn-docs/pull/208 .  ML discussion triggered by @tbouron (TY) and you are completely correct, that should have been done in the first instance.  In my defence I was exhausted after fixing the deadlock, and the (massive) problems with the removed code (including no tests as you'll note I haven't removed any tests in this PR!) should have been found during a code review when this was added.  But it is a poor excuse.  TY for policing this.

As for why:  Consider if an entity is added to two groups.  It is unclear then what `cluster.first: true` means.  A blueprint relying on that in one group may suddenly start failing if you use dynamic membership to sort entities in another way.  Or if you have a group of groups, the problem gets very bad as `cluster.first` would mean both the first in your parent's group and the first in your group.  On top of that the deadlock with parent setting child's sensors while holding the members lock was common and bad.  I think it's not good practise to update someone else's sensors in the best of times, but to do it while holding a synch lock on an object related to the child is definitely asking for trouble.  And not to have _any_ tests....

tbouron:
@ahgittin NP, I don't contest the why, I'm sure there was a good reason (confirmed by the above explanation) I was just surprised that this wasn't flagged before for downstream project or blueprints, that's all.

Anyway, thank you for the release notes

PR #778, opened 2017-07-25
[WIP] Adds ‘deploymentUid’

aledsage:
As discussed on the mailing list: https://lists.apache.org/thread.html/8eb58818d3ba50cfa5d5111363258cfe8529a76d17d45dc01823e0ba@%3Cdev.brooklyn.apache.org%3E

This PR adds support for deploying YAML apps via the rest api, to include `&deploymentUid=...`. This gets turned into a config key named `org.apache.brooklyn.deploymentUid` set on the app being deployed.

If there is already an app in existence with that deploymentUid, then the deployment fails (*before* creating the entity).

---
An alternative (simpler) implementation would be to allow the internal entity-id to be specified via such a query parameter. There is already code to fail-fast if an entity with a given id already exists.

However, we removed (the deprecated) support for supplying the entity id in the `EntitySpec`. We'd presumably have to add that back in (perhaps as "internal", somehow, rather than on the main `EntitySpec` class).


aledsage:
Closing this PR - consensus on the mailing list is that we'll instead support setting the app-id. I'll work on that and submit a new PR.

aledsage:
Replaced by https://github.com/apache/brooklyn-server/pull/782

PR #779, opened 2017-07-26
Exclude Vault headers trace log

bostko:


bostko:
Closing in favor of #780

PR #780, opened 2017-07-26
Removes all logging of vault responses

m4rkmckenna:
Previously body & headers were logged now only the URI and status code is logged

(builds upon #774 )


bostko:
LGTM.

PR #781, opened 2017-07-27
Support DSL in template options

geomacy:
At the moment it's not possible to specify DSL in template options - for example you can't do 

      - type: some-type-publishing-subnet-id
        id: ent1

      - type: some-service
        id: service
        brooklyn.config:
          provisioning.properties:
            templateOptions:
              subnetId: $brooklyn:entity("ent1").attributeWhenReady("subnet.id")

see new test [here](https://github.com/geomacy/brooklyn-server/commit/d2f5bf9e0a5038208ff98aedb16795c217d01fc7#diff-99cb6184bb386646b61f6678346ec63bR75).

This adds support in `TemplateOptionsOption` for resolving the values supplied for the options. This allows the approach in the blueprint above to be used.

### DIFF
--- a/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/templates/customize/TemplateOptionsOption.java
+++ b/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/templates/customize/TemplateOptionsOption.java
@@ -20,15 +20,25 @@
 package org.apache.brooklyn.location.jclouds.templates.customize;
 
 import java.util.Map;
+import java.util.concurrent.ExecutionException;
 
+import org.apache.brooklyn.api.mgmt.ExecutionContext;
+import org.apache.brooklyn.api.mgmt.ManagementContext;
+import org.apache.brooklyn.api.mgmt.Task;
+import org.apache.brooklyn.core.entity.EntityInternal;
+import org.apache.brooklyn.core.mgmt.BrooklynTaskTags;
+import org.apache.brooklyn.core.mgmt.internal.LocalManagementContext;
 import org.apache.brooklyn.util.core.config.ConfigBag;
 import org.apache.brooklyn.util.core.flags.MethodCoercions;
+import org.apache.brooklyn.util.core.task.BasicExecutionContext;
+import org.apache.brooklyn.util.core.task.Tasks;
+import org.apache.brooklyn.util.exceptions.Exceptions;
 import org.apache.brooklyn.util.guava.Maybe;
 import org.jclouds.compute.options.TemplateOptions;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-class TemplateOptionsOption implements TemplateOptionCustomizer {
+public class TemplateOptionsOption implements TemplateOptionCustomizer {
     private static final Logger LOG = LoggerFactory.getLogger(TemplateOptionsOption.class);
 
     @Override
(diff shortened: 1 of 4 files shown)


locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/templates/customize/TemplateOptionsOption.java
```diff
@@ -39,8 +45,17 @@ public void apply(TemplateOptions options, ConfigBag config, Object v) {
...
+            if (optionValue != null) {
+
+                if (optionValue instanceof TaskFactory<?>) {
+                    final TaskAdaptable<?> taskA = ((TaskFactory<?>) optionValue).newTask();
+                    optionValue = getValueFromTask(taskA.asTask());
+                } else if (optionValue instanceof TaskAdaptable<?>) {
+                    optionValue = getValueFromTask(((TaskAdaptable<?>) optionValue).asTask());
+                }
+
+                Maybe<?> result = MethodCoercions.tryFindAndInvokeBestMatchingMethod(options, option.getKey(), optionValue);
```
neykov: Can you use `ValueResolver` instead? For example:
`Tasks.resolveDeepValue(value, Object.class, mgmt.getServerExecutionContext());` (or alternatively the context entity's execution context).

geomacy: Ah! Don't know why I didn't think of that. Nicer. I think I can pass null as the context and it will use the current exectution context?

neykov: Don't think you can use `null` here. Seems the thread context is something used around immediate values only. cc @ahgittin.

Looking at the code the canonical usage is with: `((EntityInternal)BrooklynTaskTags.getTargetOrContextEntity(Tasks.current())).getExecutionContext()`.

geomacy: ok that's a new commit added now with the `null` replaced with the `getTargetOrContextEntity` context

geomacy:
Retest this please.

geomacy:
Build failure is Jenkins related, I think. 

     # There is insufficient memory for the Java Runtime Environment to continue.
    # Cannot create GC thread. Out of system resources.

Retest this please

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/templates/customize/TemplateOptionsOption.java
```diff
@@ -39,8 +49,22 @@ public void apply(TemplateOptions options, ConfigBag config, Object v) {
...
-                Maybe<?> result = MethodCoercions.tryFindAndInvokeBestMatchingMethod(options, option.getKey(), option.getValue());
+            Object optionValue = option.getValue();
+            if (optionValue != null) {
+
+                try {
+                    final EntityInternal entity = (EntityInternal) BrooklynTaskTags.getTargetOrContextEntity(Tasks.current());
+                    final ExecutionContext exec = (null != entity
+                        ? entity.getExecutionContext()
+                        : BasicExecutionContext.getCurrentExecutionContext());
+                    if (exec != null) {
```
aledsage: I wonder if there is some way we could log.warn if `exec != null` and if the value is a deferred supplier. But not sure how easy it is to do for the latter (looking at `ValueResolver`, it handles `v instanceof ImmediateSupplier || v instanceof DeferredSupplier || v instanceof TaskFactory || v instanceof TaskAdaptable || v instanceof Task || v instanceof Future`, but it also recursively looks inside iterables and maps.

Therefore maybe just ignore this comment :-(

geomacy:
Thanks @aledsage, have updated message per your first comment. Will squash and merge.

PR #783, opened 2017-07-28
Restore HTTP annotations to deprecated API methods

sjcorbett:
The missing annotations mean that brooklyn-client is unable to proxy the affected classes (since Resteasy expects to work with every method), and by extension that every build of the project fails. I think the simplest way to solve this is to expose the deprecated methods at `.../deprecated`.

### DIFF
--- a/karaf/itests/pom.xml
+++ b/karaf/itests/pom.xml
@@ -73,13 +73,6 @@
         </dependency>
         <dependency>
             <groupId>org.ops4j.pax.exam</groupId>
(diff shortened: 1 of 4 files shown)


sjcorbett:
cc @ahgittin since you deprecated the methods in https://github.com/apache/brooklyn-server/commit/1e615f55c71f23dfc8f65d8ddd2bb79254fce71a and https://github.com/apache/brooklyn-server/commit/36fd318c21a54493042cc985c13efeaeaa91a715

geomacy:
hm merge conflicts

PR #784, opened 2017-08-02
Upgrade json-path to 2.4.0 (from 2.0.0)

aledsage:
This supports `.size()`
Also fix the json-path dependency, to exclude conflicting versions.

The `.size()` is very useful for things like a CouchDB cluster - we can easily add a sensor that retrieves json for the members of the cluster. We'd also like a sensor for the size of that list (so we can check that all CouchDB nodes know about each other at startup).


duncangrant:
LGTM - needs minidev missing requirement issue fixed

aledsage:
Thanks @duncangrant - fixed the `minidev` issue; itest now passes. Merging.

PR #787, opened 2017-08-03
Don't validate an empty list of types.

geomacy:
This is so as to avoid the side effect of the "getCatalog()" in the
validate step, which does an unofficial catalog load that takes up
unnecessary time on startup.


aledsage:
Code LGTM, but I've not tested it. Assuming you've run through the startup + rebind, then I'm happy for you to merge this @geomacy

geomacy:
I have run through startup and rebind, will merge.

PR #788, opened 2017-08-03
UpdatingMap: support configuring ‘producer’

aledsage:
And adds a test for using `UpdatingMap` to set the service-not-up-indicators


aledsage:
Thanks @duncangrant - merging.

PR #789, opened 2017-08-03
Adds DynamicCluster.START_TIMEOUT

aledsage:
Similar to `SoftwareProcess`, this allows a dynamic cluster to wait for
`service.isUp` before returning from `start()`. It defaults to not waiting
(for backwards compatibility).

Previously, it would return as soon as all the initial member’s start()
effectors had finished executing.

---
Note that the new test `testReportsServiceUpAsSoonAsQuorumSize()` just demonstrates existing behaviour - this PR does not change that behaviour.

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/entity/group/DynamicClusterImpl.java
+++ b/core/src/main/java/org/apache/brooklyn/entity/group/DynamicClusterImpl.java
@@ -33,6 +33,7 @@ import java.util.concurrent.Callable;
 import java.util.concurrent.Semaphore;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
+
 import javax.annotation.Nullable;
 
 import org.apache.brooklyn.api.effector.Effector;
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/apache/brooklyn/entity/group/DynamicClusterImpl.java
```diff
@@ -485,6 +491,17 @@ protected void doStart() {
         }
     }
 
+    public void waitForServiceUp() {
```
duncangrant: why public?

aledsage: Good point - will change to protected (I copied it from `SoftwareProcessImpl`, where these are public).

aledsage:
build failure is unrelated, but I'll add a commit here to try to fix it:
```
2017-08-03 15:53:13,507 INFO  TESTNG FAILED: "Surefire test" - org.apache.brooklyn.camp.brooklyn.catalog.CatalogMakeOsgiBundleTest.testCatalogBomFromBundleWithManifest() finished in 116 ms
java.lang.AssertionError: attribute=Sensor: a.sensor (java.lang.String); val=null
	at org.apache.brooklyn.test.Asserts.fail(Asserts.java:758)
	at org.apache.brooklyn.test.Asserts.assertTrue(Asserts.java:737)
	at org.apache.brooklyn.core.entity.EntityAsserts.assertAttribute(EntityAsserts.java:103)
	at org.apache.brooklyn.camp.brooklyn.catalog.CatalogMakeOsgiBundleTest.assertBasic1DeploysAndHasSensor(CatalogMakeOsgiBundleTest.java:123)
	at org.apache.brooklyn.camp.brooklyn.catalog.CatalogMakeOsgiBundleTest.testCatalogBomFromBundleWithManifest(CatalogMakeOsgiBundleTest.java:114)
```

PR #790, opened 2017-08-03
Adds HttpFeed.preemptiveBasicAuth

aledsage:
For a description of pre-emptive basic auth, see https://hc.apache.org/httpcomponents-client-ga/tutorial/html/authentication.html (section 4.6. Preemptive authentication). As is described in that link, pre-emptive basic auth should be used with caution!

However, if you really do need to use basic auth then some servers don't send back the right challenge-response in the headers of the 401 response. For example, this seems to be the case with CouchDB!

We've seen in some downstream project people working around it by auto-generating the basic-auth header, and configuring the `HttpRequestSensor` accordingly. That is a bad pattern! (e.g. we lose the fact that this is a credential, so would never be able to use externalised configuration to ensure the passwords are not logged or written to persisted state).

This PR will allow those users to delete a bunch of lines of ugly yaml configuration from their blueprints. It will also help people knocking together a proof of concept - they can use `preemptiveBaicAuth` while they ask the people responsible for the servers to fix it properly!

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/feed/http/HttpFeed.java
+++ b/core/src/main/java/org/apache/brooklyn/feed/http/HttpFeed.java
@@ -24,6 +24,7 @@ import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.net.URI;
 import java.net.URL;
+import java.nio.charset.StandardCharsets;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
@@ -47,11 +48,11 @@ import org.apache.brooklyn.core.location.Machines;
 import org.apache.brooklyn.util.executor.HttpExecutorFactory;
 import org.apache.brooklyn.util.guava.Maybe;
 import org.apache.brooklyn.util.http.HttpToolResponse;
-import org.apache.brooklyn.util.http.executor.UsernamePassword;
 import org.apache.brooklyn.util.http.executor.HttpConfig;
 import org.apache.brooklyn.util.http.executor.HttpExecutor;
 import org.apache.brooklyn.util.http.executor.HttpRequest;
 import org.apache.brooklyn.util.http.executor.HttpResponse;
+import org.apache.brooklyn.util.http.executor.UsernamePassword;
 import org.apache.brooklyn.util.http.executor.apacheclient.HttpExecutorImpl;
 import org.apache.brooklyn.util.stream.Streams;
 import org.apache.brooklyn.util.time.Duration;
@@ -71,6 +72,7 @@ import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import com.google.common.collect.SetMultimap;
 import com.google.common.collect.Sets;
+import com.google.common.io.BaseEncoding;
 import com.google.common.io.ByteStreams;
 import com.google.common.reflect.TypeToken;
 
@@ -144,6 +146,7 @@ public class HttpFeed extends AbstractFeed {
         private Credentials credentials;
         private String uniqueTag;
         private HttpExecutor httpExecutor;
+        private Boolean preemptiveBasicAuth;
         private volatile boolean built;
 
         public Builder entity(Entity val) {
@@ -220,6 +223,10 @@ public class HttpFeed extends AbstractFeed {
             }
             return this;
         }
+        public Builder preemptiveBasicAuth(Boolean val) {
+            this.preemptiveBasicAuth = val;
+            return this;
+        }
         public Builder uniqueTag(String uniqueTag) {
             this.uniqueTag = uniqueTag;
             return this;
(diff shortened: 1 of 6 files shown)


duncangrant:
Could you update the description to explain what this does in a bit more detail please?

core/src/main/java/org/apache/brooklyn/feed/http/HttpFeed.java
```diff
@@ -228,6 +235,25 @@ public Builder httpExecutor(HttpExecutor val) {
             this.httpExecutor = val;
             return this;
         }
+        public Map<String, String> buildBaseHeaders() {
+            if (Boolean.TRUE.equals(preemptiveBasicAuth)) {
+                Credentials creds = credentials;
+                if (creds == null) {
+                    throw new IllegalArgumentException("Must not enable preemptiveBasicAuth when there are no credentials, in feed for "+baseUri);
```
duncangrant: We do this a lot in brooklyn and I'm not sure why.  Given that we're using a typed language why not create a MisconfiguredPreemptiveAuthException?  We can check the type of the exception thrown in tests and a tweak to the error text doesn't break the tests.

core/src/test/java/org/apache/brooklyn/feed/http/HttpFeedTest.java
```diff
@@ -360,7 +364,90 @@ public void testPollsMultiClearsOnSubsequentFailure() throws Exception {
...
+    }
 
+    // Expected behaviour of o.a.http.client is that it first sends the request without credentials,
+    // and then when given a challenge for basic-auth it re-sends the request with the basic-auth header.
+    @Test
+    public void testNonPreemptiveBasicAuth() throws Exception {
+        final String username = "brooklyn";
+        final String password = "hunter2";
+        
+        if (server != null) server.shutdown();
```
duncangrant: The mock server stuff in this file is in a bit of a mess.  I'd suggest we clean it up but not in this PR.

core/src/test/java/org/apache/brooklyn/feed/http/HttpFeedTest.java
```diff
@@ -360,7 +364,90 @@ public void testPollsMultiClearsOnSubsequentFailure() throws Exception {
...
+        assertEquals(req.getHeader("Authorization"), null);
+        
+        RecordedRequest req2 = server.takeRequest();
+        String headerVal = req2.getHeader("Authorization");
+        String expected = getBasicAuthHeaderVal(username, password);
+        assertEquals(headerVal, expected);
+    }
+
+    public static String getBasicAuthHeaderVal(String username, String password) {
+        String toencode = username + (password == null ? "" : ":"+password);
```
geomacy: same comment about not dropping the `:`

core/src/main/java/org/apache/brooklyn/feed/http/HttpFeed.java
aledsage: My personal preference is to create new exception types when it is useful to the user. But saying that, if it extends `IllegalArgumentException` then there's not much difference from their perspective.

If we created things as specific as `MisconfiguredPreemptiveAuthException`, I worry that this leads towards the extreme of an exception per config key!

Maybe a middle ground is to have an `IllegalConfigurationException` or some such?

I'm personally ok with us asserting exception message snippets in unit tests (they change rarely, and it confirms the message does appear at the top level rather than only being buried far down in the "caused by").

However, the text in this particular example is far too long!

Perhaps we should discuss this separately, and figure out our approach to such exception across-the-board.

core/src/main/java/org/apache/brooklyn/feed/http/HttpFeed.java
```diff
@@ -228,6 +235,25 @@ public Builder httpExecutor(HttpExecutor val) {
...
+                }
+                String username = checkNotNull(creds.getUserPrincipal().getName(), "username");
+                String password = creds.getPassword();
+                String toencode = username + (password == null ? "" : ":"+password);
```
aledsage: Ah, thanks!

core/src/main/java/org/apache/brooklyn/feed/http/HttpFeed.java
```diff
@@ -228,6 +235,25 @@ public Builder httpExecutor(HttpExecutor val) {
             this.httpExecutor = val;
             return this;
         }
+        public Map<String, String> buildBaseHeaders() {
+            if (Boolean.TRUE.equals(preemptiveBasicAuth)) {
+                Credentials creds = credentials;
+                if (creds == null) {
+                    throw new IllegalArgumentException("Must not enable preemptiveBasicAuth when there are no credentials, in feed for "+baseUri);
+                }
+                String username = checkNotNull(creds.getUserPrincipal().getName(), "username");
```
aledsage: I think we leave the name/password as-is. If we want to trim, or warn on extra white space, that should be done earlier (e.g. at the level of the config key's validation rules?).

Good point about colon in the username. I'll change it to fail-fast if it has a colon.

For the record, from RFC 7617 section 2:
```
Furthermore, a user-id containing a colon character is invalid, as
   the first colon in a user-pass string separates user-id and password
   from one another; text after the first colon is part of the password.
   User-ids containing colons cannot be encoded in user-pass strings.

   Note that many user agents produce user-pass strings without checking
   that user-ids supplied by users do not contain colons; recipients
   will then treat part of the username input as part of the password.
```

core/src/main/java/org/apache/brooklyn/feed/http/HttpFeed.java
geomacy: I tend to find the more-situation-specific Exceptions most useful from the developer point of view, as it can be handy to add breakpoints on specific types when debugging; if you want to break when the exception is thrown it can sometimes be a bother to set it on the more generic type, as it may get thrown in other places.  

I agree there's no compelling reason to change the `IllegalArgumentException` here, however.  A separate discussion could be had about possible more useful base exceptions.  `IllegalBrooklynConfigurationException`, anyone? :-)

aledsage:
Thanks @geomacy @duncangrant - comments addressed (in a second commit).

The jenkins build failure was real - I've also fixed that; it's building again.

aledsage:
Thanks @duncangrant - merging.

PR #793, opened 2017-08-04
HttpRequestSensor: handling non-json

aledsage:
Previously the `HttpRequestSensor` failed if no `jsonPath: ` is suppiled in the yaml. I've added a failing test, but not worked specifically on that. It doesn't pick up the default value of `$`.

Added support for setting `jsonPath` to blank, so that it just takes the pure-string value (i.e. no json parsing).

Also added a few more (unrelated!) tests.

### DIFF
--- a/camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/AggregatorYamlTest.java
+++ b/camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/AggregatorYamlTest.java
@@ -184,6 +184,36 @@ public class AggregatorYamlTest extends AbstractYamlTest {
     }
     
     @Test
+    public void testListExcludesBlank() throws Exception {
+        Entity app = createAndStartApplication(
+                "services:",
+                "- type: " + BasicApplication.class.getName(),
+                "  brooklyn.children:",
+                "    - type: " + TestEntity.class.getName(),
+                "    - type: " + TestEntity.class.getName(),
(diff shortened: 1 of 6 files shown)


camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/AggregatorYamlTest.java
```diff
@@ -184,6 +184,36 @@ public void testList() throws Exception {
...
+        EntityAsserts.assertAttributeEqualsEventually(app, myResult, ImmutableList.of());
+
+        child1.sensors().set(myVal, "val1");
+        EntityAsserts.assertAttributeEqualsEventually(app, myResult, ImmutableList.of("val1"));
+        
+        child2.sensors().set(myVal, "val2");
+        EntityAsserts.assertAttributeEqualsEventually(app, myResult, ImmutableList.of("val1", "val2"));
+        
+        child1.sensors().set(myVal, null);
+        EntityAsserts.assertAttributeEqualsEventually(app, myResult, MutableList.of("val2"));
```
m4rkmckenna: Id also add

```
child1.sensors().set(myVal, "");
EntityAsserts.assertAttributeEqualsEventually(app, myResult, MutableList.of("val2"));

child1.sensors().set(myVal, "                        ");
EntityAsserts.assertAttributeEqualsEventually(app, myResult, MutableList.of("val2"));
```

PR #796, opened 2017-08-14
Disabled AUTO_WRAP_CATALOG_YAML_AS_BUNDLE

Graeme-Miller:
There is currently a bug in brooklyn-server surrounding auto wrap. Under certain conditions, templates which have been added to brooklyn are not able to be deployed. 

I am still investigating the issue, but have raised a PR to open a discussion about turning off temporarily whilst we investigate the issue.


geomacy:
@Graeme-Miller what's the motivation for this?

Graeme-Miller:
no longer required

PR #800, opened 2017-08-17
AutoScalerPolicy to resize to limits on expunge

nakomis:
Currently, if a user resizes a cluster with an `AutoScalerPolicy` such that the pool size is now outside the min and max pool size for the policy, the cluster is not resized to the bounds

### DIFF
--- a/core/src/test/java/org/apache/brooklyn/core/test/entity/TestSizeRecordingClusterImpl.java
+++ b/core/src/test/java/org/apache/brooklyn/core/test/entity/TestSizeRecordingClusterImpl.java
@@ -1,0 +1,59 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.brooklyn.core.test.entity;
+
+import com.google.common.collect.ImmutableList;
+import org.apache.brooklyn.entity.group.DynamicClusterImpl;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+public class TestSizeRecordingClusterImpl extends DynamicClusterImpl implements TestSizeRecordingCluster {
+
+    public TestSizeRecordingClusterImpl () {
+    }
+
+    private final List<Integer> sizeHistory = Collections.synchronizedList(new ArrayList<Integer>());
+
+    @Override
+    public Integer resize(Integer desiredSize) {
+        int existingSize = getCurrentSize();
+        int newSize = super.resize(desiredSize);
+        if (newSize != existingSize) {
+            addSizeHistory(newSize);
+        }
+        return newSize;
+    }
+
+    @Override
+    public List<Integer> getSizeHistory() {
+        synchronized (sizeHistory) {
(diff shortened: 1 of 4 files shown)


nakomis:
On second thoughts, these tests seem a little fragile and may have non-deterministic failures. Please DO NOT MERGE (yet)

nakomis:
Should now be good to go

policy/src/main/java/org/apache/brooklyn/policy/autoscaling/AutoScalerPolicy.java
```diff
@@ -458,6 +459,25 @@ public void onEvent(SensorEvent<Number> event) {
...
+    /**
+     * This should usually be a no-op as the min and max pool sizes are taken into account when
+     * an automatic resize event occurs, however this covers the special case where the user
+     * has manually resized the pool
+     */
+    private SensorEventListener<? super Integer> poolEventHandler = new SensorEventListener<Integer>() {
+        @Override
+        public void onEvent(SensorEvent<Integer> event) {
+            Sensor<Integer> sensor = event.getSensor();
+            assert sensor.equals(DynamicCluster.GROUP_SIZE);
```
m4rkmckenna: Is `assert` not removed at runtime??

nakomis:
PR comment addressed

core/src/test/java/org/apache/brooklyn/core/test/entity/TestSizeRecordingClusterImpl.java
```diff
@@ -0,0 +1,59 @@
...
+package org.apache.brooklyn.core.test.entity;
+
+import com.google.common.collect.ImmutableList;
+import org.apache.brooklyn.entity.group.DynamicClusterImpl;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+public class TestSizeRecordingClusterImpl extends DynamicClusterImpl implements TestSizeRecordingCluster {
```
aledsage: Can we instead just use the existing `org.apache.brooklyn.core.test.entity.TestCluster`?

In your test, you're configuring `TestCluster.MEMBER_SPEC` so it calling `super.resize()` should be fine I'd have thought.

policy/src/test/java/org/apache/brooklyn/policy/autoscaling/AutoScalerPolicyPoolSizeTest.java
```diff
@@ -0,0 +1,110 @@
...
+    @Override
+    public void setUp() throws Exception {
+        super.setUp();
+        LOG.info("resetting " + getClass().getSimpleName());
+        cluster = app.createAndManageChild(EntitySpec.create(TestSizeRecordingCluster.class)
+                .configure(TestCluster.INITIAL_SIZE, CLUSTER_INIITIAL_SIZE)
+                .configure(TestCluster.MEMBER_SPEC, EntitySpec.create(BasicStartable.class))
+        );
+        PolicySpec<AutoScalerPolicy> policySpec = PolicySpec.create(AutoScalerPolicy.class)
+                .configure(AutoScalerPolicy.RESIZE_OPERATOR, new ResizeOperator() {
```
aledsage: Why supply a `RESIZE_OPERATOR` (I'm guessing it's because that's what `AutoScalerPolicyTest` does)?

Better to just rely on the default `RESIZE_OPERATOR`, which does exactly this - then we are testing a more realistic way of using the `AutoScalerPolicy` . You never use the `resizes` field anyway, and we can live without the logging.

policy/src/test/java/org/apache/brooklyn/policy/autoscaling/AutoScalerPolicyPoolSizeTest.java
```diff
@@ -0,0 +1,110 @@
...
+            cluster = null;
+            policy = null;
+        }
+    }
+
+    @Test
+    public void testResizeUp() throws Exception {
+        EntityAsserts.assertAttributeEqualsEventually(cluster, TestCluster.GROUP_SIZE, CLUSTER_INIITIAL_SIZE);
+        // Simulate user expunging the entities manually
+        for (int i = 0; i < CLUSTER_MAX_SIZE - CLUSTER_MIN_SIZE; i++) {
```
aledsage: This sum seems weird: for your test it deletes `4 - 2` entities, which takes us from initial size of 3 down to 1. But that feels more like coincidence than design. Do you not want to delete `CLUSTER_INITIAL_SIZE - CLUSTER_MIN_SIZE + 1`?

PR #802, opened 2017-08-18
Adds karaf application-list & enity-info commands

m4rkmckenna:
`brooklyn:application-list`
Display a list of deployed applications and their state

`brooklyn:entity-info <id>`
Display detailed information about a single entity

new-commands


karaf/commands/src/main/java/org/apache/brooklyn/karaf/commands/EntityInfo.java
```diff
@@ -0,0 +1,137 @@
...
+
+    @Override
+    public Object execute() throws Exception {
+        final Entity entity = managementContext.getEntityManager().getEntity(id);
+        printHeader("Basic Information");
+        final ShellTable infoTable = new ShellTable();
+        infoTable.column("").bold();
+        infoTable.column("");
+        infoTable.noHeaders();
+        infoTable.addRow().addContent("Application Id", entity.getApplicationId());
```
drigodwin: If the entity id is invalid, this throws a NPE
```
karaf@brooklyn()> brooklyn:entity-info invalid-id

Basic Information
-----------------
Error executing command: java.lang.NullPointerException
```
Catch this and report an error message?

karaf/commands/src/main/java/org/apache/brooklyn/karaf/commands/EntityInfo.java
```diff
@@ -0,0 +1,137 @@
...
+import org.apache.karaf.shell.api.action.Option;
+import org.apache.karaf.shell.api.action.lifecycle.Reference;
+import org.apache.karaf.shell.api.action.lifecycle.Service;
+import org.apache.karaf.shell.support.ansi.SimpleAnsi;
+import org.apache.karaf.shell.support.table.ShellTable;
+
+import java.util.Collections;
+import java.util.Optional;
+
+@Command(scope = "brooklyn", name = "entity-info", description = "Display entity info")
```
drigodwin: I'm not sure if this is clear that it's a running entity as opposed to one in the catalog

m4rkmckenna: What change do you suggest?

PR #803, opened 2017-08-22
Adds cluster.max.size to dynamic cluster

nakomis:
Adds cluster.max.size to dynamic cluster, and prevents the cluster from being manually increased beyond that size

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/entity/group/DynamicClusterImpl.java
+++ b/core/src/main/java/org/apache/brooklyn/entity/group/DynamicClusterImpl.java
@@ -595,6 +595,10 @@ public class DynamicClusterImpl extends AbstractGroupImpl implements DynamicClus
     @Override
     public Integer resize(Integer desiredSize) {
         synchronized (mutex) {
+            Optional<Integer> optionalMaxSize = Optional.fromNullable(config().get(MAX_SIZE));
+            if (optionalMaxSize.isPresent() && desiredSize > optionalMaxSize.get()) {
+                throw new Resizable.InsufficientCapacityException("Desired cluster size " + desiredSize + " exceeds maximum size of " + optionalMaxSize.get());
+            }
             int originalSize = getCurrentSize();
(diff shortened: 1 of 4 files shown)


core/src/test/java/org/apache/brooklyn/entity/group/DynamicClusterTest.java
```diff
@@ -1378,6 +1378,24 @@ public void testChildCommandPermitNotReleasedWhenMemberStartTaskCancelledBeforeS
...
+                .configure(DynamicCluster.MAX_SIZE, 6));
+        cluster.start(ImmutableList.of(loc));
+
+        cluster.resize(4);
+        EntityAsserts.assertAttributeEqualsEventually(cluster, DynamicCluster.GROUP_SIZE, 4);
+        try {
+            cluster.resize(8);
+            Asserts.shouldHaveFailedPreviously("Cluster resize should have failed because max size was exceeded");
+        } catch (Exception e) {
+            // ignored.
```
m4rkmckenna: You could assert that the expected exception is thown ... Empty catch blocks is a code smell

core/src/main/java/org/apache/brooklyn/entity/group/DynamicClusterImpl.java
```diff
@@ -595,6 +595,10 @@ public void restart() {
     @Override
     public Integer resize(Integer desiredSize) {
         synchronized (mutex) {
+            Optional<Integer> optionalMaxSize = Optional.fromNullable(config().get(MAX_SIZE));
+            if (optionalMaxSize.isPresent() && desiredSize > optionalMaxSize.get()) {
+                throw new IllegalArgumentException("Desired cluster size " + desiredSize + " exceeds maximum size of " + optionalMaxSize.get());
```
m4rkmckenna: Should this not be `InsufficientCapacityException `

nakomis:
PR comments addressed

PR #805, opened 2017-08-25
Modify jcloudsLocation.releaseNode

andreaturli:
Workaround waiting for a fix for https://issues.apache.org/jira/browse/JCLOUDS-1332

Uses `destroyNodesMatchingId` instead of `destroyNode` as it consistently invokes `cleanUpIncidentalResourcesOfDeadNodes`

### DIFF
--- a/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
+++ b/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
@@ -522,7 +522,9 @@ public class JcloudsLocation extends AbstractCloudMachineProvisioningLocation im
 
     @Override
     public void killMachine(String cloudServiceId) {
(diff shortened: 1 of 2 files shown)


duncangrant:
Should the same change be applied to killMachine( as well?

aledsage:
@andreaturli code LGTM - thanks! Nice and simple fix for an annoying problem.

I've not tested it myself. If you or someone else has done so on appropriate cloud(s), then very happy for you to do so.

drigodwin:
retest this please

drigodwin:
This PR swaps the mechanism Brooklyn is using to destroy a node to one which is less tested within Brooklyn. The problem it addresses has been fixed upstream in jclouds so assuming we can incorporate that in before the next release we should not need this PR. 

I think however we should leave this PR open and unmerged in the meantime in case the jclouds release is not available and this fix has not been incorporated before the next release.

PR #806, opened 2017-08-25
Osgis.bundleFinder: support version range

aledsage:
Useful for `brooklyn.libraries` when name:version of pre-installed bundle is specified.

@m4rkmckenna as discussed - can you take a look please?

### DIFF
--- a/core/src/test/java/org/apache/brooklyn/util/core/osgi/OsgisTest.java
+++ b/core/src/test/java/org/apache/brooklyn/util/core/osgi/OsgisTest.java
@@ -1,0 +1,140 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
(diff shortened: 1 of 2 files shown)


core/src/test/java/org/apache/brooklyn/util/core/osgi/OsgisTest.java
```diff
@@ -0,0 +1,117 @@
...
+                .symbolicName("mybundle")
+                .version("[1.0.0, 1.1.0)")
+                .findAll();
+        assertEquals(result, ImmutableList.of(myBundle1_0_0));
+        
+        List<Bundle> result2 = Osgis.bundleFinder(framework)
+                .symbolicName("mybundle")
+                .version("[1.0.0, 2.0.0)")
+                .findAll();
+        assertEquals(result2, ImmutableList.of(myBundle1_0_0, myBundle1_1_0));
```
m4rkmckenna: I think its worth adding the versions `2.0.0.SNAPSHOT` & `2.0.0` and making assertions around this boundary

Alos add checks for the range `[1.0.0, 2.0.0]`

aledsage:
Thanks @m4rkmckenna - I've added a couple more tests, squashed it into a single commit, and rebased against master. Once jenkins confirms this builds, I'll merge it.

PR #808, opened 2017-08-31
BROOKLYN-528: same proxy class for all entities of type

aledsage:
Fixes https://issues.apache.org/jira/browse/BROOKLYN-528

As well as fixing the space leak and decreasing the memory usage, this improves performance of entity creation from about 90 per second to about 230 per second (on my laptop).

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/objs/proxy/ClassLoaderCache.java
+++ b/core/src/main/java/org/apache/brooklyn/core/objs/proxy/ClassLoaderCache.java
@@ -1,0 +1,121 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/apache/brooklyn/core/objs/proxy/ClassLoaderCache.java
```diff
@@ -0,0 +1,121 @@
...
+ * The classloader is created with references to ClassLoaders for the given class and interfaces
+ * (including all super-classes and super-interfaces of those). Also see comment in
+ * {@link InternalEntityFactory#createEntityProxy(Iterable, org.apache.brooklyn.api.entity.Entity)}.
+ * 
+ * If a classloader has already been created for the given class+interfaces, then it guarantees
+ * to return that same classloader instance. This is very important when using {@code newProxyInstance}
+ * because if a different ClassLoader instance is used then it will reflectively create a new Proxy class.
+ * See https://issues.apache.org/jira/browse/BROOKLYN-528.
+ */
+class ClassLoaderCache {
```
andreaturli: [minor] @aledsage cache looks exactly what we need here.
FYI @robertgmoss has used a `computeIfAbsent` in https://github.com/apache/brooklyn-server/blob/master/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/AbstractComputeServiceRegistry.java#L79 which may save some boilerplate code.

geomacy:
LGTM, and the integration tests pass for me too. Will merge.

PR #810, opened 2017-09-07
REST API for bundles, types, and subtypes

ahgittin:
also tweaks to other classes to make them more usable, eg VersionedName comparables and bundle removal giving an OsgiBundleInstallationResult

test in BundleAndTypeAndSubtypeResourcesTest based on CatalogResourceTest, basically giving parity in terms of functionality and test coverage


ahgittin:
bundle-rest-api

rest/rest-api/src/main/java/org/apache/brooklyn/rest/api/CatalogApi.java
```diff
@@ -47,6 +47,9 @@
 import io.swagger.annotations.ApiResponse;
 import io.swagger.annotations.ApiResponses;
 
+/** @deprecated since 0.12.0 use /bundle, /type, and /subtype */
+// but we will probably keep this around for a while as many places use it
+@Deprecated
```
tbouron: I strongly disagree with deprecating the entire `/catalog` endpoint. I think we should rather deprecate `/catalog/(entities|applications|policies|enrichers)` and add `/bundle`, `/type` and `/subtype` within `/catalog` as we have a "catalog" of bundles, types, etc. (that is effectivement what you did with the `/type` endpoint, by duplicating `/application`, `/entities`, etc) Promoting them to the root level feel way too loose.

Also, what is the difference between `type` and `subtype`? It's not clear to me and it looks like a subtype is just a subset of type. If that is a case, why bother having a endpoint for this? Cannot we just have a filter on the `/type` endpoint?

Regarding the bundle additions, the `POST /catalog/create` can already deal with bundles so I just reuse that. For the deletion, I would use the and `DELETE /catalog/bundles/<symbolicName>/<version>`.

rest/rest-api/src/main/java/org/apache/brooklyn/rest/api/SubtypeApi.java
```diff
@@ -0,0 +1,96 @@
...
+import io.swagger.annotations.Api;
+import io.swagger.annotations.ApiOperation;
+import io.swagger.annotations.ApiParam;
+
+@Path("/subtypes")
+@Api("Subtypes")
+@Consumes(MediaType.APPLICATION_JSON)
+@Produces(MediaType.APPLICATION_JSON)
+public interface SubtypeApi {
+
```
m4rkmckenna: DRY just add the below filters to the `TypeApi` as a query param

rest/rest-api/src/main/java/org/apache/brooklyn/rest/domain/TypeDetail.java
```diff
@@ -0,0 +1,86 @@
...
+ */
+package org.apache.brooklyn.rest.domain;
+
+import org.apache.brooklyn.api.typereg.RegisteredType;
+import org.apache.brooklyn.api.typereg.RegisteredType.TypeImplementationPlan;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonInclude.Include;
+
+public class TypeDetail extends TypeSummary {
```
m4rkmckenna: Public api ... add java doc

rest/rest-api/src/main/java/org/apache/brooklyn/rest/domain/TypeSummary.java
```diff
@@ -0,0 +1,287 @@
...
+import org.apache.brooklyn.util.text.VersionComparator;
+
+import com.fasterxml.jackson.annotation.JsonAnyGetter;
+import com.fasterxml.jackson.annotation.JsonAnySetter;
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonInclude.Include;
+import com.google.common.collect.ComparisonChain;
+
+public class TypeSummary implements Comparable<TypeSummary> {
```
m4rkmckenna: java doc

rest/rest-api/src/main/java/org/apache/brooklyn/rest/api/TypeApi.java
```diff
@@ -0,0 +1,94 @@
...
+        @QueryParam("regex") @DefaultValue("") String regex,
+        @ApiParam(name = "fragment", value = "Substring case-insensitive to search for")
+        @QueryParam("fragment") @DefaultValue("") String fragment);
+
+    @Path("/{nameOrAlias}")
+    @GET
+    @ApiOperation(value = "Get summaries for all versions and instances of a given type or alias, with best match first", 
+            response = TypeSummary.class,
+            responseContainer = "List")
+    public List<TypeSummary> listVersions(
```
m4rkmckenna: The response type seems odd to me ... I would expect it to only return 1 summary

If aliases can be re used i think this is a big mistake

rest/rest-api/src/main/java/org/apache/brooklyn/rest/api/SubtypeApi.java
```diff
@@ -0,0 +1,96 @@
...
+    public List<TypeSummary> list(
+        @ApiParam(name = "supertype", value = "Supertype to query", required = true)
+        @PathParam("supertype")
+        String supertype,
+        @ApiParam(name = "versions", value = "Whether to list 'latest' of each symbolic-name or 'all' versions", required = false, defaultValue = "latest")
+        @QueryParam("versions")
+        String versions,
+        @ApiParam(name = "regex", value = "Regular expression to search for")
+        @QueryParam("regex") @DefaultValue("") String regex,
+        @ApiParam(name = "fragment", value = "Substring case-insensitive to search for")
```
m4rkmckenna: Is regex and fragment not a duplication

`"some-fragment" == ".*some-fragment.*"`

rest/rest-api/src/main/java/org/apache/brooklyn/rest/api/SubtypeApi.java
```diff
@@ -0,0 +1,96 @@
...
+    @GET
+    @ApiOperation(value = "Get all known types which declare the given argument as a supertype", 
+            response = TypeSummary.class, responseContainer = "List")
+    public List<TypeSummary> list(
+        @ApiParam(name = "supertype", value = "Supertype to query", required = true)
+        @PathParam("supertype")
+        String supertype,
+        @ApiParam(name = "versions", value = "Whether to list 'latest' of each symbolic-name or 'all' versions", required = false, defaultValue = "latest")
+        @QueryParam("versions")
+        String versions,
```
m4rkmckenna: Should be singular

aledsage:
TL;DR: I propose we split this PR into two. The `/bundles` part we can merge pretty quickly. However, the `/types` and `/subtypes` is too controversial - it probably deserves a `/v2/` of the REST api.

---
I don't want to lose the word "catalog" in the REST api - it's so good at getting across the meaning for users! The alternative `/type` is just not as good, in my opinion.

The multiple endpoints of `/types` and `/subtypes` is confusing. I'd model the latter as just a filter on `/type`. It would be better achieved with an additional query parameter rather than a separate endpoint.

If designing a `/v2` REST api, we could use `/catalog` instead of `/type`. However, it will likely take a while to get to a stable and good `/v2` api. There are other cleanup/improvements we should probably do to the REST api if we're releasing a new version of it (e.g. exclude the deprecated stuff, get rid of `/locations` but figure out if we really need to support locations from brooklyn.properties, find out from the community about other inconsistencies or hard-to-understand parts of the api).

The meaning of `GET /subtypes/application` looks completely different from `GET /catalog/applications`. The latter retrieves the catalog items marked as `template`, but the new api returns everything that implements `Application`. Perhaps this is an opportunity to get rid of the "entity" versus "template" distinction (at least in the REST api). The original meaning/intent of "template" has entirely changed / been misused! I believe it was originally intended as a partially-complete YAML blueprint that someone would retrieve via the REST api, and then modify. They'd then POST their .yaml file to deploy their app. It has now been used as the list of apps to include in a "quick launch" view. If designing a new `/v2` api, I'd explicitly support a "quick launch" list and would get rid of the "template" versus "application" versus "entity" distinction in the REST api (anywhere you can use an entity, you can use an app; anywhere you need an app then a non-app entity will be auto-wrapped in a basic-app).

ahgittin:
As per ML:

Aled, Thomas, Mark - thanks for the good feedback.

All - There are some good suggestions which I agree deserve discussion.  Specific points.

(1) Should /types and /bundles be top-level or under a /catalog prefix ?  Thomas suggested the latter which also fits.  My reason for doing top-level is simply that most REST APIs these days seem to have more top-level resources.  /catalog is not necessary and in some ways it's cleaner having separate endpoints.  On the other hand the /catalog prefix is nice to orient consumers, as Aled says:  `bundles` and `types` on their own aren't as self-evident.  And it means we could continue to have `POST /catalog` as the main way to install.

(2) Should we deprecate `/catalog` ?  Thomas suggests maybe not yet.  I think we should as equivalent functionality and more is available at the new endpoints.  (Though if we go with (1), obviously we wouldn't deprecate the whole endpoint, just the old calls.).  Also worth noting:  I don't think we should remove the deprecated endpoint anytime soon however.  

(3) Should we switch to /v2 ?  Aled has suggested we do as the generic `types` support is a significant shift from the old more restrictive `catalog`.  I don't think we should yet, however:  I'd prefer to make that move when we are ready to remove all old endpoints so v2 is clutter-free.  To the extent v2 can look like a subset of v1 we make life easier for users and ourselves.  Also there is significant work to add a /v2 endpoint and I don't think there is benefit yet to justify this work.

(4) Should `/subtypes` be an endpoint peer of `/types` ?  It has been noted the same functionality as `/subtypes/entity` could be done with `/types?super=entity` (with appropriate query paramter).  My reasoning for making it a separate endpoint is that this is an activity I expect people will want to do a lot, avoiding the query parameter makes for a slightly nicer API.  But it is repetition in the code.

(There are some other minor issues but I think I agree with the points made there.)

My answers:

(1) slight preference for the `/catalog` prefix
(2) strong pref to deprecate old calls - they are redundant and multiple is confusing
(3) strong pref to stay with `/v1` for now
(4) slight pref for explicit `[/catalog]/subtypes` endpoint

ahgittin:
Regarding other point about template v application - should that be addressed in this PR @aledsage or could it be a subsequent one?

rest/rest-api/src/main/java/org/apache/brooklyn/rest/api/CatalogApi.java
ahgittin: Good points @tbouron - discussed on ML

rest/rest-api/src/main/java/org/apache/brooklyn/rest/api/SubtypeApi.java
ahgittin: Discussed on ML

rest/rest-api/src/main/java/org/apache/brooklyn/rest/api/SubtypeApi.java
ahgittin: I think plural as `version=all` doesn't make sense and it implies wanting to do things like `version=1.0`.  It could be `boolean allVersions` but my thinking was in future this might take a range.

rest/rest-api/src/main/java/org/apache/brooklyn/rest/api/SubtypeApi.java
ahgittin: Not if `some-fragment` contains a `.` or other special character.  Globs are more user friendly and search friendly.  (This is borrowed from elsewhere in API incidentally where we do the same thing and there have been times I've been grateful for the easier `fragment`.)

aledsage:
For template versus application, we can defer that (as long as we're not making pain for ourselves in this api; I don't think we are). Good to separate things into more PRs. UIs that want to show a quick launch will continue to use `/v1/catalog/application`, which will not be supported by the new endpoint(s).

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/transform/TypeTransformer.java
```diff
@@ -0,0 +1,194 @@
...
+    public static <T extends Entity> TypeSummary summary(BrooklynRestResourceUtils b, RegisteredType item, UriBuilder ub) {
+        return embellish(new TypeSummary(item), item, false, b, ub);
+    }
+
+    public static TypeDetail detail(BrooklynRestResourceUtils b, RegisteredType item, UriBuilder ub) {
+        return embellish(new TypeDetail(item), item, true, b, ub);
+    }
+
+    private static <T extends TypeSummary> T embellish(T result, RegisteredType item, boolean detail, BrooklynRestResourceUtils b, UriBuilder ub) {
+        result.getExtraFields().put("links", makeLinks(item, ub));
```
aledsage: This use of `getXyz().put(key, val)` is a really bad pattern, especially when there is a `setExtraField(key, val)` method you can call instead.

It makes it a lot less clear to people supporting the code how the `BundleSummary` is populated.

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/transform/TypeTransformer.java
```diff
@@ -0,0 +1,194 @@
...
+   
+            AtomicInteger paramPriorityCnt = new AtomicInteger();
+            for (SpecParameter<?> input: spec.getParameters())
+                config.add(EntityTransformer.entityConfigSummary(input, paramPriorityCnt));
+            for (Sensor<?> x: type.getSensors())
+                sensors.add(SensorTransformer.sensorSummaryForCatalog(x));
+            for (Effector<?> x: type.getEffectors())
+                effectors.add(EffectorTransformer.effectorSummaryForCatalog(x));
+            
+            result.getExtraFields().put("config", config);
```
aledsage: Where do you imagine we'll document the names "config", "sensors" and "effectors" (or do people have to discover it by making some live calls and figuring out what looks to be there consistently for everything of a given type)?

Where would we document the type of the values for something like an effector?

I think the looseness won't be an issue for Brooklyn developers (because we can easily navigate the code base to find the `EffectorTransformer`, when working on the CLI or UI).

However, I think it makes it feel more scary for a user of Brooklyn who is thinking of writing an integration with our REST api. They won't know what fields to expect. They can make a few calls and then guess at what is optional versus mandatory, and then code as defensively as possible in case their guess was wrong.

rest/rest-api/src/main/java/org/apache/brooklyn/rest/api/BundleApi.java
```diff
@@ -0,0 +1,158 @@
...
+@Api("Bundles")
+@Consumes(MediaType.APPLICATION_JSON)
+@Produces(MediaType.APPLICATION_JSON)
+public interface BundleApi {
+
+    @GET
+    @ApiOperation(value = "List bundles registered in the system including their types", 
+            response = BundleSummary.class,
+            responseContainer = "List")
+    public List<BundleSummary> list(
```
aledsage: Feels like this will return a huge amount of data (giving all the types that are inside the bundle).

Do we want a way to just list the bundles? I think we want to have a `BundleSummary` and a `BundleDetail`, where only the latter includes all the types. The `list` method could include support for `?detail=true` if we think that's important for a UI to retrieve everything it needs to show in one go.

rest/rest-api/src/main/java/org/apache/brooklyn/rest/api/BundleApi.java
```diff
@@ -0,0 +1,158 @@
...
+    @Consumes // anything (if doesn't match other methods with specific content types
+    @ApiOperation(
+            value = "Adds types to the registry from the given item, autodetecting type as ZIP/JAR or BOM YAML",
+            response = BundleInstallationRestResult.class
+    )
+    @ApiResponses(value = {
+            @ApiResponse(code = 400, message = "Error processing the given archive, or the catalog.bom is invalid"),
+            @ApiResponse(code = 201, message = "Catalog items added successfully")
+    })
+    public Response createAutodetecting(
```
aledsage: I lean towards not doing this - instead rely on `Content-Type: ...` being supplied.

I say this because of the annoying behaviour in the existing api for things like `POST /v1/catalog`. If you post an invalid .bom file, it first tries to parse it as a .bom which fails, and then tries to parse it as a different format. The error that gets returned is often misleading (e.g. it gives you details of the second attempt, rather than the first attempt). I've hit this problem while trying to help real newbee (but smart) users of Brooklyn.

Looking at this new code, it would be quite easy to lose the error message that tells you your yaml was invalid, with it instead telling you that it's not a zip file.

What does this really buy us? The UI and CLI will always supply it. If using `curl` then you have to type a little bit more, but I think that's fine.

rest/rest-api/src/main/java/org/apache/brooklyn/rest/domain/TypeDetail.java
```diff
@@ -0,0 +1,86 @@
...
+            data = p.getPlanData();
+        }
+        public String getFormat() {
+            return format;
+        }
+        public Object getData() {
+            return data;
+        }
+    }
+    private final TypeImplementationPlanSummary plan;
```
aledsage: I'd make this non-final, so that json-deserialization doesn't have to do nasty stuff.

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/resources/BundleResource.java
```diff
@@ -0,0 +1,186 @@
...
+                }
+            }
+        }
+        return toBundleSummary(bundles.values());
+    }
+
+    private List<BundleSummary> toBundleSummary(Iterable<ManagedBundle> sortedItems) {
+        List<BundleSummary> result = MutableList.of();
+        for (ManagedBundle t: sortedItems) {
+            result.add(TypeTransformer.bundleSummary(brooklyn(), t, ui.getBaseUriBuilder(), mgmt()));
```
aledsage: Does this use anything like `filters.add(RegisteredTypePredicates.entitledToSee(mgmt()));`?

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/resources/BundleResource.java
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    @Override
+    public BundleSummary detail(String symbolicName, String version) {
+        ManagedBundle b = lookup(symbolicName, version);
+        return TypeTransformer.bundleDetails(brooklyn(), b, ui.getBaseUriBuilder(), mgmt());
+    }
+
+    protected ManagedBundle lookup(String symbolicName, String version) {
+        // TODO entitlements for bundles
```
aledsage: Perhaps we could just use the `catalog.see` entitlement for now?
```
        if (!Entitlements.isEntitled(mgmt().getEntitlementManager(), Entitlements.SEE_CATALOG_ITEM, symbolicName+(Strings.isBlank(version)?"":":"+version))) {
            throw WebResourceUtils.forbidden("User '%s' is not authorized to see catalog bundle",
                Entitlements.getEntitlementContext().user());
        }
```

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/resources/BundleResource.java
```diff
@@ -0,0 +1,186 @@
...
+        ManagedBundle b = ((ManagementContextInternal)mgmt()).getOsgiManager().get().getManagedBundle(new VersionedName(symbolicName, version));
+        if (b==null) {
+            throw WebResourceUtils.notFound("Bundle with id '%s:%s' not found", symbolicName, version);
+        }
+        return b;
+    }
+
+    @Override
+    public BundleInstallationRestResult remove(String symbolicName, String version, Boolean force) {
+        ManagedBundle b = lookup(symbolicName, version);
```
aledsage: Really want an entitlements check (as per your previous TODOs). Otherwise the presence of the `/bundles` endpoint will be a horrible backdoor for allowing a user to damage other people's apps.

rest/rest-api/src/main/java/org/apache/brooklyn/rest/api/SubtypeApi.java
aledsage: No strong feelings, but I wonder if we should err towards the smaller api (simpler to support and for someone to get their head around). You can achieve everything with regex. A UI could offer a fragment search on top of that, escaping things as needed.

aledsage: Could do with more description of what the fragment will search (think it's just the name or alias?).

Somewhere in the existing `/v1/catalog` api the fragment will search the entire yaml plan - that's surprising, something useful, and sometimes annoying when it returns a bunch of seemingly unrelated things!

rest/rest-api/src/main/java/org/apache/brooklyn/rest/api/TypeApi.java
aledsage: @m4rkmckenna I think it's a list because it's giving you info about each version (e.g. an entry for v1.0.0 and another for v1.1.0).

aledsage: Interesting point about reusing aliases. Hopefully that's not the case!

rest/rest-api/src/main/java/org/apache/brooklyn/rest/api/CatalogApi.java
ahgittin: update:  i think this is how we are going to do it now

rest/rest-api/src/main/java/org/apache/brooklyn/rest/domain/TypeSummary.java
ahgittin: done

rest/rest-api/src/main/java/org/apache/brooklyn/rest/domain/TypeSummary.java
```diff
@@ -0,0 +1,287 @@
...
+    private final String displayName;
+    @JsonInclude(value=Include.NON_EMPTY)
+    private final String description;
+    @JsonInclude(value=Include.NON_EMPTY)
+    private String iconUrl;
+
+    @JsonInclude(value=Include.NON_EMPTY)
+    private Set<String> aliases;
+    @JsonInclude(value=Include.NON_EMPTY)
+    private Set<Object> supertypes;
```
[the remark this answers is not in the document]
ahgittin: agree, done

rest/rest-api/src/main/java/org/apache/brooklyn/rest/api/SubtypeApi.java
ahgittin: fragment is v useful as regex-escaping is tedious (esp for case insensitive!).  it could be called `glob` to be clearer.  agree we should say what is searched, will do that.

ahgittin: (change to `glob` could be `/v2beta` item)

rest/rest-api/src/main/java/org/apache/brooklyn/rest/domain/TypeDetail.java
ahgittin: done

rest/rest-api/src/main/java/org/apache/brooklyn/rest/api/TypeApi.java
ahgittin: yes this is for all verisons

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/transform/TypeTransformer.java
ahgittin: for now there is no doc for extra info for specific types.  agree this is a gap but it is a lot more classes and maintenance to have all that detail completely type specific.  i'm quite sure run-time discoverability for extra fields will be sufficient for a while -- and it's not at all uncommon in pragmatic REST APIs i have used -- and when users start wanting more precision then i think it's worth doing the extra work (and preference for extra fields added in this class with docs to say they are only for selected types, rather than lots of subtypes).

rest/rest-api/src/main/java/org/apache/brooklyn/rest/api/BundleApi.java
ahgittin: agree

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/transform/TypeTransformer.java
ahgittin: agree

rest/rest-api/src/main/java/org/apache/brooklyn/rest/api/BundleApi.java
ahgittin: this is all being removed as we will just have `POST /catalog` -- but i agree with you

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/resources/BundleResource.java
ahgittin: using this for now, adding TODO on entitlements that it should ideally distinguish between types and bundles

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/resources/BundleResource.java
ahgittin: agree, as above

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/resources/BundleResource.java
ahgittin: agree, as above

ahgittin:
all updated as discussed, running tests now.

misc notes:

* entitlements reuses see_catalog_item for now, for bundles and types; that should be tidied up at some point
* `POST /catalog/bundles` has a different return type to `POST /catalog` so i've kept that; in v2 we should switch
* create autodetecting removed from `/catalog/bundles`

core/src/main/java/org/apache/brooklyn/core/typereg/RegisteredTypePredicates.java
```diff
@@ -194,27 +219,42 @@ public boolean apply(@Nullable RegisteredType item) {
...
+            public boolean apply(Object input) {
+                if (input instanceof Class) input = ((Class<?>)input).getName();
+                return filter.equals(input);
+            }
+        });
     }
     
     private static class AnySuperTypeMatches implements Predicate<RegisteredType> {
-        private final Predicate<Class<?>> filter;
+        private final Predicate<Object> filter;
```
aledsage: Is this definitely not persisted anywhere? If it is, then changing the generics could cause weird errors later.

core/src/main/java/org/apache/brooklyn/core/mgmt/ha/OsgiArchiveInstaller.java
```diff
@@ -60,6 +60,7 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.google.common.base.Function;
```
aledsage: I only just noticed that a bunch of this code is in `core/mgmt/ha`. Why is that? I'd have thought it would equally apply to non-ha (e.g. persisting your state locally, and then restarting Brooklyn).

rest/rest-api/src/main/java/org/apache/brooklyn/rest/api/CatalogApi.java
```diff
@@ -153,6 +153,9 @@ public Response createFromUpload(
             @QueryParam("forceUpdate") @DefaultValue("false")
                     boolean forceUpdate);
     
+    /** @deprecated since 0.13.0 delete the bundle via DELETE /catalog/bundles/xxx */
+    // but we will probably keep this around for a while as many places use it
+    @Deprecated
```
aledsage: I'm still uncomfortable about the way we're handling our versioning, and with deprecating `/v1/catalog` methods. To me, the old `/catalog` is the "v1 way of doing it". Once we do a properly versioned `/v2` then it will be clear that v1 is deprecated (and we'll document it as such).

But we can go with this for now.

For future work, it would be great if we can get a process in place by which we version the api in a sensible way: to help people understand what they are using, what is replacing it, and how/when it will be deprecated.

A big problem with `@Deprecated` here is that I don't think an api user is told about it at all (it's not even in the swagger), so from the perspective of a rest-api-user it's as though it's not deprecated.

rest/rest-api/src/main/java/org/apache/brooklyn/rest/domain/EnricherConfigSummary.java
```diff
@@ -20,60 +20,24 @@
...
-import java.util.Objects;
 
 import org.apache.brooklyn.config.ConfigKey;
 
-import com.fasterxml.jackson.annotation.JsonProperty;
-import com.fasterxml.jackson.databind.annotation.JsonSerialize;
-import com.google.common.collect.ImmutableMap;
-
-public class EnricherConfigSummary extends ConfigSummary {
+// TODO remove? this class has no value over its super
```
aledsage: Agreed - I think we can safely delete it. The json returned by `AdjunctConfigSummary` will be identical to `EnricherConfigSummary` etc. But we can do that in a future PR.

Not sure what the difference is between `EntityConfigSummary` and `AdjunctConfigSummary` (it has "pinned" and "constraints"). Adjuncts can have constraints, I believe. For "pinned", the `CatalogConfig` javadoc says "a pinned configuration means that the config key will always be displayed when presenting as editable in the catalog", so I think that also applies to adjuncts (well, most adjuncts and eventually all).

rest/rest-resources/src/test/java/org/apache/brooklyn/rest/resources/BundleAndTypeResourcesTest.java
```diff
@@ -0,0 +1,1098 @@
...
+import org.testng.Assert;
+import org.testng.annotations.Test;
+import org.testng.reporters.Files;
+
+import com.google.common.base.Joiner;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Iterables;
+
+public class BundleAndTypeResourcesTest extends BrooklynRestResourceTest {
```
aledsage: Great tests!

Class is getting pretty big. Maybe worth splitting into separate classes for bundle and type tests (or separating out the invalid jar/bundle/bom tests). But leave as-is for now.

core/src/main/java/org/apache/brooklyn/core/typereg/RegisteredTypePredicates.java
ahgittin: didn't think generics would affect persistence, but i've restored it and made a new class, just in case.

also removed inner classes elsewhere here where they might be returned to the caller.

rest/rest-api/src/main/java/org/apache/brooklyn/rest/domain/EnricherConfigSummary.java
ahgittin: agree, there is no reason to have more than one `ConfigSummary`

rest/rest-api/src/main/java/org/apache/brooklyn/rest/api/CatalogApi.java
ahgittin: agree

core/src/main/java/org/apache/brooklyn/core/mgmt/ha/OsgiArchiveInstaller.java
ahgittin: indeed, longstanding misplacement.  would be good to move `OsgiManager` and the related classes.

ahgittin:
addressed comments - tidied `RegisteredTypePredicates` for persistence, and deprecated all pointless descendants of `ConfigSummary`

PR #811, opened 2017-09-08
fix npe that can occur with WinRM when no subnet IP

ahgittin:
do a better job of returning public IP but above all don't NPE


ahgittin:
note:

* 32285720a098eef1769c202cadd12390d4ef1a7c fixed this for SshML but missed this
* https://github.com/apache/brooklyn-server/pull/536 backported a fix including WinRM to 0.9 but missed master

so this just fills those gaps

PR #813, opened 2017-09-11
Added highlights to policy endpoint

Graeme-Miller:
Policies now return highlights from the REST API. Highlights are a map of ids -> (description, time, task id). They are used to expose more information about a policy, for example the last action or the triggers for a policy.

I have also updated the rebind code such that policies can have there highlights re-bound.

In the future highlights will be extended to all adjuncts.

### DIFF
--- a/api/src/main/java/org/apache/brooklyn/api/mgmt/rebind/mementos/PolicyMemento.java
+++ b/api/src/main/java/org/apache/brooklyn/api/mgmt/rebind/mementos/PolicyMemento.java
@@ -21,6 +21,7 @@ package org.apache.brooklyn.api.mgmt.rebind.mementos;
 import java.util.Map;
 
 import org.apache.brooklyn.api.mgmt.rebind.RebindSupport;
+import org.apache.brooklyn.api.objs.HighlightTuple;
 
 /**
  * Represents the state of an policy, so that it can be reconstructed (e.g. after restarting brooklyn).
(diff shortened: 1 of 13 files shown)


ahgittin:
looks good, nice and straightforward.  will run some tests then merge assuming all good.

ahgittin:
works a treat, did a sample impl on `ServiceRestarter`, and tested persistence too.  merging this.  i'll open a follow-on PR which will address the points here and show how it's used in that class.

ahgittin:
@m4rkmckenna re your general comment, a human client will use this to understand what's going on.  currently adjuncts (policies etc) are opaque unless you look at logs or tasks (or code).  highlights are an arbitrary set of summary info -- although with some conventions -- which let an adjunct communicate to a user at a high level what it has been doing.  on the premise that a little bit of sensibly distilled info is better than lots of raw info -- but sensible distillation does depend on the adjunct author a bit.  (i'd like to have activities properly tagged to adjuncts as well, so we can also expose lots of "raw" info.)

your point about javadoc is a good one so people know how to use these -- added in #818.

i prefer the name `HighlightTuple` as there is nothing adjunct-specific about them, just they are only used for adjuncts. 

i also prefer `Objects.equal` and `hashCode` _but_ if your IDE generates this for you or it has already been written i don't think it's worth the time to change ... doesn't create any value for user!

PR #815, opened 2017-09-11
BROOKLYN-532: karaf separate mgmt init and start

aledsage:
See https://issues.apache.org/jira/browse/BROOKLYN-532. This implements the last of the suggestions in https://issues.apache.org/jira/browse/BROOKLYN-532?focusedCommentId=16161370&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-16161370

### DIFF
--- a/karaf/start/src/main/resources/OSGI-INF/blueprint/blueprint.xml
+++ b/karaf/start/src/main/resources/OSGI-INF/blueprint/blueprint.xml
@@ -1,0 +1,36 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+Copyright 2015 The Apache Software Foundation.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+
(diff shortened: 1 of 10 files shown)


aledsage:
I tested this by using `brooklyn-dist/karaf/apache-brooklyn/target/assembly/`, running `./bin/start`.

I added a `Thread.sleep()` into the code to check that the rest-api and web-console started up as soon as karaf-init was done (they did).
(Oops - I left that config option in `BasicLauncher.startPartTwo()`, I'd delete that!)

I also did a `bundle:refresh 123` to refresh `brooklyn-core` and confirmed that it started up correctly.

If one just does a `bundle:restart` on only the `brooklyn-karaf-start` then it fails (because can't call `start()` multiple times), but if `brooklyn-karaf-init` is first refreshed then it works as expected.

karaf/start/src/main/resources/OSGI-INF/blueprint/blueprint.xml
```diff
@@ -0,0 +1,36 @@
...
+<blueprint xmlns="http://www.osgi.org/xmlns/blueprint/v1.0.0"
+           xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+           xmlns:cm="http://aries.apache.org/blueprint/xmlns/blueprint-cm/v1.2.0"
+           xsi:schemaLocation="
+             http://www.osgi.org/xmlns/blueprint/v1.0.0 http://www.osgi.org/xmlns/blueprint/v1.0.0/blueprint.xsd
+             http://aries.apache.org/blueprint/xmlns/blueprint-cm/v1.2.0 http://aries.apache.org/schemas/blueprint-cm/blueprint-cm-1.2.0.xsd
+             ">
+
+    <!-- From brooklyn-karaf-init -->
+    <reference id="localLauncher"
```
m4rkmckenna: @aledsage Nice ... Took me a bit to figure out how we made sure this didnt get started before part one

aledsage:
retest this please

PR #817, opened 2017-09-12
Optimizations

ahgittin:
Minor things I noticed which were chewing up CPU and could be easily fixed

### DIFF
--- a/utils/common/src/main/java/org/apache/brooklyn/util/guava/TypeTokens.java
+++ b/utils/common/src/main/java/org/apache/brooklyn/util/guava/TypeTokens.java
@@ -68,5 +68,19 @@ public class TypeTokens {
     public static <T> Class<T> getRawRawType(TypeToken<T> token) {
         return (Class)token.getRawType();
     }
+
+    /** Checks that if both type and token are supplied, either exactly one is null, or 
+     * they both refer to the same non-null type */
+    public static <T> void checkCompatibleOneNonNull(Class<? super T> type, TypeToken<T> typeToken) {
+        if ((type==null && typeToken!=null) || (type!=null && typeToken==null)) {
+            return;
+        }
+        if (type==null && typeToken==null) {
+            throw new NullPointerException("Type not set (neither class or type token)");
+        }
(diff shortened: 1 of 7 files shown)


utils/common/src/main/java/org/apache/brooklyn/util/guava/TypeTokens.java
```diff
@@ -68,5 +85,19 @@
...
+     * they both refer to the same non-null type */
+    public static <T> void checkCompatibleOneNonNull(Class<? super T> type, TypeToken<T> typeToken) {
+        if ((type==null && typeToken!=null) || (type!=null && typeToken==null)) {
+            return;
+        }
+        if (type==null && typeToken==null) {
+            throw new NullPointerException("Type not set (neither class or type token)");
+        }
+        if (!type.equals(typeToken.getRawType())) {
+            throw new NullPointerException("Invalid types, token is "+typeToken+" (raw "+typeToken.getRawType()+") but class is "+type);
```
aledsage: This should be `IllegalArgumentException` or `IllegalStateException`, rather than NPE.

utils/common/src/main/java/org/apache/brooklyn/util/guava/TypeTokens.java
```diff
@@ -24,6 +24,14 @@
 
 public class TypeTokens {
 
+    // creating TypeToken is surprisingly expensive so cache these common ones
+    public static TypeToken<String> STRING = TypeToken.of(String.class);
```
aledsage: If leaving as-is, then these should be `final`.

I'd have gone for:
```
public static final Map<Class<?>, TypeToken<?>> COMMON_TYPE_TOKENS = ImmutableMap.<Class<?>, TypeToken<?>>builder()
        .put(String.class, TypeToken.of(String.class))
        ...
        .build();
```

And:
```
TypeToken<?> result = COMMON_TYPE_TOKENS.get(raw);
if (result == null) result = TypeToken.of((Class<T>)raw);
return (TypeToken<T>) result;
```

ahgittin:
good comments @aledsage - thx - will apply both and merge

PR #818, opened 2017-09-13
More on adjunct highlights

ahgittin:
Applies #813 to a policy (`ServiceRestarter`) to see how it works.  As reported there it works good, REST and persistence acting as expected.

The API for policy authors was a little awkward, and there were some PR review comments in #813.  This addresses both of these.

I will be adding highlight info to more policies and adjuncts.  If this is reviewed in time I'll do it in a different branch, else I'll add them here for simplicity.

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/objs/AbstractEntityAdjunct.java
+++ b/core/src/main/java/org/apache/brooklyn/core/objs/AbstractEntityAdjunct.java
@@ -29,11 +29,14 @@ import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicBoolean;
 
+import javax.annotation.Nullable;
+
 import org.apache.brooklyn.api.entity.Entity;
 import org.apache.brooklyn.api.entity.EntityLocal;
 import org.apache.brooklyn.api.entity.Group;
 import org.apache.brooklyn.api.mgmt.ExecutionContext;
 import org.apache.brooklyn.api.mgmt.SubscriptionHandle;
+import org.apache.brooklyn.api.mgmt.Task;
 import org.apache.brooklyn.api.objs.BrooklynObject;
 import org.apache.brooklyn.api.objs.Configurable;
 import org.apache.brooklyn.api.objs.EntityAdjunct;
@@ -42,7 +45,6 @@ import org.apache.brooklyn.api.sensor.AttributeSensor;
 import org.apache.brooklyn.api.sensor.Sensor;
 import org.apache.brooklyn.api.sensor.SensorEventListener;
 import org.apache.brooklyn.config.ConfigKey;
-import org.apache.brooklyn.core.config.BasicConfigKey;
 import org.apache.brooklyn.core.config.ConfigConstraints;
 import org.apache.brooklyn.core.config.ConfigKeys;
 import org.apache.brooklyn.core.config.internal.AbstractConfigMapImpl;
@@ -113,9 +115,13 @@ public abstract class AbstractEntityAdjunct extends AbstractBrooklynObject imple
 
     private Map<String, HighlightTuple> highlights = new HashMap<>();
 
+    /** Name of a highlight that indicates the last action taken by this adjunct. */
     public static String HIGHLIGHT_NAME_LAST_ACTION = "lastAction";
+    /** Name of a highlight that indicates the last confirmation detected by this adjunct. */
     public static String HIGHLIGHT_NAME_LAST_CONFIRMATION= "lastConfirmation";
+    /** Name of a highlight that indicates the last violation detected by this adjunct. */
     public static String HIGHLIGHT_NAME_LAST_VIOLATION= "lastViolation";
+    /** Name of a highlight that indicates the triggers for this adjunct. */
     public static String HIGHLIGHT_NAME_TRIGGERS = "triggers";
 
     public AbstractEntityAdjunct() {
@@ -135,10 +141,6 @@ public abstract class AbstractEntityAdjunct extends AbstractBrooklynObject imple
                 FlagUtils.checkRequiredFields(this);
             }
         }
-    }
-
-    protected void addHighlight(String name, HighlightTuple tuple) {
-        highlights.put(name, tuple);
     }
 
     /**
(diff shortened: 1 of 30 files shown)


ahgittin:
See https://github.com/apache/brooklyn-server/pull/818/commits/5ae9c41c1dc8d2884598a2375581d2c4dc14ce14#diff-0724222c7ef44a5337e1e8f44d9d5faeR114 for illustration of what I mean by an easier adjunct java author API (just some convenience methods to allow DRY).

ahgittin:
Now has highlights on most things.

core/src/main/java/org/apache/brooklyn/core/objs/AbstractEntityAdjunct.java
```diff
@@ -568,6 +570,98 @@ public void setUniqueTag(String uniqueTag) {
...
+     * Also see convenience methods eg  {@link #highlightOngoing(String, String)} and {@link #highlight(String, String, Task)}
+     * and {@link HighlightTuple}. 
+     */
+    protected void setHighlight(String name, HighlightTuple tuple) {
+        highlights.put(name, tuple);
+    }
+
+    /** As {@link #setHighlight(String, HighlightTuple)}, convenience for recording an item which is intended to be ongoing. */
+    protected void highlightOngoing(String name, String description) {
+        highlights.put(name, new HighlightTuple(description, 0, null));
```
Graeme-Miller: Should be calling setHighlight, rather than putting directly into the map. This would allow these methods to remain uncoupled from how we store the highlights. Same for the below highlight methods.

core/src/main/java/org/apache/brooklyn/core/objs/AbstractEntityAdjunct.java
```diff
@@ -568,6 +570,98 @@ public void setUniqueTag(String uniqueTag) {
...
+    protected <T> void highlightTriggers(Sensor<?> s, Object source) {
+        highlightTriggers(Collections.singleton(s), Collections.singleton(source));
+    }
+    protected <T> void highlightTriggers(Iterable<? extends Sensor<? extends T>> s, Object source) {
+        highlightTriggers(s, (Iterable<?>) (source instanceof Iterable ? (Iterable<?>)source : Collections.singleton(source)));
+    }
+    protected <U> void highlightTriggers(Sensor<?> s, Iterable<U> sources) {
+        highlightTriggers(Collections.singleton(s), sources);
+    }
+    protected <T,U> void highlightTriggers(Iterable<? extends Sensor<? extends T>> sensors, Iterable<U> sources) {
```
Graeme-Miller: This method would benefit from comments to explain what it is doing, with multiple loops and nested ifs it isn't easy to tell at a glance.

core/src/main/java/org/apache/brooklyn/core/objs/AbstractEntityAdjunct.java
```diff
@@ -568,6 +570,98 @@ public void setUniqueTag(String uniqueTag) {
...
+        boolean firstWord = true;
+        if (sensors!=null) for (Object s: sensors) {
+            if (s==null) continue;
+            if (!firstWord) { msg.append(", "); } else { firstWord = false; }
+            // s is normally a sensor but forgive other things if caller cheated generics
+            msg.append(s instanceof Sensor ? ((Sensor<?>)s).getName() : s.toString());
+        }
+        if (firstWord) msg.append("<nothing>");
+        
+        firstWord = true;
```
Graeme-Miller: lines 620 -> 639 are confusing and it is difficult to extract what the code is supposed to do. I had a go at what I think is a simpler solution:
```
        if(!myList.isEmpty() && !(myList.size() == 1 && myList.iterator().next().equals(getEntity())) {
            msg.append(" on ");
            List<String> stringList = myList.stream().map((Sensor s) -> {
                if (s.equals(getEntity())) {
                    return "self";
                } else if (s instanceof Sensor) {
                    return ((Sensor<?>) s).getName();
                } else {
                    return s.toString();
                }
            }).collect(Collectors.toList());
            msg.append(String.join(",", stringList));
        }
```

grkvlt: Interesting @Graeme-Miller - Not sure that you need the second clause in your `if`, as the body handles the `self` output the same as the original code. I'd prefer the `Collectors.joining(", ")` collector, to get a `String` out directly... Also, I'm not sure you should try and chest the lambda type inference like that, just go `s -> { /* stuff */ }` and it'll be fine. Redoing the whole method with lambdas gave me this, which is almost exactly the same length ;)

```
    protected <T,U> void highlightTriggers(Iterable<? extends Sensor<? extends T>> sensors, Iterable<U> sources) {
        StringBuilder msg = new StringBuilder("Listening for ");
        if (Iterables.isEmpty(sensors)) {
            msg.append("<nothing>");
        } else {
            String sensorsText = StreamSupport.stream(sensors.spliterator(), false)
                    .filter(s -> s != null)
                    .map(s -> {
                        if (s instanceof Sensor) {
                            return ((Sensor<?>) s).getName();
                        } else {
                            return s.toString();
                        }
                    })
                    .collect(Collectors.joining(", "));
            msg.append(sensorsText);
        }
        if (!Iterables.isEmpty(sources)) {
            String sourcesText = StreamSupport.stream(sources.spliterator(), false)
                    .filter(s -> s != null)
                    .map(s -> {
                        if (s.equals(getEntity())) {
                            return "self";
                        } else if (s instanceof Sensor) {
                            return ((Sensor<?>) s).getName();
                        } else {
                            return s.toString();
                        }
                    })
                    .collect(Collectors.joining(", "));
            msg.append(" on ")
               .append(sourcesText);
        }
        highlightTriggers(msg.toString());
    }
```

core/src/main/java/org/apache/brooklyn/core/objs/AbstractEntityAdjunct.java
```diff
@@ -568,6 +570,98 @@ public void setUniqueTag(String uniqueTag) {
...
+    protected <T> void highlightTriggers(Iterable<? extends Sensor<? extends T>> s, Object source) {
+        highlightTriggers(s, (Iterable<?>) (source instanceof Iterable ? (Iterable<?>)source : Collections.singleton(source)));
+    }
+    protected <U> void highlightTriggers(Sensor<?> s, Iterable<U> sources) {
+        highlightTriggers(Collections.singleton(s), sources);
+    }
+    protected <T,U> void highlightTriggers(Iterable<? extends Sensor<? extends T>> sensors, Iterable<U> sources) {
+        StringBuilder msg = new StringBuilder("Listening for ");
+        boolean firstWord = true;
+        if (sensors!=null) for (Object s: sensors) {
```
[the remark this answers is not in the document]
m4rkmckenna: Agree for the sake of readability please add braces and a line break

core/src/main/java/org/apache/brooklyn/core/objs/AbstractEntityAdjunct.java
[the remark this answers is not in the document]
ahgittin: agree

core/src/main/java/org/apache/brooklyn/core/objs/AbstractEntityAdjunct.java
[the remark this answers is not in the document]
ahgittin: clearly bad style, no argument!  i added the guard when debugging, forgot to format it nicely...

core/src/main/java/org/apache/brooklyn/core/objs/AbstractEntityAdjunct.java
Graeme-Miller: @grkvlt very interesting, didn't know about Collectors.joining(", "), that's very handy.

re. the second clause in the if. My understanding of Alex's code is that, if there is only one entry, and that entry is self, then we shouldn't add "on...". The second clause in the if is to guard against that. Maybe Alex can clarify?

ahgittin: great suggestions -- so used to using lambas in JS but not in java :)

using choice operator to make more concise, going with:

```
        if (sensors==null || Iterables.isEmpty(sensors)) {
            msg.append("<nothing>");
        } else {
            String sensorsText = StreamSupport.stream(sensors.spliterator(), false)
                    .filter(s -> s != null)
                    .map(s -> (s instanceof Sensor ? ((Sensor<?>) s).getName() : s.toString()))
                    .collect(Collectors.joining(", "));
            msg.append(sensorsText);
        }

        if (sources!=null && !Iterables.isEmpty(sources)) {
            String sourcesText = StreamSupport.stream(sources.spliterator(), false)
                    .filter(s -> s != null)
                    .map(s -> (s.equals(getEntity()) ? "self" : s.toString()))
                    .collect(Collectors.joining(", "));
            if (!"self".equals(sourcesText)) {
                msg.append(" on ").append(sourcesText);
            }
        }
```

ahgittin:
all addressed, thanks.  if build passes then let's merge.

core/src/main/java/org/apache/brooklyn/core/objs/AbstractEntityAdjunct.java
grkvlt: Looks nice @ahgittin - it's taken me a while to get used to operating in this style with Java as well, but it does makes things both concise and still readable...

ahgittin:
fixed test failure as streaming with lambdas enforces generic information (did not know that - kinda cool actually)

PR #820, opened 2017-09-13
use type registry when confirming catalog on startup

ahgittin:
prevents errors trying to check legacy catalog items

#814 caused legacy CI instances to be created for TR items which could cause errors; this limits search to legacy CIs to prevent error, and confirms TR items also


ahgittin:
brooklyn CLI gives error in some startup configurations with #814 and this is minor so committing before review but review welcome

m4rkmckenna:
@ahgittin Please do not merge your own PRs without someone reviewing them. 

Its really bad form

PR #822, opened 2017-09-13
New periodic and scheduled effector policies

grkvlt:
Adds policies to execute effectors at specific times, either singly or repeated.

### DIFF
--- a/policy/src/main/java/org/apache/brooklyn/policy/action/PeriodicEffectorPolicy.java
+++ b/policy/src/main/java/org/apache/brooklyn/policy/action/PeriodicEffectorPolicy.java
@@ -18,18 +18,17 @@
  */
 package org.apache.brooklyn.policy.action;
 
-import java.util.Map;
-import java.util.concurrent.TimeUnit;
+import java.util.List;
 
 import org.apache.brooklyn.api.effector.Effector;
 import org.apache.brooklyn.api.entity.EntityLocal;
 import org.apache.brooklyn.api.policy.Policy;
 import org.apache.brooklyn.config.ConfigKey;
 import org.apache.brooklyn.core.config.ConfigKeys;
-import org.apache.brooklyn.util.collections.MutableMap;
+import org.apache.brooklyn.util.exceptions.Exceptions;
+import org.apache.brooklyn.util.text.Strings;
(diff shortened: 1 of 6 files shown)


ahgittin:
cool @grkvlt - wonder if you could do two things:

* merge #818 here for testing and follow the pattern there to set trigger and action highlights
* include a YAML illustrating use of these policies

ahgittin:
(yaml could be used in a test :nudge:)

policy/src/main/java/org/apache/brooklyn/policy/action/PeriodicEffectorPolicy.java
```diff
@@ -0,0 +1,151 @@
...
+ *       period: 1 day
+ *       time: 08:00:00
+ *   - type: org.apache.brooklyn.policy.action.ScheduledEffectorPolicy
+ *     brooklyn.config:
+ *       effector: resize
+ *       args:
+ *         desiredSize: 1
+ *       period: 1 day
+ *       time: 18:00:00
+ * }</pre>
```
drigodwin: This example is for the `ScheduledEffectorPolicy` not the `PeriodicEffectorPolicy`

policy/src/main/java/org/apache/brooklyn/policy/action/PeriodicEffectorPolicy.java
```diff
@@ -0,0 +1,151 @@
...
+    public void setEntity(final EntityLocal entity) {
+        super.setEntity(entity);
+
+        AttributeSensor<Boolean> sensor = config().get(START_SENSOR);
+        subscriptions().subscribe(entity, sensor, handler);
+    }
+
+    @Override
+    public void rebind() {
+        // XXX Will the AtomicBoolean field be saved during rebind?
```
drigodwin: Probably resolve this question before merging...

drigodwin:
I have added a PR for the documentation for this [here](https://github.com/apache/brooklyn-docs/pull/213).

grkvlt:
@ahgittin I will do some testing of how this interacts with #818 but it will be in a separate PR so that this can get into the 0.12 RC

PR #823, opened 2017-09-14
Fix creation script

Graeme-Miller:
An issue was created by https://github.com/apache/brooklyn-server/pull/812.

That PR removed the password needed to create the user brooklyn'@'localhost

### DIFF
--- a/camp/camp-brooklyn/src/test/resources/visitors-creation-script.sql
+++ b/camp/camp-brooklyn/src/test/resources/visitors-creation-script.sql
@@ -20,13 +20,12 @@ create database visitors;
 use visitors;
 
 
-# the below will create user (and note create user not supported in some dialects)
+# Create two users: 'brooklyn'@'%' and 'brooklyn'@'localhost' with full access
 # default password specified for compatibility with older blueprints that don't provide config;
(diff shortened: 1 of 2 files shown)


Graeme-Miller:
I have tested this on a MySQLNode and it works

ahgittin:
@Graeme-Miller @drigodwin odd - local seemed to work for me but sometimes mysql is weird which NIC/socket it uses for localhost, and different versions are different.  good fix.

two things--

* is the same update made to brooklyn-library ?

* is there a reason we're not consistent what perms we give to the two users?  (this PR gives `usage` to the users @ `%` but `all` to @ `localhost` -- previously was `all on visitors` and `usage` elsewhere, to both users) ... in other words should the file read

```
grant usage on *.* to 'brooklyn'@'%' identified by '${config["creation.script.password"]!"br00k11n"}';
grant all privileges on visitors.* to 'brooklyn'@'%';
# useful if sockets work also
grant all privileges on *.* to 'brooklyn'@'localhost' identified by '${config["creation.script.password"]!"br00k11n"}';
grant all privileges on visitors.* to 'brooklyn'@'localhost';
 ```

PR #824, opened 2017-09-15
Config for VanillaSoftwareProcess disabling ssh polling

aledsage:
See individual commits for more details.

The `FunctionSensor` I primarily added to make it easier to write my `VanillaSoftwareProessYamlTest`, but I think it's a useful addition to allow you to do in yaml what you can do in java entities (i.e. to add a `FunctionFeed` to an entity).

### DIFF
--- a/software/base/src/main/java/org/apache/brooklyn/entity/software/base/VanillaSoftwareProcessImpl.java
+++ b/software/base/src/main/java/org/apache/brooklyn/entity/software/base/VanillaSoftwareProcessImpl.java
@@ -18,20 +18,34 @@
  */
 package org.apache.brooklyn.entity.software.base;
 
+import org.apache.brooklyn.core.entity.lifecycle.ServiceStateLogic.ServiceNotUpLogic;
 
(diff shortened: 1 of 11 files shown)


software/base/src/main/java/org/apache/brooklyn/entity/software/base/VanillaSoftwareProcessImpl.java
```diff
@@ -18,20 +18,40 @@
  */
 package org.apache.brooklyn.entity.software.base;
 
+import org.apache.brooklyn.api.sensor.AttributeSensor;
+import org.apache.brooklyn.core.entity.lifecycle.ServiceStateLogic.ServiceNotUpLogic;
+import org.apache.brooklyn.core.sensor.Sensors;
 
 public class VanillaSoftwareProcessImpl extends SoftwareProcessImpl implements VanillaSoftwareProcess {
+    
+    AttributeSensor<Boolean> SERVICE_LAUNCHED = Sensors.newBooleanSensor("service.process.isLaunched",
```
aledsage: This should be deleted.

aledsage:
Thanks @drigodwin - I'll merge this, and will update the brooklyn docs.

PR #825, opened 2017-09-16
fix non-det failure in scaling down test

ahgittin:
minor test bug - ensure that policy doesn't start before size reaches 6,
and fix execution context bug found in autoscaler changing pool size

### DIFF
--- a/policy/src/test/java/org/apache/brooklyn/policy/autoscaling/AutoScalerPolicyPoolSizeTest.java
+++ b/policy/src/test/java/org/apache/brooklyn/policy/autoscaling/AutoScalerPolicyPoolSizeTest.java
@@ -30,6 +30,8 @@ import org.apache.brooklyn.core.test.entity.TestCluster;
 import org.apache.brooklyn.core.test.entity.TestSizeRecordingCluster;
 import org.apache.brooklyn.entity.stock.BasicStartable;
 import org.apache.brooklyn.util.collections.MutableList;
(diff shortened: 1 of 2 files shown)


policy/src/test/java/org/apache/brooklyn/policy/autoscaling/AutoScalerPolicyPoolSizeTest.java
```diff
@@ -101,10 +103,18 @@ public void testResizeUp() throws Exception {
     @Test
     public void testResizeDown() throws Exception {
         EntityAsserts.assertAttributeEqualsEventually(cluster, TestCluster.GROUP_SIZE, CLUSTER_INIITIAL_SIZE);
+        // temporarily increase, otherwise the policy can sometimes kick in to resize down before size 6 is reached
+        policy.config().set(AutoScalerPolicy.MAX_POOL_SIZE, CLUSTER_MAX_SIZE + 2);
```
aledsage: Looks fine to me - happy to merge.

However, I'd probably have fixe this by changing the `TestSizeRecordingCluster` so that we can get a list of the `desiredSize` values that had been passed into resize. Assuming this is to test the policy, then that would be a good way to ensure it was called with the right values. These changes have removed the test that it automatically scales down when cluster grows beyond the allowed limit - it's only testing that it will automatically scale down when MAX_POOL_SIZE config value changes.

PR #829, opened 2017-09-18
[WIP] BROOKLYN-535: restarter policy when entity stopping

aledsage:
See https://issues.apache.org/jira/browse/BROOKLYN-534

### DIFF
--- a/policy/src/main/java/org/apache/brooklyn/policy/ha/ServiceRestarter.java
+++ b/policy/src/main/java/org/apache/brooklyn/policy/ha/ServiceRestarter.java
@@ -28,9 +28,11 @@ import org.apache.brooklyn.api.sensor.SensorEvent;
 import org.apache.brooklyn.api.sensor.SensorEventListener;
 import org.apache.brooklyn.config.ConfigKey;
 import org.apache.brooklyn.core.config.ConfigKeys;
+import org.apache.brooklyn.core.entity.Attributes;
 import org.apache.brooklyn.core.entity.Entities;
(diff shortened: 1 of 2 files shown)


policy/src/main/java/org/apache/brooklyn/policy/ha/ServiceRestarter.java
```diff
@@ -133,6 +135,10 @@ protected synchronized void onDetectedFailure(SensorEvent<Object> event) {
             LOG.warn("ServiceRestarter suspended, so not acting on failure detected at "+entity+" ("+event.getValue()+")");
             return;
         }
+        if (isEntityStopping()) {
+            highlightViolation("Failure detected but entity stopping");
```
tbouron: Should we highlight the violation here? I guess the question is more: is it really a violation when we are in a stopping state?

policy/src/main/java/org/apache/brooklyn/policy/ha/ServiceRestarter.java
```diff
@@ -133,6 +135,10 @@ protected synchronized void onDetectedFailure(SensorEvent<Object> event) {
...
+            LOG.info("Entity stopping, so ServiceRestarter not acting on failure detected at "+entity+" ("+event.getValue()+")");
```
tbouron: You are missing a `return;`

tbouron:
@aledsage Do you think you can look at my comments above?

nakomis:
@aledsage Can you take a look at this to see if it is still relevant, and address @tbouron 's comment

Thanks

PR #830, opened 2017-09-19
Add support for kubernetes helm

andreaturli:


andreaturli:
retest this please

drigodwin:
retest this please

tbouron:
@andreaturli Might be completely off-base here but wouldn't it make more sense for this to be in the clocker project

andreaturli:
good question @tbouron -- I think Helm is a package manager for Kubernets and helps you deploy application to kubernetes. Seemed to me the `location/container` is a good fit, as it is where brooklyn implements the support to deployTo kubernetes. No strong feelings though

locations/container/pom.xml
```diff
@@ -32,10 +32,50 @@
...
     <dependencies>
+        <dependency>
+            <groupId>org.microbean</groupId>
+            <artifactId>microbean-helm</artifactId>
+            <version>2.5.0.1-SNAPSHOT</version>
+            <exclusions>
+                <!--<exclusion>-->
+                    <!--<groupId>io.grpc</groupId>-->
+                    <!--<artifactId>grpc-netty</artifactId>-->
+                <!--</exclusion>-->
```
tbouron: Can be removed

locations/container/pom.xml
```diff
@@ -32,10 +32,50 @@
...
+                <exclusion>
+                    <groupId>com.google.api.grpc</groupId>
+                    <artifactId>proto-google-common-protos</artifactId>
+                </exclusion>
+                <!--<exclusion>-->
+                    <!--<groupId>com.google.instrumentation</groupId>-->
+                    <!--<artifactId>instrumentation-api</artifactId>-->
+                <!--</exclusion>-->
```
tbouron: Same as above

locations/container/src/main/java/org/apache/brooklyn/container/location/kubernetes/KubernetesLocation.java
```diff
@@ -347,8 +386,8 @@ public Boolean call() {
                 if (result.isEmpty()) {
                     return false;
                 }
-                List<HasMetadata> check = client.resource(result.get(0)).inNamespace(result.get(0).getMetadata().getNamespace()).get();
-                if (result.size() > 1 || check.size() != 1 || check.get(0).getMetadata() == null) {
+                HasMetadata check = client.resource(result.get(0)).inNamespace(result.get(0).getMetadata().getNamespace()).get();
```
tbouron: Is that normal the signature changed from `List<HasMetadata>` to `HasMetadata` ?

locations/container/pom.xml
andreaturli: thanks @tbouron 
also now there is an official release http://search.maven.org/#search%7Cga%7C1%7Ca%3Amicrobean-helm which may be better and needs testing

locations/container/src/main/java/org/apache/brooklyn/container/location/kubernetes/KubernetesLocation.java
andreaturli: think so, but need to double-check as I don't recall the details

andreaturli:
@tbouron I remember I used minikube to test it. Any k8s cluster should work though

locations/container/pom.xml
tbouron: Ah ok, I'll let you decide which one is best then ;)

ahgittin:
would be great to have this merged!  nearly there IMO.

re clocker, deploy _to_ i think of as Brooklyn, and this is squarely in that camp ... deploy _of_ {k8s,docker,etc} is more clocker.

would be nice to have a yaml example -- ideally a docs addition that references an example tested here (is there a brooklyn-docs PR ?).  also nice to have another pair of eyes on this but given the time and value, and i've looked at the code and am happy with it, i'm happy for this to be merged.

andreaturli:
retest this please

andreaturli:
@tbouron as per our review, I've add the creation of some `.helm` dirs required to run on a fresh installation: thanks for spotting the issue!

it basically mimic the same approach taken by the `microbeam-helm` project - https://github.com/microbean/microbean-helm/blob/ae51ed6571af204df9b764f0b2b59f2f8e2c2aa4/.travis.yml#L6

tbouron:
Just tested, works like a charm, thanks @andreaturli !

Just need to make Jenkins happy again and I'll merge it 👍

PR #832, opened 2017-09-19
Add ManagementNodeStateListener support

aledsage:
As discussed on the dev@brooklyn mailing list, subject `UsageListener: notified of promotion to master?`...

Adds support for registering a `ManagementNodeStateListener`, to be notified of the Brooklyn server's state changes such as when it becomes "master".

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/mgmt/internal/ManagementNodeStateListenerManager.java
+++ b/core/src/main/java/org/apache/brooklyn/core/mgmt/internal/ManagementNodeStateListenerManager.java
@@ -1,0 +1,175 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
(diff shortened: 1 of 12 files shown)


core/src/main/java/org/apache/brooklyn/core/mgmt/internal/ManagementNodeStateListenerManager.java
```diff
@@ -0,0 +1,175 @@
...
+                        ((ManagementContextInjectable) listener).setManagementContext(managementContext);
+                    }
+                    listeners.add((ManagementNodeStateListener)listener);
+                }
+            }
+        }
+    }
+
+    @Override
+    public void onStateChange(ManagementNodeState state) {
```
Graeme-Miller: seems a bit strange that we can call onStateChange with the same state and have the ManagementNodeStateListenerManager filter it out. Feel like it should be the responsibility of the HighAvailabilityManagerImpl, but it is neat to include it in ManagementNodeStateListenerManager

aledsage:
retest this please

core/src/main/java/org/apache/brooklyn/core/mgmt/internal/ManagementNodeStateListenerManager.java
aledsage: Agreed it's a bit strange. But I didn't want to complicate the `HighAvailabilityManagerImpl` more. It would only be a few lines in there (a new field to store the last value, etc) - I could move it there if you think it's better @Graeme-Miller?

Graeme-Miller: not sure, but I think it is fine as it is. Your call.

aledsage:
jenkins build failed with:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-install-plugin:2.5.2:install (default-install) on project brooklyn-core: Failed to install artifact org.apache.brooklyn:brooklyn-core:jar:tests:0.13.0-SNAPSHOT: No space left on device -> [Help 1]
```

aledsage:
retest this please

PR #833, opened 2017-09-19
Fix PeriodicEffectorPolicy

aledsage:
The `PeriodicEffectorPolicy` rebind test(s) were failing non-deterministically in the jenkins master build, and same for me locally.

This PR makes a bunch of changes to fix those, along with various other improvements to the policy implementations. The config / documented behaviour of the policies should be unaffected by these changes.

### DIFF
--- a/policy/src/main/java/org/apache/brooklyn/policy/action/AbstractScheduledEffectorPolicy.java
+++ b/policy/src/main/java/org/apache/brooklyn/policy/action/AbstractScheduledEffectorPolicy.java
@@ -40,6 +40,7 @@ import org.apache.brooklyn.core.config.ConfigKeys;
 import org.apache.brooklyn.core.entity.EntityInitializers;
 import org.apache.brooklyn.core.entity.trait.Startable;
 import org.apache.brooklyn.core.policy.AbstractPolicy;
+import org.apache.brooklyn.util.collections.MutableList;
 import org.apache.brooklyn.util.core.config.ConfigBag;
 import org.apache.brooklyn.util.core.config.ResolvingConfigBag;
 import org.apache.brooklyn.util.core.task.Tasks;
@@ -53,9 +54,9 @@ import org.slf4j.LoggerFactory;
 
 import com.google.common.annotations.Beta;
 import com.google.common.base.Predicates;
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Iterables;
-import com.google.common.collect.Lists;
 import com.google.common.reflect.TypeToken;
 
 @Beta
@@ -101,7 +102,7 @@ public abstract class AbstractScheduledEffectorPolicy extends AbstractPolicy imp
 
     public static final ConfigKey<Boolean> RUNNING = ConfigKeys.builder(Boolean.class)
             .name("running")
-            .description("Set if the executor has started")
+            .description("[INTERNAL] Set if the executor has started")
             .defaultValue(Boolean.FALSE)
             .reconfigurable(true)
             .build();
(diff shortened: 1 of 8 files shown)


policy/src/main/java/org/apache/brooklyn/policy/action/ScheduledEffectorPolicy.java
```diff
@@ -59,8 +60,8 @@ public ScheduledEffectorPolicy() {
     public void setEntity(final EntityLocal entity) {
         super.setEntity(entity);
 
-        subscriptions().subscribe(entity, INVOKE_IMMEDIATELY, this);
-        subscriptions().subscribe(entity, INVOKE_AT, this);
+        subscriptions().subscribe(ImmutableMap.of("notifyOfInitialValue", true), entity, INVOKE_IMMEDIATELY, this);
```
grkvlt: This makes sense, good idea

policy/src/main/java/org/apache/brooklyn/policy/action/AbstractScheduledEffectorPolicy.java
```diff
@@ -233,8 +269,8 @@ public synchronized void run() {
             Object result = entity.invoke(effector, resolved).getUnchecked();
             LOG.debug("{}: Effector {} returned {}", new Object[] { this, effector.getName(), result });
         } catch (RuntimeInterruptedException rie) {
-            Thread.interrupted();
-            // TODO sometimes this seems to hang the executor?
+            // Gracefully stop
+            Thread.currentThread().interrupt();
```
grkvlt: Thanks, I couldn't remember the right way to handle this.

policy/src/main/java/org/apache/brooklyn/policy/action/AbstractScheduledEffectorPolicy.java
```diff
@@ -101,15 +102,15 @@
...
+            .description("[INTERNAL] Set if the executor has started")
             .defaultValue(Boolean.FALSE)
             .reconfigurable(true)
             .build();
 
     public static final ConfigKey<List<Long>> SCHEDULED = ConfigKeys.builder(new TypeToken<List<Long>>() { })
             .name("scheduled")
             .description("List of all scheduled execution start times")
-            .defaultValue(Lists.newCopyOnWriteArrayList())
+            .defaultValue(ImmutableList.of())
```
grkvlt: We can use the non-COW version because you just re-set the config with a mutated copy, right?

policy/src/main/java/org/apache/brooklyn/policy/action/AbstractScheduledEffectorPolicy.java
```diff
@@ -210,16 +227,35 @@ protected Duration getWaitUntil(String time) {
         }
     }
 
+    protected Date parseTime(String time) throws ParseException {
+        boolean formatted = time.contains(":"); // FIXME deprecated TimeDuration coercion
+        if (formatted) {
+            synchronized (FORMATTER) {
+                // DateFormat is not thread-safe; docs say to use one-per-thread, or to synchronize externally
+                return FORMATTER.parse(time);
```
grkvlt: Good catch

policy/src/test/java/org/apache/brooklyn/policy/action/AbstractEffectorPolicyTest.java
```diff
@@ -0,0 +1,81 @@
...
+import org.apache.brooklyn.core.test.BrooklynAppUnitTestSupport;
+import org.apache.brooklyn.core.test.entity.TestEntity;
+import org.apache.brooklyn.test.Asserts;
+import org.apache.brooklyn.util.time.Duration;
+
+import com.google.common.base.Predicates;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Iterables;
+
+public class AbstractEffectorPolicyTest extends BrooklynAppUnitTestSupport {
```
grkvlt: I think this looks better, I should probably have done the same and lifted shared parts up out of the tests.

policy/src/main/java/org/apache/brooklyn/policy/action/AbstractScheduledEffectorPolicy.java
aledsage: If you set the defaultValue to something mutable, then if two different policy instances retrieve and modify it, they'll see each other's changes!

And yes, because we're modifying a copy we can use the non-COW version.

PR #835, opened 2017-09-20
Tasks subsystem improvements - adjuncts, names, efficiency, GC

ahgittin:
* Adjuncts' tasks run in their context with tag so can be found more easily
* Better naming of tasks across the board (so the look nicer in the UI)
* More tidies to same-thread execution and run more things in same thread
* Tidied task GC code, and changed to GC less aggressively
* Fix race in initialization tasks

This increases the number of tasks kept around because sensor events are actually interesting (previously they were always garbage collected).  However this may impact performance at scale: that deserves some investigation.  There may be a balance to be had re which tasks to keep around and for how long.

NB this builds on #816, review that first


ahgittin:
Completed this round of improvements.  A few more improvements:

* many dynamic tasks run in same thread where possible, reducing overhead and supporting immediate execution
* fix bug where subscriptions could be delivered out of order, where last subscription might not correspond to last value

The last item, in https://github.com/apache/brooklyn-server/pull/835/commits/9213f0e25288114718f2a06a50ee206bf0120561, does a bit more synching so is slightly risky, but I've minimized this and I think it is the correct thing to do, consistent at least with other code paths.  Testing so far is good with it and there are ample notes on why.

What I have _not_ done here is deprecation or warning in `getImmediately(Task)`, as suggested in #816.  That's turning in to a big job so I'll do it in another PR from which I'll reference this (so you should see a message about it below shortly.)

core/src/main/java/org/apache/brooklyn/core/mgmt/internal/NonDeploymentManagementContext.java
```diff
@@ -254,7 +250,19 @@ public synchronized SubscriptionContext getSubscriptionContext(Entity entity) {
         if (!this.entity.equals(entity)) throw new IllegalStateException("Non-deployment context "+this+" can only use a single Entity: has "+this.entity+", but passed "+entity);
         if (mode==NonDeploymentManagementContextMode.MANAGEMENT_STOPPED)
             throw new IllegalStateException("Entity "+entity+" is no longer managed; subscription context not available");
-        return subscriptionContext;
+        // see also AbstractManagementContext.getSubscriptionContext - needed for callbacks, to ensure the correct entity context is set
+        Map<String, ?> subscriptionFlags = ImmutableMap.of("tags", ImmutableList.of(BrooklynTaskTags.tagForContextEntity(entity)));
+        return new BasicSubscriptionContext(subscriptionFlags, qsm, entity);
```
aledsage: Why instantiate this in the `getSubscriptionContext` rather than the constructor?
Do you really want a new instance every time it is called, rather than lazy-instantiation and returning the same one?

For the record, I don't think it should really matter (i.e. fine as-is) - at some point we should be able to delete lots of this code. It was originally written when folk could call entity constructors directly and then call methods on it (so the entity would not have a usable management context).

core/src/main/java/org/apache/brooklyn/core/mgmt/rebind/RebindManagerImpl.java
```diff
@@ -490,15 +490,7 @@ public ChangeListener getChangeListener() {
...
-            Task<List<Application>> task = ec.submit(new Callable<List<Application>>() {
-                @Override public List<Application> call() throws Exception {
-                    return rebindImpl(classLoader, exceptionHandler, mode);
-                }});
-            try {
-                return task.get();
-            } catch (Exception e) {
-                throw Exceptions.propagate(e);
-            }
+            return ec.get(Tasks.<List<Application>>builder().displayName("rebind").dynamic(false).body(() -> rebindImpl(classLoader, exceptionHandler, mode)).build());
```
aledsage: Minor: I prefer splitting it onto multiple lines (this line is now 168 chars long, the first 12 being the indent).

core/src/main/java/org/apache/brooklyn/core/objs/AbstractEntityAdjunct.java
```diff
@@ -402,10 +415,21 @@ public void setDisplayName(String name) {
...
+            return ((EntityInternal)entity).getManagementContext();
+        }
+        return null;
+    }
+    
     public void setEntity(EntityLocal entity) {
         if (destroyed.get()) throw new IllegalStateException("Cannot set entity on a destroyed entity adjunct");
         this.entity = entity;
-        this.execution = ((EntityInternal) entity).getExecutionContext();
+        this.execution = new BasicExecutionContext( getManagementContext().getExecutionManager(),
```
aledsage: I'd prefer us to add `managementContext.getExecutionContext(entity, adjunct)` (so it follows the same pattern as for `getSubscriptionContext`).

core/src/main/java/org/apache/brooklyn/util/core/task/BasicExecutionManager.java
```diff
@@ -797,6 +800,7 @@ public TaskLookup(BasicExecutionManager mgr, Task<?> t) {
                 .description("Details of the original task have been forgotten.")
                 .body(Callables.returning((T)null)).build();
             ((BasicTask<T>)t).ignoreIfNotRun();
+            ((BasicTask<T>)t).cancelled = true;
```
aledsage: Worth a comment saying why you're setting `cancelled = true`. I presume it's because we don't really want anyone executing the "gone" task, and more importantly that if we are GC'ing tasks then marking the "gone" task as cancelled will help with cleanup of sub-tasks that have lost their submitted-by-task reference?

core/src/test/java/org/apache/brooklyn/core/entity/EntityAssertsTest.java
```diff
@@ -85,12 +83,7 @@ public void shouldAssertAttributeEqualsEventually() throws Exception {
         entity.sensors().set(TestEntity.NAME, "before");
         final String after = "after";
 
-        Task<?> assertValue = entity.getExecutionContext().submit(new Runnable() {
-            @Override
-            public void run() {
-                EntityAsserts.assertAttributeEqualsEventually(entity, TestEntity.NAME, after);
-            }
-        });
+        Task<?> assertValue = entity.getExecutionContext().submit("assert attr equals", () -> EntityAsserts.assertAttributeEqualsEventually(entity, TestEntity.NAME, after));
```
aledsage: Preference breaking a long line across multiple lines.

core/src/test/java/org/apache/brooklyn/core/mgmt/internal/LocalSubscriptionManagerTest.java
```diff
@@ -170,4 +181,84 @@ public void run() {
...
+        RecordingSensorEventListener<Object> listener = new RecordingSensorEventListener<>();
+        
+        entity.sensors().set(TestEntity.NAME, "myname");
+        entity.sensors().set(TestEntity.SEQUENCE, 123);
+        entity.sensors().emit(TestEntity.MY_NOTIF, -1);
+
+        // delivery should be in subscription order, so 123 then 456
+        entity.subscriptions().subscribe(ImmutableMap.of("notifyOfInitialValue", true), entity, TestEntity.SEQUENCE, listener);
+        // wait for the above delivery - otherwise it might get dropped
+        Asserts.succeedsEventually(MutableMap.of("timeout", Duration.seconds(5)), () -> {
```
aledsage: Don't bother setting timeouts - just rely on the defaults. If you want things to run faster locally, then set the Java system property `-Dbrooklyn.test.defaultTimeout=5s`.

core/src/test/java/org/apache/brooklyn/core/mgmt/internal/LocalSubscriptionManagerTest.java
```diff
@@ -170,4 +181,84 @@ public void run() {
...
+            synchronized (count) { 
+                entity.sensors().set(TestEntity.SEQUENCE, count.incrementAndGet()); 
+            } 
+        };
+        entity.subscriptions().subscribe(ImmutableMap.of(), entity, TestEntity.SEQUENCE, listener);
+        for (int i=0; i<10; i++) {
+            new Thread(set).start();
+        }
+        
+        Asserts.succeedsEventually(MutableMap.of("timeout", Duration.seconds(5)), () -> {
```
aledsage: As above, don't bother with bespoke timeouts. (Same below).

core/src/test/java/org/apache/brooklyn/core/mgmt/internal/LocalSubscriptionManagerTest.java
ahgittin: ooh handy forgot about this!

ahgittin:
thanks @aledsage - great comments, will adress then merge this and #836

core/src/main/java/org/apache/brooklyn/core/config/ConfigConstraints.java
```diff
@@ -112,12 +115,21 @@ public ConfigConstraints(T brooklynObject) {
     abstract Iterable<ConfigKey<?>> getBrooklynObjectTypeConfigKeys();
 
     public Iterable<ConfigKey<?>> getViolations() {
-        // TODO in new task
-        return validateAll();
+        ExecutionContext exec = 
+            getBrooklynObject() instanceof EntityInternal ? ((EntityInternal)getBrooklynObject()).getExecutionContext() :
+            // getBrooklynObject() instanceof AbstractEntityAdjunct ? ((AbstractEntityAdjunct)getBrooklynObject()).getExecutionContext() :
```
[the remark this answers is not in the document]
ahgittin: yep, in this case it should have been uncommented as part of this PR

core/src/main/java/org/apache/brooklyn/core/mgmt/internal/NonDeploymentManagementContext.java
ahgittin: it's a thin shim so doesn't matter if we cache or recreate

would be nice if we could remove the whole non-deployment mgmt context stage.  suspect there are things like initializers and maybe rebind steps that we do want to run before other things or be queued to be the first things run when management context is attached so it may be that this complexity is still necessary.

core/src/main/java/org/apache/brooklyn/core/mgmt/rebind/RebindManagerImpl.java
ahgittin: definitely, line getting too long to see what it really does

core/src/main/java/org/apache/brooklyn/core/objs/AbstractEntityAdjunct.java
ahgittin: yes, that's cleaner

core/src/main/java/org/apache/brooklyn/util/core/task/BasicExecutionManager.java
ahgittin: never occurred to me that marking cancelled might help with GC.  have added comments.

the only thing i was optimizing was the surprising expense of soft references (noted 30 lines earlier)

ahgittin:
good comments, and straightforward, all addressed

PR #837, opened 2017-09-20
[IMPORTANT] Fix FunctionSensor’s rebind

aledsage:
It was trying to serialise the TypeCoercions function. We don’t need
that - it coerces automatically.

This is important: the brooklyn master build will fail without this fix!


aledsage:
Merging this now to fix the build (and because it's very low risk). But if someone can review it later that would be good!

ahgittin:
LGTM

PR #840, opened 2017-09-21
Fix catalog YAML

drigodwin:
Square brackets within the yaml string meant it was interpreted as an array and broke the initial catalog load. This puts speach marks around so it's interpreted as a string.


aledsage:
Thanks for spotting and fixing! Merging now.

PR #841, opened 2017-09-21
BROOKLYN-538: add failing test

aledsage:
Adds a test that demonstrates https://issues.apache.org/jira/browse/BROOKLYN-538

I'm going to mark issue as "won't fix" for the following reasons:
* The exception is benign - the sensor gets the expected value of `null`
* The exception happens deep within the `JsonPath` library - it might well be the right behaviour for it to fail if you ask for path `$` on a null object.
* Working around it in Brooklyn code would be added complexity for no real benefit.


aledsage:
Closing this PR - it is linked to from https://issues.apache.org/jira/browse/BROOKLYN-538, so we can refer back to it if we ever decide to do anything with BROOkLYN-538 other than "won't fix".

PR #842, opened 2017-09-21
Improve Brooklyn-management of OSGi-already-managed bundles

ahgittin:
Previously could claim it was already installed but not brooklyn-manage it.

Now it is much smarter about checking whether what is being supplied now (by reference or by contents) is an exact match and if so, don't uninstall and reinstall it.

Some annoying issues fixing tests due to how karaf/java deals with `target/classes/` bundles in dev, but tests are all working too now.

NB this requires #810 due to conflicts in osgi bundle removal semantics


core/src/main/java/org/apache/brooklyn/core/mgmt/ha/OsgiArchiveInstaller.java
```diff
@@ -129,58 +133,115 @@ private synchronized void init() {
...
+                            log.debug("Detected bundle "+suppliedKnownBundleMetadata+" installed to Brooklyn already; no URL or stream supplied, so re-using existing installation");
+                            // if bundle is brooklyn-managed simply say "already installed"
+                            result.metadata = existingBrooklynInstalledBundle.get();
+                            result.setIgnoringAlreadyInstalled();
+                            return;
+                            
+                        } else {
+                            // if bundle is not brooklyn-managed we want to make it be so
+                            // and for that we need to find a URL.
+                            // some things declare usable locations, though these might be maven and
```
geomacy: text ends just with `and` - did you mean to put something more?

core/src/main/java/org/apache/brooklyn/core/mgmt/ha/OsgiArchiveInstaller.java
```diff
@@ -129,58 +133,115 @@ private synchronized void init() {
...
-            if (zipIn==null) {
-                if (installedBundle.isPresent()) {
-                    // no way to install (no url), or no need to install (not forced); just ignore it
-                    result.metadata = osgiManager.getManagedBundle(new VersionedName(installedBundle.get()));
-                    if (result.metadata==null) {
-                        // low-level installed bundle
-                        result.metadata = new BasicManagedBundle(installedBundle.get().getSymbolicName(), installedBundle.get().getVersion().toString(),
-                            suppliedKnownBundleMetadata!=null ? suppliedKnownBundleMetadata.getUrl() : null);
+                
+                assert zipIn!=null || suppliedKnownBundleMetadata.getUrl()!=null : "should have found a stream or inferred a URL";
```
geomacy: Do you really mean to use normal Java `assert` here? If it's worth asserting here is it not worth asserting normally at run time ie. without having to configure JVM `-ea`?

core/src/main/java/org/apache/brooklyn/core/mgmt/ha/OsgiArchiveInstaller.java
```diff
@@ -129,58 +133,115 @@ private synchronized void init() {
...
+                    String url = suppliedKnownBundleMetadata.getUrl();
+                    if (BrooklynVersion.isDevelopmentEnvironment() && url.startsWith("system:file:")) {
+                        // in live dists the url is usually mvn: but in dev/test karaf will prefix it with system;
+                        // leave the url alone so we correctly dedupe when considering whether to update, but create a zip file
+                        // so that things work consistently in dev/test (in particular ClassLoaderUtilsTest passes).
+                        // pretty sure we have to do this, even if not replacing the osgi bundle, because we need to
+                        // get a handle on the zip file (although we could skip if not doing persistence - but that feels even worse than this!)
+                        try {
+                            url = Strings.removeFromStart(url, "system:");
+                            File zipTemp = new BundleMaker(ResourceUtils.create()).createJarFromClasspathDir(url);
```
geomacy: I take it that `createJarFromClasspathDir` works ok in this case, where the url will start with `file:`.

core/src/main/java/org/apache/brooklyn/core/mgmt/ha/OsgiArchiveInstaller.java
ahgittin: the assertion was more for a person reading it

core/src/main/java/org/apache/brooklyn/core/mgmt/ha/OsgiArchiveInstaller.java
ahgittin: yep, `ClassLoaderUtilsTest` depends on this and it passes.  pretty cool.

ahgittin:
all addressed -- thx @geomacy, merging

PR #843, opened 2017-09-22
limit reset/persistence to legacy items

ahgittin:
following the change in https://github.com/apache/brooklyn-server/pull/814,
legacy compatibility caused persistence of registered types from bundles, unnecessarily (and error-prone on rebind where it expects a real CatalogItemDto)


aledsage:
Note this fixes https://issues.apache.org/jira/browse/BROOKLYN-539

aledsage:
Thanks @ahgittin - confirmed this fixes it for me in Karaf. Merging now.

PR #849, opened 2017-10-02
Retrieve the right AWS metadata for private hostname

tbouron:
In most conditions for AWS deployments (machine sshable) Brooklyn was retrieving the `public-hostname` [metadata](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html) for both public **AND** private hostname sensors.

### DIFF
--- a/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
+++ b/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
@@ -2829,16 +2829,11 @@ public class JcloudsLocation extends AbstractCloudMachineProvisioningLocation im
             return getPublicHostname(node, sshHostAndPort, userCredentials, setup);
         }
 
-        String provider = (setup != null) ? setup.get(CLOUD_PROVIDER) : null;
-        Boolean lookupAwsHostname = (setup != null) ? setup.get(LOOKUP_AWS_HOSTNAME) : null;
-        if (provider == null) provider = getProvider();
-
-        // TODO Discouraged to do cloud-specific things; think of this code for aws as an
-        // exceptional situation rather than a pattern to follow. We need a better way to
-        // do cloud-specific things.
-        if ("aws-ec2".equals(provider) && Boolean.TRUE.equals(lookupAwsHostname)) {
-            Maybe<String> result = getHostnameAws(node, sshHostAndPort, userCredentials, setup);
-            if (result.isPresent()) return result.get();
+        // NodeMetadata.getHostname() is supposed to return the private hostname. If it's not null, we want to prioritise
+        // this, otherwise we call getPrivateHostnameGeneric()
+        final String hostname = node.getHostname();
+        if (hostname != null) {
+            return hostname;
         }
 
         Optional<String> preferredAddress = sshHostAndPort.isPresent() ? Optional.of(sshHostAndPort.get().getHostText()) : Optional.<String>absent();


locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -2957,7 +2957,7 @@ String getHostnameAws(HostAndPort hostAndPort, LoginCredentials userCredentials,
                     "get public AWS hostname",
                     ImmutableList.of(
                             BashCommands.INSTALL_CURL,
-                            "echo `curl --silent --retry 20 http://169.254.169.254/latest/meta-data/public-hostname`; exit"));
+                            "echo `curl --silent --retry 20 http://169.254.169.254/latest/meta-data/"+(privateHost ? "local-hostname" : "public-hostname")+"`; exit"));
```
andreaturli: thanks @tbouron, could we avoid the extra boolean parameter if we use something like
```
response=$(curl --write-out %{http_code} --silent --output /dev/null http://169.254.169.254/latest/meta-data/local-hostname)
if [ $(response) == "200" ] ; 
then
echo `curl --silent --retry 20 http://169.254.169.254/latest/meta-data/privateHost"`; exit"));
else
echo `curl --silent --retry 20 http://169.254.169.254/latest/meta-data/public-hostname"`; exit"));
```
or something similar?

tbouron: @andreaturli I don't think so. `http://169.254.169.254/latest/meta-data/local-hostname` 
 will always return some data and a 200 as AWS assign private IP/host for every single EC2 instance.

`http://169.254.169.254/latest/meta-data/public-hostname` could return a 404 (if the machine has been provisioned without an elastic IP, not entirely sure though) but in this situation, but that's not what we want

andreaturli: not sure I follow, @tbouron -- I think it would be easier to read if we have 2 methods
`getPrivateAwsHostname` and `getPublicAwsHostname` what do you think?

tbouron: @andreaturli In you code example, `http://169.254.169.254/latest/meta-data/local-hostname` will always return a `200` (as an EC2 instance always as a private IP) so your if statement will always go one way which is not what we want here.

It would may be easier to read (although I'm not sure about that) but will also introduce some code duplication which I would be reluctant do to, especially in this huge file.

andreaturli: sorry @tbouron, `String getHostnameAws(...)` is already very confusing, not mentioning the fact that is AWS specific which is another smell, IMHO.
Which hostname is going to return that method? 
I think we have the chance to improve this area of the code and I think another (boolean) parameter is not going to help in that respect. wdyt?

tbouron: If you are ok to add even more code + duplication then sure, no problem, I'll do that :D

Agree this is a code smell (which [has already been identified based on this comment](https://github.com/tbouron/brooklyn-server/blob/2455e054a62cbfff874c07c70f6cdf897f19769e/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java#L2836-L2838)) but I don't have a magical solution to improve this particular area. If you have any idea, I'm opened to suggestions

andreaturli: @tbouron if we can rely on jclouds for getting the private hostname `NodeMetadata.getHostname`, Brooklyn doesn’t need to calculate a new one especially for nodes created using JcloudsLocation. Notice `NodeMetadata.getHostname` is nullable so I think JcloudsLocation should manage that or assign the value to `host.addresses.private`

tbouron: @andreaturli I did what you suggested: using `NodeMetadata.getHostname` or reusing the `getPrivateHostnameGeneric()` method for private hostname. The public hostname is calculated as before.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -2829,16 +2829,11 @@ protected String getPrivateHostname(NodeMetadata node, Optional<HostAndPort> ssh
...
 
-        String provider = (setup != null) ? setup.get(CLOUD_PROVIDER) : null;
-        Boolean lookupAwsHostname = (setup != null) ? setup.get(LOOKUP_AWS_HOSTNAME) : null;
-        if (provider == null) provider = getProvider();
-
-        // TODO Discouraged to do cloud-specific things; think of this code for aws as an
-        // exceptional situation rather than a pattern to follow. We need a better way to
-        // do cloud-specific things.
-        if ("aws-ec2".equals(provider) && Boolean.TRUE.equals(lookupAwsHostname)) {
-            Maybe<String> result = getHostnameAws(node, sshHostAndPort, userCredentials, setup);
```
andreaturli: looks much easier @tbouron -- can we remove `getHostnameAws` or is it used elsewhere?

tbouron: @andreaturli It is unfortunately used by the `DefaultConnectivityResolver` class

tbouron:
@andreaturli Looks like it works as expected :)

screen shot 2017-10-03 at 11 17 06

andreaturli:
lgtm thanks @tbouron

PR #819, opened 2017-09-13
Improve config key descriptions

aledsage:
See individual commits for details.

Note the last two commits make significant changes to config:
* remove the default for `launch.command` of `VanillaSoftwareProcess`.
* Rename some config keys (deprecating the old names) for `latch.*` and `skip.*`.

PR #851, opened 2017-10-02
Update config keys' deprecated names

tbouron:
Since Brooklyn 0.12.0 ([this PR](https://github.com/apache/brooklyn-server/pull/819) precisely) some config keys' name have been deprecated generated a warning when installing a blueprint using those. 

This fixes it by using the new names. It also fixes a typo for `latch.install.resources`


tbouron:
retest this please

PR #852, opened 2017-10-04
Merge initial catalog and persisted catalog

aledsage:
When starting Brooklyn, the catalog is now populated by merging the initial catalog (from the default .bom file) with the persisted catalog.

This PR is pretty much a re-write of the `CatalogInitialization` class to do this.

---
It also cleans up some code, some code paths and some strange things we've previously supported. Of particular interest...

**No longer auto-populates catalog on `localManagementContext.getCatalog()`**
Previously `localManagementContext.getCatalog()` would cause the catalog to be populated if it had not already been done so. This was important for tests that did not go through the full intialization lifecycle. I've moved that into the class `LocalManagementContextForTests` - we should restrict the code-paths for production usage to just those we expect!

The consequence is that any tests relying on it must create the `LocalManagementContextForTests` rather than a full-blown `LocalManagementContext` (almost all Brooklyn tests do this already), but this might affect downstream projects.

**Fewer ways catalog initialization can happen**
We now only support 2 production ways of populating the catalog:
* `populateInitialCatalogOnly`: called if persistence is disabled
* `populateInitialAndPersistedCatalog`: called on rebind

The second way can be called multiple times (e.g. if repeatedly doing it for `hot-standby`, or if subsequently promoting to `master`).

The method `unofficialPopulateInitialCatalog` is to be only called by tests (though unfortunately not enforced as such).

**Tidy BasicLauncher's persistence initialization**

* `confirmCatalog` is moved from `Main` to `CatalogInitialization`
* `BasicLauncher.checkConfiguration` now checks that we never enable HA mode if persistence is disabled. Moves handling of non-persistence out of `initPersistence`.


core/src/test/java/org/apache/brooklyn/core/mgmt/rebind/RebindCatalogEntityTest.java
```diff
@@ -1,153 +0,0 @@
...
-import org.apache.brooklyn.test.support.TestResourceUnavailableException;
-import org.apache.brooklyn.util.core.javalang.UrlClassLoader;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-import org.testng.annotations.BeforeMethod;
-import org.testng.annotations.Test;
-
-import com.google.common.base.Function;
-
-public class RebindCatalogEntityTest extends RebindTestFixture<StartableApplication> {
```
aledsage: See description in this individual commit:
```
Was testing use of catalog class loader, not OSGi, with catalog item
requiring custom class loader being added programmatically via 
catalog.addItem() and then fiddling with catalog on rebind to add it 
back in!

No user should be doing that; the test is not something we want to
support.
```

launcher/src/test/java/org/apache/brooklyn/launcher/AbstractBrooklynLauncherRebindTest.java
```diff
@@ -0,0 +1,183 @@
...
+                Constants.BUNDLE_SYMBOLICNAME, bundleName.getSymbolicName(),
+                Constants.BUNDLE_VERSION, bundleName.getOsgiVersion().toString()));
+            tmpFiles.add(bf);
+        }
+        return bf;
+        
+//        ReferenceWithError<OsgiBundleInstallationResult> b = ((ManagementContextInternal)mgmt).getOsgiManager().get().install(
+//                new FileInputStream(bf) );
+//
+//            b.checkNoError();
```
tbouron: Will it be used in the future? If not, I would suggest to remove it

aledsage:
Thanks @tbouron - incorporated comments; once jenkins confirms it builds then I'll merge it.

PR #853, opened 2017-10-04
[WIP] deprecate Task as values in config and getImmediately(Task) in favour of TaskFactory

ahgittin:
As discussed in #816 and suggested to be fixed in #835 but pushed out to here:

There are bad things that can happen if an unsubmitted `Task` is set as a config value:  attempts to read it change the behaviour for subsequent reads.  In particular a call to `getImmediately` (which can be done during validation, or maybe even during a REST read) can cause it to be submitted in an interrupted state, which damages it for all future use.  But more generally, once it has been evaluated it will always and only return that value, where a user may sensibly expect it to return the current value on each read.

Changing those methods -- `config().set(key, task)` and `getImmediately(task)` -- to take `TaskFactory` instances instead removes this confusion and problems, making it clear that a new task is run on each execution.

However this is a big job due to the prevalence of `Task` being used where a `TaskFactory` is now wanted.  In particular `DependentConfiguration` returns `Task` instances!  (There is a long-standing comment that `DslComponent` should be used instead, as that generates factories internal, and YAML use behaves as proposed to implement here across the board, so it is mainly a concern internally and for people using the Java API.  Still it is a clean-up worth doing.)

I've committed one set of changes so far (it is just one commit beyond the merge of #835), signposting the direction this will go and logging warnings in the deprecated code paths.  Next I will actually deprecate these and update usages.

In particular I will likely deprecate all `Task`-returning `DependentConfig` with `TaskFactory` alternatives; then in a subsequent version we can switch the deprecated methods to return `TaskFactory`:  this means code such as `config().set(key, attributeWhenReady(...))` although it will be deprecated for one release will be migrated with no effort from a user to the new behaviour.

Comments at this early stage very welcome cc @aledsage @sjcorbett @grkvlt .

### DIFF
--- a/api/src/main/java/org/apache/brooklyn/api/objs/Configurable.java
+++ b/api/src/main/java/org/apache/brooklyn/api/objs/Configurable.java
@@ -20,10 +20,13 @@ package org.apache.brooklyn.api.objs;
 
 import java.util.Set;
 
+import org.apache.brooklyn.api.mgmt.ExecutionContext;
 import org.apache.brooklyn.api.mgmt.Task;
+import org.apache.brooklyn.api.mgmt.TaskAdaptable;
+import org.apache.brooklyn.api.mgmt.TaskFactory;
 import org.apache.brooklyn.config.ConfigKey;
-import org.apache.brooklyn.config.ConfigMap;
 import org.apache.brooklyn.config.ConfigKey.HasConfigKey;
+import org.apache.brooklyn.config.ConfigMap;
 
 import com.google.common.annotations.Beta;
 import com.google.common.base.Predicate;
@@ -85,19 +88,39 @@ public interface Configurable {
         <T> T set(HasConfigKey<T> key, T val);
         
         /**
-         * Sets the config to the value returned by the task.
+         * Sets the config to return the value returned by the task this generates,
+         * on demand when invoked.
+         * <p>
+         * Returns immediately without blocking; subsequent calls to {@link #getConfig(ConfigKey)} 
+         * will create and execute an instance of the task, blocking unless otherwise configured.
          * 
+         * @see {@link #setConfig(ConfigKey, Object)}
+         */
+        <T> T set(ConfigKey<T> key, TaskFactory<? extends TaskAdaptable<T>> val);
+        
+        /**
+         * Sets the config to the value returned by the task.
+         * <p>
+         * It is recommended to use {@link #set(ConfigKey, TaskFactory)} instead so that calls to validate 
+         * items via {@link ExecutionContext#getImmediately(Object)} do not interrupt and wreck the task set here. 
+         * <p>
          * Returns immediately without blocking; subsequent calls to {@link #getConfig(ConfigKey)} 
          * will execute the task, and block until the task completes.
          * 
          * @see {@link #setConfig(ConfigKey, Object)}
          */
-        <T> T set(ConfigKey<T> key, Task<T> val);
+        <T> T set(ConfigKey<T> key, TaskAdaptable<T> val);
+        
+        /**
+         * @see {@link #setConfig(ConfigKey, TaskFactory)}
+         */
+        <T> T set(HasConfigKey<T> key, TaskFactory<? extends TaskAdaptable<T>> val);
         
         /**
          * @see {@link #setConfig(ConfigKey, Task)}
          */
-        <T> T set(HasConfigKey<T> key, Task<T> val);
+        // TODO deprecate
+        <T> T set(HasConfigKey<T> key, TaskAdaptable<T> val);
         
         /** @deprecated since 0.11.0 see {@link ConfigMap#findKeys(Predicate)} */
         @Deprecated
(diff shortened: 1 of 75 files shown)


api/src/main/java/org/apache/brooklyn/api/objs/Configurable.java
```diff
@@ -85,19 +88,39 @@
...
+         * It is recommended to use {@link #set(ConfigKey, TaskFactory)} instead so that calls to validate 
+         * items via {@link ExecutionContext#getImmediately(Object)} do not interrupt and wreck the task set here. 
+         * <p>
          * Returns immediately without blocking; subsequent calls to {@link #getConfig(ConfigKey)} 
          * will execute the task, and block until the task completes.
          * 
          * @see {@link #setConfig(ConfigKey, Object)}
          */
-        <T> T set(ConfigKey<T> key, Task<T> val);
+        <T> T set(ConfigKey<T> key, TaskAdaptable<T> val);
```
aledsage: Add `TODO Deprecate` here as well.

Some power-users might implement `ConfigurationSupport`, which this will impact, but that's fine: it's marked `@Beta`.

core/src/main/java/org/apache/brooklyn/core/effector/EffectorTasks.java
```diff
@@ -57,9 +58,14 @@
     private static final Logger log = LoggerFactory.getLogger(EffectorTasks.class);
     
     public interface EffectorTaskFactory<T> {
+        // TODO deprecate
```
aledsage: How does this relate to config of type Task? I thought that `EffectorTaskFactory` would be used in an entirely different context.

core/src/main/java/org/apache/brooklyn/core/sensor/DependentConfiguration.java
```diff
@@ -102,8 +102,15 @@
...
+ * and should now normally be used instead. If using these methods that return {@link Task} instances
+ * you should either ensure it is submitted or that validation and any other attempts to 
+ * {@link ExecutionContext#getImmediately(TaskAdaptable)} on this task is blocked;
+ * it is all to easy otherwise for an innocuous immediate-get to render this task interrupted
  */
+// possibly even the REST API just looking at config could cancel?
+// TODO should we deprecate these and provide variants that return TaskFactory instances?
+// maybe even ones that are nicely persistable? i (alex) think so, 2017-10.
+// see https://github.com/apache/brooklyn-server/pull/816#issuecomment-333858098
 public class DependentConfiguration {
```
aledsage: In `DslComponent`, it calls to things like `DependentConfiguration.attributeWhenReady(targetEntity, (AttributeSensor<?>)targetSensor)` for its implementation of `newTask`.

I think we want to keep this code, but for people to not use it directly! I presume the way to do that (sticking to our deprecation policy) would be to create a new internal class (e.g. that delegates to it), and to deprecate this class.

I've always wondered about our `DslComponent` class. It feels like there's a big chunk of code in `brooklyn-camp` that could live in `core`. Things like the `DslComponent.AttributeWhenReady` class no longer has anything to do with the original yaml.  We could move that to core (preserving backwards compatibility with persisted state obviously).

I don't think there's a need for any other new code that returns `TaskFactory` instances - we should re-use the existing `DslComponent` code as much as possible.

core/src/main/java/org/apache/brooklyn/util/core/task/BasicExecutionContext.java
```diff
@@ -242,16 +244,40 @@ public synchronized T get(long timeout, TimeUnit unit) throws InterruptedExcepti
...
+    @Override
+    public <T> Maybe<T> getImmediately(TaskFactory<Task<T>> callableOrSupplier) {
+        return getImmediatelyInternal(callableOrSupplier);
+    }
+    
+    @Override
+    public <T> Maybe<T> getImmediately(TaskAdaptable<T> callableOrSupplier) {
+        if (!(callableOrSupplier instanceof TaskFactory)) {
+            Task<T> t = callableOrSupplier.asTask();
+            if (!t.isSubmitted()) {
```
aledsage: I'd like us to go further and say that all calls that pass in a task are deprecated. But this is a good first step.

I'd also want us to warn about deprecated usage if someone called `getImmediately((Object)task)`

core/src/main/java/org/apache/brooklyn/util/core/task/BasicExecutionContext.java
```diff
@@ -242,16 +244,40 @@ public synchronized T get(long timeout, TimeUnit unit) throws InterruptedExcepti
...
+     * currently supports {@link Supplier}, {@link Callable}, {@link Runnable}, 
+     * or {@link Task} (recommended to be {@link TaskFactory}) instances; 
      * with tasks if it is submitted or in progress,
      * it fails if not completed; with unsubmitted, unqueued tasks, it gets the {@link Callable} job and 
      * uses that; with such a job, or any other callable/supplier/runnable, it runs that
      * in an {@link InterruptingImmediateSupplier}, with as much metadata as possible (eg task name if
      * given a task) set <i>temporarily</i> in the current thread context */
-    @SuppressWarnings("unchecked")
     @Override
     public <T> Maybe<T> getImmediately(Object callableOrSupplier) {
```
aledsage: Would like to say that passing in a `Task` is deprecated.

nakomis:
@ahgittin Can you take a look at this please to see if it is still relevant, and address the comments above if appropriate

Thanks

PR #855, opened 2017-10-05
[BROOKLYN-542] add test for freemarker builtins hashes

andreaturli:


aledsage:
Retest this please

core/src/test/java/org/apache/brooklyn/util/core/text/TemplateProcessorTest.java
```diff
@@ -247,4 +251,40 @@ public void testDotSeparatedKeyCollisionFailure() {
...
+        entity.sensors().set(sensor1, ImmutableMap.of("name", "mouse", "price", 50));
+        String keys = Joiner.on("\n").join(
+                "<#list attribute['map']?keys as k>",
+                "${k}",
+                "</#list>");
+        String resultForKeys = TemplateProcessor.processTemplateContents(keys, entity, ImmutableMap.of());
+        assertEquals(resultForKeys.trim(), "name\nprice");
+
+        String values = Joiner.on("\n").join(
+                "<#list attribute['map']?keys as k>",
```
aledsage: I think this should say `?values...`

andreaturli:
yes @aledsage this fails because of that jira issue.

it's only the first commit to demonstrate the problem. I'll add a fix as soon as I have it. ( still trying to understand freemarker API)

andreaturli:
Close this for long inactivity - thanks @aledsage for your help

PR #856, opened 2017-10-05
[WIP] clean up startup logging around catalog

ahgittin:
just looking at logs and tidying up stuff that is wrong or misleading

have done one so far but i think there are others.  best hold off reviewing until all done.

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/catalog/internal/CatalogBundleLoader.java
+++ b/core/src/main/java/org/apache/brooklyn/core/catalog/internal/CatalogBundleLoader.java
@@ -103,8 +103,8 @@ public class CatalogBundleLoader {
                 this.managementContext.getCatalog().addTypesFromBundleBom(bomText, mb, force, result.mapOfNewToReplaced);
                 if (validate) {
                     Set<RegisteredType> matches = MutableSet.copyOf(this.managementContext.getTypeRegistry().getMatching(RegisteredTypePredicates.containingBundle(mb.getVersionedName())));
-                    if (!(matches.containsAll(result.mapOfNewToReplaced.keySet()) && result.mapOfNewToReplaced.keySet().containsAll(matches))) {
-                        // sanity check
+                    if (matches.size()!=result.mapOfNewToReplaced.size()) {
+                        // sanity check - should have same types, except result keys might not be fully resolved, so don't do set equality
                         LOG.warn("Discrepancy in list of Brooklyn items found for "+mb.getVersionedName()+": "+
                             "installer said "+result.mapOfNewToReplaced.keySet()+" ("+result.mapOfNewToReplaced.keySet().size()+") "+
                             "but registry search found "+matches+" ("+matches.size()+")");
@@ -128,10 +128,12 @@ public class CatalogBundleLoader {
     }
 
     /**
-     * Remove the given items from the catalog.
+     * Remove the given item from the catalog.
      *
-     * @param item Catalog items to remove
+     * @param item Catalog item to remove
+     * @deprecated since 0.13.0 remove bundles
      */
+    @Deprecated
     public void removeFromCatalog(CatalogItem<?, ?> item) {
         try {
             this.managementContext.getCatalog().deleteCatalogItem(item.getSymbolicName(), item.getVersion());
@@ -143,6 +145,13 @@ public class CatalogBundleLoader {
         }
     }
     
+    /**
+     * Remove the given item from the catalog.
+     *
+     * @param item Catalog item to remove
+     * @deprecated since 0.13.0 remove bundles
+     */
+    @Deprecated
     public void removeFromCatalog(VersionedName n) {
         ((ManagementContextInternal)managementContext).getOsgiManager().get().uninstallCatalogItemsFromBundle(n);
     }


core/src/main/java/org/apache/brooklyn/core/catalog/internal/CatalogBundleLoader.java
```diff
@@ -103,8 +103,8 @@ private TemporaryInternalScanResult scanForCatalogInternal(Bundle bundle, boolea
                 this.managementContext.getCatalog().addTypesFromBundleBom(bomText, mb, force, result.mapOfNewToReplaced);
                 if (validate) {
                     Set<RegisteredType> matches = MutableSet.copyOf(this.managementContext.getTypeRegistry().getMatching(RegisteredTypePredicates.containingBundle(mb.getVersionedName())));
-                    if (!(matches.containsAll(result.mapOfNewToReplaced.keySet()) && result.mapOfNewToReplaced.keySet().containsAll(matches))) {
-                        // sanity check
+                    if (matches.size()!=result.mapOfNewToReplaced.size()) {
```
tbouron: I'm afraid this will be too loose: the 2 sets have the same size but we don't know if the items are "the same", i.e. same `symbolicName` and `version`.

Could we have a middle ground here? Something like looping over each items in `result.mapOfNewToReplaced`, checking if `matches` contains one with the same `symbolicName` and `version`. Those fields should be resolved and would provide sufficient equality in this case. WDYT?

core/src/main/java/org/apache/brooklyn/core/catalog/internal/CatalogBundleLoader.java
```diff
@@ -128,10 +128,12 @@ private TemporaryInternalScanResult scanForCatalogInternal(Bundle bundle, boolea
...
 
     /**
-     * Remove the given items from the catalog.
+     * Remove the given item from the catalog.
      *
-     * @param item Catalog items to remove
+     * @param item Catalog item to remove
+     * @deprecated since 0.13.0 remove bundles
      */
+    @Deprecated
```
tbouron: Does not seem in the scope of this PR. Also not sure why deprecating it?

core/src/main/java/org/apache/brooklyn/core/catalog/internal/CatalogBundleLoader.java
```diff
@@ -143,6 +145,13 @@ public void removeFromCatalog(CatalogItem<?, ?> item) {
         }
     }
     
+    /**
+     * Remove the given item from the catalog.
+     *
+     * @param item Catalog item to remove
+     * @deprecated since 0.13.0 remove bundles
+     */
+    @Deprecated
```
tbouron: Same as above

core/src/main/java/org/apache/brooklyn/core/catalog/internal/CatalogBundleLoader.java
```diff
@@ -128,10 +128,12 @@ private TemporaryInternalScanResult scanForCatalogInternal(Bundle bundle, boolea
     }
 
     /**
-     * Remove the given items from the catalog.
+     * Remove the given item from the catalog.
      *
-     * @param item Catalog items to remove
+     * @param item Catalog item to remove
+     * @deprecated since 0.13.0 remove bundles
```
tbouron: Should be update to `1.0.0`

core/src/main/java/org/apache/brooklyn/core/catalog/internal/CatalogBundleLoader.java
```diff
@@ -143,6 +145,13 @@ public void removeFromCatalog(CatalogItem<?, ?> item) {
         }
     }
     
+    /**
+     * Remove the given item from the catalog.
+     *
+     * @param item Catalog item to remove
+     * @deprecated since 0.13.0 remove bundles
```
tbouron: Should be update to `1.0.0`

core/src/main/java/org/apache/brooklyn/core/catalog/internal/CatalogBundleLoader.java
aledsage: Looks like code is not called, and it's in an internal package. Therefore I'm ahppy with it being deprecated (or even deleted immediately). Same for below.

core/src/main/java/org/apache/brooklyn/core/catalog/internal/CatalogBundleLoader.java
aledsage: @ahgittin can you explain more the purpose/context of this check, and why you only want to log.warn if the sizes are different?

I recall seeing such warnings in the log when I suspect we shouldn't have, so think you're right to guard against it better.

For @tbouron's suggestion of checking the symbolicName:version, that makes sense - or does this get called a lot or with really big things so performance matters?

tbouron:
@ahgittin Do you think you can look at the comments above?

tbouron:
@ahgittin Any news on this PR?

nakomis:
@ahgittin Can you take a look at this please to see if it is still relevant, and address the comments above if appropriate

Thanks

PR #857, opened 2017-10-09
Deprecate `@Catalog` annotation

aledsage:
As discussed on dev@brooklyn mailing list (subject "Deprecating @Catalog annotation?")


core/src/main/java/org/apache/brooklyn/core/catalog/internal/BasicBrooklynCatalog.java
```diff
@@ -671,6 +671,8 @@ private void collectCatalogItemsFromItemMetadataBlock(String sourceYaml, Managed
         if (scanJavaAnnotations==null || !scanJavaAnnotations) {
             // don't scan
         } else {
+            log.warn("Deprecated use of scanJavaAnnotations" + (containingBundle != null ? " in bundle " + containingBundle.getVersionedName() : ""));
```
tbouron: Shouldn't we say `Use catalog.bom instead` here?

PR #641, opened 2017-04-19
[WIP] Fix sensor changes event ordering 

neykov:
Makes sure that events are fired (and thus handled) in the order the attribute is set. Without syncronyzing the last event fired might not match the last value of the attribute breaking the logic in the ComputeServiceIndicatorsFromChildrenAndMembers enricher.

Possibly related:
* https://issues.apache.org/jira/browse/BROOKLYN-518
* https://issues.apache.org/jira/browse/BROOKLYN-243

PR #860, opened 2017-10-11
improve locking for sensor values

ahgittin:
fixes https://issues.apache.org/jira/browse/BROOKLYN-544

introducing a public Lock for writes and callers, and the weaker uglier monitor for very short internal synchronized blocks.
also adds a Locks convenience class for running code with a lock.


core/src/main/java/org/apache/brooklyn/core/sensor/AttributeMap.java
```diff
@@ -115,8 +138,12 @@ public Object getSynchObjectInternal() {
      * @param newValue the new value
      * @return the old value.
      * @throws IllegalArgumentException if path is null or empty
+     * 
+     * @deprecated since 0.13.0 becoming private; callers should only ever use {@link #update(AttributeSensor, Object)}
```
aledsage: Should be `since 1.0.0`

core/src/main/java/org/apache/brooklyn/core/sensor/AttributeMap.java
```diff
@@ -138,62 +165,91 @@ private void checkPath(Collection<String> path) {
...
+        // ands gets here won't call any other code at all
+        return withLock(() -> {
+            T oldValue = updateInternalWithoutLockOrPublish(attribute, newValue);
             entity.emitInternal(attribute, newValue);
             return oldValue;
-        }
+        });
     }
     
+    /** @deprecated since 0.13.0 this is becoming an internal method, {@link #updateInternalWithoutLockOrPublish(AttributeSensor, Object)} */
```
aledsage: Should be `since 1.0.0`

core/src/main/java/org/apache/brooklyn/core/sensor/AttributeMap.java
```diff
@@ -138,62 +165,91 @@ private void checkPath(Collection<String> path) {
...
     public void remove(AttributeSensor<?> attribute) {
         BrooklynLogging.log(log, BrooklynLogging.levelDebugOrTraceIfReadOnly(entity),
             "removing attribute {} on {}", attribute.getName(), entity);
-
-        remove(attribute.getNameParts());
+        withLock(() -> remove(attribute.getNameParts()) );
     }
 
     // TODO path must be ordered(and legal to contain duplicates like "a.b.a"; list would be better
+    /** @deprecated since 0.13.0 becoming private; callers should only ever use {@link #remove(AttributeSensor)}
```
aledsage: `since 1.0.0`

core/src/main/java/org/apache/brooklyn/core/sensor/AttributeMap.java
```diff
@@ -212,9 +268,6 @@ public void remove(Collection<String> path) {
      * @throws IllegalArgumentException path is null or empty.
      */
     public Object getValue(Collection<String> path) {
```
aledsage: Why not deprecate this as well?

ahgittin:
agree with all comments.  updated and merging.

neykov:
Seems this fixes [BROOKLYN-525](https://issues.apache.org/jira/browse/BROOKLYN-525) as well.
Worth checking by enabling the corresponding tests as done in https://github.com/apache/brooklyn-server/pull/641.

PR #862, opened 2017-10-13
Fix npr when reading streamed bundle

duncangrant:
This happens when there is an existing unmanaged bundle with the same id
and a replacement bundle is uploaded without the force flag

Tested manually using boms and jars (yaml only and with java)
using curl and br tool.


aledsage:
LGTM.

For the record, we hit a NullPointerException when there was already a bundle added to karaf (via the initial feature.xml), which we then tried to add to the catalog via `br catalog add file.jar`.

tbouron:
retest this please

PR #863, opened 2017-10-17
Fix ManagementNodeStateListener

aledsage:
Wait for rebind to complete before notifying of ‘MASTER’. Otherwise if the listener does things like listing the apps, it might get back an empty list.


duncangrant:
This looks reasonable to me and improves testing.  I think we should consider how much of this should be covered in the user docs as we make it all seem quite simple here https://brooklyn.apache.org/v/latest/ops/high-availability/index.html but I think some of this behaviour could surprise people.

PR #864, opened 2017-10-17
Separate dumper class

aledsage:
Move all `Entities.dumpInfo` to `Dumper`, deprecating the methods in `Entities`.

Previously, `Entities` was too much of a dumping ground for lots of methods related to entities, and has become huge! We've previously moved some stuff from it, such as `sanitize` to `Sanitizer`. This PR continues that trend.

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/entity/Dumper.java
+++ b/core/src/main/java/org/apache/brooklyn/core/entity/Dumper.java
@@ -1,0 +1,376 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.brooklyn.core.entity;
+
+import static org.apache.brooklyn.core.entity.Entities.isSecret;
+import static org.apache.brooklyn.core.entity.Entities.isTrivial;
+
+import java.io.IOException;
+import java.io.PrintWriter;
+import java.io.Writer;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.ExecutionException;
+
+import org.apache.brooklyn.api.entity.Entity;
+import org.apache.brooklyn.api.entity.Group;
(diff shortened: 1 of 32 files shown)


core/src/main/java/org/apache/brooklyn/core/entity/Dumper.java
```diff
@@ -0,0 +1,379 @@
...
+import org.apache.brooklyn.util.core.flags.FlagUtils;
+import org.apache.brooklyn.util.guava.Maybe;
+import org.apache.brooklyn.util.text.Strings;
+
+import com.google.common.collect.Lists;
+import com.google.common.collect.Maps;
+import com.google.common.collect.Sets;
+
+/**
+ * Convenience methods for dumping info about entities etc.
```
tbouron: This class has method to dump info not only for entities, but locations, enrichers, etc. Maybe it would be better to say:
`Convenience methods for dumping info about BrooklynObject etc.` ?

core/src/main/java/org/apache/brooklyn/core/entity/Entities.java
```diff
@@ -302,305 +292,190 @@ public static boolean isTrivial(Object v) {
...
-        List result = new ArrayList(sensors);
-        Collections.sort(result, new Comparator<Sensor>() {
-                    @Override
-                    public int compare(Sensor arg0, Sensor arg1) {
-                        return arg0.getName().compareTo(arg1.getName());
-                    }
-
-        });
-        return result;
+        return Dumper.sortSensors(sensors);
```
tbouron: Should this be in the `Dumper` class? It does not seem to have a relation to dumping data

core/src/main/java/org/apache/brooklyn/core/entity/Entities.java
```diff
@@ -302,305 +292,190 @@ public static boolean isTrivial(Object v) {
...
-        List result = new ArrayList(configs);
-        Collections.sort(result, new Comparator<ConfigKey>() {
-                    @Override
-                    public int compare(ConfigKey arg0, ConfigKey arg1) {
-                        return arg0.getName().compareTo(arg1.getName());
-                    }
-
-        });
-        return result;
+        return Dumper.sortConfigKeys(configs);
```
tbouron: Same as above

core/src/main/java/org/apache/brooklyn/core/entity/Entities.java
```diff
@@ -302,305 +292,190 @@ public static boolean isTrivial(Object v) {
...
     public static <T> Map<String, T> sortMap(Map<String, T> map) {
-        Map<String,T> result = Maps.newLinkedHashMap();
-        List<String> order = Lists.newArrayList(map.keySet());
-        Collections.sort(order, String.CASE_INSENSITIVE_ORDER);
-
-        for (String key : order) {
-            result.put(key, map.get(key));
-        }
-        return result;
+        return Dumper.sortMap(map);
```
tbouron: Same as above

core/src/main/java/org/apache/brooklyn/core/entity/Entities.java
aledsage: It's now a package-private method of `Dumper`, used to sort the sensors when we're going to dump out the entity's details. It therefore needs to be in (or accessible from) the `Dumper` class. I don't think it should ever have been a public method (i.e. part of our sprawling api).

This is a way to deprecate it from the public api.

tbouron: Fair enough

PR #867, opened 2017-10-23
BROOKLYN-546: fix rebind of system bundles

aledsage:


core/src/main/java/org/apache/brooklyn/core/mgmt/ha/OsgiArchiveInstaller.java
```diff
@@ -379,8 +379,15 @@ private synchronized void close() {
                 if (canUpdate()) { 
                     result.bundle = osgiManager.framework.getBundleContext().getBundle(result.getMetadata().getOsgiUniqueUrl());
                     if (result.getBundle()==null) {
-                        log.warn("Brooklyn thought is was already managing bundle "+result.getMetadata().getVersionedName()+" but it's not installed to framework; reinstalling it");
-                        updating = false;
+                        if (hasBundleOnInstall(osgiManager, inferredMetadata, zipFile)) {
+                            // e.g. happens if system-bundle is brought under management, and then try to add it again:
+                            // if added from "initial catalog" has it, and then added from persisted state as well.
+                            result.setIgnoringAlreadyInstalled();
+                            return ReferenceWithError.newInstanceWithoutError(result);
```
ahgittin: This return path only takes effect if passing the `canUpdate()` check on line 379 but presumably behaviour should be the same as for a non-snapshot version?

Is the checksum check we do in the other block (line 397 in new money) is equivalent or if we need a super-set check combining that line with `hasBundleOnInstall()`.

Probably those checks and the two `ignoringAlreadyInstalled` returns should be moved to before line 379.

I'd also rename the new `hasBundleOnInstall` method to `isEquivalentBundleAlreadyOsgiInstalled` or something else a little more descriptive.

aledsage:
@ahgittin Still requires some more testing, but wanted to share this now for feedback.

----
As discussed with @ahgittin, there was an additional rebind issue when loading the initial catalog (and subsequently loading persisted bundles). Karaf will have cached OSGi bundles for that initial catalog. The initial catalog would duplicate some of these - it couldn't tell that the karaf bundle installed on the previous brooklyn run was identical to the bundle it wanted to install (it can't yet see the persisted brooklyn-managed records, and it can't understand osgi location `brooklyn:ID`).

Possible solutions for that included:
1) make `brooklyn:ID` osgi location be a checksum (so the same osgi location would be used for two identical brooklyn-managed bundles).
2) clear the karaf cache of brooklyn-managed bundles (e.g. all those with location in the form `brooklyn:ID`), in the knowledge that the initial catalog and persisted state will reinstall them.
3) load initial catalog _after_ persistence (but we expect the initial catalog to tell us whether to remove some items from persistence so this isn't a good option without major re-work).
4) if we're not certain that it's identical, then uninstall and reinstall the bundle (which is what happens currently). This might be a problem if it's a core brooklyn module, but those should have `mvn:` urls so it _can_ confirm whether it's a duplicate).
5) rework rebind / initial catalog load so that we load placeholders for bundles that have been persisted (enough to do the deduping or install things referenced in initial catalog load).

We opted for doing (1) and (4).

---
Another issue (FAO @ahgittin) is that every time Brooklyn was started, we'd add to the persisted state another copy of each core brooklyn bundle!

This is because the initial catalog references things like `org.apache.brooklyn.policy`. The `ManagedBundle` we create to represent it gets an auto-generated unique id. We see that the bundle is already installed, so we don't re-install it in karaf (good). But we do still persist it to the `/bundles` persisted state sub-directory.

I've worked around that by black-listing `org.apache.brooklyn.*` from being persisted (see `OsgiArchiveInstaller.isBlacklistedForPersistence(ManagedBundle)`. The assumption is that those will always ship with Brooklyn - we should use just the version that is shipped with the given release of Brooklyn.

---
Another issue (similar, not yet fixed) is that we persist a new identical copy of `brooklyn-default-karaf-catalog` every time we restart Brooklyn. This is because the `ManagedBundle` for the initial catalog has an auto-generated unique id. As above, we do not install the duplicate bundle, but we still persist it.

Luckily rebind is not as bad as you might think! It spots that an identical bundle is already installed, so doesn't install it or parse the catalog.bom of every copy that is in persisted state.

Possible solutions for this include:
1) changing the persistence file/object names to use the `symbolicName_version.jar` format (similar naming to the old `/catalog` directory).
2) only persist the bundle if we actually installed it (i.e. it wasn't a duplicate).
3) search persisted state for a duplicate, and don't add if it already exists.

Option (1) is very tempting, given that is it's logical unique identity: we shouldn't be persisting (and thus installing) multiple things with the same name:version. We'd need to be careful with backwards compatibility if someone has `/bundles` that uses the unique ids. We could later add additional checks to only replace the persisted file if its checksum had actually changed.

Option (2) feels wrong - what is persisted depends on what is in your karaf cache! For example, if you change the persisted state directory then we wouldn't write out the initial catalog to it because karaf would have cached those bundles from the previous run.

ahgittin:
Re the last problem, catalog being persisted, I think I'm good with option (1).  I was leery of using something that might have collisions (`name_version`, as opposed to the unique ID) but I've grown more comfortable that this isn't an issue in this case.  I don't see backwards compatibility being too much of a problem, apart from the obvious.

I'm not convinced by your argument against 2.  We could say to delete from persistence something if it is a duplicate of a different ID _brooklyn_-managed bundle, then being simply a karaf cache jar wouldn't be enough to block persistence.  The slight mess is still that the persisted ID for the default catalog changes on every restart but could live with this if it's easier/safer than (1).

aledsage:
retest this please

core/src/main/java/org/apache/brooklyn/core/typereg/BasicManagedBundle.java
```diff
@@ -135,6 +145,9 @@ public boolean equals(Object obj) {
             // this makes equality with other OsgiBundleWithUrl items symmetric,
             // but for two MB's we look additionally at checksum
             if (!Objects.equal(checksum, ((ManagedBundle)other).getChecksum())) return false;
+            
+            // only equal if have the same ManagedBundle uid; important for persistence.changeListener().unmanage()
+            if (!Objects.equal(getId(), ((ManagedBundle)other).getId())) return false;
```
aledsage: I don't like this: it breaks transitivity for the equals method: if `x.equals(y) && `y.equals(z)` then `x.equals(z)` should be true. It fails if `x` and `z` are `BasicManagedBundle` with different checksums or ids, but `y` is a `OsgiBundleWithUrl` with the same name:version and url.

However, that was already broken because of the checksum comparison (if an `OsgiBundleWithUrl` impl behaves in the way indicated in the pre-existing comments).

Longer term, I lean towards not trying to have equality with `OsgiBundleWithUrl`.

aledsage:
@ahgittin can you take another look please: gone for option (2) as discussed: only persist the bundle if we actually installed it (so delete the bundle from persisted state if it is a duplicate of another brooklyn-managed-bundle.

This was harder than I expected: promotion of an HA hot-standby behaved differently. See unit test and comments for more info.

aledsage:
Test failure is real: `org.apache.brooklyn.core.mgmt.ha.HotStandbyTest.testChangeMode`.
I've begun to look into that, and will fix it in the morning.

core/src/main/java/org/apache/brooklyn/core/typereg/BasicManagedBundle.java
ahgittin: agree, should have explicit methods to check or compare the `VersionedName`s where that is the item of interest

ahgittin:
@aledsage conflicts with #866 ^ :(

#WhyBigPullReqsArentSoBadWhenCodeIsIntertwined  (this is causing pain for the reviewer as well as the author)

ahgittin:
interesting subtlety on ha.  nice that black/white list is configurable.

minor comment re possibility of unmanaging new bundle rather than old.

otherwise good, once conflicts resolved and test fixed.

ahgittin:
retest this please

aledsage:
retest this please

failure was the unrelated non-deterministic test `org.apache.brooklyn.core.entity.lifecycle.ServiceStateLogicTest.testStopsNicelyToo`

aledsage:
retest this please

Excellent, jenkins build successful @ahgittin - I'm getting it to build again to make sure the new tests are reasonably reliable.

ahgittin:
good stuff, merging

PR #868, opened 2017-10-24
allow types from different bundles if equivalent

ahgittin:
previously this was only allowed if older bundle was a wrapper, or if it was forced;
this could cause problems if such a disallowed addition was forced and then you tried to rebind!

now it stores the bundle from which a registered type comes, so we can sensibly have several types.
we could even allow them to be different - which would of course cause chaos for users,
but if we allow multi-tenant catalogs it will be essential (we simply have to filter for visible
types when looking at the registry, and prevent bundle name collisions - eg by prefixing with tenant)
though for now we don't allow them to be different.

needs testing and test cases still

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/catalog/internal/BasicBrooklynCatalog.java
+++ b/core/src/main/java/org/apache/brooklyn/core/catalog/internal/BasicBrooklynCatalog.java
@@ -950,8 +950,13 @@ public class BasicBrooklynCatalog implements BrooklynCatalog {
             }
             
             if (version==null) {
-                // use this as default version when nothing specified
-                version = BasicBrooklynCatalog.NO_VERSION;
+                if (containingBundle!=null) {
+                    version = containingBundle.getVersionedName().getVersionString();
+                }
+                if (version==null) {
+                    // use this as default version when nothing specified or inferrable from containing bundle
+                    version = BasicBrooklynCatalog.NO_VERSION;
+                }
             }
             
             if (sourcePlanYaml==null) {
--- a/core/src/main/java/org/apache/brooklyn/core/catalog/internal/BasicBrooklynCatalog.java
+++ b/core/src/main/java/org/apache/brooklyn/core/catalog/internal/BasicBrooklynCatalog.java
@@ -1440,7 +1440,7 @@ public class BasicBrooklynCatalog implements BrooklynCatalog {
         Maybe<OsgiManager> osgiManager = ((ManagementContextInternal)mgmt).getOsgiManager();
         if (osgiManager.isPresent() && AUTO_WRAP_CATALOG_YAML_AS_BUNDLE) {
             // wrap in a bundle to be managed; need to get bundle and version from yaml
-            OsgiBundleInstallationResult result = addItemsOsgi(yaml, forceUpdate, osgiManager);
+            OsgiBundleInstallationResult result = addItemsOsgi(yaml, forceUpdate, osgiManager.get());
             return toLegacyCatalogItems(result.getTypesInstalled());
 
             // if all items pertaining to an older anonymous catalog.bom bundle have been overridden
@@ -1457,7 +1457,7 @@ public class BasicBrooklynCatalog implements BrooklynCatalog {
         Maybe<OsgiManager> osgiManager = ((ManagementContextInternal)mgmt).getOsgiManager();
         if (osgiManager.isPresent() && AUTO_WRAP_CATALOG_YAML_AS_BUNDLE) {
             // wrap in a bundle to be managed; need to get bundle and version from yaml
-            return addItemsOsgi(yaml, forceUpdate, osgiManager);
+            return addItemsOsgi(yaml, forceUpdate, osgiManager.get());
 
             // if all items pertaining to an older anonymous catalog.bom bundle have been overridden
             // we delete those later; see list of wrapper bundles kept in OsgiManager
@@ -1472,7 +1472,7 @@ public class BasicBrooklynCatalog implements BrooklynCatalog {
         return result;
     }
 
-    protected OsgiBundleInstallationResult addItemsOsgi(String yaml, boolean forceUpdate, Maybe<OsgiManager> osgiManager) {
+    protected OsgiBundleInstallationResult addItemsOsgi(String yaml, boolean forceUpdate, OsgiManager osgiManager) {
         Map<?, ?> cm = BasicBrooklynCatalog.getCatalogMetadata(yaml);
 
         if(cm == null) {
@@ -1503,7 +1503,7 @@ public class BasicBrooklynCatalog implements BrooklynCatalog {
 
         OsgiBundleInstallationResult result = null;
         try {
-            result = osgiManager.get().install(new BasicManagedBundle(vn.getSymbolicName(), vn.getVersionString(), null), new FileInputStream(bf), true, true, forceUpdate).get();
+            result = osgiManager.install(new BasicManagedBundle(vn.getSymbolicName(), vn.getVersionString(), null), new FileInputStream(bf), true, true, forceUpdate).get();
         } catch (FileNotFoundException e) {
             throw Exceptions.propagate(e);
         } finally {
--- a/core/src/main/java/org/apache/brooklyn/core/catalog/internal/BasicBrooklynCatalog.java
+++ b/core/src/main/java/org/apache/brooklyn/core/catalog/internal/BasicBrooklynCatalog.java
@@ -62,6 +62,7 @@ import org.apache.brooklyn.core.mgmt.ha.OsgiManager;
 import org.apache.brooklyn.core.mgmt.internal.CampYamlParser;
 import org.apache.brooklyn.core.mgmt.internal.ManagementContextInternal;
 import org.apache.brooklyn.core.typereg.BasicBrooklynTypeRegistry;
+import org.apache.brooklyn.core.typereg.BasicManagedBundle;
 import org.apache.brooklyn.core.typereg.BasicRegisteredType;
 import org.apache.brooklyn.core.typereg.BasicTypeImplementationPlan;
 import org.apache.brooklyn.core.typereg.BrooklynTypePlanTransformer;
@@ -1502,7 +1503,7 @@ public class BasicBrooklynCatalog implements BrooklynCatalog {
 
         OsgiBundleInstallationResult result = null;
         try {
-            result = osgiManager.get().install(null, new FileInputStream(bf), true, true, forceUpdate).get();
+            result = osgiManager.get().install(new BasicManagedBundle(vn.getSymbolicName(), vn.getVersionString(), null), new FileInputStream(bf), true, true, forceUpdate).get();
         } catch (FileNotFoundException e) {
             throw Exceptions.propagate(e);
         } finally {
--- a/core/src/main/java/org/apache/brooklyn/core/catalog/internal/BasicBrooklynCatalog.java
+++ b/core/src/main/java/org/apache/brooklyn/core/catalog/internal/BasicBrooklynCatalog.java
@@ -1041,7 +1041,12 @@ public class BasicBrooklynCatalog implements BrooklynCatalog {
             // (originally seen during a race where the empty-remover ran while we were installing)
             throw new IllegalStateException("Loading from a bundle which is not installed");
         }
-        String wrapped = bb.get().getHeaders().get(BROOKLYN_WRAPPED_BOM_BUNDLE);
+        return isWrapperBundle(bb.get());
+    }
+    
+    @Beta
+    public static boolean isWrapperBundle(Bundle b) {
+        String wrapped = b.getHeaders().get(BROOKLYN_WRAPPED_BOM_BUNDLE);
         return wrapped!=null && wrapped.equalsIgnoreCase("true");
     }
 
--- a/core/src/main/java/org/apache/brooklyn/core/typereg/BasicBrooklynTypeRegistry.java
+++ b/core/src/main/java/org/apache/brooklyn/core/typereg/BasicBrooklynTypeRegistry.java
@@ -41,6 +41,8 @@ import org.apache.brooklyn.api.typereg.RegisteredTypeLoadingContext;
 import org.apache.brooklyn.core.catalog.internal.BasicBrooklynCatalog;
 import org.apache.brooklyn.core.catalog.internal.CatalogItemBuilder;
 import org.apache.brooklyn.core.catalog.internal.CatalogUtils;
+import org.apache.brooklyn.core.mgmt.ha.OsgiManager;
+import org.apache.brooklyn.core.mgmt.internal.ManagementContextInternal;
 import org.apache.brooklyn.test.Asserts;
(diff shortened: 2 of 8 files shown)


ahgittin:
there are failing tests this needs to revisit

```
  CatalogOsgiVersionMoreEntityTest.testMoreEntityV1ThenV2GivesV2:181->AbstractYamlTest.addCatalogItems:248 » IllegalState
  CatalogOsgiVersionMoreEntityTest.testMoreEntityV2ThenV1GivesV1:162->AbstractYamlTest.addCatalogItems:248 » IllegalState
  CatalogOsgiYamlVersioningTest.testAddSameVersionWithoutBundleWorksIfItemIsSnapshot:62->assertJustOneBundle:97 Expected one bundle after installing the same; but had: [BasicManagedBundle{symbolicName=brooklyn-catalog-bom-lh1iWaI0, version=0.0.0.SNAPSHOT, url=null}, BasicManagedBundle{symbolicName=brooklyn-catalog-bom-XL3tI4t4, version=0.0.0.SNAPSHOT, url=null}] expected [true] but found [false]
  CatalogYamlEntityOsgiTypeRegistryTest>CatalogYamlEntityTest.testReplacementFailureLeavesPreviousItemFromAnonymousBundleIntact:685->doTestReplacementFailureLeavesPreviousIntact:74->CatalogYamlEntityTest.doTestReplacementFailureLeavesPreviousIntact:725 Failed containsLiteral(DeliberatelyMissing): Bundle BasicManagedBundle{symbolicName=brooklyn-catalog-bom-tfLpTgDV, version=0.0.0.SNAPSHOT, url=null} failed installation: Error installing catalog items: Cannot add BasicRegisteredType[my.catalog.app.id.load:0.1.2-SNAPSHOT;brooklyn-catalog-bom-tfLpTgDV:0.0.0.SNAPSHOT] to catalog; it is different to BasicRegisteredType[my.catalog.app.id.load:0.1.2-SNAPSHOT;brooklyn-catalog-bom-hWIMdfBx:0.0.0.SNAPSHOT], and my.catalog.app.id.load:0.1.2-SNAPSHOT is defined in different bundle (throwing)
  CatalogYamlRebindTest.testDeleteEmptyBundleRemovedFromPersistence:423 Collection has wrong size 4 (expected 2): [hcb07jab3e, hcb07jab3e.jar, jtom2vc1rq, jtom2vc1rq.jar]
```

core/src/main/java/org/apache/brooklyn/core/typereg/BasicBrooklynTypeRegistry.java
```diff
@@ -368,12 +370,51 @@ public void addToLocalUnpersistedTypeRegistry(RegisteredType type, boolean canFo
...
         
-        RegisteredType oldType = mgmt.getTypeRegistry().get(type.getId());
-        if (oldType==null || canForce || BrooklynVersionSyntax.isSnapshot(oldType.getVersion())) {
-            log.debug("Inserting "+type+" into "+this);
-            localRegisteredTypes.put(type.getId(), type);
-        } else {
-            assertSameEnoughToAllowReplacing(oldType, type);
+        Map<String, RegisteredType> knownMatchingTypesByBundles = localRegisteredTypesAndContainingBundles.get(type.getId());
+        if (knownMatchingTypesByBundles!=null && !knownMatchingTypesByBundles.isEmpty()) {
+            for (RegisteredType existingT: knownMatchingTypesByBundles.values()) {
```
aledsage: Risks `ConcurrentModificationException`. Elsewhere it synchronizes on the (non-concurrent) `Map<String, RegisteredType` to do `put`, but here we iterate over it without synchronizing. Perhaps take a copy while synchronized on it, and then iterate over that copy? Or probably safer to just synchronize on the map for all of this block.

core/src/main/java/org/apache/brooklyn/core/typereg/BasicBrooklynTypeRegistry.java
```diff
@@ -498,10 +524,34 @@ public void delete(VersionedName type) {
...
+                    }
+                }
+                if (removedItem==null) {
+                    throw new NoSuchElementException("Requested to delete "+type+" from "+type.getContainingBundle()+", "
+                        + "but that type was not known in that bundle, it is in "+m.keySet()+" instead");
+                }
+            }
+        } else {
+            // try legacy delete
+            delete(type.getVersionedName());
```
aledsage: Feels dangerous (but a small race): another thread might be adding a `RegisteredType` (with same id, from another bundle) `localRegisteredTypesAndContainingBundles`. The `delete(VersionedName)` will then delete the thing added by the other thread.

Can we instead separate out the legacy deletion, so it doesn't try to also again delete from `localRegisteredTypesAndContainingBundles`?

ahgittin:
what do you mean by forbidding "the force-delete of part of a bundle" ?  You mean "delete type X" ?  it's only used by legacy rest call (which won't work very well, deleted item comes back after rebind), and by calling code iterating over a bundle to delete it, which is messy as it's unsynched, but good enough, and can be cleaned up later IMO (i considered adding a "delete bundle" method here but thought it out of scope).

--

i didn't worry about CME's but you're right, they'll be no fun if they happen IRL

ahgittin:
there's a few more subtle fixes in progress, but one i want to record here.  consider:

1) i upload a BOM containing `type1` and `type2` without a bundle name, say it becomes `anonymous-brooklyn-wrapped-bom-bundle-11111:0.0.0-SNAPSHOT`

2) i update this and re-upload, with new definitions for the same types, say now the bundle gets given a name `anonymous-brooklyn-wrapped-bom-bundle-22222:0.0.0-SNAPSHOT`

in this case we _don't_ want to have multiple instances, and the expectation is that the previous types are replaced by this one, then `anonymous-bundle-11111` is noted as being empty (and it was noted as a wrapper bundle) so it is removed.  so we should allow replacement of types even from different bundles _if_ the old one is a wrapper bundle.

but say now we

3) put `type1` in a full-fledged bundle call it `bund` (not a wrapper) and install that, with again a different plan body

if this is allowed to replace `type1` then we get into a potentially dangerous state on rebind, that if `bund` is read first and then `anonymous-bundle-22222`, the latter bundle will fail because it can't install an incompatible `type1`.  we could try to load anonymous items first but not sure we can guarantee this (eg if `type1` depends on something that is loaded later).  but if this _isn't_ allowed to replace `type1` then we make it hard to move away from anonymous bundles.

if `bund` includes both `type1` and `type2`, which it normally would, then there is no problem because the anonymous bundles will be deleted when we did (3), but it's quite hard to check if `bund` redeclares everything or just a subset from anonymous bundles.

what i think the resolution should be is that _only_ anonymous wrapper bundles can give different definitions for types from other anonymous wrapper bundles.  this has two consequences:

(A) if you deploy anonymous BOMs which contain conflicting definitions of types and some of them don't contain all the types then you may see different definitions after rebind

(B) if you want to transition from a wrapped bundle to a full bundle, you must keep the types the same and change in a subsequent deployment (or bump the type's version which is probably wise anyway)

consqeuence (B) is fine i think.  consequence (A) might be irritating but it doesn't break rebind and we discourage anonymous bundles so I think we can live with it.

@aledsage ?

ahgittin:
above notes are implemented, and tests fixed, and concurrency issues resolved

given that we had several tests fail i'm comfortable with the level of testing this gets

ready for real review @aledsage

aledsage:
retest this please

aledsage:
@ahgittin I wonder if it's time to "break backwards compatibility", in time for major release (Brooklyn 1.0.0) so that `bundle: ` is required when deploying an auto-wrapped .bom file! It's a simple thing for users to add to their existing .bom files, with the addition their .bom file would still work with 0.x, and we can make the error message very clear when they do a `br catalog add`.

If we did that, our lives would become a lot simpler I believe. If we don't do it now, then it's going to be harder to justify breaking backwards compatibility (to require `bundle: `) until we're doing a 2.0 release.

Thoughts? Would that make life a lot easier for us? (If you agree, then this is definitely one for the mailing list!).

core/src/main/java/org/apache/brooklyn/core/typereg/BasicBrooklynTypeRegistry.java
```diff
@@ -368,13 +399,77 @@ public void addToLocalUnpersistedTypeRegistry(RegisteredType type, boolean canFo
...
+                    } else {
+                        reasonForDetailedCheck = type.getId()+" is defined in different bundle";
+                    }
+                    assertSameEnoughToAllowReplacing(existingT, type, reasonForDetailedCheck);
+                }
+            
+                log.debug("Inserting "+type+" into "+this+
+                    (oldContainingBundlesToRemove.isEmpty() ? "" : " (removing entry from "+oldContainingBundlesToRemove+")"));
+                for (String oldContainingBundle: oldContainingBundlesToRemove) {
+                    knownMatchingTypesByBundles.remove(oldContainingBundle);
```
aledsage: This is hard to follow - it looks reasonable, but it's hard to reason about because there are so many situations. Also the context/granularity of how this is used feels strange - other code like `BasicBrooklynCatalog.collectCatalogItemsFromItemMetadataBlock` is calling this repeatedly for each thing in a bundle (. If something goes wrong, it's then up to the caller to revert those actions. But this call side-effects what we had recorded against for other bundles. So hard to see how we get back to the previous state on such a rollback.

I come back to the idea of making `bundle: ` compulsory. That would simplify this code.

Longer term, I also think we want a "working view" of the `BrooklynTypeRegistry` that can apply a bunch of changes and then commit or abort them. But that's a bigger topic of conversation (completely separate from this PR!).

core/src/main/java/org/apache/brooklyn/core/typereg/BasicBrooklynTypeRegistry.java
```diff
@@ -368,13 +399,77 @@ public void addToLocalUnpersistedTypeRegistry(RegisteredType type, boolean canFo
...
+                }
+                knownMatchingTypesByBundles.put(type.getContainingBundle(), type);
+            });
+    }
+
+    private boolean isWrapperBundle(String bundleNameVersion) { 
+        if (bundleNameVersion==null) return true;
+        Maybe<OsgiManager> osgi = ((ManagementContextInternal)mgmt).getOsgiManager();
+        // if not osgi, everything is treated as a wrapper bundle
+        if (osgi.isAbsent()) return true;
```
aledsage: [bigger topic than this PR, but raising it here while I'm thinking about it].

Now that osgi is the "only" way to run brooklyn (except for tests, and for developers using a quick launch in their IDE?), I wonder how we want to handle `osgi.isAbsent()` long term. For example, instead of handling it in our code in lots of places, perhaps tests should inject a pseudo-osgi-manager that just tracks bundle install/delete/find, and does a simple peek inside it for a catalog.bom at the root. Anything more complicated would fail-fast.

Let's discuss that separately, but it does feel like a way to simplify the production code (making our test framework code slightly more complicated instead).

core/src/main/java/org/apache/brooklyn/core/catalog/internal/BasicBrooklynCatalog.java
```diff
@@ -1492,7 +1503,7 @@ protected OsgiBundleInstallationResult addItemsOsgi(String yaml, boolean forceUp
 
         OsgiBundleInstallationResult result = null;
         try {
-            result = osgiManager.get().install(null, new FileInputStream(bf), true, true, forceUpdate).get();
+            result = osgiManager.get().install(new BasicManagedBundle(vn.getSymbolicName(), vn.getVersionString(), null), new FileInputStream(bf), true, true, forceUpdate).get();
```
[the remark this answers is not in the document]
ahgittin: has to be non-osgi version string -- that's the reason we pass in.  previously it was inferring version from osgi bundle, then using it where the non-osgi version string was wanted.

core/src/main/java/org/apache/brooklyn/core/typereg/BasicBrooklynTypeRegistry.java
```diff
@@ -74,28 +92,40 @@ public BasicBrooklynTypeRegistry(ManagementContext mgmt) {
...
+        if (m==null) return null;
+        if (m.isEmpty()) return null;
+        if (m.size()==1) return m.values().iterator().next();
+        // get the highest version of first alphabetical - to have a canonical order
+        return m.get( Ordering.from(VersionedNameStringComparator.INSTANCE).max(m.keySet()) );
+    }
+
     @SuppressWarnings("deprecation")
     @Override
     public Iterable<RegisteredType> getMatching(Predicate<? super RegisteredType> filter) {
```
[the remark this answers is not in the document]
ahgittin: It's not so much the semantics of the method as what is allowed in catalog

i'm updating release notes:

```
1. BOM files that do not declare a version now give the version of the bundle to their entities,
   rather than the default 0.0.0-SNAPSHOT version.

1. Some catalog methods may return the same type multiple times, if contained in multiple bundles.
   Previously only one of the bundle's definition of the type was returned. 
   Except for anonymous bundles it is no longer allowed to have give items with the same name and version.
   (This is required to prevent Brooklyn from getting in to a state where it cannot rebind.)
```

core/src/main/java/org/apache/brooklyn/core/typereg/BasicBrooklynTypeRegistry.java
ahgittin: Reverting / rolling-back to undo side effects is no different to as before, is it?  (Except there may be fewer types which need to be restored.)

Agree with complexity creep but feels better we take that pain than give it to users who would have to switch and change code for new mental model which still feels a little premature as discussed on ML.  Plus I think once we nail this I expect we can leave it alone.

+1 to snapshotting type registry for rollback and other purposes, that's one of the reasons I prefer this approach:  once legacy catalog is removed it should be much easier than with previous legacy catalog as it's just the local map of maps defined in this class which needs to be copied, two-levels deep; that removes a lot of pain around rollback (plus it could allow per-tenant views of types based on which bundles they are allowed to access)

core/src/main/java/org/apache/brooklyn/core/typereg/BasicBrooklynTypeRegistry.java
ahgittin: that would be nice.  yes, pseudo-osgi `OsgiManager` might be cleaner than checks throughout our code, depending how much smarts are needed there.  alternatively it could spin up felix on demand.

other thing to consider is _only_ osgi, though that make make running individual IDE tests too cumbersome to set up and/or slow to run.

core/src/main/java/org/apache/brooklyn/core/catalog/internal/BasicBrooklynCatalog.java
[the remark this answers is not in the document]
ahgittin: agree re passing `OsgiManager` rather than it wrapped in `Maybe`, i've done that now

ahgittin:
all comments addressed @aledsage - one minor code change re `OsgiManager`

build seems stuck.  passes locally (just checked) so i think merge now?

aledsage:
Thanks @ahgittin - merging.

PR #869, opened 2017-10-25
Reverted code for selecting subnet hostname for non aws clouds

Graeme-Miller:
A change was made to how Brooklyn retrieves the subnet hostname in https://github.com/apache/brooklyn-server/pull/849.

The intention of that PR was to remove AWS specific functionality, but instead it affected all clouds as it moved all clouds to using ```node.getHostname()```. This functionality cant be relied on for a number of clouds. Openstack, for instance, were getting subnet hostnames like 'qa-tomcat-ser-lc5dhz8sm3-b80' which do not work. Other affected clouds include Azure and Softlayer.

This PR changes the code back to before 849, whilst still removing the AWS specific functionality. However, when testing this code, I discovered that for most clouds we do not actually get a subnet hostname. Instead we use the subnet IP. It's only really AWS that would get a subnet hostname as it uses the AWS specific code. If people prefer, we can simply revert https://github.com/apache/brooklyn-server/pull/849 instead.

### DIFF
--- a/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
+++ b/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
@@ -2827,13 +2827,6 @@ public class JcloudsLocation extends AbstractCloudMachineProvisioningLocation im
             return getPublicHostname(node, sshHostAndPort, userCredentials, setup);
         }
 
-        // NodeMetadata.getHostname() is supposed to return the private hostname. If it's not null, we want to prioritise
-        // this, otherwise we call getPrivateHostnameGeneric()
-        final String hostname = node.getHostname();
-        if (hostname != null) {
-            return hostname;
-        }
-
         Optional<String> preferredAddress = sshHostAndPort.isPresent() ? Optional.of(sshHostAndPort.get().getHostText()) : Optional.<String>absent();
         return getPrivateHostnameGeneric(node, setup, preferredAddress);
     }


Graeme-Miller:
@tbouron  and @andreaturli  it would be good to get your opinion on this.

aledsage:
For #849's change, the reason the code used the "public-hostname" for both public and private hostname sensor in AWS was because that hostname is resolvable in both (inside the region it resolves to the private ip, and outside the region it resolves to the public ip). That was true for aws-classic. I suspect that for VPCs and private subnets, it's less true - but probably worked because we'd fail to get the hostname and would log.warn and fallback to using the IPs.

Completely agree with your observation @Graeme-Miller that we can't trust `node.getHostname()` (unless we start putting cloud-specific code back in again).

---
The concern with these changes is that people must know which sensor to use: for example, will the app be used from inside the same region or from outside. For example, that's a problem when you run a demo in GCE - the URL advertised at the top-level app is often the subnet ip. Do these changes mean we'll have the same problem demo'ing in AWS?

Arguably we should solve that problem at the app-level, to say explicitly which URL(s) to advertise (as described in Alex's proposal for `.public` sensor name suffixes etc).

Overall I'm fine with this change, as long as it doesn't make demos / getting-started harder by advertising "main.url" as the internal IP of nginx in a three-tier app deployment to AWS!

andreaturli:
As this code is part of the `JcloudsLocation` we should rely on jclouds to report the right details from the provisioned node, so if `node.getHostname()` is not accurate we should open an issue on jclouds. My two cents

Graeme-Miller:
@aledsage did a quick investigation and it looks like you are right.
In AMP 4.8, our main.uri (which is highlighted in the UI) would be based off of the host.subnet.hostname so would be publicly accessible.

With the change in https://github.com/apache/brooklyn-server/pull/849 host.subnet.hostname would be populated correctly, and would not be publicly accessible. 

I am happy to close this PR and revert https://github.com/apache/brooklyn-server/pull/849 if you think that is the right route?

Graeme-Miller:
Spoke to @aledsage and we have agreed reverting #849 is the way to go. I have created #871 to do this

PR #874, opened 2017-11-02
avoid UpdatingMapTest non-det failure

ahgittin:
need to allow a first default event as that transformation may be running slowly


ahgittin:
fixes non-deterministic failure as observed in #872

aledsage:
retest this please

PR #876, opened 2017-11-03
Stop memberSpecs being inherited

drigodwin:
If I create a cluster inside a cluster currently the memberSpec is inherited by the second cluster. When that memberSpec is inherited it loops creating an infinite structure. This isn't useful and therefore memberSpecs shouldn't be inherited.


aledsage:
LGTM; merging.

PR #877, opened 2017-11-06
update maven-bundle-plugin to include headers for catalog bundles rebind after upgrade

robertgmoss:
depends on #873

### DIFF
--- a/parent/pom.xml
+++ b/parent/pom.xml
@@ -1315,6 +1315,8 @@
                                 <Export-Package>brooklyn.*,org.apache.brooklyn.*</Export-Package>
                                 <Implementation-SHA-1>${buildNumber}</Implementation-SHA-1>
                                 <Implementation-Branch>${scmBranch}</Implementation-Branch>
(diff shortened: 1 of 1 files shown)


robertgmoss:
It should be safe to add these headers here since it will not be inherited by `brooklyn-downstream-parent`

aledsage:
Agreed this should be safe, and agree it's the right default. Merging.

---
For the record, my reasoning is...

bundles that are shipped in brooklyn will be automatically updated when a new brooklyn release is shipped. The new distro version will contain the new bundle versions of everything in brooklyn. They are extremely likely to contain java code, rather than pure yaml .bom files (for pure-yaml, we can much better handle having multiple versions - we don't need to worry about java binary compatibility, for which version of Brooklyn it was compiled against). These bundles should be automatically upgraded.

We should be careful in Brooklyn about moving entities/policies around between bundles, or renaming bundles. If we do, then that bundle may need to override the defaults.

Blueprints are increasingly written by the community (e.g. see contributions to github.com/brooklyncentral). Their release cadence is different from that of core brooklyn. Moving forwards, the core of brooklyn is not where we should be writing and maintaining a lot of entity integrations.

PR #878, opened 2017-11-06
Misc enhancements - warnings and more lookup conveniences

ahgittin:
Builds on #875 (review that first) adding just:

* a `Task.warning(msg, false)` that doesn't `LOG.warning`
* `mgmt.lookup(...)` methods supporting adjuncts and more predicates
* removing unnecessary task finalization warnings
* routine to `assertEntityHealthyEventually`

### DIFF
--- a/api/src/main/java/org/apache/brooklyn/api/mgmt/ManagementContext.java
+++ b/api/src/main/java/org/apache/brooklyn/api/mgmt/ManagementContext.java
@@ -41,6 +41,7 @@ import org.apache.brooklyn.config.StringConfigMap;
 import org.apache.brooklyn.util.guava.Maybe;
 
 import com.google.common.annotations.Beta;
+import com.google.common.base.Predicate;
 
 /**
  * This is the entry point for accessing and interacting with a realm of applications and their entities in Brooklyn.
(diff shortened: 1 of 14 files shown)


ahgittin:
retest this please

expect that failure is spurious - PR included in #879 which passes

ahgittin:
still working on rebind, but note that this now also depends on https://github.com/apache/brooklyn-server/pull/887

aledsage:
retest this please

api/src/main/java/org/apache/brooklyn/api/mgmt/ManagementContext.java
```diff
@@ -309,8 +310,14 @@
...
-    /** Finds an entity with the given ID known at this management context */
-    // TODO in future support policies etc
+    /** As {@link #lookup(Predicate)} comparing the ID of the object with the given string */
     public <T extends BrooklynObject> T lookup(String id, Class<T> type); 
 
-}
\ No newline at end of file
+    /** Finds a {@link BrooklynObject} known in this management context 
+     * satisfying the given predicate, or null */
+    public <T extends BrooklynObject> T lookup(Predicate<? super T> filter);
```
aledsage: I reviewed these same commits in https://github.com/apache/brooklyn-server/pull/879, and comments addressed there!
* https://github.com/apache/brooklyn-server/pull/879#discussion_r149616384 for marking these `@Beta` and discussion of `tryLookup` instead.
* https://github.com/apache/brooklyn-server/pull/879#discussion_r149616593 for discussion of performance of lookup (agree performance is probably ok - takes approx 1ms to look up a non-existent entity id when there are 1000 entities). However, saying "only used for debugging and tests" is concerning - this is part of our public api so we should treat it as though users and other brooklyn developers will call it in all sorts of situations that we haven't thought of yet.

ahgittin:
@aledsage suggest merge this when #879 is ready to go, which should be imminent

#890 will then only really have the 2 relevant commits as this branch is merged in there - no need for a rebase.  (the github ui doesn't always update that, but if you make a fresh PR unchanged i think it will show.)

aledsage:
LGTM; merging this, then #879, then #890.

@ahgittin I think we'll agree to disagree about the use of merge commits in private branches that have not yet been merged to the public repo(s), versus the use of rebase so that each PR has a clear list of commits only for the changes in that PR.

PR #880, opened 2017-11-06
Fix rebind EntitySpec with policies ref

aledsage:


api/src/main/java/org/apache/brooklyn/api/entity/EntitySpec.java
```diff
@@ -112,12 +118,29 @@
...
+    private List<Policy> policies;
+    private List<Enricher> enrichers;
+
     public EntitySpec(Class<T> type) {
         super(type);
     }
 
     @Override
+    protected Object readResolve() {
+        if (policies != null && policies.size() > 0) {
```
ahgittin: personal preference for `&& !policies.isEmpty()`

ahgittin:
Good thorough fix.

One item needing fixed -- should be `enrichers = null` in second `readResolve` block.

Are there known instances where a policy/enricher would have been persisted and is now being ignored?  Or was the rebind error simply happening for empty lists and relatively unimportant items?

aledsage:
@ahgittin thanks. The rebind error was for an empty list of policies (which I think is the default, so rebind should have been failing for any upgrade from old to new persisted state!).

I don't see how we could have had a policy/enricher instance being referenced by an EntitySpec in historic persisted state - it would have failed to rebind in older versions of Brooklyn because the dangling-policy-ref error I described in javadoc of the test.

ahgittin:
aha gotcha - do the `s/policy/enricher` if you haven't already then good to merge

aledsage:
retest this please

failure is unrelated:
```
java.lang.AssertionError: result=Present[value=Present[value=foo]] expected [true] but found [false]
	at org.apache.brooklyn.util.core.task.ValueResolverTest.testUnsubmittedTaskWithExecutionContextTimesOutWhenImmediate(ValueResolverTest.java:159)
```

PR #882, opened 2017-11-06
Reuse bundle's catalog.bom from server-cli module

tbouron:
Some items were missing when I did the update of `catalog.bom` files to use bundles. This fixes it by including this bundle within the `karaf-init`'s `catalog.bom`


ahgittin:
Are the demo items `1-server-template` ... `4-` still included?  Looks like they are deleted here.  I think docs, tutorials, etc refer to them and they are a good getting-started sequence.

If a sub-project wants to take only some of these I might suggest refactoring so stock things like `server` and `cluster` and `vanilla-...` are in `brooklyn/cli/components.bom` and `1-server-template` through `4-...` are in `brooklyn/cli/learning-templates.bom`, then the root `catalog.bom` includes those two `classpath://... .bom` files.

The brooklyn dist should then include this catalog.  But other dists might only refer to the `components.bom`.

ahgittin:
aha never mind i think this has been done by having the templates in https://github.com/apache/brooklyn-dist/pull/112 .

feels cleaner not to have blueprints introduced in the `dist/karaf/config/...` IE using the refactoring mentioned earlier.  but what you've done is already an improvement removing a lot of the reptetition - so whichever you think best.

tbouron:
@ahgittin demo items are not remove, there are added to the catalog in `brooklyn-dist`. BTW, I did add them, there were always there, I just remove the duplication with catalog items' definition within bundles.

I think it's cleaner to have `template` items being defined in `brooklyn-dist` rather than here as your final catalog should be tailored per distribution. Putting them in the `catalog.bom` bundle means that this will show up in any downstream project/distribution without possibly to remove them

ahgittin:
your comment must have came in at the same time as mine noting you removed the repetition & weren't responsible.  i'll buy the idea that tutorial templates live in dist.  good to merge IMO.

PR #883, opened 2017-11-07
Deprecates use of Task in ConfigKey

aledsage:
Also avoid `StackOverflowError` in `ExecutionContext.getImmediately(task)`
When a non-`BasicTask` is passed in.

---
The motivation for deprecating setting a `ConfigKey` as a `Task` is that we really shouldn't be doing it! It can get very messed up if one tries to `getImmediately` or get non-blocking (e.g. by a `Transformer` that is trying to evaluate its result when triggered). That can leave the task in a cancelled state, so all subsequent calls to `task.get()` (i.e. all calls to `config().get(mykey)`) will fail!

Instead, we should be using `TaskFactory` (which is how the DSL evaluation works).

It's good to deprecate this before a major release of 1.0.0.

It would be good to add an overloaded method for setting config to `set(ConfigKey<T>, TaskFactory<T>)` so that such usage doesn't involve ugly casting to work around the strongly typed generics. We can do that as a stage two, separate from this PR.

### DIFF
--- a/api/src/main/java/org/apache/brooklyn/api/mgmt/ExecutionContext.java
+++ b/api/src/main/java/org/apache/brooklyn/api/mgmt/ExecutionContext.java
@@ -85,20 +85,26 @@ public interface ExecutionContext extends Executor {
      * Implementations will typically act like {@link #get(TaskAdaptable)} with additional
      * tricks to attempt to be non-blocking, such as recognizing some "immediate" markers.  
      * <p>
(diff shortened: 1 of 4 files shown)


ahgittin:
Looks good.  Merging.

Agree re `set(ConfigKey<T>, TaskFactory<T>)`.  Would also be nice to warn but similarly we can do that in a separate PR, once we've removed other uses.  In particular Java calls to `DependentConfiguration` eg `attributeWhenReady` will make `Task` instances so that is an obvious item to fix.

There is related work in #853 which should be built upon as this continues.

PR #884, opened 2017-11-07
Add brooklyn-cli as a feature to karaf

Graeme-Miller:
https://github.com/apache/brooklyn-server/pull/882 added brooklyn-cli to a catalog.bom. This causes brooklyn to fail to start, as karaf can't find brooklyn-cli. This PR adds brooklyn-cli to karaf

### DIFF
--- a/karaf/features/src/main/feature/feature.xml
+++ b/karaf/features/src/main/feature/feature.xml
@@ -327,11 +327,16 @@
         <bundle>mvn:org.apache.brooklyn/brooklyn-test-framework/${project.version}</bundle>
     </feature>
 
+    <feature name="brooklyn-cli" version="${project.version}" description="Brooklyn CLI" >
+        <bundle>mvn:org.apache.brooklyn/brooklyn-cli/${project.version}</bundle>
+    </feature>
+
     <feature name="brooklyn-server-software-all" version="${project.version}" description="Brooklyn All Core Entities">
         <feature>brooklyn-software-base</feature>
         <feature>brooklyn-jmxmp-agent</feature>
         <feature>brooklyn-jmxrmi-agent</feature>
         <feature>brooklyn-test-framework</feature>
+        <feature>brooklyn-cli</feature>
     </feature>
 
     <feature name="brooklyn-osgi-launcher" version="${project.version}" description="Brooklyn init">


aledsage:
LGTM; merging.

aledsage:
Turns out this might only be working because of an error in https://github.com/apache/brooklyn-server/blob/master/server-cli/pom.xml#L118 - there's a missing comma, so this seems to behave the same as `<Import-Package>!*</ImportPackage>`. The `org.apache.brooklyn.cli` OSGi bundle has no `Import-Package` in its manifest!

If it did generate that correctly, then it would include an import package dependency on `org.apache.brooklyn.camp.brooklyn`, which comes from `brooklyn-camp`. The order the features are declared means it might try to install `brooklyn-cli` before `brooklyn-camp`, so its dependencies wouldn't resolve. Or maybe it would all be fine, with karaf correctly figuring out all the dependencies at startup.

---
However, the fundamentally concern is that there is no reason to include the `brooklyn-cli` module in karaf. Its java code is pretty-much unusable (we've excluded the `airlift` package import because that is not in karaf).

If we're just doing this to get the contents of its catalog.bom, maybe we should move its catalog.bom somewhere else instead.

As was discussed on the mailing list previously, we agreed to delete from production-code the brooklyn classic mode (i.e. the `Main` class way of launching brooklyn that is in `brooklyn-cli`). The suggestion was to move it to tests-only so developers could still launch it from their IDE if they really wanted. Therefore it would be nice to consider the `brooklyn-cli` module as deprecated.

Thoughts @Graeme-Miller @ahgittin @tbouron?

tbouron:
I think you are right @aledsage, I clearly didn't think this through with #882. I would remove this dependency and add those item into `brooklyn-library` or `brooklyn-core` instead

aledsage:
I tried correcting the `Package-Import`, building brooklyn karaf dist, and starting it. It failed to install the `brooklyn-cli` bundle because `brooklyn-launcher` is not included in the dist:
```
2017-11-08T08:51:51,254 ERROR   9 o.a.k.f.i.s.BootFeaturesInstaller [pool-1-thread-2] Error installing boot features
org.osgi.service.resolver.ResolutionException: Unable to resolve root: missing requirement [root] osgi.identity; osgi.identity=brooklyn-headless; type=karaf.feature; version=0; filter:="(&(osgi.identity=brooklyn-headless)(type=karaf.feature)(version>=0.0.0))" [caused by: Unable to resolve brooklyn-headless/1.0.0.SNAPSHOT: missing requirement [brooklyn-headless/1.0.0.SNAPSHOT] osgi.identity; osgi.identity=brooklyn-server-software-all; type=karaf.feature [caused by: Unable to resolve brooklyn-server-software-all/1.0.0.SNAPSHOT: missing requirement [brooklyn-server-software-all/1.0.0.SNAPSHOT] osgi.identity; osgi.identity=brooklyn-cli; type=karaf.feature [caused by: Unable to resolve brooklyn-cli/1.0.0.SNAPSHOT: missing requirement [brooklyn-cli/1.0.0.SNAPSHOT] osgi.identity; osgi.identity=org.apache.brooklyn.cli; type=osgi.bundle; version="[1.0.0.SNAPSHOT,1.0.0.SNAPSHOT]"; resolution:=mandatory [caused by: Unable to resolve org.apache.brooklyn.cli/1.0.0.SNAPSHOT: missing requirement [org.apache.brooklyn.cli/1.0.0.SNAPSHOT] osgi.wiring.package; filter:="(&(osgi.wiring.package=org.apache.brooklyn.launcher)(version>=1.0.0)(!(version>=2.0.0)))"]]]]
```

I think we should back off from including `brooklyn-cli`, and just move the catalog items.

I'm torn as to where the best place is - they are really just aliases for other catalog items, including for nginx in `brooklyn-software-webapp` in the `brooklyn-library` repo. Options include:
1. Create a new module, such as `org.apache.brooklyn:brooklyn-catalog-aliases` in `brooklyn-library/catalog-aliases` (and include it in `brooklyn-library/karaf/catalog/src/main/resources/catalog.bom`)
2. Add it to an existing bundle, such as `brooklyn-core` (_don't like because that shouldn't know about nginx) or `brooklyn-software-webapp` (_don't like because `server` etc is nothing to do with webapps_)
3. Create a new `catalog-aliases.bom` in `brooklyn-dist/karaf/config/src/main/resources/catalog/` (i.e. next to `catalog-templates.bom`). _don't like that, because it should be in a bundle._

On balance, I favour option 1: the new module.

Thoughts @Graeme-Miller @ahgittin @tbouron?

tbouron:
@aledsage I agree with your assessment, option 1 seems to be the best for all the reasons you mentioned. I can do those changes if you want?

PR #885, opened 2017-11-08
Icons for adjuncts, and descriptions

ahgittin:
These weren't returned by API; this PR fixes that, and adds icons for selected policies.


tbouron:
@ahgittin Are we still using the @catalog annotation? I thought it was deprecated.

Also, karaf uses exclusively the BOM file if I'm not mistaken

ahgittin:
thanks @tbouron - i hadn't realised that work was done.  cool!  i've added to the `policy/../catalog.bom` also.

rest/rest-api/src/main/java/org/apache/brooklyn/rest/api/AdjunctApi.java
```diff
@@ -253,4 +253,12 @@ public Response setConfig(
...
             @QueryParam("recurse") @DefaultValue("false") Boolean recurse);
 
+    @GET
+    @ApiOperation(value = "Returns an icon for the adjunct, if defined")
+    @Path("/{adjunct}/icon")
+    public Response getIcon(
+        @ApiParam(value = "Application ID or name", required = true) @PathParam("application") String applicationId,
+        @ApiParam(value = "Entity ID or name", required = true) @PathParam("entity") String entityId,
+        @ApiParam(value = "Adjunct ID or name", required = true) @PathParam("adjunct") String adjunctToken);
+
```
tbouron: Is it necessary to have this endpoint if the icon is returned within the `AdjunctSummary`?

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/transform/AdjunctTransformer.java
```diff
@@ -45,17 +45,24 @@
...
+                // route to server if it is a server-side url
+                adjunctSummary.iconUrl(adjunctUri(entity, adjunct, ub)+"/icon");
+        }
+
         return (T) adjunctSummary.state(inferStatus(adjunct)).links( buildLinks(entity, adjunct, ub, adjunctSummary instanceof AdjunctDetail) );
     }
 
-    public static AdjunctDetail adjunctDetail(BrooklynRestResourceUtils utils, Entity entity, EntityAdjunct adjunct, UriBuilder ub) {
-        AdjunctDetail result = embellish(new AdjunctDetail(adjunct), entity, adjunct, ub);
+    public static AdjunctDetail adjunctDetail(Entity entity, EntityAdjunct adjunct, UriBuilder ub, BrooklynRestResourceUtils utils) {
```
tbouron: Why changing the signature? This is public API and a breaking change for downstream projects

rest/rest-api/src/main/java/org/apache/brooklyn/rest/api/AdjunctApi.java
ahgittin: it's only the `iconUrl` returned in adjunct summary.  the url might be a server-side `classpath:`.  this endpoint returns the actual image.

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/transform/AdjunctTransformer.java
ahgittin: the adjunct api is new since last release so i think this is okay?

ahgittin:
icons are all CC with url in commit history so don't think we need to add to our uber-license file ... @rdowner ?

ahgittin:
non-det failure unrelated (though it is one of the policies mentioned here):

`org.apache.brooklyn.policy.ha.ServiceRestarterTest.testEmitsFailureEventWhenRestarterFails`

```
expected [on-fire] but found [stopping]
Stacktrace

java.lang.AssertionError: expected [on-fire] but found [stopping]
	at org.apache.brooklyn.policy.ha.ServiceRestarterTest.testEmitsFailureEventWhenRestarterFails(ServiceRestarterTest.java:116)
Standard Output

2017-11-08 09:26:55,884 INFO  TESTNG INVOKING CONFIGURATION: "Surefire test" - @AfterMethod org.apache.brooklyn.core.test.BrooklynMgmtUnitTestSupport.tearDown()
2017-11-08 09:26:55,898 INFO  TESTNG PASSED CONFIGURATION: "Surefire test" - @AfterMethod org.apache.brooklyn.core.test.BrooklynMgmtUnitTestSupport.tearDown() finished in 14 ms
2017-11-08 09:26:55,898 INFO  TESTNG INVOKING CONFIGURATION: "Surefire test" - @BeforeMethod org.apache.brooklyn.policy.ha.ServiceRestarterTest.setUp()
2017-11-08 09:26:55,899 INFO  Added external config supplier named 'brooklyn-demo-sample': org.apache.brooklyn.core.config.external.InPlaceExternalConfigSupplier@eca6a74
2017-11-08 09:26:55,939 INFO  TESTNG PASSED CONFIGURATION: "Surefire test" - @BeforeMethod org.apache.brooklyn.policy.ha.ServiceRestarterTest.setUp() finished in 41 ms
2017-11-08 09:26:55,939 INFO  TESTNG INVOKING: "Surefire test" - org.apache.brooklyn.policy.ha.ServiceRestarterTest.testEmitsFailureEventWhenRestarterFails()
2017-11-08 09:26:55,963 INFO  ServiceRestarter notified; dispatching job for FailingEntityImpl{id=hlb2cvug7i} (FailureDescriptor{component=FailingEntityImpl{id=hlb2cvug7i}, description=simulate failure})
2017-11-08 09:26:55,963 WARN  ServiceRestarter acting on failure detected at FailingEntityImpl{id=hlb2cvug7i} (FailureDescriptor{component=FailingEntityImpl{id=hlb2cvug7i}, description=simulate failure})
2017-11-08 09:26:56,040 WARN  Setting FailingEntityImpl{id=hlb2cvug7i} on-fire due to problems when expected running, up=false, not-up-indicators: null
2017-11-08 09:26:56,041 WARN  Error invoking restart at FailingEntityImpl{id=hlb2cvug7i}: Simulating entity restart failure for test
2017-11-08 09:26:56,044 WARN  Task Task[sensor r726t791bn:entity.children.removed]@ZblNlUgv is being finalized before completion
2017-11-08 09:26:56,044 WARN  Task Task[sensor r726t791bn:entity.children.removed]@Xui3nJyq is being finalized before completion
2017-11-08 09:26:56,044 WARN  Task Task[sensor r726t791bn:service.state.expected]@hYA9kKdc is being finalized before completion
2017-11-08 09:26:56,046 WARN  ServiceRestarter failed for FailingEntityImpl{id=hlb2cvug7i}: Restart failure (error java.util.concurrent.ExecutionException: org.apache.brooklyn.core.mgmt.internal.EffectorUtils$EffectorCallPropagatedRuntimeException: Error invoking restart at FailingEntityImpl{id=hlb2cvug7i}: Simulating entity restart failure for test) at FailingEntityImpl{id=hlb2cvug7i}: FailureDescriptor{component=FailingEntityImpl{id=hlb2cvug7i}, description=simulate failure}
2017-11-08 09:26:56,054 INFO  TESTNG FAILED: "Surefire test" - org.apache.brooklyn.policy.ha.ServiceRestarterTest.testEmitsFailureEventWhenRestarterFails() finished in 112 ms
java.lang.AssertionError: expected [on-fire] but found [stopping]
	at org.apache.brooklyn.policy.ha.ServiceRestarterTest.testEmitsFailureEventWhenRestarterFails(ServiceRestarterTest.java:116)
```

rest/rest-api/src/main/java/org/apache/brooklyn/rest/api/AdjunctApi.java
tbouron: @ahgittin Hum, so I just checked and the way the `applications` and `entities` endpoints work is as follow: the `iconUrl` field returned by the `EntitySummary` is a relative link to the related icon endpoint, i.e. `/v1/entities/icon/symbolicName/version`

So indeed, we need this icons endpoint for adjunct. But more importantly, we should mirror the same behaviour.

tbouron:
@ahgittin My understanding was that if the licence is different than `Apache-2.0`, then it needs to be included.

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/transform/AdjunctTransformer.java
tbouron: My question is more: do we really need this change? It just shuffles the arguments, there is no benefits compare to the hassle of potential breaking things and informing users. I don't think it is worth the effort. 

If we are going forward with this change, it should be added to the list of breaking changes in the release notes.

ahgittin: it isn't a breaking change as there has been no release since it was introduced.  changes are common within a release as new things evolve.  this makes the method consistent with others and puts salient fields first, context fields after.

i'm pretty sure nothing external is using any of these transformers and almost certain not this very new one.  i'm pretty sure you're talking a very low theoretical-only risk and as we haven't done a release we're not breaking any guarantees we offer.

release notes are for breaking changes since previous release so would be very odd to put it there.  especially as it would be trivial to leave the old method in place deprecated instead -- that would be the sensible fix -- but for the reasons above don't even think that's worth it.

i'll take this all back if you can point me to anything that might actually be impacted however!

rest/rest-api/src/main/java/org/apache/brooklyn/rest/api/AdjunctApi.java
ahgittin: @tbouron what are you looking at?  `/v1/entities/icon/symbolicName/version` isn't a path i'm familiar with.

i've based this on `EntityApi.getIcon` which corresponds to `/v1/applications/XXX/entities/XXX/icon` exactly as does this.

`CatalogResource.getIcon` is different but we should mirror entity instances.  (you could also use the `/v1/catalog/icon/XXX/VVV` endpoint for this but that won't look at icon tags on the instances.)

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/transform/AdjunctTransformer.java
tbouron: @ahgittin adjuncts API is part of 0.12.0 isn't it? if yes, this is a breaking change. If it was introduced after 0.12.0 then point taken.

ahgittin:
@tbouron i thought CC was similar to public domain but attribution is nice so i'll add that

rest/rest-api/src/main/java/org/apache/brooklyn/rest/api/AdjunctApi.java
tbouron: @ahgittin Sorry, I meant `iconUrl: /v1/catalog/icon/symbolicName/version`.

For example, `/v1/applications` returns the above `iconUrl` if the catalog items have icons

tbouron:
@ahgittin attribution is nice indeed. Bear in ming that I'm not an expert in licensing, @rdowner knows more than me on this topic, I'm sure he will be able to provide a better answer.

ahgittin:
license for icons added, last commit.  references detail in https://github.com/apache/brooklyn-dist/pull/113

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/transform/AdjunctTransformer.java
ahgittin: it's not in 0.12.0 that's the whole point - see the first comment in this thread

adjunct api was added in 92b9cc638653b53effa20a696e69cb0d9c5c9a9c after 0.13.0-SNAPSHOT version bump d5f9ae48b5c212090abc374c7c0732193f42a254

tbouron: Sorry, I misunderstood you comment. I'm fine with it then

rest/rest-api/src/main/java/org/apache/brooklyn/rest/api/AdjunctApi.java
ahgittin: as we're dealing with adjunct instances here best to copy the entity instance api as i've done, no?

whether an entity returns `/v1/applications/XXX/entities/XXX/icon` or `/v1/catalog/icon/symbolicName/version` as the URL is not guaranteed.  in fact the latter would sometimes be wrong if an entity has a custom icon set in a tag, and the code suggests entity instances returns the former, at least for `EntityDetail`, in `ApplicationResource.fromEntity`:

    iconUrl = EntityTransformer.entityUri(entity, ui.getBaseUriBuilder())+"/icon";

the `EntitySummary` class doesn't include `iconUrl`.  you probably mean `CatalogEntitySummary` which is for types not instances in which case the non-instance link would be the correct one.

but for this purpose -- adjunct _instances_ - i think what i've done, mirroring entity instances, is correct.

tbouron: Yeah, I probably mixed up the name of REST endpoints with the multiple methods of getting an icon. There are too many of them, it's really confusing and hard to follow our REST API!

This looks fine to me.

ahgittin: Thx.

We should introduce `/v1/catalog/type/XXX/V/icon` for consistency with `/v1/applications/XXX/icon` and `/v1/applications/XXX/entities/XXX/icon` and `/v1/applications/XXX/entities/XXX/adjuncts/XXX/icon` and deprecate /v1/catalog/icon/XXX/V`.  But I think our multiple methods are sensible, you can get an icon for instances, and for a type.

removing deprecated items like `CatalogEntitySummary` and others in favour of `TypeSummary` should make it simpler

think on any other confusing things and let me know

ahgittin:
@tbouron good to merge this and https://github.com/apache/brooklyn-dist/pull/113 now?

PR #887, opened 2017-11-09
Config on entity dynamic goes on type after persistence

ahgittin:
Fixes two issues noted with rebind of dynamic config.  See first commit first for failing test, then the others which fix it and comment.


ahgittin:
agree with comment - addressed, merging

PR #889, opened 2017-11-09
Revert @JsonInclude annotations to what it was in 0.12.0 for backward compatibility

tbouron:
As per as [the discussion on the ML](https://lists.apache.org/thread.html/274c9b60e5fc15a7f25f0311dd0139667d0c7a5536ff86523010089d@%3Cdev.brooklyn.apache.org%3E), this revert the use of `@JsonInclude` annotation for certain fields of domain objects to insure backward compatibility.


ahgittin:
LGTM

PR #890, opened 2017-11-09
REST `/applications/fetch` can request selected sensors

ahgittin:
This allows clients to efficiently get selected sensors they are interested in.

Does this by adding a `@JsonAnyGetter` map to `EntitySummary`, as done for `TypeSummary` and elsewhere, returning data in `sensors` key.

Re-enables the `fetch` test (disabled in #156) to assert this, but with better logging if the problem recurs.

Also sets batch config read task as transient so it doesn't clutter the activities view.


ahgittin:
Note it's just the last two commits here; the others come from #878 (we now use the `lookup(Predicate)` convenience in the test)

ahgittin:
K - good to merge after #879

PR #900, opened 2017-11-17
Fix VSP stop/start when ssh monitoring disabled

duncangrant:
When stop / starting a VanillaSoftwareProcess with ssh monitoring the
process will never be marked as healthy after a stop start.

On a restart the feeds are not disconnected so a service.notup.indicator
is set.  When this is cleared on start the ServiceNotUpLogic will detect
the change and set the isUp sensor true.  However when stopping the
process feeds are disconnected so no service.notup indicators are set.
This means the ServiceNotUp logic will not detect a change on start and
won't set the service isUp sensor to true.

Fixed by setting a service.notUp.indicator on stop.


asfgit:
FAILURE
 3247 tests run, 1 skipped, 1 failed.
--none--

asfgit:
SUCCESS 
 4934 tests run, 0 skipped, 0 failed.
--none--

asfgit:
SUCCESS 
 4934 tests run, 0 skipped, 0 failed.
--none--

PR #902, opened 2017-11-21
BROOKLYN-545: Fix DynamicCluster.zoneFailureDetector default val

aledsage:
Fixes https://issues.apache.org/jira/browse/BROOKLYN-545

Below is an extract from the comments in the code, explaining it - see other code comments for more info:
```
    // Previously, when there was a mutable default value for ZONE_FAILURE_DETECTOR, the second cluster
    // would include failure info for the first cluster's zone-locations. When these were unmanaged,
    // the zone-locations would become dangling referenes within the ZoneFailureDetector. This caused
    // NPE on rebind.
    //
    // Config key persistence was then changed so that statically defined keys (such as ZONE_FAILURE_DETECTOR,
    // declared on the DynamicCluster class) would not be persisted. This meant that the ZoneFailureDetector
    // state was no longer persisted for either location. The NPE on rebind went away, but the config key
    // default value instance was still shared between all clusters.
    //
    // Now the detector has a null default value. We implicitly set a transient field for the zoneFailureDetector
    // so if using defaults then nothing is persisted. But if a custom ZoneFailureDetector is supplied, then
    // that will be persisted and used.
```


drigodwin:
retest this please

PR #903, opened 2017-11-21
Delete bundle: handle pre-installed karaf bundles

aledsage:
Previously, if you tried to delete a bundle that was pre-installed in brooklyn karaf then the REST pai call would return a 500 response:
```
curl -v -X DELETE -u xxxxxxxx:xxxxxxxx http://localhost:8081/v1/catalog/bundles/org.example.my-pre-installed-bundle/1.0.0
```
gave the exception:
```
"No such bundle installed in OSGi when uninstalling: BasicManagedBundle{symbolicName=org.example.my-pre-installed-bundle, version=1.0.0, url=mvn:org.example/my-pre-installed-bundle/1.0.0}"...
```

e.g. to reproduce, try removing a bundle like `http://localhost:8081/v1/catalog/bundles/org.apache.brooklyn.software-database/1.0.0.SNAPSHOT`

This PR changes our bundle-lookup so that we don't just use the `bundleMetadata.getOsgiUniqueUrl()`, but instead use lookup by url etc (in the same way as we use elsewhere).

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/mgmt/ha/OsgiManager.java
+++ b/core/src/main/java/org/apache/brooklyn/core/mgmt/ha/OsgiManager.java
@@ -472,17 +472,17 @@ public class OsgiManager {
             }
             
             if (!leaveInOsgi) {
-                Bundle bundle = framework.getBundleContext().getBundle(bundleMetadata.getOsgiUniqueUrl());
-                result.bundle = bundle;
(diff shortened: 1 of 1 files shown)


ahgittin:
non-det test failure:

    org.apache.brooklyn.core.entity.lifecycle.ServiceStateLogicTest.testStopsNicelyToo

LGTM, merging

PR #905, opened 2017-11-25
BROOKLYN-566: fix SaltEntity rebind to historic state

aledsage:
Fixes https://issues.apache.org/jira/browse/BROOKLYN-566

These were renamed in https://github.com/apache/brooklyn-library/pull/123, which broke rebind if the persisted state contained a SaltEntity from 0.11.0.

Also see https://github.com/apache/brooklyn-library/pull/144


geomacy:
LGTM, merging

PR #907, opened 2017-12-03
BROOKLYN-567 Prevent mutating EntitySpecSupplier's flag field

bostko:
I believe this fixes the issue.
Glad to do a little Sunday hiking in Apache Brooklyn.


ahgittin:
good to see you hiking these trails!  is #911 a safe further simplification of this?

(merging that will merge this; if there are subtleties that it misses then good to explain them here and then we can merge this and close that)

PR #911, opened 2017-12-06
Different fix for entity spec mutated bug BROOKLYN-657

ahgittin:


ahgittin:
@bostko thanks for #907 - i think this is a further simplification but can you confirm?

(BTW merging this will also merge #907)

camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/creation/BrooklynComponentTemplateResolver.java
```diff
@@ -509,13 +509,11 @@ public EntitySpecSupplier(EntitySpecConfiguration flag) {
                 this.flag = flag;
             }
             @Override public EntitySpec<?> get() {
-                EntitySpecConfiguration specConfig = (EntitySpecConfiguration) flag;
                 // TODO: This should called from BrooklynAssemblyTemplateInstantiator.configureEntityConfig
                 // And have transformSpecialFlags(Object flag, ManagementContext mgmt) drill into the Object flag if it's a map or iterable?
                 @SuppressWarnings("unchecked")
-                Map<String, Object> resolvedConfig = (Map<String, Object>)transformSpecialFlags(specConfig.getSpecConfiguration());
-                specConfig.setSpecConfiguration(resolvedConfig);
```
bostko: Good fix as well. Now I think `setSpecConfiguration` method should be also removed.
BROOKLYN-567

aledsage:
LGTM, and confirmed that the example test in https://issues.apache.org/jira/browse/BROOKLYN-567 now works (with the persisted state files not increasing in size).

PR #913, opened 2017-12-08
AbstractMachineLocation base class

alasdairhodge:
Introduces `AbstractMachineLocation` base class to hold responsibilities that may otherwise be duplicated between `SshMachineLocation` and `WinRmMachineLocation`.

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/location/AbstractMachineLocation.java
+++ b/core/src/main/java/org/apache/brooklyn/core/location/AbstractMachineLocation.java
@@ -20,10 +20,32 @@ package org.apache.brooklyn.core.location;
 
 import java.util.Map;
 
+import org.apache.brooklyn.api.location.MachineDetails;
 import org.apache.brooklyn.api.location.MachineLocation;
+import org.apache.brooklyn.api.location.OsDetails;
+import org.apache.brooklyn.config.ConfigKey;
+import org.apache.brooklyn.core.config.ConfigKeys;
 import org.apache.brooklyn.util.collections.MutableMap;
 
+import com.google.common.base.Optional;
+
 public abstract class AbstractMachineLocation extends AbstractLocation implements MachineLocation {
+
+    public static final ConfigKey<MachineDetails> MACHINE_DETAILS = ConfigKeys.newConfigKey(
+            MachineDetails.class,
+            "machineDetails");
+
+    public static final ConfigKey<Boolean> DETECT_MACHINE_DETAILS = ConfigKeys.newBooleanConfigKey("detectMachineDetails",
+            "Attempt to detect machine details automatically.", true);
+
+    protected static final MachineDetails UNKNOWN_MACHINE_DETAILS = new BasicMachineDetails(
+            new BasicHardwareDetails(-1, -1),
+            new BasicOsDetails("UNKNOWN", "UNKNOWN", "UNKNOWN")
+    );
+
+    private volatile MachineDetails machineDetails;
+    private final Object machineDetailsLock = new Object();
+
 
     public AbstractMachineLocation() {
         this(MutableMap.of());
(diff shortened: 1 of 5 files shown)


aledsage:
Build failure is real:
```
[ERROR] /home/jenkins/jenkins-slave/workspace/brooklyn-server-pull-requests/core/src/main/java/org/apache/brooklyn/location/localhost/LocalhostMachineProvisioningLocation.java:[284,8] error: method does not override or implement a method from a supertype
```
This is for `@Override LocalhostMachineProvisioningLocation.getMutexSupport()`, I believe.

core/src/main/java/org/apache/brooklyn/core/location/AbstractMachineLocation.java
```diff
@@ -0,0 +1,139 @@
...
+            mutexes().acquireMutex(mutexId, description);
+        } catch (InterruptedException ie) {
+            throw new RuntimeInterruptedException("Interrupted waiting for mutex: " + mutexId, ie);
+        }
+    }
+
+    /** @deprecated Mutex-related methods are now accessible via {@link #mutexes()} */
+    @Override
+    @Deprecated
+    public boolean tryAcquireMutex(String mutexId, String description) {
```
aledsage: Why have these on the new abstract super class? If we want people to use `mutexes().tryAcquireMutex()` instead, then we should leave these methods only on `SshMachineLocation`, and also `AbstractMachineLocation` should not implement `WithMutexes`. It seems wrong to add deprecated methods to the new super-class.

alasdairhodge:
Argh, forgot that interim push woiuld trigger rebuild... in progress.

core/src/main/java/org/apache/brooklyn/core/location/AbstractMachineLocation.java
alasdairhodge: Agree.

core/src/main/java/org/apache/brooklyn/core/location/AbstractMachineLocation.java
```diff
@@ -0,0 +1,131 @@
...
+
+    @Override
+    public OsDetails getOsDetails() {
+        return getMachineDetails().getOsDetails();
+    }
+
+    protected abstract MachineDetails detectMachineDetails();
+
+    @Override
+    public void acquireMutex(String mutexId, String description) throws RuntimeInterruptedException {
```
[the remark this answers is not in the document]
alasdairhodge: Now done. Annoyingly, `WithMutexes.acquireMutex()` forces all callers to handle `InterruptedException`. IMO it'd be cleaner to throw an unchecked exception instead, but I appreciate that may be contentious. I've preserved calls to the now-deprecated `SshMachineLocation.acquireMutex()` until this is agreed.

core/src/main/java/org/apache/brooklyn/location/localhost/LocalhostMachineProvisioningLocation.java
```diff
@@ -268,22 +266,15 @@ public void release(SshMachineLocation machine) {
                 BrooklynConfigKeys.SKIP_ON_BOX_BASE_DIR_RESOLUTION, 
                 true);
 
-        private static final WithMutexes mutexSupport = new MutexSupport();
```
aledsage: You've change the semantics for localhost - note that this field was static. To preserve that, you can override `WithMutexes mutexes()` to return the static field.

I think you'll be fine removing from the constructor the bit that sets it in the map (note that `SshMachineLocation.mutexSupport` was marked with `@SetFromFlag`). By just overriding `mutexes()`, I think it should be enough.

core/src/main/java/org/apache/brooklyn/location/ssh/SshMachineLocation.java
```diff
@@ -1104,4 +1013,36 @@ public String findPassword() {
         return getConfig(SshTool.PROP_PASSWORD);
     }
 
+    /** @deprecated since 5.0.1; mutex-related methods are now accessible via {@link #mutexes()} */
```
aledsage: Should be `since 1.0.0` (same elsewhere).

core/src/main/java/org/apache/brooklyn/location/localhost/LocalhostMachineProvisioningLocation.java
alasdairhodge: Excellent catch, thanks.

core/src/main/java/org/apache/brooklyn/location/ssh/SshMachineLocation.java
alasdairhodge: D'oh!

alasdairhodge:
Silly issues now resolved, @aledsage

aledsage:
Thanks @alasdairhodge - LGTM; merging.

PR #914, opened 2017-12-11
ShellEnvSerializer: handle Duration

aledsage:
This change is useful when generating environment variables.

The use-case was a blueprint that took a map of values, turning it into an env variable that is a json string. If that map contained a `Duration`, then the env variable would contain something like `{"delay": {"nanos": 30000000}}`. With this change, it will instead generate `{"delay": "5m"}`.


duncangrant:
PR could do with a description.

aledsage:
Thanks @duncangrant - I've changed the description. I've also added another test that the environment variable is really set as expected. Once jenkins confirms the tests pass, I'll merge it.

PR #915, opened 2017-12-11
BROOKLYN-569: aggregator casts vals to numbers

aledsage:


aledsage:
Retest this please

Test failure is unrelated: `ElectPrimaryTest.testSelectionModeFailoverReelectWithPreference`. The info log from jenkins has the lines shown below. This tells us that `ElectPrimaryEffector` is being called twice - once directly int the test, and presumably concurrently by the policy's `onEvent` (it triggers that whenever a child/member is added/removed or its serviceUp/serviceState/weighting sensor changes). There is a definite race there, which can happen when the effector is being called manually.

```
2017-12-11 17:28:13,488 INFO  Test created app, and will now start BasicApplicationImpl{id=jt5d932r4f}
2017-12-11 17:28:13,515 INFO  Detected new primary TestEntityImpl{id=jt0dh78cvw} at BasicApplicationImpl{id=jt5d932r4f} (previously had null)
2017-12-11 17:28:13,515 INFO  Primary TestEntityImpl{id=jt0dh78cvw} at BasicApplicationImpl{id=jt5d932r4f} detected as healthy
2017-12-11 17:28:13,528 INFO  Started application BasicApplicationImpl{id=jt5d932r4f}
2017-12-11 17:28:13,530 INFO  Detected new primary TestEntityImpl{id=ndkegu8wyc} at BasicApplicationImpl{id=jt5d932r4f} (previously had TestEntityImpl{id=jt0dh78cvw})
2017-12-11 17:28:13,531 INFO  Detected new primary TestEntityImpl{id=jt0dh78cvw} at BasicApplicationImpl{id=jt5d932r4f} (previously had TestEntityImpl{id=ndkegu8wyc})
2017-12-11 17:28:43,544 INFO  failed succeeds-eventually, 75 attempts, 30000ms elapsed (rethrowing): java.lang.AssertionError: attribute=Sensor: primary (org.apache.brooklyn.api.entity.Entity); val=TestEntityImpl{id=jt0dh78cvw}
```

core/src/main/java/org/apache/brooklyn/enricher/stock/MathAggregatorFunctions.java
```diff
@@ -118,9 +123,10 @@ public T apply(@Nullable Collection<? extends Number> vals) {
...
             if (vals != null) {
-                for (Number val : vals) { 
-                    if (val != null) {
-                        postProcessedVals.add(val);
+                for (Object val : vals) {
+                    Maybe<Number> coercedVal = TypeCoercions.tryCoerce(val, Number.class);
+                    if (coercedVal.isPresentAndNonNull()) {
+                        postProcessedVals.add(coercedVal.get());
                         count++;
                     } else if (defaultValueForUnreportedSensors != null) {
```
ahgittin: worth a `log.warn` saying the value if not coercible?

```
} else {
  if (val!=null) {
    log.warn("Input to numeric aggregator is not a number: "+val+" ("+val.getClass+")");
  }
  if (defaultValueForUnrepoertedSensors != null) {
    ...
  }
}
```

ahgittin:
@aledsage minor comment, merge when ready

core/src/main/java/org/apache/brooklyn/enricher/stock/MathAggregatorFunctions.java
aledsage: Done - but also guarding so we don't log repeatedly at warn. The aggregator might well get called every few seconds per enricher, so must avoid flooding the log with such warn messages.

tbouron:
Tested, LGTM, thanks @aledsage

PR #917, opened 2017-12-16
BROOKLYN-570: fix entity synchronizing on self

aledsage:
Fixes https://issues.apache.org/jira/browse/BROOKLYN-570

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/entity/AbstractEntity.java
+++ b/core/src/main/java/org/apache/brooklyn/core/entity/AbstractEntity.java
@@ -214,6 +214,13 @@ public abstract class AbstractEntity extends AbstractBrooklynObject implements E
     private Entity selfProxy;
     private volatile Application application;
     
+    /**
+     * Lock to be held when setting the application.
+     * Must not synchronize on `this`, as then entity authors can cause horrible hanging
+     * by also synchronize on this. Instead only synchronize on private fields.
+     */
+    private final Object appMutex = new Object();
+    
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/brooklyn/core/entity/AbstractEntity.java
```diff
@@ -749,13 +756,15 @@ public Application getApplication() {
     // FIXME Can this really be deleted? Overridden by AbstractApplication; needs careful review
     /** @deprecated since 0.4.0 should not be needed / leaked outwith brooklyn internals / mgmt support? */
     @Deprecated
-    protected synchronized void setApplication(Application app) {
-        if (application != null) {
-            if (application.getId() != app.getId()) {
-                throw new IllegalStateException("Cannot change application of entity (attempted for "+this+" from "+getApplication()+" to "+app);
+    protected void setApplication(Application app) {
+        synchronized (appMutex) {
```
tbouron: @aledsage Is it the right place to synchronize as this method is `deprecated`? It looks like the answer is yes but just wanted to check.

Do you think the `@Deprecated` annotation can be removed? Or at least the `FIXME` comment?

aledsage: We don't ever want an entity-author to call `setApplication`, so I think I'm ok leaving it deprecated. Not sure what we should do long term to hide this away somewhere in framework code so that users don't accidentally call it or get confused by it.

The use of `FIXME` is too strong; I'll change it to a `TODO` - it's low priority as it does work, it's just ugly.

---
For where to put the synchronized, I think it needs to be here.

tbouron:
Thanks @aledsage. Tests passed, merging this now

PR #919, opened 2017-12-19
Stop dynamiccluster.zone.enable inheritance

drigodwin:
This is required so child clusters don't inherit this from their parent as it doesn't work.


aledsage:
LGTM; would be nice to have a unit test as well, but change is trivial so happy to merge.

aledsage:
I tested this with an app something like:
```
location:
  multi:
    targets:
      - jclouds:aws-ec2:us-east-1a
      - jclouds:aws-ec2:us-east-1b
services:
- type: org.apache.brooklyn.entity.group.DynamicCluster
  brooklyn.config:
    dynamiccluster.zone.enable: true
    initialSize: 2
    memberSpec:
      $brooklyn:entitySpec:
        type: org.apache.brooklyn.entity.group.DynamicCluster
        brooklyn.config:
          initialSize: 1
          memberSpec:
            $brooklyn:entitySpec:
              type: server
```
It failed before this change, and works after this change.

PR #921, opened 2017-12-19
Adds test for enableAvailabilityZones not re-inherited

aledsage:
Prior to @drigodwin's fix in https://github.com/apache/brooklyn-server/pull/919, this test would have failed. The child clusters are each passed a sub-location (i.e. a `SimulatedLocation`). However, because they also re-inherited the ENABLE_AVAILABILITY_ZONES=true, it tried to get the availability zone extension for the location it was given. This failed because `SimulatedLocation` does not have that extension.

With the fix in #919, the config key is not re-inherited, so the child `DynamicCluster`s do nottry to lookup the availability zone extension, and instead just create their children in the given location.


tbouron:
Thanks for the detailed description @aledsage.

LGTM, merging

PR #923, opened 2017-12-21
DynamicCluster: resize to maxSize, before throwing InsufficientCapacity

aledsage:
Building on PR #916 and the conversation in https://github.com/apache/brooklyn-server/pull/916#discussion_r157902790, this changes the behaviour of resize so that it will grow to the maxSize.

For example, imaging the extreme case where we have `initialSize=1, maxSize=10`, and an auto-scaler policy decides that the cluster should be resized to 11. Instead of throwing an exception, the behaviour now will be that the first call to `resize(11)` will cause us to resize to the max of 10 (logging a warning). A subsequent call to `resize(11)` by the policy will throw the `InsufficientCapacityException`.

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/entity/group/DynamicClusterImpl.java
+++ b/core/src/main/java/org/apache/brooklyn/entity/group/DynamicClusterImpl.java
@@ -797,10 +797,16 @@ public class DynamicClusterImpl extends AbstractGroupImpl implements DynamicClus
         Preconditions.checkArgument(delta > 0, "Must call grow with positive delta.");
         Integer maxSize = config().get(MAX_SIZE);
         if (maxSize != null) {
-            final int desiredSize = getCurrentSize() + delta;
-            if (desiredSize > maxSize) {
+            Integer currentSize = getCurrentSize();
+            final int desiredSize = currentSize + delta;
+            if (currentSize >= maxSize) {
                 throw new Resizable.InsufficientCapacityException(
-                        "Desired cluster size " + desiredSize + " exceeds maximum size of " + maxSize);
+                        "Current cluster size " + currentSize + " already at maximum permitted");
+            } else if (desiredSize > maxSize) {
+                int allowedDelta = (maxSize - currentSize);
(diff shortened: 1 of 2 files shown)


tbouron:
@aledsage Code-wise, this looks fine.

However, as I said in https://github.com/apache/brooklyn-server/pull/916#discussion_r158266008, I don't think this behaviour is the right thing to do. It will be hard to report back to the user what really happens. If one has `initialSize=5, maxSize=10` and resize by a delta of 10, how a user can know that it didn't resize to 15 because of the `maxSize` (the task won't tell this info explicitly) ? What highlight would be written if this was triggered by the `AutoScaller` policy?

Fail fast is not as smart as this but at least, it will make sense to the user. That's why I was suggesting an effector's parameter which would override the behaviour to treat bounds as suggestions, not requirements. It would be a conscious decision from a user/policy rather than an arbitrary decision from our side.

aledsage:
@tbouron The `DynamicCluster.resize(int)` effector returns the new size. In your example of `resize(15)`, it would now resize to 10 and return the value 10. In the case of calling `DynamicCluster.resizeByDelta(...)` it will return the `Collection<Entity>` that was created, so one could look at the size returned.

This is already the way it behaves if the location just has insufficient capacity. For example, if using a bring-your-own-nodes with limited capacity, it will add as many as it can and return the new size (the other members will be created and will fail, either being quarantined or deleted depending on the configuration). The same applies if your cloud has limited capacity (e.g. aws service-limits for the number of instances in a region).

I think that model is easier to program against (or at least it requires fewer changes to all the places that already use `resize()`), compared to requiring extra parameters/configuration.

I think as it stands it is hard to use - e.g. it won't play nicely with our existing auto-scaler policy. I think the change in this PR is the simplest way to make it behave closer to what people/policies desire (i.e. if one says "resize" then resize as much as you can).

ahgittin:
makes sense to resize as much as we can, but 100% agree re giving more feedback to user.

a highlight might be nice but not sure how to clear it.  the right thing feels like being able to report log messages that happen during execution.  i think accept the pain for now until we can collect and include log messages for tasks.

PR #924, opened 2017-12-21
PR #918 addendum: UsernamePassword implements legacy interface

aledsage:


aledsage:
Merging this now as it fixes broken downstream projects

PR #925, opened 2017-12-22
Fix reporting of master ID when web API used before persistence is ready

sjcorbett:
ServerResource#getHighAvailabilityPlaneStates obtains a new plane sync
record when the master node ID in the first one it obtained is null.
This generally only happens when the web-console tries to obtain the
HA status before the server has started up.

Fixes https://issues.apache.org/jira/browse/BROOKLYN-167.


ahgittin:
LGTM.  Client-side fix in https://github.com/apache/brooklyn-ui/pull/51 which could also be merged but this change is better.

PR #927, opened 2018-01-05
Bundle type endpoint

ahgittin:
It was noted that if the same type is defined in two different bundles, the API does not let you distinguish between the two.  This adds `/catalog/bundles/{symbolicName}/{version}/types/{typeSymbolicName}/{typeVersion}` to resolve that.  (It also adds the two immediate ancestor paths, with `../types` listing types in the bundle and `../types/{typeSymbolicName}` giving `TypeDetail` assuming `typeVersion==version`).

It also adds a bunch of tests for behaviour when adding the same type ID:

```
testAddSameTypeTwiceInSameBundle_SilentlyDeduped()
testAddSameTypeTwiceInSameBundleDifferentDisplayName_LastOneWins()
testAddSameTypeTwiceInSameBundleDifferentDefinition_Disallowed()
testAddSameTypeTwiceInDifferentBundleDifferentDefinition_Disallowed()
testAddSameTypeTwiceInDifferentBundleSameDefinition_AllowedAndApiMakesTheDifferentOnesClear()
```

the final test above fails after the first commit but is fixed by the second commit here, by changing the type's `self` link to use the unambiguous `bundles/.../types/...` endpoint.

### DIFF
--- a/rest/rest-api/src/main/java/org/apache/brooklyn/rest/api/BundleApi.java
+++ b/rest/rest-api/src/main/java/org/apache/brooklyn/rest/api/BundleApi.java
@@ -35,6 +35,8 @@ import javax.ws.rs.core.Response;
 
 import org.apache.brooklyn.rest.domain.BundleInstallationRestResult;
 import org.apache.brooklyn.rest.domain.BundleSummary;
+import org.apache.brooklyn.rest.domain.TypeDetail;
+import org.apache.brooklyn.rest.domain.TypeSummary;
 
 import io.swagger.annotations.Api;
 import io.swagger.annotations.ApiOperation;
@@ -88,6 +90,51 @@ public interface BundleApi {
         @PathParam("version")
         String version);
 
+
+    @Path("/{symbolicName}/{version}/types")
+    @GET
+    @ApiOperation(value = "Get all types declared in a given bundle", 
+            response = TypeDetail.class)
+    public List<TypeSummary> getTypes(
+        @ApiParam(name = "symbolicName", value = "Bundle name to query", required = true)
+        @PathParam("symbolicName")
+        String symbolicName,
+        @ApiParam(name = "version", value = "Version of bundle and of type to query", required = true)
+        @PathParam("version")
+        String version);
+    
+    @Path("/{symbolicName}/{version}/types/{typeSymbolicName}")
+    @GET
+    @ApiOperation(value = "Get detail on a given type in a given bundle", 
+            response = TypeDetail.class)
+    public TypeDetail getType(
+        @ApiParam(name = "symbolicName", value = "Bundle name to query", required = true)
+        @PathParam("symbolicName")
+        String symbolicName,
+        @ApiParam(name = "version", value = "Version of bundle and of type to query", required = true)
+        @PathParam("version")
+        String version,
+        @ApiParam(name = "typeSymbolicName", value = "Type name to query", required = true)
+        @PathParam("typeSymbolicName")
+        String typeSymbolicName);
+    
+    @Path("/{symbolicName}/{version}/types/{typeSymbolicName}/{typeVersion}")
+    @GET
+    @ApiOperation(value = "Get detail on a given type and version in a bundle (special method for unusual cases where type has different version)", 
+            response = TypeDetail.class)
+    public TypeDetail getTypeExplicitVersion(
+        @ApiParam(name = "symbolicName", value = "Bundle name to query", required = true)
+        @PathParam("symbolicName")
+        String symbolicName,
+        @ApiParam(name = "version", value = "Bundle version to query", required = true)
+        @PathParam("version")
+        String version,
+        @ApiParam(name = "typeSymbolicName", value = "Type name to query", required = true)
+        @PathParam("typeSymbolicName")
+        String typeSymbolicName,
+        @ApiParam(name = "typeVersion", value = "Version to query (if different to bundle version)", required = true)
+        @PathParam("typeVersion")
+        String typeVersion);
     
     @Path("/{symbolicName}/{version}")
     @DELETE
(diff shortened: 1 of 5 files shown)


rest/rest-api/src/main/java/org/apache/brooklyn/rest/api/BundleApi.java
```diff
@@ -88,6 +90,51 @@ public BundleSummary detail(
...
+    public TypeDetail getType(
+        @ApiParam(name = "symbolicName", value = "Bundle name to query", required = true)
+        @PathParam("symbolicName")
+        String symbolicName,
+        @ApiParam(name = "version", value = "Version of bundle and of type to query", required = true)
+        @PathParam("version")
+        String version,
+        @ApiParam(name = "typeSymbolicName", value = "Type name to query", required = true)
+        @PathParam("typeSymbolicName")
+        String typeSymbolicName);
```
tbouron: I'm not sure to understand the difference between this endpoint and the one below (i.e. `/{symbolicName}/{version}/types/{typeSymbolicName}/{typeVersion}`).

It seems to that that it will be the same as `"/{symbolicName}/{version}/types/{typeSymbolicName}/latest` so probably not needed?

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/resources/BundleResource.java
```diff
@@ -107,6 +113,42 @@ protected ManagedBundle lookup(String symbolicName, String version) {
         }
         return b;
     }
+    
+
+    @Override
+    public List<TypeSummary> getTypes(String symbolicName, String version) {
+        ManagedBundle b = lookup(symbolicName, version);
+        return TypeTransformer.bundleDetails(brooklyn(), b, ui.getBaseUriBuilder(), mgmt()).getTypes();
```
tbouron: Don't we need an entitlement check on the bundle here?

rest/rest-resources/src/test/java/org/apache/brooklyn/rest/resources/BundleAndTypeResourcesTest.java
```diff
@@ -1095,4 +1110,268 @@ private boolean checkTraits(boolean currentExpectedToBeWorking) {
...
+        Response getBundle = client().path("/catalog/bundles/" + symbolicName + "/" + TEST_VERSION).get();
+        assertEquals(getBundle.getStatus(), 404);
+    }
+    
+    // TODO might in future want to allow this if the user adding the type cannot see the other type due to entitlements;
+    // means however there might be another user who _can_ see the two different types 
+    private final static boolean DISALLOW_DIFFERENCES_IN_SAME_TYPE_ID_FROM_DIFFERENT_BUNDLES = true;
+    
+    @Test
+    public void testAddSameTypeTwiceInDifferentBundleDifferentDefinition_Disallowed() throws Exception {
```
tbouron: I don't understand why this case is not allowed. Could you elaborate on this please @ahgittin ?

rest/rest-api/src/main/java/org/apache/brooklyn/rest/api/BundleApi.java
ahgittin: it isn't `latest` - it is for case where type version = bundle version.  i think as that should be the normal case we should have an endpoint which doesn't require repeating version.

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/resources/BundleResource.java
ahgittin: `lookup` does the bundle entitlement check doesn't it?

rest/rest-resources/src/test/java/org/apache/brooklyn/rest/resources/BundleAndTypeResourcesTest.java
ahgittin: it might be allowed in future, it certainly can be behind the scenes, but it can cause oddness as it is then ambiguous what `type:version` would mean if there are two different implementations.  so there is strong argument to discourage/prevent it.

however in a multi-tenant world different users might not be allowed to see other bundles, so user 1 and user 2 might add the same `type:version` where they aren't allowed to see the other (and in fact shouldn't be allowed to know that someone else has defined that `type:version`), so in that case we _would_ allow this.  (we'd weaken restrictions to allow it, maybe e.g. user 2 can add `type:version` so long as he can't see any differing `type:version`.)  _but_ it might mean that a user 3 who is allowed to see bundles from both user 1 and user 2 is then able to see differing definition, which is not ideal and potentially confusing but i think least bad option.  cc @aledsage ?

(NB longer term discussion, not part of this PR which just clarifies and tests current behaviour in this regard)

ahgittin:
+1 to supporting `latest` on bundle version -- i'll add that.  i think all other comments addressed.  `latest` bundle support could be new PR if you're happy to merge this now @tbouron .

rest/rest-api/src/main/java/org/apache/brooklyn/rest/api/BundleApi.java
tbouron: Hum I get your PoV. I believe it's better to specify everything but why not, and it's using the same codepath so I'm fine with that

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/resources/BundleResource.java
tbouron: Indeed, I did check that but forgot about it, ignore me then.

rest/rest-resources/src/test/java/org/apache/brooklyn/rest/resources/BundleAndTypeResourcesTest.java
tbouron: Ah got it, makes sense

tbouron:
Yep, the `latest` for bundles is not in the scope of this PR so I'm happy if there is another one for this.

Merging

tbouron:
Actually @ahgittin, it seems the new endpoint is now available: I get a 404 and it does not appear on the swagger page. Not sure what's going on here

ahgittin:
v strange.  are you sure it was built correctly @tbouron ?  it maybe had a `mvn clean install` pull down a slightly-old snapshot jar?  adding a new `*Resource` class needs updates in a few places but adding methods to the `*Api` classes is usually straightforward.

tbouron:
I did but looks like I messed up my env somehow. Everything is fine now, sorry for the noise @ahgittin

PR #928, opened 2018-01-08
Allow deprecation as the catalog item level

tbouron:
In YAML, one can mark a catalog item as `deprecated` but it works only when the flag it set as the `brooklyn.catalog` level.

This allows each catalog item to use the deprecated flag as one see fit. For example, I used it for the deprecated policies/enrichers we have within Brooklyn (see changes in bom files)


tbouron:
retest this please

aledsage:
@tbouron there's a minor conflict with #929 which I just merged - could you resolve that and then happy for this to be merged.

tbouron:
@aledsage Fixed conflicts and addressed your comment

PR #929, opened 2018-01-11
Fix icons in catalog, and a few misc

ahgittin:
icons set on catalog items previously could only come in metadata, unlike name and description which could be in metadata _or_ in the `item` yaml.  as setting an `iconUrl` in the item yaml is valid (like `description` and display `name`), we should also special case check this when preparing catalog item metadata, treating all three the same.  see the most recent commit.

this fixes a bug where the icons in `policy/src/main/resource/catalog.bom` weren't getting picked up in the catalog (but they were being seen when instances were deployed).

two tiny other commits which i've noticed in the course of debugging also

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/entity/group/DynamicFabric.java
+++ b/core/src/main/java/org/apache/brooklyn/entity/group/DynamicFabric.java
@@ -55,14 +55,14 @@ public interface DynamicFabric extends AbstractGroup, Startable, Fabric {
     @SetFromFlag("memberSpec")
     ConfigKey<EntitySpec<?>> MEMBER_SPEC = ConfigKeys.newConfigKey(
             new TypeToken<EntitySpec<?>>() {}, 
-            "dynamiccfabric.memberspec", 
+            "dynamicfabric.memberspec", 
             "Entity spec for creating new members (one per location)", 
             null);
 
     @SetFromFlag("firstMemberSpec")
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/apache/brooklyn/entity/group/DynamicFabric.java
```diff
@@ -55,14 +55,14 @@
     @SetFromFlag("memberSpec")
     ConfigKey<EntitySpec<?>> MEMBER_SPEC = ConfigKeys.newConfigKey(
             new TypeToken<EntitySpec<?>>() {}, 
-            "dynamiccfabric.memberspec", 
+            "dynamicfabric.memberspec",
```
aledsage: Strictly speaking, we should use the `deprecatedNames` to leave support for the old badly spelt name. However, it's unlikely that anyone is using the misspelt name so no strong feelings.

As an example of the deprecation usage:
```
    public static final ConfigKey<Boolean> START_LATCH = ConfigKeys.builder(Boolean.class)
            .name("latch.start")
            .deprecatedNames("start.latch")
            .description("Latch for blocking start (done post-provisioning for software processes); if non-null will wait for this to resolve (normal use is with '$brooklyn:attributeWhenReady')")
            .build();
```

aledsage:
LGTM; merging.

PR #930, opened 2018-01-11
WithMutexes cleanup

alasdairhodge:
Modifies signature of `WithMutexes.acquireMutex()` to throw unchecked `RuntimeInterruptedException` to ease the burden on callers.

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/location/ssh/SshMachineLocation.java
+++ b/core/src/main/java/org/apache/brooklyn/location/ssh/SshMachineLocation.java
@@ -84,7 +84,6 @@ import org.apache.brooklyn.util.core.task.ScheduledTask;
 import org.apache.brooklyn.util.core.task.Tasks;
 import org.apache.brooklyn.util.core.task.system.internal.ExecWithLoggingHelpers;
 import org.apache.brooklyn.util.exceptions.Exceptions;
-import org.apache.brooklyn.util.exceptions.RuntimeInterruptedException;
 import org.apache.brooklyn.util.guava.KeyTransformingLoadingCache.KeyTransformingSameTypeLoadingCache;
 import org.apache.brooklyn.util.guava.Maybe;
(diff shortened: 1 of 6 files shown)


aledsage:
I agree this is an improvement, and is worth doing.

It's worth noting that it is not backwards compatible, if you consider the `org.apache.brooklyn.util.core.mutex.MutexSupport` to be part of our "public" api: code that has a `catch (InterruptedException e)` will no longer compile. It's a simple change in downstream code to either remove the catch block, or to instead catch the `RuntimeInterruptedException`.

I think it's acceptable to do this now (particularly before we release a 1.0.0).

If no-one objects, I'll merge once jenkins confirms the tests pass.

core/src/main/java/org/apache/brooklyn/location/ssh/SshMachineLocation.java
```diff
@@ -1016,12 +1015,8 @@ public String findPassword() {
     /** @deprecated since 1.0.0; mutex-related methods are now accessible via {@link #mutexes()} */
     @Override
     @Deprecated
-    public void acquireMutex(String mutexId, String description) throws RuntimeInterruptedException {
-        try {
-            mutexes().acquireMutex(mutexId, description);
-        } catch (InterruptedException ie) {
-            throw new RuntimeInterruptedException("Interrupted waiting for mutex: " + mutexId, ie);
-        }
+    public void acquireMutex(String mutexId, String description) {
```
geomacy: I would leave the `throws` in place, and change the `MutexSupport.acquireMutex` to also declare `throws RuntimeInterruptedException`.  Because it's runtime you can ignore it as a caller if you want, but the tooling support in IDEs will get some help for anyone who does want to catch it.

alasdairhodge: Effective Java (Item 62) recommends against this for unchecked exceptions, although I've no strong feelings.

PR #932, opened 2018-01-12
collect supertypes for registered types

ahgittin:
this populates the `supertypes` field of `RegisteredType` instances better, now including all java supertypes (and for specs only including what it's a spec for, ie `Entity`, now excluding the `EntitySpec` class).

this is WIP as i'd also like to include declared `RegisteredType` supers.

this is slightly messy on the API however.  specifically consider a YAML type `b` declaring its super as a java `type: a.A`, and a YAML type `c` declaring super `type: b`.  in java, the supers of `c` would then be `[ RegisteredType[b], Class[a.A] ]` (plus any supers of `a.A`).  what should this look like in REST?

currently we return a list of strings for all supers so the json response in `TypeSummary` will include `supertypes: [ b:0.0.0_SNAPSHOT, a.A ]`.

other options are available to better distinguish registered types (yaml types) from java types.  we could have eg `supertypesJava: [ a.A ]` and `supertypesYaml: [ b:0.0.0_SNAPSHOT ]`.  or we could have objects eg `supertypes: [ { kind: registered-type, name: b:0.0.0_SNAPSHOT }, { kind: java, name: a.A } ]`.

i think it's best to leave it as is and rely on the `:` character to distinguish yaml supertypes.  otherwise we're breaking the API and making the object seem more complex.  but i want to flag this bit of complexity in the semantics of the json `supertypes` field.

(opening this PR now for early review of code and to see whether anyone has strong feelings on the above.)

### DIFF
--- a/api/src/main/java/org/apache/brooklyn/api/catalog/BrooklynCatalog.java
+++ b/api/src/main/java/org/apache/brooklyn/api/catalog/BrooklynCatalog.java
@@ -63,7 +63,7 @@ public interface BrooklynCatalog {
     @Deprecated
     void deleteCatalogItem(String symbolicName, String version);
 
-    /** @return Deletes the item with the given {@link CatalogItem#getSymbolicName()
+    /** Deletes the item with the given {@link CatalogItem#getSymbolicName()
      * symbolicName} and version
      * @throws NoSuchElementException if not found 
      * @deprecated since introduced in 1.0.0, only used for transitioning */
--- a/api/src/main/java/org/apache/brooklyn/api/catalog/BrooklynCatalog.java
+++ b/api/src/main/java/org/apache/brooklyn/api/catalog/BrooklynCatalog.java
@@ -63,6 +63,13 @@ public interface BrooklynCatalog {
     @Deprecated
     void deleteCatalogItem(String symbolicName, String version);
 
+    /** @return Deletes the item with the given {@link CatalogItem#getSymbolicName()
+     * symbolicName} and version
+     * @throws NoSuchElementException if not found 
+     * @deprecated since introduced in 1.0.0, only used for transitioning */
+    @Deprecated
+    void deleteCatalogItem(String symbolicName, String version, boolean alsoCheckTypeRegistry, boolean failIfNotFound);
+    
     /** variant of {@link #getCatalogItem(String, String)} which checks (and casts) type for convenience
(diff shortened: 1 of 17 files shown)


ahgittin:
ps - `testDisabledApplicationCatalog()` and possibly others failing here because the `RegisteredType` that is created for the disabled catalog item isn't being updated; fix to come soon

ahgittin:
Removing WIP as this passes and the TODO items re including non-java registered types in list of supers is not essential and can be done as follow-on work.

This does not change the API.

ahgittin:
@aledsage right on all counts, i was sloppy with the deprecated stuff.  can't wait till we can cut all the old catalog and ideally switch to YOML too.

aledsage:
Test failure is unrelated, presumably non-deterministic, and should be investigated separately (`VanillaSoftwareProcessYamlTest.testDisableSshPolling`):
```
java.lang.AssertionError: failed succeeds-eventually, 75 attempts, 30000ms elapsed: AssertionError: actual=[starting, running]; expected=[created, starting, running]: lists don't have the same size expected [3] but found [2]
	at org.apache.brooklyn.camp.brooklyn.VanillaSoftwareProcessYamlTest.assertEventsEqualEventually(VanillaSoftwareProcessYamlTest.java:261)
	at org.apache.brooklyn.camp.brooklyn.VanillaSoftwareProcessYamlTest.testDisableSshPolling(VanillaSoftwareProcessYamlTest.java:173)
Caused by: java.lang.AssertionError: actual=[starting, running]; expected=[created, starting, running]: lists don't have the same size expected [3] but found [2]
	at org.apache.brooklyn.camp.brooklyn.VanillaSoftwareProcessYamlTest.assertIterablesEqual(VanillaSoftwareProcessYamlTest.java:278)
	at org.apache.brooklyn.camp.brooklyn.VanillaSoftwareProcessYamlTest.access$100(VanillaSoftwareProcessYamlTest.java:64)
	at org.apache.brooklyn.camp.brooklyn.VanillaSoftwareProcessYamlTest.assertEventsEqualEventually(VanillaSoftwareProcessYamlTest.java:261)
	at org.apache.brooklyn.camp.brooklyn.VanillaSoftwareProcessYamlTest.testDisableSshPolling(VanillaSoftwareProcessYamlTest.java:173)
```

PR #933, opened 2018-01-15
Log all REST api calls (when running as OSGi)

aledsage:
When we switched to OSGi (and thus jax-rs), we no longer logged the REST api calls being made to Brooklyn (see https://github.com/apache/brooklyn-server/blob/master/rest/rest-server/src/main/java/org/apache/brooklyn/rest/filter/LoggingFilter.java for the old way we logged these).

This PR adds a similar logging class for jax-rs.

### DIFF
--- a/rest/rest-resources/src/main/java/org/apache/brooklyn/rest/filter/LoggingResourceFilter.java
+++ b/rest/rest-resources/src/main/java/org/apache/brooklyn/rest/filter/LoggingResourceFilter.java
@@ -1,0 +1,235 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.brooklyn.rest.filter;
+
+import java.io.IOException;
+import java.security.Principal;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.WeakHashMap;
+
+import javax.annotation.Priority;
+import javax.servlet.http.HttpServletRequest;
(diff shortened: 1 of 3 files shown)


andreaturli:
merged at [apache/brooklyn-server master](https://git-wip-us.apache.org/repos/asf?p=brooklyn-server.git;a=commit;h=49bd6c7858aecfc44384dc628dac2bc81e84a7b9)

aledsage:
The diff of file rest/rest-resources/src/main/java/https://github.com/apache/brooklyn-server/commit/49bd6c7858aecfc44384dc628dac2bc81e84a7b9 has been merged to master.

However, this branch (and thus PR) was slightly changed compared to what was merged. The diff compared to master of org/apache/brooklyn/rest/filter/LoggingResourceFilter.java is:
```
76c76
<     private static final Set<String> CENSORED_HEADERS = ImmutableSet.of("Authorization");
---
>     private static final Set<String> CENSORED_HEADERS = ImmutableSet.of("Authorization", "Cookie");
159c159
<         boolean includeHeaders = (responseContext.getStatus() / 100 == 5) || LOG.isDebugEnabled();
---
>         boolean includeHeaders = (responseContext.getStatus() / 100 == 5) || LOG.isTraceEnabled();
```

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/filter/LoggingResourceFilter.java
```diff
@@ -0,0 +1,235 @@
...
+        String method = requestContext.getMethod();
+        String path = requestContext.getUriInfo().getPath();
+        requestContext.getSecurityContext();
+        
+        SecurityContext securityContext = requestContext.getSecurityContext();
+        Principal userPrincipal = (securityContext != null) ? requestContext.getSecurityContext().getUserPrincipal() : null;
+        String userName = (userPrincipal != null) ? userPrincipal.getName() : "<no-user>";
+        String remoteAddr = servletRequest.getRemoteAddr();
+        
+        StringBuilder message = new StringBuilder("Request received: ")
```
geomacy: Would it be worth adding a correlation value here and in the response, so the two can be easily related to each other?  Maybe just the `hashCode()` of the `servletRequest` object?

andreaturli: @geomacy @aledsage would it be useful/possible to use this library http://www.baeldung.com/spring-cloud-sleuth-single-application or at least re-use some design ideas?

geomacy: Think that looks like it's pretty closely tied to Spring.  I'm guessing it works in with Spring's request handling. Certainly some of the examples on that page use Spring wiring.

m4rkmckenna:
@aledsage Sorry im late to the party ... CXF has this functionality built in off the top of my head 

`Logging[In,Out]Inteceptor` just needs to be added to the cxf inteceptor chain

PR #939, opened 2018-01-23
BROOKLYN-576: decrease amount of debug logging

aledsage:
See https://issues.apache.org/jira/browse/BROOKLYN-576


tbouron:
LGTM, thanks @aledsage. Merging

PR #940, opened 2018-01-23
Add mapped diagnostic context with entity ids to logs

geomacy:
Adds a mapped diagnostic context to (some) log messages.

Basically if a task has a context entity this adds a field to log messages with both the task id and a list of the ids of the entity and its ancestors up to the application. For example in the output from the `ApplicationLoggingTest` below, with an application with a child and grandchild, we get

```
2018-02-02 18:00:39,351 INFO  xJTIhVuy-[ixxmgfzjos,m5it71jfiw,kw18pflnqp]     Hello from entity kw18pflnqp
2018-02-02 18:00:39,354 INFO  hqSMSJKm-[ixxmgfzjos,m5it71jfiw]   Hello from entity m5it71jfiw
2018-02-02 18:00:39,354 INFO  fCW0mP92-[ixxmgfzjos] Hello world
2018-02-02 18:00:39,365 INFO  - Stop task id is nxLdBOiG
2018-02-02 18:00:39,366 INFO  nxLdBOiG-[ixxmgfzjos] Goodbye cruel world
2018-02-02 18:00:39,367 INFO  hniq2zjE-[ixxmgfzjos,m5it71jfiw]   Goodbye from entity m5it71jfiw
2018-02-02 18:00:39,369 INFO  cFmqxDkJ-[ixxmgfzjos,m5it71jfiw,kw18pflnqp]     Goodbye from entity kw18pflnqp
```

This is basically a "proof of concept" to add entity ids to related logs.  Will remove the Do Not Merge after doing some more code review comment updates.

Also while the output from this change does add entity ids into a good number of log messages, there are other messages  where the ids don't get added.  These are probably going through paths in the code that aren't addressed here.  Nevertheless I think this is worth putting up for people to look at.

This also contains an update to `LogWatcher` so that it can make assertions on the _formatted_ output from the `PatternLayoutAppender`. 

Merge with https://github.com/apache/brooklyn-dist/pull/117


geomacy:
e.g. see attached debug log for creating a [three tier webapp](https://github.com/brooklyncentral/brooklyn-three-tier-webapp):
[brooklyn.debug.log.txt](https://github.com/apache/brooklyn-server/files/1657075/brooklyn.debug.log.txt)

ahgittin:
Cool.

It would seem better to me, rather than always setting a logging diagnostic context, to _deduce_ that information dynamically when we log, e.g. by calling to `Tasks.current()` when we log.  Is that possible?

Suggest logging without spaces and always including something in the "diagnostic" column to make it easier to strip fields.  And it's nice if columns mostly align, though with a varying-length list this won't happen.

Now turning our attention to _what_ we log:

Suggest including some task-hierarchy as well.  Full task-hierarchy and entity-hierarchy is rather much though I think.  So in light of the above, I'd go for a pattern like `TaskID@EntityID`, no ancestors, dropping `@EntityID` if no entity, and just doing `-` if no task.  So maybe (this is all pseudocode, I haven't compared with actual code):

    2018-01-23 11:52:20,118 INFO  -   Hello from something completely outside Brooklyn
    2018-01-23 11:52:20,118 INFO  abcd1111   REST call from alex@1.2.3.4 POST 0b to /v1/application/i6hk2jll5n/entity/z2scr3kcz/effector/sayHello
    2018-01-23 11:52:20,118 INFO  abcd1234@z2scr3kczj   Hello from the entity z2scr3kczj (effector sayHello, current task abcd1234)

And recording parent-child relationships at the time of child (entity and task) creation, IE something like:

    2018-01-23 11:52:20,118 DEBUG  abcd0000   Task abcd0000 created: REST call from alex@1.2.3.4 POST 780b to /v1/application
    2018-01-23 11:52:20,118 DEBUG  abcd0000   Creating new application i6hk2jll5n
    2018-01-23 11:52:20,118 DEBUG  abcd0001@i6hk2jll5n   Task abcd0001 created: initializing entity i6hk2jll5n, the root of an application
    2018-01-23 11:52:20,118 DEBUG  abcd0001@i6hk2jll5n   Blah blah setting up app i6hk2jll5n
    2018-01-23 11:52:20,118 DEBUG  abcd0001@i6hk2jll5n   Creating new entity z2scr3kczj "Tomcat Cluster" as child of i6hk2jll5n
    2018-01-23 11:52:20,118 DEBUG  abcd0002@i6hk2jll5n   Task abcd0002 created: initializing entity z2scr3kczj "Tomcat Cluster", child of i6hk2jll5n
    2018-01-23 11:52:20,118 DEBUG  abcd0002@z2scr3kczj   Blah blah blah setting up entity z2scr3kczj

General principle is:  we always know the entity and the task for any line, to filter on either of those, and we always know (earlier in the log) the location of the task and entity in the corresponding hierarchy.

However we might want exceptions for transient tasks, don't normally log them, as otherwise we'll log lots of task creation for tiny tasks.  But then we get the counter-counter-example of a transient task that logs a message, for that we _do_ want to know its hierarchy.  So I'd say we only want to log the `Task xxx created: <task title>` line for (1) non-transient tasks, and (2) those tasks that log something.  Doing (1) should be pretty easy.  For (2) it would be something like caching if we've logged the creation message, and if a task tries to log something else, but hasn't yet logged the creation message, we do that lazily (and probably include the start time in the log message as the log timestamp will be out of sequence).

Lastly FYI some of the places where we don't have an entity context I think are probably things we should fix.  EG SSH log messages I suspect are coming from an `SshMachineLocation` and I suspect _are_ using tasks nicely but are _not_ using the entity's `ExecutionContext` to launch meaning the context entity isn't being set.  The context entity might be available on an ancestor `Task` so that would be one option to "fix", at time of logging / inferring context, walk the task `submittedBy` hierarchy.  Maybe a better option would be to see whether the `submit` call done by the `SshMachineLocation` can set the entity context.

geomacy:
Great comments @ahgittin. Some thoughts in response:

> It would seem better to me, rather than always setting a logging diagnostic context, to deduce that information dynamically when we log, e.g. by calling to Tasks.current() when we log. Is that possible?

I'm sure it would be possible to do this by writing a custom version of the appenders we are currently, though we'd need to look into how best to do this (we're now using Apache log4j2 as the slf4j implementation in Karaf but still using Logback for tests).  On the other hand it seems like it would be quite an overhead to check `Tasks.current()` and work out task/entity ids for every log message, don't you think?  The beauty of the "context" approach is that you set it up once, high up in the stack. It doesn't matter too much if that's a slightly expensive operation, as  subsequent logging operations lower in the stack ignore it (it's only the output layout that takes any notice of the context).

> Suggest logging without spaces and always including something in the "diagnostic" column to make it easier to strip fields. And it's nice if columns mostly align, though with a varying-length list this won't happen.

Good ideas, I'll make those changes.

> Suggest including some task-hierarchy as well. Full task-hierarchy and entity-hierarchy is rather much though I think. So in light of the above, I'd go for a pattern like TaskID@EntityID, no ancestors, dropping @EntityID if no entity, and just doing - if no task. 

I think that's good; the one thing it loses is the ability to filter lines by the ids of any entity in the app, e.g. with a three tier app example with log `2018-01-23T17:52:27,032 [q62iiw0gil, pvvwi14b2v, zx9u0hwehk]` I can filter on `zx9u0hwehkz` to get logs for that Tomcat entity, on `pvvwi14b2v` to get logs for any entity in the web cluster, and on `q62iiw0gil` to see all logs for that given app.   How about adding task id but retaining the full stack of entity ids (which will rarely be terribly deep)?  In any case I'll add task ids.

For transient tasks I'll see if I can add in your point (1). For (2) it's more complicated, maybe that's something we can look into as a separate step. 

> Lastly FYI some of the places where we don't have an entity context I think are probably things we should fix.

That seems likely.  It would be good to identify such points and fix them up. I'll have a look at this (but again this is something that can be done in a separate commit).

geomacy:
hi @ahgittin I have taken the "do not merge" off this; check out the last two commits which add the task id to output and remove spaces.  I haven't yet included the logging of messages to track relationships  at the time of child (entity and task) creation, but I think there's enough in this at the moment to make it worth merging (keeping 'em small!), and I would aim to look at that additional stuff in another commit.

geomacy:
retest this please

geomacy:
Rebased against master

geomacy:
retest this please

geomacy:
retest this please, I'm pretty sure it's unrelated to this change

geomacy:
I have done some investigations of these build failures and I'm pretty sure they're not related to my change, e.g. this failure is the same as https://builds.apache.org/view/B/view/Brooklyn/job/brooklyn-server-master-docker/48/, whereas the previous failure on `SensorResourceTest` I have seen on a clean build of `master`.  These are things to look at but not related to this PR. Retest this please.

geomacy:
BTW @aledsage you may be interested in the changes to `LogWatcher` in here, https://github.com/apache/brooklyn-server/pull/940/files#diff-5c4439a5d92984cb96dfaf613a5af72c

geomacy:
Rebased against master

geomacy:
retest this please

geomacy:
Catch up to master, in particular get past https://github.com/apache/brooklyn-server/pull/946 which fixes the failing unit test

geomacy:
retest this please

tbouron:
Tested, works as advertised. All comments from @ahgittin have been addressed and I have to say, it's really cool to have this extra information @geomacy 🎉 👍 

Merging

PR #946, opened 2018-02-07
Remove unnecessary tearDown step with duplicate call of method

geomacy:
`tearDown` in EffectorResourceTest is copied from elsewhere but calls

        TestRendererHints.clearRegistry();

which is documented to be called from only one test in the suite (it's already called from SensorResourceTest).

This change prevents a race condition resulting in intermittent failure of SensorResourceTest


aledsage:
Thanks @geomacy - merging

PR #947, opened 2018-02-09
BROOKLYN-580 rebind machine entity feeds

aledsage:
Fixes https://issues.apache.org/jira/browse/BROOKLYN-580

### DIFF
--- a/software/base/src/main/java/org/apache/brooklyn/entity/software/base/SoftwareProcessImpl.java
+++ b/software/base/src/main/java/org/apache/brooklyn/entity/software/base/SoftwareProcessImpl.java
@@ -50,7 +50,6 @@ import org.apache.brooklyn.feed.function.FunctionPollConfig;
 import org.apache.brooklyn.location.jclouds.networking.NetworkingEffectors;
 import org.apache.brooklyn.location.ssh.SshMachineLocation;
 import org.apache.brooklyn.util.collections.MutableList;
-import org.apache.brooklyn.util.collections.MutableMap;
 import org.apache.brooklyn.util.collections.MutableSet;
 import org.apache.brooklyn.util.core.config.ConfigBag;
 import org.apache.brooklyn.util.core.task.BasicTask;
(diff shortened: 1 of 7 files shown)


geomacy:
retest this please, it's doesn't have the commit from https://github.com/apache/brooklyn-server/pull/946

software/base/src/main/java/org/apache/brooklyn/entity/software/base/SoftwareProcessImpl.java
```diff
@@ -385,33 +384,76 @@ protected void callRebindHooks() {
...
+                    if (!getManagementSupport().wasDeployed()) {
+                        if (rebindActive) {
+                            // We are still rebinding, and entity not yet managed - reschedule.
+                            Duration configuredMaxDelay = getConfig(MAXIMUM_REBIND_SENSOR_CONNECT_DELAY);
+                            if (configuredMaxDelay == null) configuredMaxDelay = Duration.millis(100);
+                            long delay = (long) (Math.random() * configuredMaxDelay.toMilliseconds());
+                            delay = Math.max(10, delay);
+                            LOG.debug("Entity {} not yet managed; re-scheduling connect-sensors on rebind in {}ms", SoftwareProcessImpl.this, delay);
+                            
+                            scheduleConnectSensorsOnRebind(Duration.millis(delay));
```
geomacy: I assume this is safe from the point of view of not risking an infinite loop of recursive calls? E.g. if somehow there were to be an exception in the entity  that meant `wasDeployed()` never became true? I take it this would be safe because sooner or later `rebindActive` will be false.  I see the comment you added below about this on `rebindTest()` as well. It seems improbable that there would be some combination of circumstances which would prevent `wasDeployed` becoming true but somehow also prevent `rebindActive` becoming false.

Nevertheless you might like to think about adding some failsafe "circuit breaker" capability here?

aledsage: Good question. I think it's safe because `rebindActive` is cleared in a finally block (https://github.com/apache/brooklyn-server/blob/master/core/src/main/java/org/apache/brooklyn/core/mgmt/rebind/RebindIteration.java#L302).

Were you thinking of a circuit breaker based on a max time (given there are already "circuit breakers" based on rebind having finished, etc)?

I don't really like setting max times because it's hard to know what a sensible limit is. For example, when I was load testing Brooklyn with 10s of thousands of entities then rebind took many minutes - but it did eventually work.

Therefore I'm going to merge this. But if you think a time-based circuit breaker is still worth it, or another kind of circuit breaker, then please do let me know.

PR #948, opened 2018-02-14
BROOKLYN-576: decrease debug logging for resolved DSL

aledsage:
We ran a test in QA that lasted 1 hour. There were 98,275 of these debug log messages (out of 226,712 total log messages at debug or above).


tbouron:
LGTM, thanks @aledsage, merging

PR #949, opened 2018-02-21
Improve JcloudsWinrmingLiveTest coverage

andreaturli:
- add Azure ARM live test
- improve error message for DefaultConnectivityResolver

---

When `DefaultConnectivityResolver.getReachableAddresses` now fails when no reachable address are found with this error message
```
2018-02-21 11:47:15,502 ERROR None of the addresses [168.63.120.198, 10.0.0.4] of node {id=brp4hwrr/brp4hwrr-9f7, providerId=/subscriptions/012e832d-7b27-4c30-9f21-22cdd9159d12/resourceGroups/brp4hwrr/providers/Microsoft.Compute/virtualMachines/brp4hwrr-9f7, name=brp4hwrr-9f7, location={scope=REGION, id=westeurope, description=West Europe, parent=azurecompute-arm, iso3166Codes=[NL]}, group=brp4hwrr, imageId=westeurope/MicrosoftWindowsServer/WindowsServer/2016-Datacenter, os={family=windows, version=2016-Datacenter, description=2016-Datacenter, is64Bit=true}, status=RUNNING[ProvisioningState/succeeded,PowerState/running], loginPort=5985, hostname=brp4hwrr-9f7, privateAddresses=[10.0.0.4], publicAddresses=[168.63.120.198], hardware={id=westeurope/Standard_B1s, providerId=Standard_B1s, name=Standard_B1s, location={scope=REGION, id=westeurope, description=West Europe, parent=azurecompute-arm, iso3166Codes=[NL]}, processors=[{cores=1.0, speed=2.0}], ram=1024, volumes=[{type=LOCAL, size=2048.0, bootDevice=false, durable=false}, {type=LOCAL, size=1047552.0, bootDevice=false, durable=false}], supportsImage=Predicates.alwaysTrue(), userMetadata={maxDataDiskCount=2}}, loginUser=jclouds, userMetadata={Name=brp4hwrr-idg4, brooklyn-user=andrea, jclouds_group=brp4hwrr}} are reachable in mode PREFER_PUBLIC
```
instead of the IllegalStateException 
```
jclouds did not return any IP addresses matching " + getNetworkMode() + " in " + location
```

### DIFF
--- a/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/DefaultConnectivityResolver.java
+++ b/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/DefaultConnectivityResolver.java
@@ -19,11 +19,14 @@
 
 package org.apache.brooklyn.location.jclouds;
 
+import java.util.Collections;
 import java.util.Iterator;
 import java.util.Map;
 
 import javax.annotation.Nullable;
 
+import com.google.common.base.Predicates;
+import com.google.common.collect.Lists;
 import org.apache.brooklyn.api.entity.Entity;
 import org.apache.brooklyn.api.entity.EntityInitializer;
 import org.apache.brooklyn.api.entity.EntityLocal;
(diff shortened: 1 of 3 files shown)


locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/DefaultConnectivityResolver.java
```diff
@@ -342,7 +343,12 @@ void publishNetworks(NodeMetadata node, Entity entity) {
     protected Iterable<HostAndPort> getReachableAddresses(NodeMetadata node, Predicate<? super HostAndPort> reachablePredicate, Duration timeout) {
         if (timeout == null) timeout = Duration.FIVE_MINUTES;
         Iterable<String> candidates = getResolvableAddressesWithMode(node);
-        return JcloudsUtil.getReachableAddresses(candidates, node.getLoginPort(), timeout, reachablePredicate);
+        Iterable<HostAndPort> result = JcloudsUtil.getReachableAddresses(candidates, node.getLoginPort(), timeout, reachablePredicate);
+        if (FluentIterable.from(result).filter(Predicates.equalTo(Optional.absent())).isEmpty()) {
+            LOG.error("None of the addresses {} of node {} are reachable in mode {}", new Object[]{Iterables.toString(candidates), node, getNetworkMode()});
+            throw new IllegalStateException("Could not find any reachable address for node: " + node + " in mode: " + getNetworkMode());
```
sjcorbett: Why not change the existing message on line 215?

andreaturli: I think it is to avoid 185 to be executed, but we could change 215 as well

sjcorbett: If you want to avoid that I think it would be cleaner to wrap 185-203 in an `if (!Iterables.isEmpty(managementCandidates))`

andreaturli: ok, although we need to check for `!FluentIterable.from(managementCandidates).filter(Predicates.equalTo(Optional.absent())).isEmpty()` as managementCandidates can't be empty

andreaturli:
@sjcorbett is it ok for you now?

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/DefaultConnectivityResolver.java
```diff
@@ -181,24 +183,27 @@ public ManagementAddressResolveResult resolve(
...
-        // Try each pair of address and credential until one succeeds.
-        if (shouldCheckCredentials() && options.pollForReachableAddresses()) {
-            for (HostAndPort hap : managementCandidates) {
-                for (LoginCredentials cred : credentialCandidates) {
-                    LOG.trace("Testing host={} with credential={}", hap, cred);
-                    if (checkCredential(location, hap, cred, config, options.isWindows())) {
-                        hapChoice = hap;
-                        credChoice = cred;
-                        break;
+        Iterable<LoginCredentials> credentialCandidates = Lists.newArrayList();
```
sjcorbett: You could use `Collections.emptyList()` or `ImmutableList.of()` here to save the redundant creation of an object. You could also assign it in an `else` block for the list so its use is a little clearer.

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/DefaultConnectivityResolver.java
```diff
@@ -212,8 +217,10 @@ public ManagementAddressResolveResult resolve(
         }
 
         if (hapChoice == null) {
-            throw new IllegalStateException("jclouds did not return any IP addresses matching " + getNetworkMode() + " in " + location);
+            LOG.error("None of the addresses of node {} are reachable in mode {}", new Object[]{node, getNetworkMode()});
+            throw new IllegalStateException("Could not find any reachable address for node: " + node + " in mode: " + getNetworkMode());
```
sjcorbett: Strictly speaking Brooklyn didn't look for a "reachable" address if ConnectivityResolverOptions.pollForReachableAddresses was false. How about "Could not determine management address for node..."?

sjcorbett:
@andreaturli looks good. I've left two minor comments that needn't stop this being merged if you wish.

PR #673, opened 2017-05-08
[WIP] [JCLOUDS 2.1.0] Upgrade to jclouds 2.1

neykov:
These are the changes I needed to implement in order to run Brooklyn against current jclouds master.
Leaving them around to be reused for the jclouds 2.1.0 upgrade once released.

PR #950, opened 2018-02-22
[BROOKLYN-583] Bump jclouds version to 2.1.0

andreaturli:
- upgrade to jclouds 2.1.0
- update feature.xml to use jclouds-karaf repository
- bump sshj version to 0.20.0


geomacy:
Looks good; have you cross checked it with https://github.com/apache/brooklyn-server/pull/673 in case there's anything there that is relevant?

andreaturli:
good point @geomacy 
I will check later and report here

andreaturli:
I think sshj.version needed to be updated too @geomacy, thanks for heads-up

PR #952, opened 2018-02-23
[BROOKLYN-584] Add support for recent jclouds's ContextLinking

andreaturli:
- add support for user-defined Guice modules
- add support for ContextLinking

### DIFF
--- a/locations/jclouds/pom.xml
+++ b/locations/jclouds/pom.xml
@@ -178,6 +178,13 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>${jclouds.groupId}.api</groupId>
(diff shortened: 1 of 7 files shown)


duncangrant:
@andreaturli Might be worth adding a docs change to https://brooklyn.apache.org/v/latest/locations/index.html#openstack?

andreaturli:
thanks @duncangrant and @geomacy 

Do you guys think the docs should be part of the same PR?

duncangrant:
@andreaturli No - shouldn't be part of this PR.  I just wrote the comment as a reminder.

geomacy:
@andreaturli @duncangrant  I agree it shouldn't be part of this PR, just think it's worth documenting

PR #953, opened 2018-04-11
Defers reading of config in SshCommandSensor

nakomis:
This allows DSL to be used in the command, env and execution dir as
a management context will then be available


core/src/main/java/org/apache/brooklyn/core/sensor/ssh/SshCommandSensor.java
```diff
@@ -136,7 +128,7 @@ public String get() {
                 .command(commandSupplier)
                 .suppressDuplicates(Boolean.TRUE.equals(suppressDuplicates))
                 .checkSuccess(SshValueFunctions.exitStatusEquals(0))
-                .onFailureOrException(Functions.constant((T) null))
+                .onFailureOrException(Functions.constant((T) "failed"))
```
aledsage: Why set this to "failed" rather than null?

The risk of "failed" is that it will trigger subsequent enrichers to try to do something with this value (e.g. if someone wrote an SshCommandSensor to get part of the url, then we might subsequently use "failed" in that url - the downstream enricher can't tell the difference between this being a valid value and being the value on failed.

If feels like the default of null is better.

Longer term, we could make this configurable so that the default value on failed can be explicitly set.

core/src/main/java/org/apache/brooklyn/core/sensor/ssh/SshCommandSensor.java
```diff
@@ -126,7 +118,7 @@ public String get() {
                 // Note that entity may be null during rebind (e.g. if this SshFeed is orphaned, with no associated entity):
                 // See https://issues.apache.org/jira/browse/BROOKLYN-568.
                 // We therefore guard against null in makeCommandExecutingInDirectory.
-                return makeCommandExecutingInDirectory(command, executionDir, entity);
+                return makeCommandExecutingInDirectory(params.get(SENSOR_COMMAND), params.get(SENSOR_EXECUTION_DIR), entity);
```
aledsage: I think we should do the `checkNotNull(params.get(SENSOR_COMMAND), ...)` in this `Supplier` method, and then pass it into the `makeCommandExecutingInDirectory`.

If this were null, then things can go very strange. For example, we might try to execute `null` when we construct the command via:
`finalCommand = "mkdir -p '"+execDir+"' && cd '"+execDir+"' && "+finalCommand`

camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/SshCommandSensorYamlTest.java
```diff
@@ -110,7 +110,46 @@ public void testSupressingDuplicates() throws Exception {
                 listener.assertEventCount(1);
             }});
     }
-    
+
+    // "Integration" because takes a couple of seconds
```
aledsage: I wonder why it takes a couple of seconds - it shouldn't be doing any real ssh (because it has a `RecordingsshTool` instead).

camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/SshCommandSensorYamlTest.java
```diff
@@ -110,7 +110,46 @@ public void testSupressingDuplicates() throws Exception {
...
+                "      suppressDuplicates: true",
+                "      period: 10ms",
+                "      onlyIfServiceUp: false");
+        waitForApplicationTasks(app);
+
+        VanillaSoftwareProcess entity = (VanillaSoftwareProcess) Iterables.getOnlyElement(app.getChildren());
+        EntityAsserts.assertAttributeEqualsEventually(entity, mySensor, "someValue");
+
+        entity.sensors().set(sourceSensor, "newValue");
+        EntityAsserts.assertAttributeEqualsEventually(entity, mySensor, "someValue");
```
aledsage: Should this not assert it equals `newValue`?

core/src/main/java/org/apache/brooklyn/core/sensor/ssh/SshCommandSensor.java
```diff
@@ -105,6 +97,7 @@ public void apply(final EntityLocal entity) {
                 Map<String, Object> env = MutableMap.copyOf(entity.getConfig(BrooklynConfigKeys.SHELL_ENVIRONMENT));
 
                 // Add the shell environment entries from our configuration
+                Map<String,Object> sensorEnv = params.get(SENSOR_SHELL_ENVIRONMENT);
```
aledsage: We should probably also do `EntityInitializers.resolve` when retrieving the env.

aledsage: Ignore this comment; later in the code it calls `Tasks.resolveDeepValue(env, ...)`

PR #957, opened 2018-04-20
SshCommandSensor.testDslWithSshSensor: convert to non-integration

aledsage:
See my comments in PR #953 

This now uses the `RecordingSshTool` to execute the ssh command sensor, so the test is quicker. If you run it 100 times, the first time still takes 2 seconds (presumably due to classloading etc), and then it takes about 200ms per test after that.

Also fixes the final assertion, that it is the `newValue` in the sensor.


nakomis:
LGTM

PR #959, opened 2018-04-27
Fix SshCommandSensor rebind

aledsage:
See the two separate commits:
* First fixes the big problem that rebind to historic persisted state is broken
* Second improves the `SshCommandSensor` so that we don't use anonymous inner classes (making our persisted state easier to support long-term)


tbouron:
This LGTM, with a lot of new tests. However, I'm not very familiar with the rebind code so it would be good if another set of eyes could also review this PR

PR #961, opened 2018-05-02
Add .m2 folder in the list of ignore folders for the RAT plugin.

tbouron:
This is required for Jenkins build as we now use a local m2 cache

### DIFF
--- a/parent/pom.xml
+++ b/parent/pom.xml
@@ -1104,6 +1104,8 @@
                       https://brooklyn.incubator.apache.org
                   -->
                   <exclude>docs/**</exclude>
+                  <!-- maven cache -->
+                  <exclude>**/.m2/**</exclude>
 
                 </excludes>
               </configuration>


parent/pom.xml
```diff
@@ -1104,6 +1104,8 @@
                       https://brooklyn.incubator.apache.org
                   -->
                   <exclude>docs/**</exclude>
+                  <!-- maven cache -->
+                  <exclude>**/.m2/**</exclude>
```
duncangrant: This doesn't feel quite right.  Why is the .m2 cache within the checked out repo?  Shouldn't it be a level up?

tbouron: When building using docker on Apache Jenkins, using the global m2 cache caused permission issues. I raised a ticket to INFRA and they told me that we should instead use a local m2 cache (https://issues.apache.org/jira/browse/INFRA-16417?focusedCommentId=16452458&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-16452458)

I update the Jenkins config to do that but now, the RAT plugin complains because the file don't have the Apache licence header, hence this change

geomacy:
@duncangrant what's the benefit of doing that?

PR #962, opened 2018-05-04
allow tags to be set on catalog items

ahgittin:
it has been supported in code and in rest api, but not possible to set from `catalog.bom` -- this fixes that

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/catalog/internal/BasicBrooklynCatalog.java
+++ b/core/src/main/java/org/apache/brooklyn/core/catalog/internal/BasicBrooklynCatalog.java
@@ -658,6 +658,9 @@ public class BasicBrooklynCatalog implements BrooklynCatalog {
                 .putIfNotNull("item", itemMetadata.get("item"))
                 .putIfNotNull("items", itemMetadata.get("items"))
                 .build();
+        // tags we treat specially to concatenate as a set (treating as config with merge might be cleaner)
+        catalogMetadata.put("tags", MutableSet.copyOf(getFirstAs(parentMetadata, Collection.class, "tags").orNull())
+            .putAll(getFirstAs(itemMetadataWithoutItemDef, Collection.class, "tags").orNull()) );
 
         // brooklyn.libraries we treat specially, to append the list, with the child's list preferred in classloading order
         // `libraries` is supported in some places as a legacy syntax; it should always be `brooklyn.libraries` for new apps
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/apache/brooklyn/core/catalog/internal/BasicBrooklynCatalog.java
```diff
@@ -934,15 +937,17 @@ private void collectCatalogItemsFromItemMetadataBlock(String sourceYaml, Managed
...
-            // just need TODO to make sure we delete previously-persisted things which now come through this path.)  
-            // NB: when everything is a bundle and we've removed all scanning then this can be the _only_ path
-            // and code can be massively simpler
-            // TODO allow these to be set in catalog.bom ?
+            // horrible API but basically `resultLegacyFormat==null` means use the new-style,
+            // adding from persisted bundles to type registry (which is not persisted)
+            // instead of old way which persisted catalog items (and not their bundles).
+            // this lets us deal with forward references, with a subsequent step to validate.
+
+            Set<Object> tags = MutableSet.of().putAll(getFirstAs(catalogMetadata, Collection.class, "tags").orNull());
```
tbouron: This was a `List<Object>` previously, why the change to a `Set<Object>`? Is that to remove duplicates? In this case, I don't think that is the right behaviour, if the user wants multiple time the same tags, we should allow that.
Also, surely this will cause backward compatibilities issues

rest/rest-api/src/main/java/org/apache/brooklyn/rest/domain/ConfigSummary.java
```diff
@@ -114,7 +116,8 @@ public ConfigSummary(ConfigKey<?> config, String label, Double priority, Boolean
                 : ImmutableList.<String>of();
         if (config.getType().isEnum()) {
             this.type = Enum.class.getName();
-            this.defaultValue = (config.getDefaultValue() == null) ? null : ((Enum<?>) config.getDefaultValue()).name();
+            this.defaultValue = config.getDefaultValue() instanceof Enum? ((Enum<?>) config.getDefaultValue()).name() :
```
tbouron: `config.getDefaultValue()` can be null. This could then throw a NPE

rest/rest-api/src/main/java/org/apache/brooklyn/rest/domain/ConfigSummary.java
```diff
@@ -114,7 +116,8 @@ public ConfigSummary(ConfigKey<?> config, String label, Double priority, Boolean
                 : ImmutableList.<String>of();
         if (config.getType().isEnum()) {
             this.type = Enum.class.getName();
-            this.defaultValue = (config.getDefaultValue() == null) ? null : ((Enum<?>) config.getDefaultValue()).name();
+            this.defaultValue = config.getDefaultValue() instanceof Enum? ((Enum<?>) config.getDefaultValue()).name() : 
+                Jsonya.convertToJsonPrimitive(config.getDefaultValue());
```
tbouron: Same a above. Also, not sure I understand why converting to JSON primitive here?

core/src/main/java/org/apache/brooklyn/core/catalog/internal/BasicBrooklynCatalog.java
ahgittin: Duplicate tags are never supported.  See e.g. `TagSupport.getTags()` which returns a `Set`..

The only reason it wasn't noticed here is that we didn't use tags much on these objects.

rest/rest-api/src/main/java/org/apache/brooklyn/rest/domain/ConfigSummary.java
ahgittin: if it's null the check returns false and it calls to `convertToJsonPrimitive` which accepts null.

rest/rest-api/src/main/java/org/apache/brooklyn/rest/domain/ConfigSummary.java
ahgittin: For consistency with code below, when it isn't an enum -- we always do that in case the default value is something complex (which it shouldn't be unless maybe a DSL but consistency with code below).

ahgittin:
Retest this please.

geomacy:
retest this please

PR #963, opened 2018-05-24
disable jclouds authorizePublicKey for extra ssh public key data

ahgittin:
we install this ourselves in addition so no need, and if we ask jclouds to do it, it seems to break sometimes (eg when keyPair is set).
there is a system property that can be set to revert for legacy behaviour.

### DIFF
--- a/locations/jclouds/src/test/java/org/apache/brooklyn/location/jclouds/provider/AbstractJcloudsLocationTest.java
+++ b/locations/jclouds/src/test/java/org/apache/brooklyn/location/jclouds/provider/AbstractJcloudsLocationTest.java
@@ -48,7 +48,7 @@ import com.google.common.collect.Maps;
 public abstract class AbstractJcloudsLocationTest {
     private static final Logger LOG = LoggerFactory.getLogger(AbstractJcloudsLocationTest.class);
 
-    private final String provider;
+    protected final String provider;
 
     protected JcloudsLocation loc;
     protected List<SshMachineLocation> machines = MutableList.of();
--- a/locations/jclouds/src/test/java/org/apache/brooklyn/location/jclouds/provider/AwsEc2LocationLiveTest.java
+++ b/locations/jclouds/src/test/java/org/apache/brooklyn/location/jclouds/provider/AwsEc2LocationLiveTest.java
@@ -18,6 +18,8 @@
  */
 package org.apache.brooklyn.location.jclouds.provider;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.testng.annotations.DataProvider;
 import org.testng.annotations.Test;
 
@@ -28,6 +30,8 @@ import org.testng.annotations.Test;
  */
 public class AwsEc2LocationLiveTest extends AbstractJcloudsLocationTest {
 
+    private static final Logger log = LoggerFactory.getLogger(AwsEc2LocationLiveTest.class);
+    
     private static final String PROVIDER = "aws-ec2";
     private static final String EUWEST_REGION_NAME = "eu-west-1";
     private static final String USEAST_REGION_NAME = "us-east-1";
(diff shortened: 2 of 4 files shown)


ahgittin:
note the code at https://github.com/apache/brooklyn-server/blob/master/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java#L1037 is unchanged, this is where we (brooklyn) install the supplied data to `authorized_keys`  ourselves, so disabling jclouds seems save

locations/jclouds/src/test/java/org/apache/brooklyn/location/jclouds/provider/AbstractJcloudsLocationTest.java
```diff
@@ -48,7 +48,7 @@
 public abstract class AbstractJcloudsLocationTest {
     private static final Logger LOG = LoggerFactory.getLogger(AbstractJcloudsLocationTest.class);
 
-    private final String provider;
+    protected final String provider;
```
tbouron: Why making this field `protected` instead of `private`. Doesn't seem to be used elsewhere, at least in this PR

locations/jclouds/src/test/java/org/apache/brooklyn/location/jclouds/provider/AwsEc2LocationLiveTest.java
```diff
@@ -28,6 +30,8 @@
  */
 public class AwsEc2LocationLiveTest extends AbstractJcloudsLocationTest {
 
+    private static final Logger log = LoggerFactory.getLogger(AwsEc2LocationLiveTest.class);
```
tbouron: The logger is not used here. Can be removed

locations/jclouds/src/test/java/org/apache/brooklyn/location/jclouds/provider/AwsEc2LocationLiveTest.java
```diff
@@ -68,4 +72,29 @@ public AwsEc2LocationLiveTest() {
...
+//            JcloudsLocationConfig.PUBLIC_KEY_FILE, "~/.ssh/aws-cloudera_rsa.pub",
+//            JcloudsLocationConfig.EXTRA_PUBLIC_KEY_DATA_TO_AUTH,
+//                aws-whirr
+//                "ssh-rsa AAAAB3NzaC1yc2EAAAABIwAAAQEAw9R7FG0pOpZ7MR+KYty+UzHerEtW9NVBJj+NmznT4zg57klqDAxitahe6cJpj8Nbt0Rmp9G4TZQzAWuoSH5ZUJpFpcVP75tMYWOP6ZymH7MZ5hXLjLHTicNyQ/EB9H2eXSK2xLnq/8oDnzKgnhIXL7tbcC7hOY9Yzu25UrO+xQDbzM3nuwlr38JJwo1fLsIiEVI3uutZW9ANZgcZg0USlFFvxGcAA2KZ322tqtQtP3YYE0IogYUjTSFj5xexFDzIcN5V2Z2tHYKW+Jl/jR98EAsq4By1L+whoX142NJGZsB1GKm4zZTh3vjfzpeGNmxrrHDGp2TOCGFJjk2seHqyyw== alex@almac.rocklynn.cognetics.org",
+//            JcloudsLocationConfig.IMAGE_ID, "us-east-1/ami-5492ba3c"
+//        ));
+//        
+//        log.info("Provisioned {} vm {}; checking if ssh'able", provider, machine);
+//        assertTrue(machine.isSshable());
+//    }
```
andreaturli: I took a first stab at it
```
    @Test(groups = "Live")
    public void testProvisionVmAndAuthPublicKey() {
        String regionName = USEAST_REGION_NAME;
        loc = (JcloudsLocation) mgmt().getLocationRegistry().getLocationManaged(provider + (regionName == null ? "" : ":" + regionName));

        EC2Api api = loc.getComputeService().getContext().unwrapApi(EC2Api.class);
        String primaryKeyName = "brooklyn-keypair-" + System.currentTimeMillis();
        try {
            Map<String, String> extra = SshKeys.generate();
            KeyPair primary = api.getKeyPairApiForRegion(regionName).get().createKeyPairInRegion(regionName, primaryKeyName);
            SshMachineLocation machine = obtainMachine(MutableMap.of(
                    JcloudsLocationConfig.KEY_PAIR, primary.getKeyName(),
                    JcloudsLocationConfig.LOGIN_USER_PRIVATE_KEY_DATA, primary.getKeyMaterial(),
                    JcloudsLocationConfig.EXTRA_PUBLIC_KEY_DATA_TO_AUTH, extra.get("public"),
                    JcloudsLocationConfig.IMAGE_ID, "us-east-1/ami-5492ba3c"
            ));

            log.info("Provisioned {} vm {}; checking if ssh'able", provider, machine);
            assertTrue(machine.isSshable());
        } finally {
            api.getKeyPairApiForRegion(USEAST_REGION_NAME).get().deleteKeyPairInRegion(USEAST_REGION_NAME, primaryKeyName);
        }
    }
```

tested on my aws account and seems fine. Are you happy with that @ahgittin ? if so, can you please enable it instead of the commented out test? thanks

locations/jclouds/src/test/java/org/apache/brooklyn/location/jclouds/provider/AwsEc2LocationLiveTest.java
ahgittin: also used in test

ahgittin:
test works perfectly -- very very nice @andreaturli !

merging

ahgittin:
Just had a report of a user who is disabling the ssh init options and using `extraSsh...` to configure the user's key (in azure, using a cloud api we think).  That set of options required the block that this PR disables.

the system property specified here should let them workaround but we will have issues longer-term if this is disabled.

a solution could be to change the check here, for the case where (1) the ssh command which installs the `extraSsh..` option is disabled and (2) this `extraSsh` field is supplied:

* check whether the cloud API supports the field; if not, fail (extraSshKey not supported when ssh disabled, for this cloud)
* if keyPair is supplied, also fail that this combo is not supported (max 1 keyPair or extraSsh when ssh disabled)
* otherwise run the code block this PR excludes

PR #966, opened 2018-05-25
Persistence efficient: unchanged and transient attributes

aledsage:
Most of the changes are to allow us to write tests, and the tests themselves. The actual important production-code changes is really just two if statements wrapping the calls to `onAttributeChanged(attribute)` and `onChanged(entity)`!

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/mgmt/persist/StoreObjectAccessorLocking.java
+++ b/core/src/main/java/org/apache/brooklyn/core/mgmt/persist/StoreObjectAccessorLocking.java
@@ -226,6 +226,21 @@ public class StoreObjectAccessorLocking implements PersistenceObjectStore.StoreO
     }
     
     @Override
+    public boolean isWriting() {
+        try {
+            boolean locked = lock.readLock().tryLock(0, TimeUnit.MILLISECONDS);
+            if (locked) {
+                lock.readLock().unlock();
+                return false;
+            } else {
+                return true;
+            }
+        } catch (InterruptedException e) {
+            throw Exceptions.propagate(e);
+        }
+    }
+    
+    @Override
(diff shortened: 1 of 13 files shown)


core/src/main/java/org/apache/brooklyn/core/mgmt/persist/StoreObjectAccessorLocking.java
```diff
@@ -225,6 +225,21 @@ public void waitForCurrentWrites(Duration timeout) throws InterruptedException,
         }
     }
     
+    @Override
+    public boolean isWriting() {
+        try {
+            boolean locked = lock.readLock().tryLock(0, TimeUnit.MILLISECONDS);
```
geomacy: Just a question - won't this fail to get the lock if another thread has a _read_ lock already - so `isWriting` will return `true` - even though nothing may have a write lock?  Even if that's true, does it matter?

geomacy:
p.s. also tested and confirm that `machine.` metrics don't get persisted

core/src/main/java/org/apache/brooklyn/core/mgmt/persist/StoreObjectAccessorLocking.java
aledsage: I don't think so - if another thread holds a read lock, then this thread will be able to acquire the read lock at the same time (sharing it with an unlimited number of readers). However, if another thread has a write lock then the attempt to immediately get a read lock will fail.

The `lock` field is a `ReentrantReadWriteLock` in "fair mode" for lock acquisition. That means the read-lock acquisition will also fail if there is a queued write-lock request.

aledsage:
Thanks @tbouron @geomacy - merging now.

PR #968, opened 2018-06-01
does a yaml parse and compares the hash of the result when checking plan equivalence

ahgittin:
means that the same plan entered twice but with different comments, spacing, quotations etc,
will not result in an error.  it only treats a plan as non-equivalent if their yaml is different
after a deserialize-then-serialize cycle.  previously a catalog push for an already-present
non-snapshot plan would fail unless the plan was identical up to a trim.  now it only fails if
the objects are different.

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/typereg/RegisteredTypes.java
+++ b/core/src/main/java/org/apache/brooklyn/core/typereg/RegisteredTypes.java
@@ -50,6 +50,7 @@ import org.apache.brooklyn.core.mgmt.BrooklynTags;
 import org.apache.brooklyn.core.mgmt.BrooklynTags.NamedStringTag;
 import org.apache.brooklyn.core.objs.BrooklynObjectInternal;
 import org.apache.brooklyn.core.typereg.JavaClassNameTypePlanTransformer.JavaClassNameTypeImplementationPlan;
+import org.apache.brooklyn.util.collections.Jsonya;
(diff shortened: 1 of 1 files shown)


duncangrant:
@ahgittin I've tested this with a few different bom files with small changes e.g. quotes, comments, etc.  Works well.  It's also nice that it's a small change. There are a couple of bits of I'm not clear on so I'll comment those individually. Also would it be worth adding some simple tests?

core/src/main/java/org/apache/brooklyn/core/typereg/RegisteredTypes.java
```diff
@@ -682,27 +683,65 @@ private static String tagForEquivalentPlan(String input) {
...
+     * this allows plans that are equivalent post-parse to be treated as equivalent.
+     * returns {@link Absent} if the input is not valid yaml.
+     */
+    private static Maybe<String> tagForEquivalentYamlPlan(String input) {
+        // plans may be trimmed by yaml parser so do that before checking equivalence
+        // it does mean a format change will be ignored
+        try {
+            Iterator<Object> plansI = Yamls.parseAll(input).iterator();
+            if (!plansI.hasNext()) {
+                return Maybe.absent("No data found");
```
duncangrant: If we returned the tagForEquivalentPlan(input) here could we then simplify the arePlansEquivalent code?

PR #971, opened 2018-06-08
DSL for $brooklyn:location()

aledsage:


aledsage:
(I've removed the commit "DslComponent: code tidy" because that broke existing tests!)

camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/dsl/methods/DslComponent.java
```diff
@@ -737,6 +720,130 @@ public String toString() {
...
+                    
+                    // this is always run in a new dedicated task (possibly a fake task if immediate), so no need to clear
+                    checkAndTagForRecursiveReference(targetEntity);
+
+                    int indexI = resolveIndex(immediate);
+                    
+                    // TODO Try repeatedly if no location(s)?
+                    Collection<Location> locations = getLocations(targetEntity);
+                    if (locations.size() < (indexI + 1)) {
+                        throw new IndexOutOfBoundsException("Target entity ("+component+") has "+locations.size()+" location(s), but requested index "+index);
```
geomacy: The behaviour is perhaps a bit odd from the app point of view. Say you set a config value to a location with a bad index,  the exception is logged in `brooklyn.debug.log` (not `info`), but the config value is still rendered in the application, which doesn't go on fire.  The value isn't a location of course, but the supplier:

```
$ br app bn ent BEN config
Key                    | Value
camp.template.id       | yJlI9Z2t
testLocN               | map[component:map[componentId: componentIdSupplier:<nil> scopeComponent:<nil> scope:THIS] index:3]
brooklyn.wrapper_app   | true
```
which users may find confusing?

nakomis:
Looks good, nothing to add to @geomacy's comments

geomacy:
hi @aledsage just checking what do you think about merging this PR?

aledsage:
Thanks @geomacy @nakomis 

I've added `testDslLocationIndexOutOfBounds`, which asserts that we get a nice exception when one tries to retrieve in anger the config value that has an invalid index.

I think this is tricky to improve, without going into bigger more general changes.

As you point out, for the blueprint:
```
location: localhost
services:
    brooklyn.config:
      config1: $brooklyn:location()
      config2: $brooklyn:location("0")
      config3: $brooklyn:location("1")
      config4: $brooklyn:attributeWhenReady("sensorDoesNotExist")
```

`br app obrjk09hec config` returns:
```
Key                  | Value   
config1              | map[type:org.apache.brooklyn.api.location.Location id:whoq9008c0]   
defaultDisplayName   | <nil>   
camp.template.id     | ClysL4Gs   
config2              | map[type:org.apache.brooklyn.api.location.Location id:whoq9008c0]   
config3              | map[component:map[componentId: componentIdSupplier:<nil> scopeComponent:<nil> scope:THIS] index:1]   
config4              | map[component:map[componentId: componentIdSupplier:<nil> scopeComponent:<nil> scope:THIS] sensorName:sensorDoesNotExist]   

```

The `config3` evaluation has only been logged at debug, because the value isn't yet being used - it's just retrieved for display purposes, so hasn't blocked. It failed to resolve immediately, so we go the DSL object back. You can see we get the same behaviour for an `attributeWhenReady` that cannot yet be evaluated.

If we wrote a blueprint that actually tried to use it (i.e. that caused `entity.config().get("config3")` to be called), then it would throw the exception, which would be logged at warn.

The two things you point out would be good more general improvements:

1. `/v1/applications/<appId>/entities/<entityId>/config/current-state?raw=false` should return nicer representation of DSL objects (e.g. calling `toString()` on it, rather than showing the values of the fields `component` and `index`).

2. Location representation in the above call doesn't give much info - just the id, and that it's a location. It would be good to give more info as well (e.g. the display name).

---
Are you ok with merging this PR as-is, and us making those more general improvements separately?

camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/DslAndRebindYamlTest.java
```diff
@@ -467,6 +467,30 @@ public void testDslLocation() throws Exception {
...
+                "services:",
+                "- type: " + BasicApplication.class.getName(),
+                "  brooklyn.config:",
+                "    config1: $brooklyn:location(\"1\")");
+
+        Application app = (Application) createStartWaitAndLogApplication(yaml);
+
+        Location appLoc = Iterables.getOnlyElement(app.getLocations());
+        try {
+            assertEquals(getConfigInTask(app, ConfigKeys.newConfigKey(Object.class, "config1")), appLoc);
```
nakomis: If an exception is not thrown here the test will pass when it should fail. This case should be checked, e.g. by throwing `IllegalStateException` after this line

nakomis:
LGTM apart from a minor comment on the test. I agree it would be nice to be able to tell whether DSL simply hasn't resolved yet or whether it's permanently failed to resolve, but I'm happy for this PR to be merged in the meantime

aledsage:
Thanks @geomacy @nakomis - merging now. Brooklyn issues are probably the best way to record and track further improvements, I think.

PR #974, opened 2018-07-26
Update Brooklyn codebase to build the newly donated UI

tbouron:
See discussion about the donated UI: https://lists.apache.org/thread.html/20d82876f7e84244c2c4d3464a1531a10af50ca3d20e9e1de97a5faf@%3Cdev.brooklyn.apache.org%3E

This updates `brooklyn-server` codebase to use the new Brooklyn UI. The Brooklyn launcher has been modified to include only the REST API sever, as the new UI is OSGi only.


ahgittin:
I think this is incomplete; these references need to be changed/removed:

```
% grep4js brooklyn.war | grep -v log:
./launcher/pom.xml:                                     <destFileName>brooklyn.war</destFileName>
./launcher/pom.xml:                                <exclude>brooklyn.war</exclude>
./launcher/src/main/java/org/apache/brooklyn/launcher/BrooklynWebServer.java:    public static final String BROOKLYN_WAR_URL = "classpath://brooklyn.war";
./launcher/src/main/java/org/apache/brooklyn/launcher/BrooklynWebServer.java:        this(managementContext, port, "brooklyn.war");
./launcher/src/main/java/org/apache/brooklyn/launcher/config/CustomResourceLocator.java:     * in particular, eclipse often does not copy WAR files as instructed by maven, so brooklyn.war might not be found */
./launcher/src/main/java/org/apache/brooklyn/launcher/config/CustomResourceLocator.java:        // brooklyn.path.override.brooklyn.war=classpath://brooklyn-replacement-webapp.war
./launcher/src/test/java/org/apache/brooklyn/launcher/WebAppRunnerTest.java: * These tests require the brooklyn.war to work. (Should be placed by maven build.)
./launcher/src/test/java/org/apache/brooklyn/launcher/WebAppRunnerTest.java:            MutableMap.of("port", "8091+", "war", "brooklyn.war", "wars", MutableMap.of("hello", "hello-world.war")) );
./launcher/src/test/java/org/apache/brooklyn/launcher/WebAppRunnerTest.java:        BrooklynWebServer server = createWebServer(MutableMap.of("port", "8091+", "war", "brooklyn.war"));
./launcher/src/test/java/org/apache/brooklyn/rest/jsgui/BrooklynJavascriptGuiLauncher.java:            : ResourceUtils.create(BrooklynJavascriptGuiLauncher.class).doesUrlExist("classpath://brooklyn.war") 
./launcher/src/test/java/org/apache/brooklyn/rest/jsgui/BrooklynJavascriptGuiLauncher.java:                ? Os.writeToTempFile(ResourceUtils.create(BrooklynJavascriptGuiLauncher.class).getResourceFromUrl("classpath://brooklyn.war"), "brooklyn", "war").getAbsolutePath()
./launcher/src/test/java/org/apache/brooklyn/rest/jsgui/BrooklynJavascriptGuiLauncher.java:            : throwingReturning("could not find jsgui war or source", "missing-brooklyn.war"),
./rest/rest-server/src/test/java/org/apache/brooklyn/rest/BrooklynRestApiLauncher.java:                : ResourceUtils.create(this).doesUrlExist("classpath://brooklyn.war") 
./rest/rest-server/src/test/java/org/apache/brooklyn/rest/BrooklynRestApiLauncher.java:                    ? Os.writeToTempFile(ResourceUtils.create(this).getResourceFromUrl("classpath://brooklyn.war"), "brooklyn", "war").getAbsolutePath()
```

I think the approach of only launching the REST API is fine, which means removing `BrooklynJavascriptGuiLauncher` and tests that expect `brooklyn.war`.

I will miss the lightweight pure-java means of launching Brooklyn with UI but I don't think there's an easy way to achieve that, given how the new UI relies on a metadata service built on OSGi services, and it's painful and not representative maintaining the non-OSGi launch pathway.  We still have the non-OSGi pathway for tests and for IDE launches with REST; there are a few places where behaviour is different than OSGi, but given the speed with which we can run non-UI tests and it still works, I think that's worth keeping as long as it's easy to do so.

We should identify in the simplest way to do IDE launches including UI, in Eclipse and in IntelliJ (but does not need to be part of this PR.)

tbouron:
@ahgittin I voluntarily didn't touch the `BrooklynLauncher` as the "classic launcher" only has been deprecated in `0.12.0`. I think remove it should be part of another PR.

I just made the bare minimum here for the build to pass

ahgittin:
I've fixed a few other places where `brooklyn.war` was mentioned, as noted.  Have left the launcher project, as discussed, even though most of it isn't really needed.

Also removed all refs to `brooklyn-jsgui` in poms, and moved JSON test fixtures which used to live there into `src/test/resources/fixtures` in the `rest-api` project.

ahgittin:
This too seems to be working nicely now.  :)

PR #979, opened 2018-08-17
karaf/features: include snakeyaml 1.17 bundle explicitly

kemitix:
The feature entry for jclouds-all-compute still requires the snakeyaml 1.17 to be explicitly
added. Now both versions of snakeyaml are available.


duncangrant:
Thanks for that @kemitix

PR #984, opened 2018-09-05
restore the field `subType` for sub-element config keys

ahgittin:
now supports generics _or_ plain, so persistence is simpler when plain,
(as done for BasicConfigKey)


aledsage:
+1, LGTM; waiting for jenkins to finish running tests.

PR #986, opened 2018-09-06
BROOKLYN-600: cleanup entities on deploy-failure

aledsage:
Fix for https://issues.apache.org/jira/browse/BROOKLYN-600

Note the `LocalEntityManager. discardPremanaged(Entity)` is a bit more complex/long that it might need to be because I want to make sure it does no harm - fail if the entity or any child is already managed (because then should call `unmanage()` instead.


duncangrant:
This looks good to me

PR #989, opened 2018-09-11
LogWatcher - don't change log level if not needed

ahgittin:
prevents the logs from missing useful output if testing for a higher-level message,
useful when investigating tests, esp SoftwareProcessStopsDuringStartTest.testStopDuringProvisionWaitsForCompletion where the log messages we want aren't showing because we reset it to warn.

also use try-close syntax for LogWatcher.

### DIFF
--- a/test-support/src/main/java/org/apache/brooklyn/test/LogWatcher.java
+++ b/test-support/src/main/java/org/apache/brooklyn/test/LogWatcher.java
@@ -19,14 +19,13 @@
 package org.apache.brooklyn.test;
 
 import static com.google.common.base.Preconditions.checkNotNull;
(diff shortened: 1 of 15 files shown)


test-support/src/main/java/org/apache/brooklyn/test/LogWatcher.java
```diff
@@ -268,17 +248,14 @@ public void assertHasEvent() {
             @Override
             public void run() {
                 assertFalse(events.isEmpty());
+                System.out.println("EVENTS: "+events);
```
aledsage: Don't include `System.out`

ahgittin:
thx @aledsage , done and merged

PR #991, opened 2018-09-14
bump jetty version in response to recent CVE's

ahgittin:
no evidence these affect brooklyn but good to stay current

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -136,7 +136,7 @@
         <sshj.version>0.20.0</sshj.version>
         <felix.framework.version>5.6.1</felix.framework.version>
         <reflections.version>0.9.10</reflections.version>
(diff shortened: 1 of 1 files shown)


ahgittin:
retest this please

aledsage:
LGTM. Merging now.

The failing test looks unrelated (`VanillaSoftwareProcessYamlTest.testAlternativeServiceUpPolling`). Confirmed it passed locally for me. Looking at the test, I could well believe it is non-deterministic. We'll look at that separately. 

For the record, the way this gets into karaf seems to be in `brooklyn-ui/modularity-server/features/src/main/feature/feature.xml`:
```
<bundle dependency="true" start-level="30">mvn:org.eclipse.jetty/jetty-proxy/${jetty.version}</bundle>
```

PR #993, opened 2018-09-14
Prevent racing persistence

ahgittin:
This _might_ fix the cause of the failure observed after #988 , and so then reinstates it, effectively unreverting #990 .

Worth having this be tested a few times in the PR to see whether it does the trick.

I suspect the FunctionFeed which is being left around is being added in a window where the entity is shutting down.  Just a guess though, not high confidence.


ahgittin:
does not fix it :( ... closing for now

PR #998, opened 2018-09-18
Pom tidy, and version upgrades

aledsage:
See individual commits (the first one just moves stuff around).


ahgittin:
looks good.  if build passes let's merge.

PR #1001, opened 2018-09-25
improve CollectionMerger to take a configurable list merge function

ahgittin:
simple extension but useful when lists need special treatment cc @aledsage


ahgittin:
merging obo @aledsage who approved this but is offline

PR #1003, opened 2018-10-01
Update IPTables save method

frogfather:
Service iptables save doesn't work with Centos7

### DIFF
--- a/utils/common/src/main/java/org/apache/brooklyn/util/ssh/IptablesCommands.java
+++ b/utils/common/src/main/java/org/apache/brooklyn/util/ssh/IptablesCommands.java
@@ -130,8 +130,8 @@ public class IptablesCommands {
      *
      */
     public static String saveIptablesRules() {
-        return alternatives(sudo("service iptables save"),
-                            chain(installPackage("iptables-persistent"), sudo("/etc/init.d/iptables-persistent save")));
+        return alternatives("if [ ${UID} -eq 0 ] ; then iptables–save > /etc/sysconfig/iptables ; else sudo iptables-save | sudo tee /etc/sysconfig/iptables ; fi",
+                chain(installPackage("iptables-persistent"), sudo("/etc/init.d/iptables-persistent save")));
     }
 
     /**


utils/common/src/main/java/org/apache/brooklyn/util/ssh/IptablesCommands.java
```diff
@@ -130,8 +130,8 @@ public static String firewalldServiceIsActive() {
      *
      */
     public static String saveIptablesRules() {
-        return alternatives(sudo("service iptables save"),
-                            chain(installPackage("iptables-persistent"), sudo("/etc/init.d/iptables-persistent save")));
+        return alternatives("if [ ${UID} -eq 0 ] ; then iptables–save > /etc/sysconfig/iptables ; else sudo iptables-save | sudo tee /etc/sysconfig/iptables ; fi",
```
grkvlt: Maybe surround this with `BashCommands.ifExecutableElse1("iptables-save", ...)` to ensure the command exists

grkvlt:
> Service iptables save doesn't work with Centos7

That is, doesn't work with the CentOS 7 AMI we use on EC2, which has a cut-down `service` command only supporting the standard _start_, _stop_, _restart_ and _status_ command verbs.

grkvlt:
Thanks @frogfather LGTM, will merge assuming tests pass

grkvlt:
I think Jenkins is failing due to permissions, maybe on the ASF side?

aledsage:
LGTM; happy for this to be merged (once confusion with the identical-looking https://github.com/apache/brooklyn-server/pull/1005 is cleared up).

utils/common/src/main/java/org/apache/brooklyn/util/ssh/IptablesCommands.java
```diff
@@ -130,8 +131,9 @@ public static String firewalldServiceIsActive() {
      *
      */
     public static String saveIptablesRules() {
-        return alternatives(sudo("service iptables save"),
-                            chain(installPackage("iptables-persistent"), sudo("/etc/init.d/iptables-persistent save")));
+        return alternatives(
+                ifExecutableElse1("iptables–save", "if [ ${UID} -eq 0 ] ; then iptables–save > /etc/sysconfig/iptables ; else sudo iptables-save | sudo tee /etc/sysconfig/iptables ; fi"),
```
ahgittin: this removes `service iptables save` altogether, is that intended?  a comment to that effect would be useful,
or if it is still useful sometimes (older OS's?) then keep it in the alternatives list

also worth checking whether `sudoNew("iptables-save > /etc/sysconfig/iptables")` works; the way it does `echo COMMAND | sudo bash` _should_ support redirect for non-root users; if not maybe refactor to add a `sudoRedirect` method in `BashCommands` capturing the conditional `tee` so that you could write `ifExecutableElse1("iptables-save", sudoRedirect("iptable-save", "/etc/sysconfig/iptables"))`

ahgittin:
as @aledsage this is largely included in #1006 .  @frogfather your call whether there is value in the recent comments here and if so update this PR, make sure to `git merge master` in to this branch, or if not just close this.

nakomis:
@frogfather Can you take a look at this please to see if it is still relevant, and address the comments above if appropriate

Thanks

frogfather:
Will do.

PR #1006, opened 2018-10-03
Update IPTables save method

grkvlt:
Service iptables save doesn't work with Centos 7, or rather doesn't work with the CentOS 7 AMI we use on EC2, which has a cut-down service command only supporting the standard start, stop, restart and status command verbs.

Generates a better bash command string that should work on all CentOS variants.


aledsage:
This looks identical to https://github.com/apache/brooklyn-server/pull/1003 - is it? Can we close your PR @grkvlt and merge @frogfather 's original?

PR #1010, opened 2018-10-22
BROOKLYN-605 rebind fails for historic state

kemitix:
The solution implemented here uses query-path values on the `<bundle>wrap:...</bundle>` entries.

For example:

`<bundle>wrap:mvn:com.amazonaws/aws-java-sdk-bundle/${aws.java.sdk.version}$Bundle-SymbolicName=com.amazonaws.aws-java-sdk-bundle&amp;Bundle-Version=${aws.java.sdk.version}&amp;Brooklyn-Catalog-Upgrade-For-Bundles=wrap_file__Users_aledsage_amp_cloudsoft-amp-karaf-5.3.0_system_com_amazonaws_aws-java-sdk-bundle_1.11.245_aws-java-sdk-bundle-1.11.245.jar:[0,1)</bundle>`

Specifically, it uses these keys:

* `Bundle-SymbolicName` - to replace the auto-generated symbolic name (with something that doesn't include the absolute path to the resource)
* `Bundle-Version` - default otherwise is zero
* `Brooklyn-Catalog-Upgrade-For-Bundles` - specify the auto-generated symbolic name of the legacy bundle

Note that the query path must start with a `$` and use `&amp;` as the delimiter between the key/value pairs. The version range should start at zero: e.g. `[0,1)` The legacy wrapped bundles were all given a version of zero.

Summary of changes:

* Adapt `o.a.b.core.catalog.internal.CatalogInitialization.gatherCatalogUpgradesInstructions` to scan non-managed bundles for catalog upgrades. Previously, only managed bundles were scanned. Wrapped bundles are not managed, so any upgrades declared wouldn't be seen. In tests this raised the scan time from 30-42ms to 103-127ms.
* Extract the `gatherCatalogUpgradesInstructions()` method into a separate `CatalogUpgradeScanner` class.
* Promote `CatalogInitialization.RegisteredTypeSupplier` to an outer class as it is used by both `CatalogInitialization` and `CatalogUpgradesScanner`.
* Moved detection and short-cutting when not in an OSGi environment out of the `CatalogUpgradesScanner` class to avoid polluting it with an edge case that isn't related to the work at-hand (Single Responsibility).
* Push out static method calls to constructor-injected `BiFunction` and `Function` parameters to simplify unit testing.
* Unit tests cover all branches on `CatalogUpgraedsScanner` except the one that only results in a log message.
* While investigating other, possible, solutions the `o.a.b.core.mgmt.rebind.RebindWithDeserializingClassRenamesTest` class was revised and `o.a.b.core.typereg.CatalogUpgradesGetBundleUpgradedIfNecessaryTest` and `o.a.b.util.javalang.ReflectionsFindMappedNameTest` were created.
* Helper methods for creating mock Bundles, that had been private within `o.a.b.core.typereg.BundleUpgradeParserTest`, were extracted to `o.a.b.core.typereg.BundleTestUtil` as public static methods. These were then used in `CatalogUpgradesScannerTest`, `CatalogUpgradesGetBundleUpgradedIfNecessaryTest` and `BundleUpgradeParserTest`.
* `assertj` was upgraded to 3.11.1 to take advantage of new features available, now that Brooklyn requires Java 8. Previously `assertj` had been held back at 2.x to avoid bringing in any Java 8 dependencies.


kemitix:
@aledsage ready for review

core/src/main/java/org/apache/brooklyn/core/catalog/internal/CatalogUpgradeScanner.java
```diff
@@ -0,0 +1,106 @@
...
+    private final Function<String, Predicate<? super RegisteredType>> unmanagedBundlePredicateSupplier;
+
+    CatalogUpgradeScanner(
+            final ManagementContextInternal managementContext,
+            final BiFunction<Bundle, RegisteredTypesSupplier, CatalogUpgrades> bundleUpgradeParser,
+            final Function<OsgiBundleWithUrl, Predicate<? super RegisteredType>> managedBundlePredicateSupplier,
+            final Function<String, Predicate<? super RegisteredType>> unmanagedBundlePredicateSupplier
+    ) {
+        this.managementContext = managementContext;
+        this.bundleUpgradeParser = bundleUpgradeParser;
```
aledsage: If subsequently we use the field without null checks, then preference for doing:
```
import static com.google.common.base.Preconditions.checkNotNull;
...
this.bundleUpgradeParser = checkNotNull(bundleUpgradeParser, "bundleUpgradeParser");
```

core/src/test/java/org/apache/brooklyn/core/mgmt/rebind/RebindWithDeserializingClassRenamesTest.java
```diff
@@ -18,76 +18,134 @@
...
+                          final ConfigInheritance inheritanceMode,
+                          final String configKeyName) {
+            this.legacyClassName = legacyClassName;
+            this.currentClassName = currentClassName;
+            this.inheritanceMode = inheritanceMode;
+            this.configKeyName = configKeyName;
+        }
+
+        static Transform of(final String legacyClassName,
+                            final ConfigInheritance configInheritance,
```
aledsage: Preference for consistent naming: the constructor calls it `inheritanceMode`, as does the field - so I'd call it that in this factory method as well.

core/src/test/java/org/apache/brooklyn/core/typereg/CatalogUpgradesGetBundleUpgradedIfNecessaryTest.java
```diff
@@ -0,0 +1,246 @@
...
+    private final String originalName = createAName("original", "1.0.0");
+    private final String updatedName = createAName("updated", "2.0.0");
+    private final VersionedName originalVersionedName = VersionedName.fromString(originalName);
+    private final VersionedName updatedVersionedName = VersionedName.fromString(updatedName);
+    private final VersionRangedName originalVersionRangedName =
+            VersionRangedName.fromString(originalName, true);
+
+    private CatalogUpgrades.Builder catalogUpgradesBuilder;
+
+    @BeforeMethod
```
aledsage: Preference for `@BeforeMethod(alwaysRun=true)`. Otherwise if someone ever marks a test as `groups="Integration"` or some such then the setUp method would not be called.

parent/pom.xml
```diff
@@ -581,6 +581,11 @@
                 <groupId>org.ops4j.pax.web</groupId>
                 <artifactId>pax-web-spi</artifactId>
                 <version>${pax-web.version}</version>
+			</dependency>
```
aledsage: Uses tabs, whereas the rest of the file uses spaces - gives weird indentation.

pom.xml
```diff
@@ -183,11 +183,13 @@
         <plantuml.version>6121</plantuml.version>
 
         <!-- Test dependencies -->
+        <testng.version>6.10</testng.version>
```
aledsage: `testng.version` and `mockito.version` now appear twice in the file - remove this.

utils/common/src/test/java/org/apache/brooklyn/util/javalang/ReflectionsFindMappedNameTest.java
```diff
@@ -0,0 +1,145 @@
...
+
+import org.assertj.core.api.WithAssertions;
+import org.testng.annotations.Test;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.UUID;
+
+/**
+ * Tests for {@link Reflections#findMappedNameAndLog(Map, String)} and
```
aledsage: Personally I don't bother with this kind of comment. I've seen them get out-of-date far too often as people modify the file to refactor or improve test coverage.

The package and class name should tell us what it is testing, and the test method names are clear.

utils/common/src/test/java/org/apache/brooklyn/util/javalang/ReflectionsTest.java
```diff
@@ -24,12 +24,12 @@
 
 import java.lang.reflect.Field;
 import java.lang.reflect.Method;
-import java.util.Arrays;
-import java.util.List;
+import java.util.*;
```
aledsage: Convention in brooklyn is to not use `.*` imports.

(If it were for everything in a nested class, such as you do in `org.apache.brooklyn.core.catalog.internal.CatalogUpgradeScannerTest.Givens.*;` then that makes complete sense. But for things like `java.util...` I'd prefer explicit imports.

core/src/main/java/org/apache/brooklyn/core/catalog/internal/CatalogUpgradeScanner.java
kemitix: JDK now offers `Objects.requireNonNull(..)` that has the same behaviour.

aledsage:
Thanks @kemitix - looks great. Merging now.

PR #1012, opened 2018-11-01
Fix for SUREFIRE-1588

kemitix:
Latest version of Java 1.8.0_191 enforces that Manifest classpath entries be relative.

https://issues.apache.org/jira/browse/SUREFIRE-1588


tbouron:
LGTM @kemitix and Jenkins is happy, thanks 👍

PR #1013, opened 2018-11-06
Partial Fix for Windows Build

kemitix:
Fixes a couple of issues with the Windows build. There are other issues beyond these fixes that are still breaking the build.

* Under Windows the recursive deletion of directories containing bundles throws an `IOException` during the test `tearDown` phase causing tests to fail. Tests now use an new static method `FileUtil.deleteDirectory()` that deletes the directory and logs, rather than throws, any `IOException`.
* `maven-compiler-plugin` was using the incorrect syntax for configuring compiling of groovy souce code. This wasn't breaking the build, but was a simple fix.
* Add a `@DisableOnWindows` annotation for TestNG that will ignore test methods with this annotation. The annotation requires a reason be given to explain why. This reason is included in log output indicating that the test is disabled.
* Disable tests in `BashCommandsIntegrationTest` that require features not typically available in a normal Windows environment: i.e. an SSH server on localhost, the `bash` command line, the `whoami` command.

### DIFF
--- a/core/src/test/java/org/apache/brooklyn/util/core/osgi/OsgiTestBase.java
+++ b/core/src/test/java/org/apache/brooklyn/util/core/osgi/OsgiTestBase.java
@@ -38,8 +38,6 @@ import org.testng.annotations.BeforeMethod;
  */
 public class OsgiTestBase {
 
-    private static final Logger log = LoggerFactory.getLogger(OsgiTestBase.class);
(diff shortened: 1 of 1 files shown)


kemitix:
@tbouron / @richardcloudsoft: Could you review this? Would tag Juan, but don't know his GitHub account.

PR #1015, opened 2018-11-15
add distinction between "application" and "template" in catalog itemType

ahgittin:
previously `template` was the only way to class something as an application, and it did both:
* generate a spec for an `Application`
* add the `catalog_template` tag so it looks llke a template

this adds support for an `itemType: application` which only does the first of these


ahgittin:
also fixes issue with resolving details for things which aren't immediately available, in first commit here

PR #1024, opened 2019-01-09
Implement REST and UI security as filter not JAAS LoginModule, start OAuth

ahgittin:
The JAAS LoginModule seems only to work with BASIC and related auth schemes; it is insufficient for things like OAuth which need a redirect and a token.

After trying many things we have concluded that going back to having our own Filter is the best way to allow the flexibility for modern security/auth approaches.  Unfortunately we need _two_ filters, one for Jersey and one for Javax (static resource) bundles.  This PR supplies all that, removing the JAAS stuff altogether, and some related tidies eg around logout.

Note that this will break downstream modules that reference the webconsole "realm" or the JAAS LoginModule.  They need to be updated to refer to the filter.  PRs for that will follow.

Anything configured with a Brooklyn `SecurityProvider` will continue to work.  However any custom `SecurityProvider` instances will need to add a new method to say whether a user/pass is required (which will trigger the `WWW_AUTHENTICATE` header which causes the browser to prompt for user/pass).

This also adds a WIP OauthSecurityProvider that does OAuth.  It needs more work to allow it to be parameterised and usable but in its current form it allows us to confirm that this approach will work for OAuth.  (The credentials in that class are useless ones only for testing.)

Finally this also simplifies the mechanism for specifying in `brooklyn.properties` / `brooklyn.cfg` that a security provider should be loaded from a custom OSGi bundle.  This can now be done by setting two additional properties for the `bundle.symbolicName` and (optionally) the `bundle.version`:

```
brooklyn.webconsole.security.provider = com.acme.MySecurityProvider
brooklyn.webconsole.security.provider.bundle.symbolicName = com.acme
brooklyn.webconsole.security.provider.bundle.version = 1.0.0
```

In addition to downstream projects, documentation needs to be updated to reflect:

* How to specify a bundle for a custom security provider
* How security is configured for the REST API (Jersey filter) and resource (Javax filter)
* (and in subsequent PR when OAuth is ready) How to configure/customize OAuth (when this is done, note it may need a new bundle)

We also need a backwards compatibility note for anyone who has relied on the JAAS LoginModule.


ahgittin:
Note:  Bumps some of the versions (see https://github.com/apache/brooklyn-server/pull/1024/files#diff-600376dffeb79835ede4a0b285078036)

ahgittin:
Note the following pull requests are needed simultaneously with this one:

* https://github.com/apache/brooklyn-ui/pull/117
* https://github.com/apache/brooklyn-dist/pull/138

ahgittin:
@jcabrerizo and I have reviewed each other's work on this.  Lots of minor version bumps and tweaks to make the above all work and tests pass but we seem to be there.  All builds and passing locally.

We can't trust downstream PR tests without this merged so going to merge all then monitor the build server.

PR #1025, opened 2019-01-16
remove WIP oauth provider

ahgittin:
will be added back when working

it's not needed for #1024 (except for testing) and probably confusing to have in, in its current state.  those who need it for testing can access it via history.

note the tokens used in here are test-only so no harm in keeping them


jcabrerizo:
LGTM. Next step, finish the implementation of the `SecurityProvider` to use Oauth authentication

PR #1026, opened 2019-01-23
Add cxf-karaf-commands dependency

duncangrant:
This is a dependency of cxf that was introduced by the version change to 3.2.7.
This adds cxf commands to the karaf shell such as listing endpoints.

### DIFF
--- a/karaf/features/src/main/feature/feature.xml
+++ b/karaf/features/src/main/feature/feature.xml
@@ -208,6 +208,7 @@
         <feature>cxf-jaxrs</feature>
         <bundle dependency="true">mvn:org.apache.cxf/cxf-rt-rs-security-cors/${cxf.version}</bundle>
         <bundle dependency="true">mvn:org.apache.cxf/cxf-rt-frontend-jaxrs/${cxf.version}</bundle>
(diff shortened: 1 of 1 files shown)


karaf/features/src/main/feature/feature.xml
```diff
@@ -208,6 +208,7 @@
         <feature>cxf-jaxrs</feature>
         <bundle dependency="true">mvn:org.apache.cxf/cxf-rt-rs-security-cors/${cxf.version}</bundle>
         <bundle dependency="true">mvn:org.apache.cxf/cxf-rt-frontend-jaxrs/${cxf.version}</bundle>
+        <bundle dependency="true">mvn:org.apache.cxf.karaf/cxf-karaf-commands/3.2.7</bundle>
```
kemitix: Could you change this to `${cxf.version}`?

tbouron: Good point. Actually, I think you can reuse `${cxf.version}` directly as these bundles are released in the same time

kemitix: @tbouron I see now that the version of cxf is a typo in the PR description. Use `${cxf.version}`

duncangrant: oops - sorry about that

PR #1027, opened 2019-01-23
Added config keys for Oauth

jcabrerizo:
Some new ConfigKeys added to mange the authentication using Oauth


tbouron:
Thanks for the PR @jcabrerizo !

However, the `OAuthProvider` introduced by #1024 was WIP and removed in #1025. So I think these config keys should be added at the same time you'll re-introduce the stable `OAuthProvider`. You could even use this PR for this.

PR #1030, opened 2019-01-24
BROOKLYN-607: prefer vanilla CentOS image on azure

aledsage:


ahgittin:
Looks good.  Will add some comments:

```
// suggest to use NaturalOrderComparator (so 10>9) then order by:
// 1) `name.replaceAll("([^0-9]+)", " ")` 
// 2) shortest non-empty name
// 3) version (NaturalOrderComparator, prefer last)
// 4) name (NaturalOrderComparator, prefer last)
// 5) other fields (NaturalOrderComparator, prefer last)
```

PR #1032, opened 2019-01-27
fix session sharing and simplify logout

ahgittin:
### have a shared session available for the various bundles 

mainly to know if each other have authorized.  see https://github.com/apache/brooklyn-server/pull/1032/commits/1a15c36ad9a5caa76bab2e140c89cc4770fa0abb for the main new code, in `MultiSessionAttributeAdapter`.
note this has some historic aborted stuff, in case we need to revisit.  that can be ignored, best to evaluate this _not_ commit-by-commit!

also...

### this makes /logout just log out and simplifies it

previously it redirected to the user-specific logout; but some clients didn't respect that, and assumed the logout had happened.
not convinced there is a good reason for /logout/{user} ... and if there is, you can now use /logout/redirect to get to it.

### calls SecurityProvider.logout and correctly invalildates everything on logout

previously that was neglected

### CSRF fail returns `403 FORBIDDEN` not `401 UNAUTHORIZED`

it's more accurate, and allows clients to distinguish problems from auth fails

### DIFF
--- a/karaf/jetty-config/src/main/resources/jetty.xml
+++ b/karaf/jetty-config/src/main/resources/jetty.xml
@@ -20,47 +20,4 @@
 <!DOCTYPE Configure PUBLIC "-//Jetty//Configure//EN" "http://www.eclipse.org/jetty/configure_9_0.dtd">
 
 <Configure id="Server" class="org.eclipse.jetty.server.Server">
-    <Call name="addBean">
-        <Arg>
-            <New class="org.eclipse.jetty.jaas.JAASLoginService">
-                <Set name="name">karaf</Set>
-                <Set name="loginModuleName">karaf</Set>
-                <Set name="roleClassNames">
-                    <Array type="java.lang.String">
-                        <Item>org.apache.karaf.jaas.boot.principal.RolePrincipal</Item>
-                    </Array>
-                </Set>
-            </New>
-        </Arg>
-    </Call>
-    <Call name="addBean">
-        <Arg>
-            <New class="org.eclipse.jetty.jaas.JAASLoginService">
-                <Set name="name">default</Set>
-                <Set name="loginModuleName">karaf</Set>
-                <Set name="roleClassNames">
-                    <Array type="java.lang.String">
-                        <Item>org.apache.karaf.jaas.boot.principal.RolePrincipal</Item>
-                    </Array>
-                </Set>
-            </New>
-        </Arg>
-    </Call>
-    <Call name="addBean">
-        <Arg>
-            <New class="org.eclipse.jetty.jaas.JAASLoginService">
-                <Set name="name">webconsole</Set>
-                <Set name="loginModuleName">webconsole</Set>
-                <Set name="roleClassNames">
-                    <Array type="java.lang.String">
-                        <Item>org.apache.brooklyn.rest.security.jaas.BrooklynLoginModule$RolePrincipal</Item>
-                    </Array>
-                </Set>
-                <Set name="identityService">
-                    <New class="org.eclipse.jetty.security.DefaultIdentityService"/>
-                </Set>
-            </New>
-        </Arg>
-    </Call>
-
 </Configure>
--- a/rest/rest-api/src/main/java/org/apache/brooklyn/rest/api/LogoutApi.java
+++ b/rest/rest-api/src/main/java/org/apache/brooklyn/rest/api/LogoutApi.java
@@ -18,14 +18,11 @@
  */
 package org.apache.brooklyn.rest.api;
 
-import javax.ws.rs.GET;
 import javax.ws.rs.POST;
 import javax.ws.rs.Path;
 import javax.ws.rs.PathParam;
 import javax.ws.rs.QueryParam;
 import javax.ws.rs.core.Response;
-
-import com.google.common.annotations.Beta;
 
 import io.swagger.annotations.Api;
 import io.swagger.annotations.ApiOperation;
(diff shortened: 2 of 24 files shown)


ahgittin:
experimenting to see what makes sense

the browser seems to keep the local cached creds, _even_ on a 401 unauthorized, so not sure if there's much point in returning that

ahgittin:
mostly fixes problem at #1033 and makes it nicer to work with from the uI, including invalidating locally-stored creds

still to do:
* push the corresponding logout UI fixes
* figure out whether the `Authorization` header being kept on the looping `/extended` call can be made to be dropped after the first call, so it doesn't cache the login, and confirm this means when you've logged out you are properly logged out
* determine what code is creating a session when the user isn't authenticated
* check if time-based invalidation (without manual cache deletion) causes the server to crash

ahgittin:
done most the list in previous comment.  confirmed logout works nicely in browser (unless the `/extended` call is done in another tab), invalidation just works without problems using new method, and no extraneous sessions are created after logout. 

one more issue i've noticed:  unauthorized requests create a session which is kept; this is a memory leak and needless.  i'll have a quick look to see if that can be undone.

remaining changes will be in PR(s) in other project (UI):

* push corresponding logout UI fixes
* figure out whether the Authorization header being kept on the looping /extended call can be made to be dropped after the first call, so it doesn't cache the creds in the app and re-log you in
* check the ui-module bundles (e.g. module registry) have the same auth/filter and session management configuration

ahgittin:
looking good.  have updated all other items in the list above at:

* https://github.com/apache/brooklyn-ui/pull/120

(i can't make basic auth fully always disabled, but that's a well-known problem; i've done what i could and added some info about it)

also see minor change needed to brooklyn-dist itests:

* https://github.com/apache/brooklyn-dist/pull/139

these two PRs should be reviewed along with this one

all ready for review and (i hope) merge

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/resources/LogoutResource.java
```diff
@@ -75,19 +67,80 @@ public Response logoutUser(String user) {
...
+        // and log with stack trace in newHttpSession
+//        HttpServletRequest jreq = req;
+//        if (jreq instanceof ThreadLocalHttpServletRequest) jreq = ((ThreadLocalHttpServletRequest)jreq).get();
+//        if (jreq instanceof ServletRequestWrapper) jreq = (HttpServletRequest) ((ServletRequestWrapper)jreq).getRequest();
+//        if (jreq instanceof Request) {
+//            log.warn("SWAPPING "+MultiSessionAttributeAdapter.info(jreq));
+//            ((Request)jreq).setSessionHandler(new TrackingSessionHandler(((Request)jreq).getSessionHandler()));
+//        } else {
+//            log.warn("UNABLE to swap request"+MultiSessionAttributeAdapter.info(jreq));
+//        }
```
tbouron: Can be removed

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/util/MultiSessionAttributeAdapter.java
```diff
@@ -0,0 +1,477 @@
...
+    
+        protected SessionHandler findPreferredBundleHandler(Session localSession, Server server, Handler[] handlers) {
+            if (PREFERRED_SYMBOLIC_NAME==null) return null;
+            
+            SessionHandler preferredHandler = null;
+            
+            if (handlers != null) {
+                for (Handler handler: handlers) {
+                    SessionHandler sh = (SessionHandler) handler;
+                    ContextHandler.Context ctx = getContext(sh);
```
tbouron: I'm a bit concerned performance wise about this call, as `getContext(sh)` uses refection. This is done within a loop of handler which can be potentially big.

If I read this code correctly, this will be executed only during the session creation, if there is no other sessions marked as preferred, so it is somewhat mitigated. But still, I'm wondering if it could be a bottleneck

karaf/jetty-config/src/main/resources/jetty.xml
```diff
@@ -20,47 +20,4 @@
 <!DOCTYPE Configure PUBLIC "-//Jetty//Configure//EN" "http://www.eclipse.org/jetty/configure_9_0.dtd">
 
 <Configure id="Server" class="org.eclipse.jetty.server.Server">
```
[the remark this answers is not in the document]
ahgittin: I'm not sure; I think it might break if we remove it.  Note that you can't configure the ServletContext / WebAppContext (where SessionHandler is set) at this jetty server-wide level.

rest/rest-api/src/main/java/org/apache/brooklyn/rest/api/LogoutApi.java
```diff
@@ -21,34 +21,40 @@
...
     @POST
-    @ApiOperation(value = "Request a logout and clean session")
-    @ApiResponses(value = {
-            @ApiResponse(code = 307, message = "Redirect to /logout/{user}, keeping the request method")
-    })
-    Response logout();
+    @ApiOperation(value = "Logout and clean session")
+    Response logout(
+            @ApiParam(value = "Instead of 200 (the default) to indicate successful logout, "
+                + "return a 401 with this value in a message key in the body (a 401 will cause browsers to clear some locally cached credentials)",
```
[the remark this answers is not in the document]
ahgittin: No, if we return 200 browsers keep the credentials.  But note that the API _does_ return 200 _unless_ this `unauthorize` parameter is supplied.  This parameter really exists to help UI clients encourage the browser clear.  Pretty sure that's the nicest pattern.

Have updated one comment in `LogoutResource` to explain better; in short there is no way that either the server or the client can 100% clear creds at present, but this appears to be the best we can do.

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/resources/LogoutResource.java
ahgittin: I'd like to keep for the reasons noted in the comment in this section

```
        // if we need to intercept session creation then can use this
        // create TrackingSessionHandler which delegates (no-op if delegate==null esp in setSessionTrackingMode)
        // and log with stack trace in newHttpSession
```

I know we could restore it from history but unless/until we are really happy with this behaviour I think it is worth making it a little more prominent.

Added

```
        //   can remove this after we've been running happily with new session management model for a while
```

PR #1033, opened 2019-01-27
Share sessions among all bundles so we can share auth

ahgittin:
this takes the first encountered session handler and caches it,
also setting the sessonPath on the cookies so cookies are valid for all bundles.

some low-level hacking to work around pain that CXF doesn't let you configure session handling;
if there is a nicer way i'd love to see it, but i think it would require rewriting chunks of CXF

### DIFF
--- a/rest/rest-resources/pom.xml
+++ b/rest/rest-resources/pom.xml
@@ -117,11 +117,6 @@
             <artifactId>cxf-rt-rs-security-cors</artifactId>
             <version>${cxf.version}</version>
         </dependency>
-        <dependency>
-            <groupId>org.apache.cxf</groupId>
-            <artifactId>cxf-rt-frontend-jaxrs</artifactId>
-            <version>${cxf.version}</version>
-        </dependency>
 
         <dependency>
             <groupId>org.apache.brooklyn</groupId>
@@ -220,10 +215,9 @@
                 <configuration>
                     <instructions>
                         <Import-Package>
-                            org.apache.karaf.jaas.config,
-                            org.eclipse.jetty.jaas,
                             *
                         </Import-Package>
+                        <Export-Package>org.apache.brooklyn.*</Export-Package>
                     </instructions>
                 </configuration>
             </plugin>
--- a/rest/rest-resources/src/main/java/org/apache/brooklyn/rest/filter/BrooklynSecurityProviderFilterHelper.java
+++ b/rest/rest-resources/src/main/java/org/apache/brooklyn/rest/filter/BrooklynSecurityProviderFilterHelper.java
@@ -18,8 +18,9 @@
  */
 package org.apache.brooklyn.rest.filter;
 
-import java.util.Set;
+import java.lang.reflect.Field;
 
+import javax.servlet.ServletRequestWrapper;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpSession;
 import javax.ws.rs.core.MediaType;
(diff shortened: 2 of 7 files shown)


aledsage:
Jenkins test failures look related, e.g. `BrooklynRestApiLauncherTest.testServletStart`:
```
java.lang.NoSuchMethodException: Security provider class org.apache.brooklyn.rest.security.provider.AnyoneSecurityProvider does not have required no-arg or 1-arg (mgmt) constructor
	at org.apache.brooklyn.rest.security.provider.DelegatingSecurityProvider.createSecurityProviderInstance(DelegatingSecurityProvider.java:184) [brooklyn-rest-resources-1.0.0-SNAPSHOT.jar:na]
```

ahgittin:
Thanks @aledsage ; fixed the obvious problem but there are still some tests that need TLC.
Top priority for the morning.

screenshot 2019-01-28 at 02 07 39

ahgittin:
Have fixed above tests now; let's see how this build does

ahgittin:
retest this please

unrelated error in polilcies

jcabrerizo:
Great job. I'm testing manually the WIP around the Google Oauth provider and seems to be working.

rest/rest-resources/pom.xml
```diff
@@ -220,10 +215,9 @@
                 <configuration>
                     <instructions>
                         <Import-Package>
-                            org.apache.karaf.jaas.config,
-                            org.eclipse.jetty.jaas,
                             *
                         </Import-Package>
+                        <Export-Package>org.apache.brooklyn.*</Export-Package>
```
aledsage: Is this belt-and-braces, or does it contain public things in packages other than `org.apache.brooklyn.*`?

jcabrerizo:
I've been testing with @richardcloudsoft the Web UI and the BR command line and it works perfectly using `AnyoneSecurityProvider` and `ExplicitUsersSecurityProvider`. The web interface also works with no more problems than the logout whit the `OauthSecurityProvider` but unfortunately, the command line cannot login in.

tbouron:
@jcabrerizo You might find this read useful. It explains how to make any CLI work within an oAuth flow (spoiler, the CLI creates a small http server when it needs to)

The article uses PHP but it applies in the same way for any other language: https://developer.okta.com/blog/2018/07/16/oauth-2-command-line

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/filter/BrooklynSecurityProviderFilterHelper.java
```diff
@@ -73,43 +83,90 @@
...
-                HttpSession s = m.getHttpSession(requestedSessionId);
-                if (s!=null) {
-                    log.trace("SESSION found for {}: {} (valid={})", webRequest.getRequestURI(), s, m.isValid(s));
-                    return s;
+        if (webRequest instanceof ServletRequestWrapper) {
+            webRequest = (HttpServletRequest) ((ServletRequestWrapper)webRequest).getRequest();
+        }
+        if (webRequest instanceof Request) {
+            if (sharedSessionHandler==null || !sharedSessionHandler.isRunning()) {
+                synchronized (BrooklynSecurityProviderFilterHelper.class) {
```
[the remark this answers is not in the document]
ahgittin: good spot, forgot to declare it volatile; done now

rest/rest-resources/pom.xml
ahgittin: not sure i understand the question; this bundle provides 5 or so packages, and most of them are things other projects might want to use, so we export those 5 packages.  does that help?

ahgittin:
This seems to be working well, except for one thing.  The original session handler is still in the handler stack, and when it detects the session we set (or when something detects it), that session handler is caching the session even though it didn't create it.  Calls to `session.invalidate()` then freeze because it cycles through all the handlers and expects the sessions all to be distinct.  This appears to be a real problem if you visit multiple webapps in the UI and then logout.

I am working on fixing it in #1032 .

(The volatile fix will go in there also FWIW, not here.)

ahgittin:
closing this in favour of #1032 -- the `invalidate()` error was a real problem.  that now does it a different way.

PR #1034, opened 2019-01-30
Fix dependency versions

aledsage:
Bump versions to match those in karaf 4.2.2, and cxf 3.2.7

And remove exclusions` for winrm4j, which are not needed and make things harder for downstream projects.

The version bumps should not change what actually happens at runtime: these higher numbered bundles will be the ones that export their packages, because they are added by cxf etc. The consequences of depending on the wrong (older) version are: 1) we compile against a different version from that used at runtime; 2) it bloats the distro size and memory usage (because both versions are present); and 3) it cab contribute towards re-wiring problems at startup (where bundles are stopped and started because different versions are being added).


ahgittin:
Also did a full build with #1032 and its downstreams and confirmed the result works well.  Did not test winrm there however.

Merging this.

PR #1036, opened 2019-02-08
Fix session cleaner

jcabrerizo:
Jetty 9.4 HouseKeeper cleans the session each 10 minutes by default, that is a problem if you are more than 10 minutes in a module browse to other even the MultiSessionAttributeAdapter tries to give you the correct previous session where the Oauth token is stored.
This increase the time scavenge interval and re-send the CSRF token each time.
Also changed the NO CONNECTION message in UI: https://github.com/apache/brooklyn-ui/pull/122


ahgittin:
two fixes.  also could you remove the `log.debug` i accidentally committed on line 218 ?  :)  thx

ahgittin:
looks good @jcabrerizo, merging (will change log level on line 218 as above -- it was my mistake anyway!)

PR #1041, opened 2019-02-11
[Security] Bump commons-compress from 1.4 to 1.18

kemitix:
Bumps commons-compress from 1.4 to 1.18. **This update includes security fixes.**
<details>
<summary>Vulnerabilities fixed</summary>

*Sourced from [The Sonatype OSS Index](https://ossindex.sonatype.org/vuln/c30fadee-a5fc-47e5-9b28-3bd160719296).*

> **[CVE-2012-2098]  Cryptographic Issues**
> Algorithmic complexity vulnerability in the sorting algorithms in bzip2 compressing stream (BZip2CompressorOutputStream) in Apache Commons Compress before 1.4.1 allows remote attackers to cause a denial of service (CPU consumption) via a file with many repeating inputs.
> 
> Affected versions: <= 1.4.0

</details>

PR #1042, opened 2019-02-12
bump commons-compress

ahgittin:
from 1.4 to 1.4.1 to avoid being flagged as a CVE

(both very old versions, not tested with 1.18, but 1.4.1 has the fix in question)


ahgittin:
safer change than #1041 but if that works i'd prefer it -- rerunning tests there now (failures look spurious)

ahgittin:
going with #1041 instead

PR #1038, opened 2019-02-11
[WIP] [Security] Bump xstream from 1.4.8 to 1.4.11.1

kemitix:
Bumps [xstream](https://github.com/x-stream/xstream) from 1.4.8 to 1.4.11.1. **This update includes security fixes.**
<details>
<summary>Vulnerabilities fixed</summary>

*Sourced from [The Sonatype OSS Index](https://ossindex.sonatype.org/vuln/764af3f0-05d8-4a8d-9421-1d51ed8f2fae).*

> **[CVE-2017-7957]  Improper Input Validation**
> XStream through 1.4.9, when a certain denyTypes workaround is not used, mishandles attempts to create an instance of the primitive type 'void' during unmarshalling, leading to a remote application crash, as demonstrated by an xstream.fromXML("<void/>") call.
> 
> Affected versions: <= 1.4.9

*Sourced from [The Sonatype OSS Index](https://ossindex.sonatype.org/vuln/325b0ce9-1324-4bb8-820d-032aaaf1a8ef).*

> **[CVE-2016-3674]  Information Exposure**
> Multiple XML external entity (XXE) vulnerabilities in the (1) Dom4JDriver, (2) DomDriver, (3) JDomDriver, (4) JDom2Driver, (5) SjsxpDriver, (6) StandardStaxDriver, and (7) WstxDriver drivers in XStream before 1.4.9 allow remote attackers to read arbitrary files via a crafted XML document.
> 
> Affected versions: <= 1.4.8

</details>
<details>
<summary>Commits</summary>

- See full diff in [compare view](https://github.com/x-stream/xstream/commits)
</details>

PR #1043, opened 2019-02-12
Use servicemix bundle of xstream at runtime

kemitix:
The updated xstream includes dependencies on libraries that are not
OSGi bundles. This servicemix repackaging, includes these libraries
within the bundle.

Follows on from #1038

### DIFF
--- a/karaf/features/src/main/feature/feature.xml
+++ b/karaf/features/src/main/feature/feature.xml
@@ -155,7 +155,7 @@
         <bundle dependency="true">mvn:org.eclipse.jetty/jetty-server/${jetty.version}</bundle>
         <bundle dependency="true">mvn:org.ops4j.pax.web/pax-web-spi/${pax-web.version}</bundle>
         <bundle dependency="true">mvn:org.apache.geronimo.specs/geronimo-ws-metadata_2.0_spec/${geronimo-ws-metadata_2.0_spec.version}</bundle>
-        <bundle dependency="true">mvn:com.thoughtworks.xstream/xstream/${xstream.version}</bundle>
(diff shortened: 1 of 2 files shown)


ahgittin:
resulting assembly works fine - great fix @kemitix , merging

jcabrerizo:
I get this message in the console: 
```
Security framework of XStream not initialized, XStream is probably vulnerable.
```
It's not more vulnerable than before, but have this trace may be a problem

kemitix:
This explains the sort of thing we would need to add to fix this: https://stackoverflow.com/a/45152845

Specifically:
```
XStream.setupDefaultSecurity(this); // to be removed after 1.5
xstream.allowTypesByWildcard(new String[] {
    "com.your.package.**"
});
```

We would need to know which packages would need to be be added, and where to set this.

jcabrerizo:
I did a fast test adding `org.apache.brooklyn` and it I couldn't run Brooklyn after that. Better leave this as is for the moment and make a detailed analysis later

duncangrant:
@kemitix @jcabrerizo please track this as a bug on issues.apache.org - I'd add it but I seem to be unable to get past the captcha.

jcabrerizo:
https://issues.apache.org/jira/secure/RapidBoard.jspa?rapidView=50&projectKey=BROOKLYN&view=detail&selectedIssue=BROOKLYN-609

ahgittin:
Solution is probably to set `*` in a `cfg` file in the dist and to read this on management initialization.  People can edit this subsequently.

To avoid the errors in test and non-osgi the code can say if there is no such `cfg` fie it sets `*` as a default with a log debug message.

PR #1044, opened 2019-02-12
Added default security config for XStream

jcabrerizo:
Added default security config for Xstream deserialization.


ahgittin:
LGTM.  Confirmed this is the only place we do `new XStream()`.

In future we may make this configurable, but as we restrict so only privileged users can specify which classes can be put into places where they will be serialized (ie adding them to catalog) we already have the protection this constraint affords.

PR #1046, opened 2019-02-13
allow the types of parameters to be extended programmatically

ahgittin:
and misc comments to clarify some immediate resolution semantics

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/objs/BasicSpecParameter.java
+++ b/core/src/main/java/org/apache/brooklyn/core/objs/BasicSpecParameter.java
@@ -27,6 +27,7 @@ import java.util.Date;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.function.BiFunction;
 
 import org.apache.brooklyn.api.catalog.CatalogConfig;
 import org.apache.brooklyn.api.entity.Entity;
@@ -53,6 +54,7 @@ import org.apache.brooklyn.util.collections.MutableList;
 import org.apache.brooklyn.util.collections.MutableMap;
 import org.apache.brooklyn.util.collections.MutableSet;
 import org.apache.brooklyn.util.guava.Maybe;
+import org.apache.brooklyn.util.text.Strings;
 import org.apache.brooklyn.util.time.Duration;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -73,7 +75,7 @@ public class BasicSpecParameter<T> implements SpecParameter<T>{
     private static final long serialVersionUID = -4728186276307619778L;
 
     private static final Logger log = LoggerFactory.getLogger(BasicSpecParameter.class);
-    
+        
     private final String label;
 
     /** pinning may become a priority or other more expansive indicator */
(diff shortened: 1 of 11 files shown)


ahgittin:
also allow order of extended type coercions to be customised

core/src/main/java/org/apache/brooklyn/core/objs/BasicSpecParameter.java
```diff
@@ -197,6 +199,15 @@ public String toString() {
         return type;
     }
 
+    /** Allows extensions to define additional types that are supported for parameters in YAML.
+     * Rules should return a TypeToken if the given name is known, null if not handled, or 
+     * throw if the name is decisively problematic. Handlers will throw if no rules match. 
+     */
+    @Beta
+    public static void addCustomTypeNameInference(String ruleName, BiFunction<String,BrooklynClassLoadingContext,TypeToken<?>> rule) {
```
aledsage: I can see why you've done this with a static field, but it feels like a code smell. Downsides are testability (e.g. a unit test can never reverse the effects of calling this method - all subsequent tests executed could be affected by what it did); interaction between entities/modules (e.g. someone could write an entity that calls this, which would affect all subsequent blueprints); and reasoning about the code (e.g. other stateful stuff in brooklyn will be stored by the management context (or some such), rather than it being a static field).

However, you've marked it `@Beta` so I'm ok with going with this approach - the code is certainly shorter (and thus simpler) than alternatives.

core/src/main/java/org/apache/brooklyn/core/objs/BasicSpecParameter.java
```diff
@@ -197,6 +199,15 @@ public String toString() {
...
+        ParseYamlInputs.customTypeNameInferencing.put(ruleName, rule);
+    }
+    
     private static final class ParseYamlInputs {
```
aledsage: Feels a bit strange to have a bunch of static methods on this class rather than in a utility class (but I see lots of static methods already exist here, so not suggesting we change it here).

Your new methods would *not* be suitable for a utility because it is stateful (updating the static field).

core/src/main/java/org/apache/brooklyn/core/objs/BasicSpecParameter.java
aledsage: I don't see anything calling this method - do we need to add unit test(s)?

utils/common/src/main/java/org/apache/brooklyn/util/javalang/coerce/CommonAdaptorTryCoercions.java
```diff
@@ -46,16 +46,16 @@ public CommonAdaptorTryCoercions(TypeCoercerExtensible coercer) {
...
 
     public CommonAdaptorTryCoercions registerAllAdapters() {
-        registerAdapter(new TryCoercerWithFromMethod());
-        registerAdapter(new TryCoercerToEnum());
-        registerAdapter(new TryCoercerToArray(coercer));
-        registerAdapter(new TryCoercerForPrimitivesAndStrings());
+        registerAdapter("11-with-from-method", new TryCoercerWithFromMethod());
+        registerAdapter("12-enum", new TryCoercerToEnum());
+        registerAdapter("13-toArray", new TryCoercerToArray(coercer));
+        registerAdapter("15-primitives", new TryCoercerForPrimitivesAndStrings());
```
aledsage: On balance, I agree with this.

However, looking at the code the numbers feel hard to understand - e.g. why is 14 missing, is that important or a distraction. Might any other class have also claimed the 10s? If another class is added that calls registerAdapter, how to know whether something has already claimed the 10s, 20s, 30s, etc.

Where such numbering conventions work well, I think everything with those numbers is in a common place (e.g. init.d type of things), so you can tell at a glance if there is a conflict and where your number fits into the sequence.

ahgittin:
Good comments -- agree with them and at some point a further cleanup might be appropriate (perhaps linking coercions to the `ManagementContext` or `BrooklynLoadingContext`) but as noted I went with the much simpler extension to what we have.  Tests passing so I'll merge.

PR #1049, opened 2019-02-26
Retry flaky unit tests

tbouron:
We know that some of our unit tests are flaky. Testng has support for retrying tests by implementing `IRetryAnalyzer` interface, then reference that for selected tests.

This does exactly that, and add the retry strategy to the usual suspects.

As a side note, these flaky tests are because of races conditions, or test environments not being properly setup/cleanup. **We should definitely fix the root cause!** But for now, retrying maximum 3 times would eliminate false positive.

### DIFF
--- a/utils/test-support/src/main/java/org/apache/brooklyn/test/support/FlakyRetryAnalyser.java
+++ b/utils/test-support/src/main/java/org/apache/brooklyn/test/support/FlakyRetryAnalyser.java
@@ -1,0 +1,37 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.brooklyn.test.support;
+
+import org.testng.IRetryAnalyzer;
+import org.testng.ITestResult;
+
+public class FlakyRetryAnalyser implements IRetryAnalyzer {
+    private int retryCount = 0;
+    private static final int maxRetryCount = 3;
+
+    @Override
+    public boolean retry(ITestResult result) {
+        if (retryCount < maxRetryCount) {
+            retryCount++;
+            return true;
+        }
+        return false;
+    }
+
+}
--- a/utils/test-support/src/main/java/org/apache/brooklyn/test/support/FlakyRetryAnalyser.java
+++ b/utils/test-support/src/main/java/org/apache/brooklyn/test/support/FlakyRetryAnalyser.java
@@ -22,13 +22,14 @@ import org.testng.IRetryAnalyzer;
 import org.testng.ITestResult;
 
 public class FlakyRetryAnalyser implements IRetryAnalyzer {
-    private int retryCount = 0;
-    private static final int maxRetryCount = 3;
+    private static final int MAX = 3;
+
+    private int count = 0;
 
     @Override
     public boolean retry(ITestResult result) {
-        if (retryCount < maxRetryCount) {
-            retryCount++;
+        if (count < MAX) {
+            count++;
             return true;
         }
         return false;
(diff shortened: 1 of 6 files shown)


utils/test-support/src/main/java/org/apache/brooklyn/test/support/FlakyRetryAnalyser.java
```diff
@@ -0,0 +1,37 @@
...
+ * under the License.
+ */
+package org.apache.brooklyn.test.support;
+
+import org.testng.IRetryAnalyzer;
+import org.testng.ITestResult;
+
+public class FlakyRetryAnalyser implements IRetryAnalyzer {
+    private int retryCount = 0;
+    private static final int maxRetryCount = 3;
```
kemitix: Constants, `private static final`, are usually all caps: `MAX_RETRY_COUNT`.
Within the context of a 'retryAnalyzer' you could get away with `count` and `MAX`. They are both private.

tbouron: Good point @kemitix, code is updated

geomacy:
`testErrorsCaughtByApiAndRestApiWorks` is still breaking for me 😭

tbouron:
Hey @geomacy. So we spent some time on this and it turns out that it is a problem specifically on MacOS. It was introduced when the version of `felix.framework.version` was bumped to `5.6.10`.

Long story short, this version changed how Felix framework handles extension bundles. It appends bundles to the classpath as they're installed.

So when e.g. a test in `org.apache.brooklyn.launcher.WebAppRunnerTest` runs all of the extension bundles are cached and added to the classpath. When the test completes the cache directory is deleted.

But when a second test runs, EmbeddedFelixFramework discovers the bundles from the last test which triggers:
```
2018-11-30 16:43:44,041 WARN  o.a.b.r.f.EmbeddedFelixFramework [main]: Unable to install manifest from jar:file:/Users/sam/.brooklyn/osgi/cache/wcx9QUag/bundle175/version0.0/bundle.jar!/META-INF/MANIFEST.MF: java.lang.IllegalStateException: Problem installing extension bundle jar:file:/Users/sam/.brooklyn/osgi/cache/wcx9QUag/bundle175/version0.0/bundle.jar!/META-INF/MANIFEST.MF: java.io.FileNotFoundException: /Users/sam/.brooklyn/osgi/cache/wcx9QUag/bundle175/version0.0/bundle.jar (No such file or directory)
java.lang.IllegalStateException: Problem installing extension bundle jar:file:/Users/sam/.brooklyn/osgi/cache/wcx9QUag/bundle175/version0.0/bundle.jar!/META-INF/MANIFEST.MF: java.io.FileNotFoundException: /Users/sam/.brooklyn/osgi/cache/wcx9QUag/bundle175/version0.0/bundle.jar (No such file or directory)
	at org.apache.brooklyn.rt.felix.EmbeddedFelixFramework.installExtensionBundle(EmbeddedFelixFramework.java:212) [brooklyn-rt-felix-1.0.0-SNAPSHOT.jar:1.0.0-SNAPSHOT]
	at org.apache.brooklyn.rt.felix.EmbeddedFelixFramework.installBootBundles(EmbeddedFelixFramework.java:143) [brooklyn-rt-felix-1.0.0-SNAPSHOT.jar:1.0.0-SNAPSHOT]
	at org.apache.brooklyn.rt.felix.EmbeddedFelixFramework.newFrameworkStarted(EmbeddedFelixFramework.java:106) [brooklyn-rt-felix-1.0.0-SNAPSHOT.jar:1.0.0-SNAPSHOT]
	at org.apache.brooklyn.rt.felix.EmbeddedFramework.getFramework(EmbeddedFramework.java:30) [brooklyn-rt-felix-1.0.0-SNAPSHOT.jar:1.0.0-SNAPSHOT]
	at org.apache.brooklyn.util.core.osgi.Osgis.getFramework(Osgis.java:301) [brooklyn-core-1.0.0-SNAPSHOT.jar:1.0.0-SNAPSHOT]
	at org.apache.brooklyn.core.mgmt.ha.OsgiManager.start(OsgiManager.java:273) [brooklyn-core-1.0.0-SNAPSHOT.jar:1.0.0-SNAPSHOT]
	at org.apache.brooklyn.core.mgmt.internal.LocalManagementContext.<init>(LocalManagementContext.java:184) [brooklyn-core-1.0.0-SNAPSHOT.jar:1.0.0-SNAPSHOT]
	at org.apache.brooklyn.launcher.common.BasicLauncher.initManagementContext(BasicLauncher.java:539) [brooklyn-launcher-common-1.0.0-SNAPSHOT.jar:1.0.0-SNAPSHOT]
	at org.apache.brooklyn.launcher.BrooklynLauncher.initManagementContext(BrooklynLauncher.java:241) [classes/:na]
	at org.apache.brooklyn.launcher.common.BasicLauncher.startPartOne(BasicLauncher.java:394) [brooklyn-launcher-common-1.0.0-SNAPSHOT.jar:1.0.0-SNAPSHOT]
	at org.apache.brooklyn.launcher.common.BasicLauncher.start(BasicLauncher.java:379) [brooklyn-launcher-common-1.0.0-SNAPSHOT.jar:1.0.0-SNAPSHOT]
	at org.apache.brooklyn.launcher.WebAppRunnerTest.testStartWithLauncher(WebAppRunnerTest.java:143) [test-classes/:na]
```

geomacy:
Wow @tbouron, that's an obscure one, good catch. It's strange, if it's to do with that felix framework change, that the problem only happens on MacOS.

What to do though? I guess we don't want to say "well just don't build on Mac".

PR #1056, opened 2019-03-18
SECURITY Bump jackson-databind to 2.9.8

kemitix:
* [SECURITY] Bump `jackson-databind` from 2.9.7 to 2.9.8

Due to transitive dependencies involving `jackson-databind` the following upgrades were also required:
* Bump `cxf.version` from 3.2.7 to 3.2.8
* Bump `jetty.version` from 9.4.12.v20180830 to 9.4.14.v20181114 
* Bump `winrm4` from 0.6.1 to 0.7.0

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -163,7 +163,7 @@
         <jax-rs-api.version>2.1.1</jax-rs-api.version>
         <maxmind.version>2.8.0-rc1</maxmind.version>
         <maxmind-db.version>1.2.1</maxmind-db.version>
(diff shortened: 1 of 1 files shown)


ahgittin:
alternatively we could do excludes in `jetty`, `cxf`, and `winrm4j` -- we've been hit with big changes though admittedly in minor bumps not patch bumps

as these are patch bumps and tests pass i'm merging, let's see how the `dist` works with it cc @aledsage

PR #1062, opened 2019-04-01
[SECURITY] change Executors.newCachedThreadPool() to newFixedThreadPool(int)

bd2019us:
Since the size of 'sockets' can be large, cached thread pool has a high risk in running out of memory and causes the program crashing. Using newFixedThreadPool() is more secure and should be preferred.


tbouron:
Thanks for the PR @bd2019us, LGTM.

Although, I would prefer having another opinion from someone like @aledsage or @ahgittin before merging.

aledsage:
Thanks @bd2019us and @tbouron - LGTM as well. Merging.

PR #1064, opened 2019-08-26
Winrm improvements

ahgittin:
pass `shell.env` fields to winrm machines, and use `INSTALL_DIR` and `RUN_DIR` similar to as done for linux


nakomis:
LGTM; reviewed and (live) tested

PR #1065, opened 2019-09-06
take build number from submodules' git sha1

ahgittin:
instead of the root module, so that the SHA-1 written to manifests accurately reflects the state of the code that built it

### DIFF
--- a/parent/pom.xml
+++ b/parent/pom.xml
@@ -775,7 +775,8 @@
                     <artifactId>buildnumber-maven-plugin</artifactId>
                     <version>1.3</version>
                     <configuration>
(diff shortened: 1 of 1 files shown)


ahgittin:
in particular this means that things which inherit `brooklyn-downstream-parent` and use or are used as submodules get the right SHA-1 in their JAR's `META-INF/MANIFEST.MF`

ahgittin:
build failure is that `mvn` home dir not found on build server -- config problem w jenkins https://builds.apache.org/job/brooklyn-server-pull-requests/3125/ cc @asf-ci as there is nothing odd or different about our config from when it has worked in the past AFAIK

duncangrant:
retest this please

PR #1066, opened 2019-10-02
JcloudsLocation: don’t expose cloud identity in toString()

alasdairhodge:
Although it's not strictly secret, IMO it'd be preferable not to expose the `identity` field in `JcloudsLocation.toString()`, which does pop up in a few random places in the web UI. Display name and ID ought to be sufficient.

### DIFF
--- a/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
+++ b/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
@@ -312,12 +312,21 @@ public class JcloudsLocation extends AbstractCloudMachineProvisioningLocation im
 
     @Override
     public String toString() {
-        Object identity = getIdentity();
+        String className = getClass().getSimpleName();
         String configDescription = config().getLocalBag().getDescription();
-        if (configDescription!=null && configDescription.startsWith(getClass().getSimpleName()))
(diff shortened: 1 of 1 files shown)


ahgittin:
What about instead if `Strings.isNonBlank(getDisplayName())` then that's the only thing we show inside the `[`...`]`, whereas if it is blank we show all the other details (identity, config description, id)?

This way in a well-configured env user will only see nice things and with good names it will be unambiguous, but in other uses we aren't hiding info that is useful to identify the location -- specifically if I have multiple AWS locations and I haven't given them names I think the identify is useful.

geomacy:
> What about instead if `Strings.isNonBlank(getDisplayName())` then that's the only thing we show inside the `[`...`]`, whereas if it is blank we show all the other details (identity, config description, id)?
> 
> This way in a well-configured env user will only see nice things and with good names it will be unambiguous, but in other uses we aren't hiding info that is useful to identify the location -- specifically if I have multiple AWS locations and I haven't given them names I think the identify is useful.

+1, sounds good to me

geomacy:
nudge ꜛ @alasdairhodge ... :-)

nakomis:
@alasdairhodge Can you take a look at this please to see if it is still relevant, and address the comments above if appropriate

Thanks

ludofarine:
@alasdairhodge - will you get a chance to take a look at this, this week?
Thanks

PR #1067, opened 2019-11-18
improved templates support, and jackson bump

ahgittin:
templates can be given a string, for cases where it isn't valid YAML.
(note even without that, if templates include comments even not as a string, they are preserved).

also bump jackson, exclude some deps in case versions conflict.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -122,7 +122,7 @@
         <xpp3.servicemix.version>1.1.4c_7</xpp3.servicemix.version>
         <kxml2.servicemix.version>2.3.0_3</kxml2.servicemix.version>
         <!-- double-check downstream projects before changing jackson version -->
-        <fasterxml.jackson.version>2.10.0.pr2</fasterxml.jackson.version> <!-- 2.9.9 matches cxf-jackson (from cxf-jaxrs), but 2.10 has better yaml support -->
+        <fasterxml.jackson.version>2.10.1</fasterxml.jackson.version> <!-- 2.9.9 matches cxf-jackson (from cxf-jaxrs), but 2.10 has better yaml support -->
         <cxf.version>3.2.8</cxf.version>
         <httpcomponents.httpclient.version>4.5.6</httpcomponents.httpclient.version> <!-- To match cxf-http-async -->
         <httpcomponents.httpcore.version>4.4.9</httpcomponents.httpcore.version> <!-- To match cxf-http-async -->
--- a/pom.xml
+++ b/pom.xml
@@ -122,7 +122,7 @@
         <xpp3.servicemix.version>1.1.4c_7</xpp3.servicemix.version>
         <kxml2.servicemix.version>2.3.0_3</kxml2.servicemix.version>
         <!-- double-check downstream projects before changing jackson version -->
-        <fasterxml.jackson.version>2.9.8</fasterxml.jackson.version> <!-- To match cxf-jackson (from cxf-jaxrs) -->
+        <fasterxml.jackson.version>2.10.0.pr2</fasterxml.jackson.version> <!-- 2.9.9 matches cxf-jackson (from cxf-jaxrs), but 2.10 has better yaml support -->
         <cxf.version>3.2.8</cxf.version>
         <httpcomponents.httpclient.version>4.5.6</httpcomponents.httpclient.version> <!-- To match cxf-http-async -->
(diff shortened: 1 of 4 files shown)


nakomis:
@ahgittin LGTM, tests pass. Happy for this to be merged

pom.xml
```diff
@@ -122,7 +122,7 @@
         <xpp3.servicemix.version>1.1.4c_7</xpp3.servicemix.version>
         <kxml2.servicemix.version>2.3.0_3</kxml2.servicemix.version>
         <!-- double-check downstream projects before changing jackson version -->
-        <fasterxml.jackson.version>2.9.8</fasterxml.jackson.version> <!-- To match cxf-jackson (from cxf-jaxrs) -->
+        <fasterxml.jackson.version>2.10.0.pr2</fasterxml.jackson.version> <!-- 2.9.9 matches cxf-jackson (from cxf-jaxrs), but 2.10 has better yaml support -->
```
aledsage: What better yaml support does 2.10 give?

Do we really need to upgrade `fasterxml.jackson.version` and `snakeyaml.version`? Things are simpler when our versions match the versions of things pre-shipped in karaf, or included in low-level features such as cxf.

It reduces the risk of weird startup problems where us loading a newer version causes lots of low-level bundles to be refreshed/rewired. Those startup problems can cause Karaf to fail to start if they are too slow.

CXF 3.3.2 (which is the version that goes with Karaf 4.2.7) ships with fasterxml 2.9.8 and with snakeyaml 1.24.

utils/rest-swagger/pom.xml
```diff
@@ -105,6 +105,24 @@
         <dependency>
             <groupId>com.fasterxml.jackson.dataformat</groupId>
             <artifactId>jackson-dataformat-xml</artifactId>
+            <exclusions>
+                <exclusion>
+                    <!-- jackson above v 2.10.0 wants 6.0 of woodstox, but cxf used elsewhere is only at 5.x -->
+                    <groupId>com.fasterxml.woodstox</groupId>
```
aledsage: We should be really careful about using these kinds of excludes. They only apply to the maven build.

Excluding this tell us that jackson 2.10.0 has been compiled against a different version of woodstox than is shipped in karaf (and in the Brooklyn distro), so we risk `NoSuchMethoError`s etc.

These kind of exclusions are a worry to me: it's safer to just use the same version, if possible.

Also, they only apply to the maven build and not to karaf. Depending on what the feature.xml says, we might end up with both or just the pre-installed version.

pom.xml
aledsage: Why `2.10.0.pr2`? Don't know what "pr" means, but it sounds like a release candidate.

`2.10.0` or `2.10.1` looks newer than `2.10.pr2` - https://search.maven.org/artifact/com.fasterxml.jackson.dataformat/jackson-dataformat-xml

ahgittin:
the new snakeyaml+jackson version fixes bugs around the handling of quotes and nulls which could affect blueprints

`2.10.pr2` was the best with new snakeyaml support when i opened this.  updated to `2.10.1`.

i agree there is a minor risk of incompatible woodstox xml methods but we don't actually use jackson for xml except possibly in swagger where this method is done so i view it as small risk and very easily validated.  problems with quotes and nulls being handled wrong were real so i hope this can be approved (esp now that we're on an official jackson release)

ahgittin:
note if karaf 4.2.7 uses CXF 3.3.2 we should bump that version as part of c4b8dbc60698828bd78a38315b4877ecc3829de2 ?

aledsage:
Thanks for the explanations @ahgittin .

We probably also use jackson for xml in WinRM (via winrm4j) - it uses CXF, so I think it uses cxf-jackson and thus code that is compiled against fasterxml.jackson 2.9.x. I'd want us to thoroughly test winrm4j (in Karaf) with these changes. But we could do that after merging. If it breaks, it will be a difficult judgement call to decide what to do - fight with winrm a lot or live with the blueprint-parsing bugs/limitations.

For upgrading CXF to 3.3.2, I've been looking into that. See email thread on brooklyn mailing list. We'll need a new release of winrm4j for that to work, which feels do'able. Whether it's definitely worth it is another matter, but the "hard work" of figuring out how to make it work is done.

By merging this PR first, it means we can test Brooklyn against the winrm4j release candidate.

Merging now.

PR #1073, opened 2019-12-05
Fix session error

jcabrerizo:
- This marks as used all the session with the same ID on the different session handlers
- Fix bug on the loop for setting and removing attributes on all session detected by Denis


ahgittin:
looks good to me

PR #1074, opened 2019-12-05
Upgrade jclouds to 2.1.2 (from 2.1.0)

aledsage:
jclouds 2.1.2 is the latest with a jclouds-karaf build (there isn't one yet for 2.1.3 or 2.2.0).

The biggest improvement (for me) in jclouds 2.1.2 [1] is support for "EC2 r5, t3, and x1 instance types".

I've smoke-tested this in the karaf distro by deploying a Windows entity to aws-ec2.

Note that `org.apache.jclouds.labs/azurecompute/${jclouds.version}` was removed in jclouds 2.1.2. That's fine - we only really want to support azure ARM rather than the ancient azure-classic.

[1] https://jclouds.apache.org/releasenotes/2.1.2/

### DIFF
--- a/karaf/features/src/main/feature/feature.xml
+++ b/karaf/features/src/main/feature/feature.xml
@@ -288,7 +288,6 @@
     </feature>
 
     <feature name="brooklyn-locations-jclouds-labs" version="${project.version}" description="Brooklyn JClouds Labs Location Targets">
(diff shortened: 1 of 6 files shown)


tbouron:
Thank you @aledsage. I tested basic VM deployments on AWS, Azure ARM and GCE and it works as expected so merging.

PR #1076, opened 2019-12-16
add time data to returned data

frogfather:
Adds creationTime key to returned data to allow ui to sort applications by creation time.
PR# 150 in brooklyn-ui implements the sorting.


ahgittin:
useful.  merging.

PR #1080, opened 2019-12-19
add check for missing endpoint

frogfather:
Azure-ARM locations with no defined `endpoint` led to extreme server performance issues at app deployment time. Eventually causes OOME, Jackson serialisation timeouts, can even hang Chrome sub-process.
Added a check for missing endpoint parameter when deploying to Azure-ARM locations to avoid this issue on the server.

### DIFF
--- a/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/AbstractComputeServiceRegistry.java
+++ b/locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/AbstractComputeServiceRegistry.java
@@ -57,7 +57,7 @@ public abstract class AbstractComputeServiceRegistry implements ComputeServiceRe
     private final Map<Map<?, ?>, ComputeService> cachedComputeServices = new ConcurrentHashMap<>();
 
     @Override
-    public ComputeService findComputeService(ConfigBag conf, boolean allowReuse) throws IllegalArgumentException {
+    public ComputeService findComputeService(ConfigBag conf, boolean allowReuse) {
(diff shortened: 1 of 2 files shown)


locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/AbstractComputeServiceRegistry.java
```diff
@@ -57,7 +57,7 @@
     private final Map<Map<?, ?>, ComputeService> cachedComputeServices = new ConcurrentHashMap<>();
 
     @Override
-    public ComputeService findComputeService(ConfigBag conf, boolean allowReuse) {
+    public ComputeService findComputeService(ConfigBag conf, boolean allowReuse) throws IllegalArgumentException {
```
tbouron: This method override the one from the super class, the signature shouldn't change. Actually, it doesn't need to as `IllegalArgumentException` is a runtime exception

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -612,6 +612,9 @@ public MachineLocation obtain(Map<?,?> flags) throws NoMachinesAvailableExceptio
         for (int i = 1; i <= attempts; i++) {
             try {
                 return obtainOnce(setup);
+            } catch (IllegalArgumentException e) {
+                LOG.warn("Attempt #{}/{} to obtain machine threw error: {}", new Object[]{i, attempts, e});
+                exceptions.add(e);
```
tbouron: Same as above, the catch block is unnecessary as `RuntimeException` is already caught below

PR #1081, opened 2020-01-06
Add support for use of DSL when specifying the pattern for $brooklyn:formatString()

tbouron:
Previously, the `$brooklyn:formatString()` didn't accept nested DSL for its pattern, i.e. the first parameter of the function. This fixes it.


tbouron:
Great, thanks @iuliana. Merging...

PR #1083, opened 2020-01-15
Tidy instantiation

ahgittin:
means it now tries the new "transformers" when installing things to catalog.
if that fails, it still does the legacy "plan-to-spec converters" so should be safe.

it also adds better logging for when the catalog is loading.


ahgittin:
Most changes originally in this PR were made via an alternative means, but there are some logging and test strictness improvements, and code tidy.  Merging.

PR #1085, opened 2020-02-10
Don’t prompt for confirmation with apt-get

kemitix:
`apt-get install` will try to prompt the user for confirmation when installing a package unless the `-y` parameter is provided. This causes the install to abort is there is no attached terminal.


tbouron:
LGTM, tried on Ubuntu 18.

PR #1091, opened 2020-04-20
Update library versions due to CVEs

duncangrant:
Fixes the following CVEs
https://nvd.nist.gov/vuln/detail/CVE-2017-5929
https://nvd.nist.gov/vuln/detail/CVE-2019-10086

### DIFF
--- a/test-support/src/main/java/org/apache/brooklyn/test/LogWatcher.java
+++ b/test-support/src/main/java/org/apache/brooklyn/test/LogWatcher.java
@@ -158,7 +158,7 @@ public class LogWatcher implements Closeable {
 
         PatternLayoutEncoder ple = new PatternLayoutEncoder() {
             @Override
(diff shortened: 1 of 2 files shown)


test-support/src/main/java/org/apache/brooklyn/test/LogWatcher.java
```diff
@@ -182,7 +182,7 @@ public void doEncode(ILoggingEvent event) throws IOException {
                 }
                 LOG.trace("level="+event.getLevel()+"; event="+event+"; msg="+event.getFormattedMessage());
```
iuliana, line 183: Any reason we are not using the recommended syntax like `LOG.trace("level= {} ; event={} ; msg={}", event.getLevel(), event,event.getFormattedMessage() );` ?

iuliana:
Approved. :)

PR #1096, opened 2020-05-06
Bumps cxf versions

nakomis:


rest/rest-resources/src/main/java/org/apache/brooklyn/rest/util/MultiSessionAttributeAdapter.java
```diff
@@ -217,7 +219,11 @@ private void invalidateAllSession(HttpSession preferredSession, HttpSession loca
         private HttpSession findValidPreferredSession(HttpSession localSession, HttpServletRequest optionalRequest) {
             if (localSession instanceof Session) {
                 SessionHandler preferredHandler = getPreferredJettyHandler((Session)localSession, true, true);
-                HttpSession preferredSession = preferredHandler==null ? null : preferredHandler.getHttpSession(localSession.getId());
+
+                SessionIdManager idManager = ((DefaultSessionIdManager)preferredHandler.getSessionIdManager());
```
aledsage, line 223: Why do we need to cast it to `DefaultSessionIdManager`?

aledsage: I see this code is mostly a copy of the impl of `sessionHandler.getHttpSession(String)`, which is now protected in jetty 9.4.22.

But your code does not call `session.setIdChanged(true)`, which that method does when the id doesn't match. I've no idea what the implications are of that difference!

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/util/MultiSessionAttributeAdapter.java
```diff
@@ -217,7 +219,14 @@ private void invalidateAllSession(HttpSession preferredSession, HttpSession loca
         private HttpSession findValidPreferredSession(HttpSession localSession, HttpServletRequest optionalRequest) {
             if (localSession instanceof Session) {
                 SessionHandler preferredHandler = getPreferredJettyHandler((Session)localSession, true, true);
-                HttpSession preferredSession = preferredHandler==null ? null : preferredHandler.getHttpSession(localSession.getId());
+
```
duncangrant, line 222: Needs and if preferredHAndler != null here

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/util/MultiSessionAttributeAdapter.java
```diff
@@ -217,7 +219,14 @@ private void invalidateAllSession(HttpSession preferredSession, HttpSession loca
...
+                SessionIdManager idManager = preferredHandler.getSessionIdManager();
+                String id = idManager.getId(localSession.getId());
+                HttpSession preferredSession = preferredHandler.getSession(id);
+
+                if (preferredSession != null && !((Session)preferredSession).getExtendedId().equals(id))
```
duncangrant, line 227: I think you're comparing to the wrong id here?  the localSession.getId returns the extended id and the idManager.getId returns the non-extended id

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/util/MultiSessionAttributeAdapter.java
```diff
@@ -217,7 +219,17 @@ private void invalidateAllSession(HttpSession preferredSession, HttpSession loca
         private HttpSession findValidPreferredSession(HttpSession localSession, HttpServletRequest optionalRequest) {
             if (localSession instanceof Session) {
                 SessionHandler preferredHandler = getPreferredJettyHandler((Session)localSession, true, true);
-                HttpSession preferredSession = preferredHandler==null ? null : preferredHandler.getHttpSession(localSession.getId());
+
+                HttpSession preferredSession
+                if (preferredHandler == null) {
```
duncangrant, line 224: should be !=null

nakomis: Doh! I'd written it for `== null` then changed the logic and forgot to change the check...

pom.xml
```diff
@@ -161,20 +161,21 @@
         <aopalliance.version>1.0</aopalliance.version>
         <commons-configuration.version>1.7</commons-configuration.version>
         <commons-lang.version>2.4</commons-lang.version>
-        <jax-rs-api.version>2.1.1</jax-rs-api.version>
+        <jax-rs-api.version>2.1.1</jax-rs-api.version> <!-- differs from jclouds 2.1.2, which depends on v2.0.1 -->
         <maxmind.version>2.8.0-rc1</maxmind.version>
         <maxmind-db.version>1.2.1</maxmind-db.version>
-        <winrm4j.version>0.8.0</winrm4j.version>
+        <winrm4j.version>0.9.0-SNAPSHOT</winrm4j.version> <!--  FIXME NO CHECK IN -->
```
aledsage, line 167: @nakomis please change this to 0.9.0, which has just been released (to be compatible with cxf 3.3.5)

PR #1099, opened 2020-05-13
BROOKLYN-627: fix okhttp/kubernetes startup issue

aledsage:
See description in https://issues.apache.org/jira/browse/BROOKLYN-627, which this is intended to fix.

I wrote a little script to loop round, starting Brooklyn and tearing it down again (deleting the data dir each time) to see if I could reproduce the error reported in BROOKLYN-627. I could not do so locally, and it continued to work after this fix!

I think it's worth including this fix, speculatively, as it could plausibly fix the non-deterministic issue.


iuliana:
LGTM

PR #1104, opened 2020-06-02
policy/sensor-feed flexibility and highlights

ahgittin:
be more forgiving of http response data and sensor type, and tidy highlights and logging


ahgittin:
retest this please

PR #1105, opened 2020-07-08
Helm CLI based kubernetes deployment

duncangrant:
This is a hopefully simple first attempt at using the helm cli to do a kubernetes deployment via Brooklyn.
It requires the helm cli be available on the classpath and uses the existing KubernetesLocation.

In the future I hope to integrate better with our current kubernetes work to let us use some of the existing entities and extend those to match other kubernetes types.


ahgittin:
Nice!  Merging.

PR #1106, opened 2020-08-10
New ChildrenBatchEffector

jcabrerizo:
Adding a new effector to call a inner effector in all the children entities where the effector is inserted in batches of a parametrized size.
It accept also a couple of parameters for define if it should continue or fail when the child effector doesn't exist and for failing and skipt next if one of the inner effector taks fails. An of course a parameter for define the name of the child effector o be called
This example blueprint shows the idea
```
services:
    name: Test ChildrenBatchEffector

    brooklyn.initializers:
      - type: org.apache.brooklyn.core.effector.util.ChildrenBatchEffector
        brooklyn.config:
          name: ChildrenBatchEffectorInParent
          effectorToInvoke: testEffector
          failOnMissingEffector: true

    brooklyn.children:
      - type: org.apache.brooklyn.entity.stock.BasicEntity
        name: child one
        brooklyn.initializers:
          - type: org.apache.brooklyn.core.effector.http.HttpCommandEffector
            brooklyn.config:
              name: testEffector
              uri: "http://localhost.io:8081/v1/applications"
              httpUsername: admin
              httpPassword: password
              httpVerb: get

      - type: org.apache.brooklyn.entity.stock.BasicEntity
        name: child two
        brooklyn.initializers:
          - type: org.apache.brooklyn.core.effector.http.HttpCommandEffector
            brooklyn.config:
              name: testEffector
              uri: "http://localhost.io:8081/v1/applications"
              httpUsername: admin
              httpPassword: password
              httpVerb: get
```

If you call the effector from the UI it will show you this dialog with the fields filled with the information from the blueprint
image

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/effector/util/ChildrenBatchEffector.java
+++ b/core/src/main/java/org/apache/brooklyn/core/effector/util/ChildrenBatchEffector.java
@@ -1,0 +1,183 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/brooklyn/core/effector/util/ChildrenBatchEffector.java
```diff
@@ -0,0 +1,188 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Effector to  invoking an effector across children or members,
+ */
+public class ChildrenBatchEffector implements EntityInitializer {
```
ahgittin, line 50: extending `AddEffectorInitializerAbstract` is better, provides standard way to define name, description (missing here), and parameters.  also it ensures things are serialized and deserialized nicely.

ahgittin:
:+1:

PR #1108, opened 2020-09-07
Unwrap entity specs when used from the DSL

ahgittin:
At present if you refer to an entity which is trivially wrapped to make an application, in a `$brooklyn:entitySpec` DSL, the application is added without being unwrapped.  This is inconsistent with behaviour elsewhere, eg in the UI for "add child" and when a trivially-app-wrapped entity is referred to in `brooklyn.children`.

(FYI a "trivially-app-wrapped" entity is what you get if you say `services: [ { type: my-entity } ]` or in other places where an entity is supplied but the code needs an application and so constructs a trivial BasicApplication instance with the entity as a child.)

This PR corrects the above behaviour, so that eg such a trivially-wrapped-app can be used in a dynamic cluster and the entity is used as the spec, not the application.

Note that it is straightforward to suppress this unwrapping, just add a no-op initializer or enricher to the `Application`, and that will always prevent unwrapping.


duncangrant:
I've tested  this in a few scenarios and works well - thanks @ahgittin

PR #1110, opened 2020-09-21
There are shared utility functions no longer public

duncangrant:
This makes AddEffectorInitializerAbstractProto public as a recent change
had made it private.  Alternatively we could have pushed the utility
functions to the implementation.


iuliana (approved):

ahgittin:
Can we do the same for the other new `*Proto` classes (if needed) and make any private or package-private functions `protected` (eg `newEffectorBuilder`).

Otherwise LGTM - good catch.

iuliana (approved):

PR #1111, opened 2020-09-29
ensure good errors about beans, and some other transforms, are returned to caller

ahgittin:


duncangrant (approved):
LGTM thanks @ahgittin

PR #1114, opened 2020-10-05
add a post-init hook interface for entities

ahgittin:
For use to add extra validation logic after all initializers have run.


iuliana (approved):
Approved. 
But, I would preffer we start using default methods to avoid testing entity types and conversions.

PR #1116, opened 2020-10-20
More on catalog bundle resolvers - support format, compatibility fixes

ahgittin:
support `format` on the REST API for catalog bundles and deployment plans.

this can give better error messages than auto-detect, if auto-detect picks the wrong type.

also make the osgi bundle version logic compatible with logic where a BOM was added, allowing `id` to indicate bundle name.


iuliana (approved):
LGTM

PR #1117, opened 2020-10-22
Improve search path logic for edge cases

ahgittin:
See first commit for the failing tests, then second commit for the fix and clarifications.


iuliana (approved):
LGTM

PR #1119, opened 2020-10-27
DSL serialization and deserialization

ahgittin:
brooklyn DSL expressions could get mangled when used within a bean-with-type; this improves some small details of how DSL is serialized and deserialized by Jackson for bean with type.  it also improves `WrappedValue` deserialization and handling of `$brooklyn:literal` (deferred until needed)


iuliana (approved):
LGTM

PR #1121, opened 2020-11-09
Resolve some edge cases in dealing with bean with type config 

ahgittin:
Allow bean-with-type conversion as part of type coercion.  Maintain data as maps in config a little bit longer.

Major rework of `CustomTypeConfigYamlTest` to do a much better job of enumerating different test cases.


iuliana (commented):
LGTM

PR #1122, opened 2020-11-10
Winrmtaskfactory - Similar to SshTaskFactory

duncangrant:
So this gives us a WinrmTaskFactory which can be used in the same way as the sshtaskfactory. Some generalisations of MachineLocation etc were required which I think is a good thing.

e.g.

                WinRmTasks.newWinrmExecTaskFactory(winRmMachineLocation, "my command").runAsCommand().summary();


duncangrant (commented):
This is still WIP (because I've found a couple of mistakes and github won't let me request changes of myself)

PR #1123, opened 2020-11-11
Misc minor - recording SSH track task, and $brooklyn:object accept string

ahgittin:
the latter allows us to supply custom objects and then chain additional DSL syntax eg

`$brooklyn:object("[my.CustomType]").attributeWhenReady(...)`

using the normal map argument this is impossible:

```
$brooklyn:object:
  type: my.CustomType
# where does attributeWhenReady(...) hang?
```


ahgittin:
One more tweak, now when deserialization any `WrappedValue` that contains a DSL expression, we parse the DSL.

PR #1124, opened 2020-11-12
Winrmtaskfactory

duncangrant:
So this gives us a WinrmTaskFactory which can be used in the same way as the sshtaskfactory. Some generalisations of MachineLocation etc were required which I think is a good thing.

e.g.

                WinRmTasks.newWinrmExecTaskFactory(winRmMachineLocation, "my command").runAsCommand().summary();


ahgittin:
some great improvements here.  merging.

PR #1125, opened 2020-11-17
Better error reporting for bad bundles, and fix some minor serialization errors revealed

ahgittin:
Introduces scoring conventions to reduce noise and to pick the best error(s)


duncangrant (approved):
LGTM
I built and tested this locally

PR #1127, opened 2020-11-25
allow location registry to listen for resolvers

ahgittin:
using osgi ServiceListener -- previously it loaded it once so wouldn't find things added after it was initialized


duncangrant (approved):
LGTM thanks @ahgittin

PR #1128, opened 2020-11-30
add interfaces to MachineManagementMixins for shutdown/startup/reboot…

ahgittin:
… and for getting metrics

implementation using jclouds does suspend, but as a placeholder for other providers to do true stops/starts

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/location/MachineLifecycleUtils.java
+++ b/core/src/main/java/org/apache/brooklyn/core/location/MachineLifecycleUtils.java
@@ -1,0 +1,288 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.brooklyn.core.location;
+
+import com.google.common.base.Preconditions;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.function.Function;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import org.apache.brooklyn.api.location.*;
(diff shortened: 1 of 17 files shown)


locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -20,6 +20,7 @@
 
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
+import org.apache.brooklyn.api.location.MachineManagementMixins.MachineMetadata;
```
grkvlt, line 23: Reorg imports?

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
```diff
@@ -1256,6 +1257,39 @@ public JcloudsMachineLocation resumeMachine(Map<?, ?> flags) {
...
+        if (!(l instanceof JcloudsSshMachineLocation)) {
+            throw new IllegalStateException("Cannot startup machine "+l+"; wrong type");
+        }
+        getComputeService().rebootNode(((JcloudsSshMachineLocation)l).getJcloudsId());
+        return l;
+    }
+
+    @Override
+    public Map<String, Object> getMachineMetrics(MachineLocation l) {
+        return MutableMap.of();
```
grkvlt, line 1285: Why mutable? I would assume there is no expectation the recipient can modify this?

grkvlt (changes requested):
LGTM modulo two comments

locations/jclouds/src/main/java/org/apache/brooklyn/location/jclouds/JcloudsLocation.java
ahgittin: a subclass might

ahgittin:
some minor adjustments to make it better handle management/tracking of locations across suspend/resume and shutdown/restart cycles

@grkvlt thx for the early PR comments, addressed/responded

core/src/main/java/org/apache/brooklyn/core/location/MachineLifecycleUtils.java
```diff
@@ -0,0 +1,288 @@
...
+                String m1id = ((GivesMachineMetadata)p1).getMachineMetadata(m1).getId();
+                String m2id = ((GivesMachineMetadata)p2).getMachineMetadata(m2).getId();
+                if (Objects.equals(m1id, m2id)) {
+                    if (m1id!=null) {
+                        primarilySame = true;
+                    }  else {
+                        // indeterminate
+                    }
+                } else {
+                    primarilySame = false;
```
duncangrant, line 261: Could return false here?

duncangrant:
@ahgittin LGTM - I've created a PR to fix https://github.com/apache/brooklyn-library/pull/182 - obviously close this if you don't want it

duncangrant (approved):

PR #1130, opened 2020-12-02
improve logging esp around catalog initialization and bundle load

ahgittin:
suppress benign warnings and debug messages, esp when guessing the resolved type; add debug logging for what is being analyzed; more coercion flexibility and logback customization


iuliana (approved):
LGTM

PR #1131, opened 2020-12-03
Add support for downloading bundle as ZIP files

tbouron:
This adds a new REST API endpoint to the `BundleResource`, which returns the bundle as a ZIP file when invoked


iuliana (approved):
LGTM.

PR #1137, opened 2021-01-15
Use the jax-rs annotations to recognize correctly parameters 

tbouron:
When POST requests to `/v1/applications` are made, the beta endpoint do not behave as expected. This is because the form parameters were annotated with only the swagger annotation, not jax-rs, making them effectively useless.

Instead, jax-rs took the first parameter as the body (which is its default) and ignored the others.


iuliana (commented):
LGTM

PR #1139, opened 2021-01-19
Pretty print winrm xml

duncangrant:
Winrm writes xml to the error stream when powershell commands are run.  This adds an inline pretty printer to the xml so that when it is displayed on the UI it is readable.  Currently it's one single line.


nakomis:
LGTM; merging

PR #1141, opened 2021-01-21
Allow registered types for config etc, and support validation

ahgittin:
This allows the `BrooklynJacksonType` which implements `java.reflect.Type` (cf `Class`) via the Jackson `SimpleType` to be used not only for deserializing Brooklyn `RegisteredType` instances (which had been the case) but also as something which can be passed to a Guava `TypeToken` mostly sensibly.  (There are a few places where we need to introduce our own methods, in our `TypeTokens` class, where the Guava `TypeToken` doesn't treat the underlying `BrooklynJacksonType` quite right, and unfortunately `TypeToken` declares most of its methods as `final` so we can't easily extend it.)

This means `ConfigKey` and `Sensor` objects defined in YAML can be declared as type anything that is a Brooklyn `RegisteredType` and it will be created as that registered types.  (Previously it would be downcast to its fundamental Java type.).

This also introduces a validation subsystem allowing custom validation to be registered on any object via `BrooklynValidation` so that custom types can, if the transformer supports it, apply constraints.  Validation is run when (1) config contraints checked, or (2) WrappedValues retrieved, or (3) on coercion.


iuliana (commented):
LGTM

PR #1142, opened 2021-01-25
Better validation for config keys

ahgittin:
Previously the enforcement of config key constraints was spotty -- there was an explicit check in a few places, such as entity initialization -- but many places it was skipped.

This now does validation when individual keys are set (although not if a map is set, and not if it is a deferred supplier / future / etc) _and_ when keys are retrieved.  Where config keys are declared by a context (eg an entity) the declared key is looked up and used for constraints, so validation is performed even if anonymous (untyped) config keys are used to set or retrieved.

Extensive tests added.


duncangrant (approved):
LGTM thanks @ahgittin

PR #1144, opened 2021-02-02
Fixes issue reverting osgi bom files with SNAPSHOT versions

grkvlt:
Fix `install()` of SNAPSHOT versioned bundles so that checksums are never compared, and bundles are always updated. This is because a SNAPSHOT bundle uses the checksum of the first installation to determine the URL and saves this as metadata, which is never changed.

Signed-off-by: Andrew Donald Kennedy <andrew.kennedy@cloudsoftcorp.com>

### DIFF
--- a/camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/catalog/CatalogScanOsgiTest.java
+++ b/camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/catalog/CatalogScanOsgiTest.java
@@ -26,8 +26,12 @@ import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileNotFoundException;
 import java.io.FileOutputStream;
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
 import java.util.Map;
 import java.util.zip.ZipEntry;
+import java.util.zip.ZipFile;
+import java.util.zip.ZipOutputStream;
 
 import org.apache.brooklyn.api.mgmt.ManagementContext;
 import org.apache.brooklyn.api.typereg.ManagedBundle;
@@ -46,6 +50,7 @@ import org.apache.brooklyn.util.os.Os;
 import org.apache.brooklyn.util.osgi.OsgiTestResources;
 import org.apache.brooklyn.util.stream.Streams;
 import org.apache.brooklyn.util.text.Strings;
+import org.apache.tools.ant.taskdefs.Zip;
 import org.testng.Assert;
 import org.testng.annotations.Test;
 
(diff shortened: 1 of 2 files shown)


duncangrant (dismissed):
LGTM thanks @grkvlt

ahgittin:
hold fire on this -- failing test?

core/src/main/java/org/apache/brooklyn/util/core/osgi/BundleMaker.java
```diff
@@ -269,7 +269,8 @@ private void writeZipEntriesFromFile(ZipOutputStream zout, File existingZip, Pre
     @Deprecated
     public Bundle installBundle(File f, boolean start) {
         try {
-            Bundle b = Osgis.install( framework, "file://"+f.getAbsolutePath() );
+
+            Bundle b = Osgis.install( framework, f.toURI().toString() );
```
ahgittin, line 273: suspicious

core/src/test/java/org/apache/brooklyn/core/config/external/PropertiesFileExternalConfigSupplierTest.java
```diff
@@ -68,11 +68,6 @@ public void testFromPropertiesJavaFileUrl() throws Exception {
         doTestFromProperties(f -> f.toURI().toString());
     }
 
-    @Test
-    public void testFromPropertiesTwoSlashAndPathUrl() throws Exception {
```
ahgittin, line 72: this is a useful test

launcher/src/test/java/org/apache/brooklyn/launcher/BrooklynLauncherRebindCatalogOsgiTest.java
```diff
@@ -254,8 +254,8 @@ public void testRebindWithSystemBundleInCatalog() throws Exception {
             // Launch brooklyn, where persisted state now includes the initial catalog's bundle
             startT2(newLauncherForTests(initialBomFile.getAbsolutePath()));
             
-            // Should not have replaced the original "system bundle"
-            assertOnlyBundle(reusedFramework, bundleName, pseudoSystemBundle);
+            // FIXME Should we have replaced the original "system bundle" ???
+            assertBundleExists(reusedFramework, bundleName, pseudoSystemBundle);
```
ahgittin, line 258: feels wrong

ahgittin:
The problem here is that if we have a snapshot bundle, say `foo:1-SNAPSHOT`, and we add it when it's in state A, then add when in state B, and then try to add state A again.  The last add is ignored because it matches a previous state, when actually it should match the most recent state.

I think this needs specific logic to compare with the checksum of the most recent version of a bundle:version, bail out if those match.  And then extra logic if it matches a previous known state of a bundle:version then it restores that (this. might be a new result code).

This also needs a test.

core/src/main/java/org/apache/brooklyn/util/core/osgi/BundleMaker.java
grkvlt: removed

core/src/test/java/org/apache/brooklyn/core/config/external/PropertiesFileExternalConfigSupplierTest.java
grkvlt: reinserted

launcher/src/test/java/org/apache/brooklyn/launcher/BrooklynLauncherRebindCatalogOsgiTest.java
grkvlt: reverted test

grkvlt:
FYI, there is another way of doing this, which involves updating the metadata with the new checksum and URL each time, but this caused a test failure and also had the result of keeping old versions around but leaving them unmanaged and confusing brooklyn and osgi, so i gave up on that eventually ;(

ahgittin:
good job identifying the problem.  re solution we _do_ have to update the bundle metadata, otherwise we keep installing snapshots even when they are unchanged which is disruptive.  re-installation should be idempotent.  added a test for that and a fix for it in https://github.com/apache/brooklyn-server/pull/1176 .  merged this and that.  but @grkvlt if you get the chance to eyeball that PR post-merge that would be great.

PR #1145, opened 2021-02-03
Fixes tests failing on Windows

grkvlt:
Windows cannot just prepend `file://` to the path returned from a `File` so going via `toURI()` seems more sensible...


ahgittin:
LGTM

PR #1147, opened 2021-02-09
Pass newly created streams to winrm

duncangrant:
Extra streams are created to handle the different windows powershell streams but these were not being passed to winrm.  This fixes that.


core/src/main/java/org/apache/brooklyn/util/core/task/system/ProcessTaskWrapper.java
```diff
@@ -49,10 +47,9 @@
     private final Task<RET> task;
 
     // execution details
-    private ByteArrayOutputStream stdout;
-    private ByteArrayOutputStream stderr;
```
ahgittin, line 53: i deliberately left `stdout` and `stderr` in case they were needed for deserialization.  but we don't normally persist tasks so this should be fine -- but one to remember.

PR #1153, opened 2021-03-08
Dsl edge cases

iuliana:
Support for new CAMP DSL, using  []. 

The [x] approach =>  x is tried (in order):

-     an argument to a `get(x)` method (works for lists and maps: index or key)
-     a bean property (if x is a string, look for method called getX() or a field called  x
-     a config key, if the target is Configurable , ie getConfig(x) or config().get(x)


iuliana:
Put it back to work in progress because the algorithm fails for `$brooklyn:object(\"[brooklyn.tosca.ToscaEntityFinder,host]\").config(\"ips_container\")[\"ips\"][0]"`.

duncangrant (commented):
LGTM @iuliana thanks

PR #1154, opened 2021-03-12
Extend Java API for creation to support dry-run and to report-but-not-fail on constraint violations

ahgittin:
Useful where we want to see whether there are any runtime validation problems in the spec but not actually start stuff


duncangrant (approved):
LGTM thanks @ahgittin

PR #1157, opened 2021-03-30
Misc tidies to support building a KAR and running without errors

ahgittin:


rest/rest-resources/src/main/java/org/apache/brooklyn/rest/util/MultiSessionAttributeAdapter.java
```diff
@@ -638,12 +638,22 @@ public MultiSessionAttributeAdapter resetExpiration() {
...
+                if (ss != null) {
                     ss.setMaxInactiveInterval(maxInativeInterval);
                 }
             }
         }
         return this;
     }
+
+    private static Session getSessionSafely(Handler h, String id) {
+        if (((SessionHandler)h).getSessionCache()==null) {
```
algairim, line 651: Check if Handler is an instance of SessionHandler first, log debug/warning message "{h} is not a session handler" and return null.

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/util/MultiSessionAttributeAdapter.java
```diff
@@ -638,12 +638,22 @@ public MultiSessionAttributeAdapter resetExpiration() {
...
+            // suppress the log warning that the call to getSession can trigger, if racing during startup
+            log.debug("Skipping session reset expiration for "+id+" on "+h+" because session cache not initialized (yet)");
```
algairim, line 653: The logger should allow to parametrise the message like the following:
 `           log.debug("Skipping session reset expiration for {} on {} because session cache not initialized (yet)", id, h);`

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/util/MultiSessionAttributeAdapter.java
```diff
@@ -638,12 +638,22 @@ public MultiSessionAttributeAdapter resetExpiration() {
...
+                Session ss = getSessionSafely(h, getId());
+                if (ss != null) {
                     ss.setMaxInactiveInterval(maxInativeInterval);
                 }
             }
         }
         return this;
     }
+
+    private static Session getSessionSafely(Handler h, String id) {
```
algairim, line 650: Parameters are not expected to be modified, can be final.

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/util/MultiSessionAttributeAdapter.java
ahgittin: +1 worth adding now that it is refactored

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/util/MultiSessionAttributeAdapter.java
ahgittin: The logger does allow that, but (personal preference) I find that decreases readability and leads to bugs where the order of parameters gets out of synch with the message, so I only do that where the savings in skipping the string concatenation are likely to be significant when debug mode is off.  (Or alternatively do a `log.isDebugEnabled()` check.). This block should only happen rarely during init so not worth it IMO.

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/util/MultiSessionAttributeAdapter.java
ahgittin: compiler will spot that as an optimization; in terms of method behaviour it's not a big deal if the method changes so the variables are non-final so not worth marking IMO

ahgittin:
Thanks for good review @algairim -- one bit addressed, other two responded two.

PR #1160, opened 2021-04-09
for icons, support referencing the bundle and the type name not just the type name

ahgittin:
support icons for type within a given bundle, and prefer icon set in the item over one inherited


iuliana:
Looks good to me, I will merge it.

iuliana:
Can we add some tests?

ahgittin:
I've updated tests in 956ad1552407295cbf4985327567b844fe4d5fe0 to include the new endpoint

PR #1161, opened 2021-04-12
SMART-141 Do not split WinRM into separate streams, keep all in one

algairim:
Signed-off-by: Mykola Mandra <mykola.mandra@cloudsoftcorp.com>


ahgittin:
looks good.  will review UI side before merging for good measure.

algairim:
Test changes will follow in this PR, merge is ready after.

ahgittin:
looking at the UI PR https://github.com/apache/brooklyn-ui/pull/196 you are showing the win/powershell filter controls on any stream that looks like powershell CLI XML.  so here i think we don’t need a special stream for winrm , we can just use the default behaviour of sending it in stderr … and the UI will figure out it to render it as CLI XML.

algairim:
Adjusted unit-test and live test for a single WinRM stream.

camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/AbstractWindowsYamlTest.java
```diff
@@ -67,7 +67,7 @@ protected void assertStreams(SoftwareProcess entity, Map<String, List<String>> s
 
             String stdin = getStreamOrFail(subTask, BrooklynTaskTags.STREAM_STDIN);
             String stdout = getStreamOrFail(subTask, BrooklynTaskTags.STREAM_STDOUT);
-            String stderr = getStreamOrFail(subTask, BrooklynTaskTags.STREAM_STDERR);
+            String stderr = getStream(subTask, BrooklynTaskTags.STREAM_STDERR);
```
algairim, line 70: stderr is optional in WinRM, same as env.

software/winrm/src/test/java/org/apache/brooklyn/util/core/internal/winrm/winrm4j/PrettyXmlWriterTest.java
```diff
@@ -134,6 +134,24 @@ public void testIgnoreComment() throws IOException {
         assertEquals(recordingWriter.out.toString(), "#< CLIXML\n<t1>\n\t<t2>\n\t\t<t3/>\n\t</t2>\n</t1>");
     }
 
+    @Test
+    public void testWriteInChunks() throws IOException {
+        String xml = "<t1><t2>A</t2></t1>";
+        for (int breakAt = 0; breakAt < xml.length(); breakAt++) {
+            if (breakAt == 15) continue;
```
algairim, line 141: There is one more test-case, forgot to verify it. Code change for this to follow.

algairim: This is addressed in the new commit.

algairim:
Changes in XML formatter address issues of odd offsets in CLI XML content supplied in the WinRM stream.

PR #1162, opened 2021-04-16
Fix intermittent unmanagement race

ahgittin:
And rename tests from `clocker` to `onthefly` because the pattern is more useful than just as used in that downstream project.


ahgittin:
there could be a race between mgmt starting for an entity and our explicit unmanagement call -- only happened 1 in 50 times in the Clocker...Test.  this seems to fix it.

algairim (approved):

iuliana:
LGTM

PR #1163, opened 2021-04-16
don't interrupt tasks related to a deletion/unmanagement

ahgittin:
allow tasks to unmanage the entity where they are running; improve the logic which was meant to allow this

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/mgmt/internal/LocalEntityManager.java
+++ b/core/src/main/java/org/apache/brooklyn/core/mgmt/internal/LocalEntityManager.java
@@ -594,13 +594,25 @@ public class LocalEntityManager implements EntityManagerInternal {
         // try forcibly interrupting tasks on managed entities
         Collection<Exception> exceptions = MutableSet.of();
         try {
+            boolean inTaskForThisEntity = entity.equals(BrooklynTaskTags.getContextEntity(Tasks.current()));
+            Task rootTask = null;
             Set<Task<?>> tasksCancelled = MutableSet.of();
             for (Task<?> t: managementContext.getExecutionContext(entity).getTasks()) {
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/brooklyn/core/mgmt/internal/LocalEntityManager.java
```diff
@@ -650,6 +662,15 @@ private boolean hasTaskAsAncestor(Task<?> t, Task<?> potentialAncestor) {
         return hasTaskAsAncestor(t.getSubmittedByTask(), potentialAncestor);
     }
 
+    private Task<?> getRootTask(Task<?> t) {
+        Task<?> result = t;
+        while (t!=null) {
+            result = t;
+            t = t.getSubmittedByTask();
```
algairim, line 669: Break the loop  if `getSubmittedByTask` returns same object reference as `t`.

algairim (changes requested):

core/src/main/java/org/apache/brooklyn/core/mgmt/internal/LocalEntityManager.java
ahgittin: it is logically impossible and should be considered an illegal model for a task to claim to submit itself (or to have any recursive cycle here).  it would be ugly here if either of those happened (endless loop here) but it really shouldn't, easy to identify with `jstack`, and IMO isn't worth extra checks or tracking here to guard against an illegal model.

algairim: I look at documentation
```
    /** task which submitted this task, if was submitted by a task */`
    public Task<?> getSubmittedByTask();
```
If task cannot submit itself then this should be fine of course.

algairim (approved):

PR #1174, opened 2021-05-13
Ldap login improvements

duncangrant:
This is a set of, what I hope are, improvements around using ldap / a domain controller for sign on.  It provides a login page, rather than the basic auth log in popup, handles multiple ldap realms, and fixes some odd issues in how we handle username/password based authentication.


rest/rest-resources/src/main/java/org/apache/brooklyn/rest/filter/BrooklynSecurityProviderFilterJersey.java
```diff
@@ -19,6 +19,7 @@
 package org.apache.brooklyn.rest.filter;
 
 import java.io.IOException;
+import java.net.URI;
```
jcabrerizo, line 22: This import is not used

jcabrerizo (commented):
LGTM @duncangrant good work

jcabrerizo (approved):

PR #1175, opened 2021-05-21
Added decorateWithHierarchySpecTag(..) method to be used to create 

iuliana:
the hierarchical spec tag when deploying blueprints.

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/typereg/RegisteredTypes.java
+++ b/core/src/main/java/org/apache/brooklyn/core/typereg/RegisteredTypes.java
@@ -333,7 +333,20 @@ public class RegisteredTypes {
     @Beta
     public static RegisteredType addTag(RegisteredType type, Object tag) {
         if (tag!=null) {
-            ((BasicRegisteredType)type).tags.add( tag );
+            if (tag instanceof Map) {
+                Map<String,Object> mapTag = (Map<String,Object>) tag;
+                if(mapTag.containsKey(BrooklynTags.YAML_SPEC_HIERARCHY)) {
+                    Map<String,String> hierarchySpecTag = (Map<String,String>) mapTag.get(BrooklynTags.YAML_SPEC_HIERARCHY);
+                    BrooklynTags.SpecTag currentSpecTag = new BrooklynTags.HierarchySpecTagBuilder()
+                            .format(hierarchySpecTag.get("format"))
+                            .summary(hierarchySpecTag.get("summary"))
+                            .contents(hierarchySpecTag.get("contents"))
+                            .build();
+                    ((BasicRegisteredType)type).tags.add( currentSpecTag );
+                }
+            } else {
+                ((BasicRegisteredType) type).tags.add(tag);
+            }
         }
         return type;
     }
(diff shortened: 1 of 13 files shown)


duncangrant (approved):

core/src/main/java/org/apache/brooklyn/core/typereg/RegisteredTypes.java
```diff
@@ -333,7 +332,20 @@ public static RegisteredType addAliases(RegisteredType type, Iterable<String> al
     @Beta
     public static RegisteredType addTag(RegisteredType type, Object tag) {
         if (tag!=null) {
-            ((BasicRegisteredType)type).tags.add( tag );
+            if (tag instanceof Map &&( (Map<String,Object>) tag).containsKey(BrooklynTags.YAML_SPEC_HIERARCHY)) {
```
ahgittin, line 335: special-casing this tag here feels wrong.  feels like the logic belongs in `BrooklynTags` or `HierarchySpecTag`.

ahgittin:
nice work.  minor comment can be dealt with later.

not 100% sure we have all the data we want for context but we can iterate from here.  there is one a closely related feature i've wanted which is to indicate on a spec which ancestor brought it in; this could be handled by a `level_in_parent_yaml` field in your hierarchy spec tag.  might add that.

ahgittin:
i get two related test failures though:

```
Failed tests: 
  EntityResourceTest.testGetSpecAcceptingXYaml:93->testGetSpec:126 expected [name: ent-with-spec
services:
- type: org.apache.brooklyn.entity.stock.BasicEntity] but found []
  EntityResourceTest.testGetSpecWithNoExplicitAccepts:98->testGetSpec:126 expected [name: ent-with-spec
services:
- type: org.apache.brooklyn.entity.stock.BasicEntity] but found [[{"format":"brooklyn-camp","summary":"brooklyn-camp plan for null","contents":"name: ent-with-spec\nservices:\n- type: org.apache.brooklyn.entity.stock.BasicEntity"}]]

Tests run: 251, Failures: 2, Errors: 0, Skipped: 0
```

rest/rest-resources/src/test/java/org/apache/brooklyn/rest/resources/EntityResourceTest.java
```diff
@@ -114,11 +114,11 @@ protected void testGetSpec(String acceptMimeType) throws Exception {
         // Retrieve the yaml spec, and confirm it is as expected (not wrapped in quotes, and treating \n sensibly)
         Response response;
         if (acceptMimeType != null) {
-            response = client().path("/applications/" + appName + "/entities/" + appName + "/spec")
+            response = client().path("/applications/" + appName + "/entities/" + appName + "/speclist")
```
ahgittin, line 117: this is used in tests which expect the response from `/spec`; that's the test failures.  easy to fix of course!  i'll do that.

iuliana: @ahgittin I might have alread done that and not pushed. I reemember running the build and noticed the test failure too. I did that temporary modification to check that my API was called.

ahgittin:
yeah we've all done that!  simple fix, included in merge to master.  :)

PR #1176, opened 2021-05-26
Fix/snapshot bom update

ahgittin:
Build on https://github.com/apache/brooklyn-server/pull/1144 expanding the tests to cover one missed case, then fixing that case.


ahgittin:
a few minor fixes on #1144 - merging this along with that

PR #1178, opened 2021-06-01
Fix issue where vault not ready on init

duncangrant:
If vault is used for security provider and is not ready on startup then
the security provider will fail and will be replaced by the
BlackholeSecurityProvider. This change means that whenever a security
provider fails, the DelegatingSecurityProvider will attempt to
re-initialize it on subsequent authorization attempts.

This change does mean that, when failing over to the BlackholeSecurityProvider, it could be slower to let no one authenticate.

I have added a second change as I noticed a fairly impressive memory leak in how the security provider was used.  Again I'm pretty sure no behaviour was changed here.


algairim:
Looks good.

PR #1179, opened 2021-06-02
Fixed implementation to re-order spec tag values

iuliana:
Added code and explanatory comments.


ahgittin:
aha, this makes sense, because any spec_hierarchy tags provided for a catalog item represent what the registered type plan is converted _from_ -- as opposed to that tag's meaning for an instantiated entity where that list represents what the type plan is converted _to_ or _derived from_.  good spot.  merging.

PR #1180, opened 2021-06-03
Fix login form

duncangrant:
Logout was not redirecting correctly.  This adds a header to the
login form to the 401 api response so that it can be a) ignored by the
cli and b) followed by the UI.

### DIFF
--- a/rest/rest-resources/src/main/java/org/apache/brooklyn/rest/filter/BrooklynSecurityProviderFilterJersey.java
+++ b/rest/rest-resources/src/main/java/org/apache/brooklyn/rest/filter/BrooklynSecurityProviderFilterJersey.java
@@ -44,6 +44,7 @@ import org.slf4j.LoggerFactory;
 public class BrooklynSecurityProviderFilterJersey implements ContainerRequestFilter {
 
     private static final Logger log = LoggerFactory.getLogger(BrooklynSecurityProviderFilterJersey.class);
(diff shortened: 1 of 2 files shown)


iuliana (approved):
lgtm

algairim:
Looks ok

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/filter/BrooklynSecurityProviderFilterJersey.java
```diff
@@ -71,6 +74,11 @@ public void filter(ContainerRequestContext requestContext) throws IOException {
                     rin = Response.status(Status.UNAUTHORIZED).entity("Authentication is required").build();
                 }
             }
+            if (rin.getStatus()==Status.UNAUTHORIZED.getStatusCode() &&
+                    Strings.isNonBlank(mgmt.getConfig().getConfig(BrooklynSecurityProviderFilterJavax.LOGIN_FORM))) {
+                rin = Response.status(Status.UNAUTHORIZED).entity("Authentication is required").header(LOGIN_PAGE_HEADER, mgmt.getConfig().getConfig(BrooklynSecurityProviderFilterJavax.LOGIN_FORM)).build();
```
ahgittin, line 79: maybe cleaner if we make `rin` the `ResponseBuilder` and `build()` it when we return?  though the builder API might not support that.

ahgittin:
merging

PR #1181, opened 2021-06-07
Fix search path order and disallow duplicate symbolic names in a bundle

ahgittin:


duncangrant (approved):
LGTM @ahgittin thanks

PR #1182, opened 2021-06-07
Improve extra header name, cleanup NPE

duncangrant:
Additional header for when using login form now uses X_ to show it is
non-standard.

Handles NPE when server still starting.


iuliana (approved):
LGTM.

PR #1183, opened 2021-06-08
Feature - Terminated Node Deletion from persistence on startup

zan-mateusz:
Changes related to automatic terminated node removal on startup.
Requirements:
- new configuration property to specify threshold of inactivity after which terminated nodes are to be removed
- nodes removed from persistence on startup of the server

Implementation Notes:
- property in 'brooklyn.cfg' added: brooklyn.ha.timeoutForInactiveNodeRemovalOnStartup
- format of property is a duration, e.g 1h30m, 2d12h, 30s, etc.
- loadSyncRecord() loads the management node data from persistence store. If property above is specified (not ZERO) and node is terminated, compares if its inactivity period is higher than threshold and if so, removes from the persistence store.
- loadSyncRecord() is also periodically called during run so isStartup property introduced to ensure this code is fired only during startup

Testing:
- Tested locally:
-- terminated nodes correctly removed after inactivity period expires
-- terminated nodes correctly retained if inactivity period not exceeded
-- no terminated nodes removed in case without property present in brooklyn.cfg
-- non-terminated nodes correctly retained
-- other active servers correctly updated in the HA Status table (about page) 
-- tested as master and non-master nodes
-- nodes are not removed as server is running (startup only)

TODO
- Would be nice to introduce a unit test for this. I think this would involve creating a persistence store and making up some nodes in terminated state and simulating startup? Not sure how feasible is that and need to investigate further, but some pointers would be helpful.

### DIFF
--- a/api/src/main/java/org/apache/brooklyn/api/mgmt/ha/ManagementPlaneSyncRecordPersister.java
+++ b/api/src/main/java/org/apache/brooklyn/api/mgmt/ha/ManagementPlaneSyncRecordPersister.java
@@ -45,6 +45,7 @@ public interface ManagementPlaneSyncRecordPersister {
      * Note that this method is *not* thread safe.
      */
     ManagementPlaneSyncRecord loadSyncRecord() throws IOException;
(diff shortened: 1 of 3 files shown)


iuliana (approved):
LGTM

PR #1184, opened 2021-06-08
code for functionality to remove a particular node from persistence via API

zan-mateusz:
Back end side of Node Management enhancements

REQUIREMENTS
* UI to support setting state and priority for management servers (currently exposed via REST API only), eg to trigger a new master to run
* UI to show date of last seen for terminated servers (thought we had that but I didn't see it in screen shot) [only if this is easy to add!]
* invocation of code to remove files for terminated servers via explicit REST operation with UI button (TBC)

IMPLEMENTATION

1. Extra information (last active, priority) added to the table for each management node, status column renamed to ‘Information’. 
     * ‘Last active’ is based on remoteTimestamp, now passed in HA Summary.
     * Timestamp parsed into a ‘x ago’ with help of a custom filter using moment’s fromNow() function.

2. ‘Remove terminated nodes’ button and ‘remove’ button for a particular management node on row hover.
     * Remove button present for each node whose status is “TERMINATED”, when the row is hovered on with the mouse (can’t see in the screenshot as pointer is hidden).
     * For removing all terminated nodes, API call to “ha/states/clear” is made which clear all nodes apart from master and then repopulates the active nodes (essentially removing just the terminated ones).
     * Because of the above, in the refresh logic I introduced a concept of expectedNodeCounter, which ensures that we repopulate the active nodes before the table is refreshed (otherwise the operation would wipe all nodes apart from master and page would need to be refreshed to refresh the table correctly).
     * For removing a single node, I introduced another API “ha/states/clear/node” which takes a parameter - nodeId, ID of the node to be removed. In such case only 1 node is removed.

3. ‘Manage’ button available on hover over for the current management node (node can only manage itself).
     * Button click opens the modal that allows to modify the node (see below)
4. Modal view for node management once ‘manage’ button is clicked. 
     * List of parameters of class list-group-items and labels of class badge used for the node data (to be uniform with other sections in about page).
     * Edit button that allows to make changes and Cancel button to dismiss the modal and not apply any changes.
     * Clicking edit changes the labels into editable components for Status and Priority. Status is a drop down  select a new status from list ["MASTER", "STANDBY", "HOT_STANDBY", "HOT_BACKUP"], priority is a number-type input
     * Apply button disabled if no changes are made.
     * Once a change to either property gets introduced, the colour of the label changes to indicate that. Also ‘Apply’ button will then be enabled, which sends the API calls to make the changes in persistence.
     * When switching from non-MASTER to MASTER (or vice-versa), appropriate message is displayed with a warning message
     * Changing state to and from MASTER results in change of MASTER in the management context (note: this might take around a minute to load)
     * All changes are automatically reflected in the HA Status table with no need to refresh page.
5. For operations that take a little bit longer (removal of all nodes, changing status to MASTER), a spinner is introduced under the HA Status part of about page to indicate that data is loading. It is present on all operations.
6. All operations carried out update the persistence, so removing nodes, changing the properties is correctly saved back to persistence files

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/mgmt/ha/HighAvailabilityManagerImpl.java
+++ b/core/src/main/java/org/apache/brooklyn/core/mgmt/ha/HighAvailabilityManagerImpl.java
@@ -1019,13 +1019,8 @@ public class HighAvailabilityManagerImpl implements HighAvailabilityManager {
         return record; 
     }
 
-    private boolean updateLastManagementPlaneSyncRecordWithLocalKnowledge() {
-        if (lastSyncRecord != null) {
-            lastSyncRecord = updateManagementPlaneSyncRecordWithLocalKnowledge(lastSyncRecord);
-            return true;
-        } else {
-            return false;
-        }
+    private void updateLastManagementPlaneSyncRecordWithLocalKnowledge() {
+        if (lastSyncRecord != null) updateManagementPlaneSyncRecordWithLocalKnowledge(lastSyncRecord);
     }
 
     @Override
@@ -1059,7 +1054,7 @@ public class HighAvailabilityManagerImpl implements HighAvailabilityManager {
                 ManagementPlaneSyncRecord result = persister.loadSyncRecord(managementContext.getBrooklynProperties().getConfig(TIMEOUT_FOR_INACTIVE_NODE_REMOVAL_ON_STARTUP));
                 
                 if (useLocalKnowledgeForThisNode) {
-                    result = updateManagementPlaneSyncRecordWithLocalKnowledge(result);
+                    updateManagementPlaneSyncRecordWithLocalKnowledge(result);
                 }
                 
                 if (i>0) {
@@ -1081,7 +1076,7 @@ public class HighAvailabilityManagerImpl implements HighAvailabilityManager {
         throw new IllegalStateException(message, lastException);
     }
 
-    private ManagementPlaneSyncRecord updateManagementPlaneSyncRecordWithLocalKnowledge(ManagementPlaneSyncRecord result) {
+    private void updateManagementPlaneSyncRecordWithLocalKnowledge(ManagementPlaneSyncRecord result) {
         // Report this node's most recent state, and detect AWOL nodes
         ManagementNodeSyncRecord me = BasicManagementNodeSyncRecord.builder()
                 .from(result.getManagementNodes().get(ownNodeId), true)
@@ -1098,7 +1093,6 @@ public class HighAvailabilityManagerImpl implements HighAvailabilityManager {
             builder.masterNodeId(ownNodeId);
         }
         result = builder.build();
-        return result;
     }
 
     protected ManagementNodeSyncRecord createManagementNodeSyncRecord(boolean useLocalTimestampAsRemoteTimestamp) {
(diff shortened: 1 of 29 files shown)


algairim (changes requested):
Remove logbook button that was left from the merge. It is a directive at the bottom now.

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/resources/ServerResource.java
```diff
@@ -476,6 +477,16 @@ public Response clearHighAvailabilityPlaneStates() {
         return Response.ok().build();
     }
 
+    @Override
+    public Response clearHighAvailabilityPlaneStates(String nodeId) {
```
algairim, line 481: Unit-test for this in `ServerResourceTest` please.

core/src/main/java/org/apache/brooklyn/core/mgmt/ha/HighAvailabilityManagerImpl.java
```diff
@@ -1072,6 +1081,26 @@ private ManagementPlaneSyncRecord loadManagementPlaneSyncRecordInternal(boolean
...
+            allNodes = Iterables.transform(allNodes, new MarkAwolNodes(me));
+        Builder builder = ManagementPlaneSyncRecordImpl.builder()
+                .masterNodeId(result.getMasterNodeId())
+                .nodes(allNodes);
+        builder.node(me);
+        if (getTransitionTargetNodeState() == ManagementNodeState.MASTER) {
+            builder.masterNodeId(ownNodeId);
+        }
+        result = builder.build();
+        return result;
```
algairim, line 1101: Remove the `return` condition from the method signature. It is not required.

core/src/main/java/org/apache/brooklyn/core/mgmt/ha/HighAvailabilityManagerImpl.java
```diff
@@ -1072,6 +1081,26 @@ private ManagementPlaneSyncRecord loadManagementPlaneSyncRecordInternal(boolean
         throw new IllegalStateException(message, lastException);
     }
 
+    private ManagementPlaneSyncRecord updateManagementPlaneSyncRecordWithLocalKnowledge(ManagementPlaneSyncRecord result) {
```
algairim, line 1084: I would add javadoc saying that this method re-assigns the value of the supplied parameter. Otherwise, make it final if you prefer to return a new value from this method.

core/src/main/java/org/apache/brooklyn/core/mgmt/ha/HighAvailabilityManagerImpl.java
```diff
@@ -1004,6 +1019,15 @@ public ManagementPlaneSyncRecord loadManagementPlaneSyncRecord(boolean useLocalK
...
     }
 
+    private boolean updateLastManagementPlaneSyncRecordWithLocalKnowledge() {
+        if (lastSyncRecord != null) {
+            lastSyncRecord = updateManagementPlaneSyncRecordWithLocalKnowledge(lastSyncRecord);
+            return true;
+        } else {
+            return false;
+        }
+    }
```
algairim, line 1029: Do we need this method at all? Dose not look like return value is used anywhere.

algairim (changes requested):
Some tidy-up is required. Unit-tests and javadocs would be preferable.

algairim:
Looks good now, build failure is not related to these changes.
Unit-test to follow in a separate PR.

algairim (approved):

PR #1185, opened 2021-06-09
when publishing ha health, seems it doesn't update locally

ahgittin:
this now updates the local sync record whenever we persist

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/mgmt/ha/HighAvailabilityManagerImpl.java
+++ b/core/src/main/java/org/apache/brooklyn/core/mgmt/ha/HighAvailabilityManagerImpl.java
@@ -636,6 +636,7 @@ public class HighAvailabilityManagerImpl implements HighAvailabilityManager {
             Delta delta = ManagementPlaneSyncRecordDeltaImpl.builder().node(memento).build();
             persister.delta(delta);
             managementStateWritePersistenceMetrics.noteSuccess(Duration.of(timer));
+            updateLastManagementPlaneSyncRecordWithLocalKnowledge();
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/apache/brooklyn/core/mgmt/ha/HighAvailabilityManagerImpl.java
```diff
@@ -999,6 +1000,15 @@ public ManagementPlaneSyncRecord loadManagementPlaneSyncRecord(boolean useLocalK
         return record; 
     }
 
+    private boolean updateLastManagementPlaneSyncRecordWithLocalKnowledge() {
+        if (lastSyncRecord!=null) {
+            lastSyncRecord = updateManagementPlaneSyncRecordWithLocalKnowledge(lastSyncRecord);
+            return true;
+        } else {
+            return false;
```
algairim, line 1008: Any plans for the return value? Javadoc would help.

zan-mateusz: We don't need to return value for the particular scenario this change addresses (as it is then retrieved via a separate API call). In any case it can be retrieved by calling getLastManagementPlaneSyncRecord()  so I think it is not necessary, happy to be proven wrong of course

ahgittin:
@zan-mateusz you seem to have incorporated this into 3f9a63161b6d8fb4ab7b45e0f3cdf788fb28cfb6 .  ideally can you `git merge` work you're incorporating (and not rebase) that way this issue will close automatically when yours goes in.  and a comment here too :).

closing, as it's in master, with improvements -- thx for those!

PR #1186, opened 2021-06-10
Ldap login changes

duncangrant:
Updated login approach was a bit rubbish and doesn't really make sense.
New approach keeps the CN with configured realm and ou as before.
If the username contains a \ or @ then the username is used as the
security principal.  This allows NETBIOS\username or username@realm for
the username field.


iuliana (commented):
LGTM

PR #1187, opened 2021-06-10
Added support for tagged bundles

iuliana:
Tags added to the blueprint in composer, are set on the bundle when added to the catalog. 
This PR should be merged together with this one: https://github.com/apache/brooklyn-ui/pull/219

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/typereg/BrooklynBomYamlCatalogBundleResolver.java
+++ b/core/src/main/java/org/apache/brooklyn/core/typereg/BrooklynBomYamlCatalogBundleResolver.java
@@ -120,7 +120,7 @@ public class BrooklynBomYamlCatalogBundleResolver extends AbstractCatalogBundleR
                     null, BrooklynBomBundleCatalogBundleResolver.FORMAT,
                     null, null);
             // if the submitted blueprint contains tags, we set them on the bundle, so they can be picked up and used to tag the plan.
-            if( cm.containsKey("tags")) {
+            if( cm.containsKey("tags") && cm.get("tags") instanceof Iterable) {
                 basicManagedBundle.tags().addTags((Iterable<?>)cm.get("tags"));
             }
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/brooklyn/core/typereg/BrooklynBomYamlCatalogBundleResolver.java
```diff
@@ -115,10 +115,16 @@ protected double scoreForNullFormat(Supplier<InputStream> f) {
 
         OsgiBundleInstallationResult result;
         try {
+
+            BasicManagedBundle basicManagedBundle = new BasicManagedBundle(vn.getSymbolicName(), vn.getVersionString(),
+                    null, BrooklynBomBundleCatalogBundleResolver.FORMAT,
+                    null, null);
+            // if the submitted blueprint contains tags, we set them on the bundle, so they can be picked up and used to tag the plan.
+            if( cm.containsKey("tags")) {
+                basicManagedBundle.tags().addTags((Iterable<?>)cm.get("tags"));
```
algairim, line 124: Can `tags` value be a non-Iterable?

iuliana: Nope.

iuliana: @algairim  Tags should not be anything else but `Iterable`, but just in case somebody has made a mistake and put something else, we do not want to prevent the bundle from deploying. Tags are not really that important after all. So I added a condition for the tags to be added only when `Iterable`.

algairim: Ok, convinced.

algairim (approved):

PR #1188, opened 2021-06-11
Basic functionality for logstore interface

jcabrerizo:
This provides the backend for expose the server logs from two different logstore implementation:
- Static logfiles
- A ElasticSearch instance

The API and the LogStore interface only exposes a query method that expect the parameters from the UI (PR TBD) and returns a list of POJO representing each entry found on the configured LogStore implementation.

The configuration for each LogStore implementation has to be added to the `brooklyn.cfg` file:

For ElasticSearch:
```
brooklyn.logbook.logStore = org.apache.brooklyn.util.core.logbook.opensearch.OpenSearchLogStore
brooklyn.logbook.openSearchLogStore.host = https://localhost:9200
brooklyn.logbook.openSearchLogStore.index = brooklyn
brooklyn.logbook.openSearchLogStore.user = admin
brooklyn.logbook.openSearchLogStore.password = admin
brooklyn.logbook.openSearchLogStore.verifySSl = false
```
For the plain log file:
```
brooklyn.logbook.logStore = org.apache.brooklyn.util.core.logbook.file.FileLogStore
brooklyn.logbook.fileLogStore.path = /var/logs/brooklyn/brooklyn.debug.log
```

There are some pending tasks to be implemented:

- [ ] Add more unit test
- [ ] Improve the plain log implementation
- [ ] Add "tail" behaviour to keep updating the ui with new entries


core/src/main/java/org/apache/brooklyn/util/core/logbook/opensearch/BrooklynOpenSearchModel.java
```diff
@@ -0,0 +1,100 @@
...
+        }
+
+        public void setHits(List<OpenSearchHit> hits) {
+            this.hits = hits;
+        }
+    }
+
+    static class OpenSearchHit {
+        String _index;
+        String _type;
```
algairim, line 68: Is underscore necessary?

core/src/main/java/org/apache/brooklyn/util/core/logbook/opensearch/OpenSearchLogStore.java
```diff
@@ -0,0 +1,227 @@
...
+            BASE_NAME_OPEN_SEARCH_LOG_STORE + ".user", "User name");
+
+    public final static ConfigKey<String> LOGBOOK_LOG_STORE_PASS = ConfigKeys.newStringConfigKey(
+            BASE_NAME_OPEN_SEARCH_LOG_STORE + ".password", "User password");
+
+    public final static ConfigKey<String> LOGBOOK_LOG_STORE_APIKEY = ConfigKeys.newStringConfigKey(
+            BASE_NAME_OPEN_SEARCH_LOG_STORE + ".apikey", "API key");
+
+    public final static ConfigKey<Boolean> LOGBOOK_LOG_STORE_VERIFY_SSL = ConfigKeys.newBooleanConfigKey(
+            BASE_NAME_OPEN_SEARCH_LOG_STORE + ".verifySSl", "Verify SSL", true);
```
algairim, line 98: SSL property must have capital `L` at the end.

algairim: Or a camel-case `Ssl`

algairim:
Looks ok in general. Some JavaDocs would be really appreciated.

core/src/main/java/org/apache/brooklyn/util/core/logbook/opensearch/BrooklynOpenSearchModel.java
jcabrerizo: I use the same field name returned by elastic search. Alternatively the fields can be annotated with ` @JsonProperty("class")` but I preferred to keep it as it is unless it's not clear enough

core/src/main/java/org/apache/brooklyn/util/core/logbook/opensearch/OpenSearchLogStore.java
jcabrerizo: Good 🔍 I fixed the typo

core/src/main/java/org/apache/brooklyn/util/core/logbook/opensearch/BrooklynOpenSearchModel.java
algairim: It is implicit. It can be forgotten overtime and renamed back to traditional filed names during some refactoring, breaking the API as a result.

I suggest adding @JsonProperty with a Javadoc at the class level explaining this.

jcabrerizo: Good point @algairim I changed the field names

algairim: Looks good now.

core/src/main/java/org/apache/brooklyn/util/core/logbook/opensearch/OpenSearchLogStore.java
```diff
@@ -0,0 +1,228 @@
...
+        initialize();
+
+        HttpClientBuilder httpClientBuilder = HttpClientBuilder.create();
+        if (!verifySsl) {
+
+            final SSLContext sslContext;
+            try {
+                sslContext = SSLContextBuilder
+                        .create()
+                        .loadTrustMaterial(new TrustSelfSignedStrategy())
```
duncangrant, line 125: Should this be configurable?

core/src/main/java/org/apache/brooklyn/util/core/logbook/BrooklynLogEntry.java
```diff
@@ -0,0 +1,142 @@
...
+
+/**
+ * This class models the log format used by Apache Brooklyn by default
+ * @see `etc/org.ops4j.pax.logging.cfg`
+ */
+public class BrooklynLogEntry {
+
+    @JsonProperty("timestamp")
+    String timestampString;
+    Date datetime;
```
ahgittin, line 36: what is relationship between `datetime` and `timestampString` ?  worth a comment here?

core/src/main/java/org/apache/brooklyn/util/core/logbook/BrooklynLogEntry.java
```diff
@@ -0,0 +1,142 @@
...
+    String taskId;
+    String entityIds;
+    String level;
+    String bundleId;
+    @JsonProperty("class")
+    String clazz;
+    String threadName;
+    String message;
+
+    public String getTimestampString() {
```
ahgittin, line 46: (personal preference) i don't think getters/setters are worth their weight where they don't do anything; but not worth changing

core/src/main/java/org/apache/brooklyn/util/core/logbook/opensearch/OpenSearchLogStore.java
ahgittin: worth a project-wide review to make this configurable.  not appropriate IMO to figure out the project-wide solution as part of this PR.

ahgittin:
This looks great @jcabrerizo .  Super work on hard stuff.  And really nice to have!  One important thing, as part of startup it looks like API resource classes get added in three places; in addition to `rest/rest-resources/src/main/resources/OSGI-INF/blueprint/service.xml` there are:

```
rest/rest-resources/src/main/java/org/apache/brooklyn/rest/BrooklynRestApi.java
rest/rest-server/src/main/webapp/WEB-INF/web.xml
```

Would be nice to simplify that but easiest thing for now probably to be safe and consistent, add this `LogbookResource` in those places.

If you're busy I can do that.  But if you have the chance to do this, and add one comment (on `datetime`) to answer my question.

core/src/main/java/org/apache/brooklyn/util/core/logbook/BrooklynLogEntry.java
jcabrerizo: @ahgittin `timestampString` is the value coming from the elasticSearch data base, but `datatime` is this time transformed to a Date datatype in the method 'FileLogStore#parseLogLine' using the format you can provide as a property on `brooklyn.cfg` that depends on the log configuration.

It has a default value so no new properties needed if the log config is not changed

core/src/main/java/org/apache/brooklyn/util/core/logbook/opensearch/OpenSearchLogStore.java
jcabrerizo: Yes you are right. This implementation is ok for the docker image provided by the Open Search project. For production this need to be configurable.
For now I think we are good addressing it in a future PR. What do you think @duncangrant

jcabrerizo:
@ahgittin I added the resources to the required files

core/src/main/java/org/apache/brooklyn/util/core/logbook/opensearch/OpenSearchLogStore.java
duncangrant: I think what you have is fine

PR #1189, opened 2021-06-23
Don't auth check when not required.

duncangrant:
Previously we checked each call and if it was authenticated we did
returned without filtering.  As the logic for initialising some security
providers has changed this means that there can be a long delay before we
return data for calls that don't need to be authenticated.

Instead we now first check whether a resource needs authorisation before
authenticating.  This speeds up the response in the above case at the
cost of some calls taking slightly longer when the user is logged in.

NB This is the most misleading diff I've ever seen - should be viewed as lines 119 to 127 being moved up the file.


ahgittin:
the diff looked sensible to me, for once.  though all too often the diff algorithm does get start/end methods totally wrong.

image

code change looks great.  merging.

PR #1191, opened 2021-06-25
Adding a feature to automatically capture a persistence export archiv…

zan-mateusz:
WIP

Automated capture of persisted state into an archive saved in the root persistence directory - invoked on persistence change.

Archive is of hardcoded name 'persistence-state-export.zip' and is saved in the root or persistence directory. In case a new one is generated, old one is removed so we keep only the most up to date version of the export.

This is applicable only to file based persistence stores


iuliana:
Pretty-fy the code a little and this can be merged.

iuliana (changes requested):

iuliana (approved):
LGTM

PR #1195, opened 2021-07-07
Feature/working peristence import api

zan-mateusz:
Introducing an API for persistence import feature.
This is intended for file based persistence stores and as a parameter, it takes the location of root of the persistence store to be imported.

Invoking the operation will merge the new data to the currently existing store. The process is as follows:
- new temporary management context is created with the persistence store to be imported
- memento of that persistence store is captured
- bundles from the persistence store are installed in the active management context - this deals with bundles/types in the catalog and locations
- contents of relevant directories (policies, enrichers, etc). are written to the active management context. These are used for the deployed applications
- rebind method adds the deployed applications to the active management context without having to reset the full management context/restart the server

Note: For the edge case when directory passed to the API is not a valid persistence, it uses prepareForSharedUse() method with REBIND parameter. As per documentation, persistMode = REBIND will rebind to the existing state, or fail if no state available. This uses methods checkPersistenceDirPlausible() and checkPersistenceDirNonEmpty() but these are not very robust checks. If we pass a dir which has a single empty file (and therefore is not a valid persistence), it will still pass those checks and eventually create a new persistence store in that location - this is a bug and will need to be fixed.

### DIFF
--- a/rest/rest-resources/src/main/java/org/apache/brooklyn/rest/resources/ServerResource.java
+++ b/rest/rest-resources/src/main/java/org/apache/brooklyn/rest/resources/ServerResource.java
@@ -18,12 +18,9 @@
  */
 package org.apache.brooklyn.rest.resources;
 
-import java.io.ByteArrayOutputStream;
-import java.io.File;
-import java.io.IOException;
-import java.io.InputStream;
+import java.io.*;
 import java.nio.file.Files;
-import java.nio.file.Path;
+import java.nio.file.StandardOpenOption;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
@@ -66,6 +63,7 @@ import org.apache.brooklyn.core.mgmt.internal.LocalManagementContext;
 import org.apache.brooklyn.core.mgmt.internal.ManagementContextInternal;
 import org.apache.brooklyn.core.mgmt.persist.*;
 import org.apache.brooklyn.core.mgmt.rebind.PersistenceExceptionHandlerImpl;
+import org.apache.brooklyn.core.server.BrooklynServerPaths;
 import org.apache.brooklyn.rest.api.ServerApi;
 import org.apache.brooklyn.rest.domain.ApiError;
 import org.apache.brooklyn.rest.domain.BrooklynFeatureSummary;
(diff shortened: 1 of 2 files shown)


rest/rest-resources/src/main/java/org/apache/brooklyn/rest/resources/ServerResource.java
```diff
@@ -528,4 +539,66 @@ protected Response exportPersistenceData(MementoCopyMode preferredOrigin) {
         }
     }
 
+    @Override
+    public Response importPersistenceData(String persistenceExportLocation) {
```
algairim, line 543: Any tests for this? More comments and/or named constants/vars would be good to have as well (just to understand what it does).

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/resources/ServerResource.java
```diff
@@ -528,4 +537,79 @@ protected Response exportPersistenceData(MementoCopyMode preferredOrigin) {
...
+            result.policies(newMementoRawData.getPolicies());
+            result.enrichers(newMementoRawData.getEnrichers());
+            result.feeds(newMementoRawData.getFeeds());
+            result.catalogItems(newMementoRawData.getCatalogItems());
+
+            PersistenceObjectStore currentPersistenceStore = ((BrooklynMementoPersisterToObjectStore) mgmt().getRebindManager().getPersister()).getObjectStore();
+            BrooklynPersistenceUtils.writeMemento(mgmt(),result.build(),currentPersistenceStore);
+            mgmt().getRebindManager().rebind(mgmt().getCatalogClassLoader(),null, mgmt().getHighAvailabilityManager().getNodeState());
+
+            // clean up the temporary management context
```
iuliana, line 600: What is the consequence if an exception is thrown and these are not called?

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/resources/ServerResource.java
```diff
@@ -549,19 +552,22 @@ protected Response exportPersistenceData(MementoCopyMode preferredOrigin) {
...
     @Override
-    public Response importPersistenceData(String persistenceExportLocation) {
+    public Response importPersistenceData(byte[] persistenceData) {
         try {
-            File persistenceLocation = new File(persistenceExportLocation);
-            if (!persistenceLocation.isDirectory()){
-                throw WebResourceUtils.badRequest("Invalid persistence directory - %s does not exist or is not a directory", persistenceExportLocation);
-            }
+            // create a temporary archive where persistence data supplied is written to
+            File tempZipFile = File.createTempFile("persistence-import",null);
```
ahgittin, line 558: should we delete this file at the end of this method?

ahgittin:
looks great - one comment which i'll add and merge

PR #1197, opened 2021-07-09
WIP Remove reverse order from logbook API, use 'tail' instead to get latest log lines. Query 'head' by default.

algairim:
Signed-off-by: Mykola Mandra <mykola.mandra@cloudsoftcorp.com>


jcabrerizo (approved):
LGTM @algairim

PR #1199, opened 2021-07-13
fix/secure-vault-values

iuliana:
Modified REST Resource to return a friendlier value when a secret value is not resolved.


ahgittin:
great - added one more commit tidying up the treatment of `raw`

PR #1200, opened 2021-07-14
Add support setup default initializers for all deployment

tbouron:
This looks up a new configuration options called `brooklyn.deployment.initializers` (comma separated list). If specified on a Brooklyn instance, all deployments will load and execute these initializers. Theses classes are expected to be `EntityInitializer`, if an error occur (either cast or anything else) then the deployment will fail.

The code will try to:
1. load the class from the default class loader.
2. if (1) fails, it will try to load the class from the `TypeRegistry`. This is to allow execution of custom initializers that might be installed in the catalog later on.
3. if (1) and (2) fails, then the deployment is aborted.

This PR contains a new initializer that give more information about the deployment (who and when currently) but it is not set by default (see https://github.com/apache/brooklyn-dist/pull/174)

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/objs/proxy/InternalEntityFactory.java
+++ b/core/src/main/java/org/apache/brooklyn/core/objs/proxy/InternalEntityFactory.java
@@ -40,6 +40,7 @@ import org.apache.brooklyn.core.entity.*;
 import org.apache.brooklyn.core.mgmt.BrooklynTags;
 import org.apache.brooklyn.core.mgmt.BrooklynTags.NamedStringTag;
 import org.apache.brooklyn.core.mgmt.BrooklynTaskTags;
+import org.apache.brooklyn.core.mgmt.classloading.JavaBrooklynClassLoadingContext;
 import org.apache.brooklyn.core.mgmt.internal.EntityManagerInternal;
 import org.apache.brooklyn.core.mgmt.internal.ManagementContextInternal;
 import org.apache.brooklyn.util.collections.MutableList;
@@ -495,17 +496,19 @@ public class InternalEntityFactory extends InternalFactory {
     private List<EntityInitializer> getGlobalDeploymentInitializers() {
         return Arrays.stream(managementContext.getConfig().getConfig(GLOBAL_DEPLOYMENT_INITIALIZER_CLASSNAMES).split(","))
                 .filter(Strings::isNonEmpty)
-                .map(className -> {
-                    try {
-                        Class<?> initializerClass = getClass().getClassLoader().loadClass(className);
-                        return (EntityInitializer) initializerClass.newInstance();
-                    } catch (ClassNotFoundException | InstantiationException | IllegalAccessException e) {
-                        RegisteredType registeredType = managementContext.getTypeRegistry()
-                                .getMaybe(className, null)
-                                .orThrow("Failed to load default initializer with class name: " + className);
-                        return managementContext.getTypeRegistry().create(registeredType, null, null);
-                    }
-                })
+                .map(className -> managementContext.getTypeRegistry()
+                        .getMaybe(className, null)
+                        .map(registeredType -> (EntityInitializer) managementContext.getTypeRegistry().create(registeredType, null, null))
+                        .or(JavaBrooklynClassLoadingContext.create(managementContext)
+                                .tryLoadClass(className)
+                                .map(aClass -> {
+                                    try {
+                                        return (EntityInitializer) aClass.newInstance();
+                                    } catch (InstantiationException | IllegalAccessException e) {
+                                        throw new IllegalStateException(e);
+                                    }
+                                }))
+                        .orNull())
                 .filter(Objects::nonNull)
                 .collect(Collectors.toList());
     }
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/brooklyn/core/objs/proxy/InternalEntityFactory.java
```diff
@@ -369,6 +385,12 @@ public void run() {
                 }
                 ((AbstractEntity)entity).addLocations(spec.getLocations());
 
+                List<EntityInitializer> initializers = Stream.concat(spec.getInitializers().stream(), getDefaultInitializers().stream())
```
ahgittin, line 388: this will run `spec.getInitializers()` _before_ default / global initializers, whereas i think default / globals should run first

also we have the lines below 394-396 which run `spec.getInitializers()` _again_ after

suggest this should just be `getDefaultInitializers().forEach(i -> i.apply(entity))` -- no need to concat streams

core/src/main/java/org/apache/brooklyn/core/objs/proxy/InternalEntityFactory.java
```diff
@@ -46,46 +48,60 @@
...
     private final InternalPolicyFactory policyFactory;
     private final ClassLoaderCache classLoaderCache;
-    
+
+    /**
+     * The initializers to be add to any application deployed by Brooklyn.
+     * e.g. <code>brooklyn.deployment.initializers=org.apache.brooklyn.core.effector.AddDeploySensorsInitializer</code>
+     * will automatically add tags to the root node of any deployed application.
+     */
+    public final static ConfigKey<String> DEFAULT_INITIALIZERS_CLASSNAMES = ConfigKeys.newStringConfigKey(
```
ahgittin, line 100: `DEFAULT` feels wrong -- should we call this `GLOBAL_DEPLOYMENT_INITIALIZERS` ?

(and change the method `getDefaultInitializers()`)

core/src/main/java/org/apache/brooklyn/core/objs/proxy/InternalEntityFactory.java
```diff
@@ -465,12 +487,30 @@ public void run() {
...
         } else {
             return entityTypeRegistry.getImplementedBy(spec.getType());
         }
     }
+
+    private List<EntityInitializer> getDefaultInitializers() {
+        return Arrays.stream(managementContext.getConfig().getConfig(DEFAULT_INITIALIZERS_CLASSNAMES).split(","))
+                .filter(Strings::isNonEmpty)
+                .map(className -> {
+                    try {
```
ahgittin, line 503: there is a canonical way we do this -- let me dig this up

core/src/main/java/org/apache/brooklyn/core/objs/proxy/InternalEntityFactory.java
tbouron: Aaah I meant to remove line 394-396 and have only one loop to be DRY, hence the concatenation. I'll update @ahgittin

core/src/main/java/org/apache/brooklyn/core/objs/proxy/InternalEntityFactory.java
tbouron: @ahgittin Should I also change the property name or are you happy with what is there already?

core/src/main/java/org/apache/brooklyn/core/objs/proxy/InternalEntityFactory.java
ahgittin: to load a java class, use `JavaBrooklynClassLoadingContext.create(mgmt).tryLoadClass(className)` -- has slightly better classpath semantics

also i wonder whether we should try the type registry first?

(longer term i'd like to clean up and standardize better how we instantiate things!)

core/src/main/java/org/apache/brooklyn/core/objs/proxy/InternalEntityFactory.java
ahgittin: `brooklyn.deployment.initializers` is a great name - i should have said!

core/src/main/java/org/apache/brooklyn/core/objs/proxy/InternalEntityFactory.java
ahgittin: no strong feelings about one loop or two, as long as globals done first, in case they want to fail fast to prevent even local initializers from running

ahgittin:
looks good - a few minor comments, then can merge

ahgittin:
pleased to see our test coverage caught the problem w initializers being run twice!

tbouron:
@ahgittin That should be all your comments addressed now

core/src/main/java/org/apache/brooklyn/core/objs/proxy/InternalEntityFactory.java
tbouron: Done!

tbouron: What I meant is that I used the canonical way to load the class and changed this order: the code now checks the `TypeRegistry` first, then tries to load from the class loader. Thank you BTW for the bit of code you shared above 👍

ahgittin:
looks great - merged

PR #1202, opened 2021-07-15
Adds an attributes map on WebEntitlementContext to handle user roles

jcabrerizo:
Adds an attributes map to WebEntitlementContext and use EntitlementContextFilter to sync the user LDAP groups from the session.

LDAP groups can only be resolved at login time, using the user password. For exposing them when checking the entitlements is necessary to add them to the `EntitlementContext`. This PR adds a map of attributes in the `WebEntitlementContext` and put there the groups when the new config key `LDAP_FETCH_USER_GROUPS` is true

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/mgmt/entitlement/WebEntitlementContext.java
+++ b/core/src/main/java/org/apache/brooklyn/core/mgmt/entitlement/WebEntitlementContext.java
@@ -19,7 +19,11 @@
 package org.apache.brooklyn.core.mgmt.entitlement;
 
 import org.apache.brooklyn.api.mgmt.entitlement.EntitlementContext;
+import org.apache.brooklyn.util.collections.MutableMap;
 import org.apache.brooklyn.util.javalang.JavaClassNames;
+
+import java.util.HashMap;
+import java.util.Map;
 
 /**
  * Indicates an authenticated web request as the entitlements context;
@@ -27,9 +31,15 @@ import org.apache.brooklyn.util.javalang.JavaClassNames;
  */
 public class WebEntitlementContext implements EntitlementContext {
 
+    public static final String ENTITLEMENTS_ATTRIBUTES = "brooklyn.entitlements.attributes";
+
+    public static final String USER_ROLES = "brooklyn.entitlements.user.roles";
+
+
     final String user;
     final String sourceIp;
     final String requestUri;
+    final Map<String, Object> attributes;
     
     /**
      * A mostly-unique identifier for the inbound request, to distinguish
@@ -38,16 +48,25 @@ public class WebEntitlementContext implements EntitlementContext {
     final String requestUniqueIdentifier;
     
     public WebEntitlementContext(String user, String sourceIp, String requestUri, String requestUniqueIdentifier) {
+        this(user, sourceIp, requestUri, requestUniqueIdentifier, new HashMap<>());
+    }
+
+    public WebEntitlementContext(String user, String sourceIp, String requestUri, String requestUniqueIdentifier, Map attributes) {
         this.user = user;
         this.sourceIp = sourceIp;
         this.requestUri = requestUri;
         this.requestUniqueIdentifier = requestUniqueIdentifier;
+        if(attributes!=null)
+            this.attributes = attributes;
+        else
+            this.attributes = MutableMap.of();
     }
-    
+
     @Override public String user() { return user; }
     public String sourceIp() { return sourceIp; }
     public String requestUri() { return requestUri; }
     public String requestUniqueIdentifier() { return requestUniqueIdentifier; }
+    public Map<String, Object> attributes() { return attributes; }
 
     @Override
     public String toString() {
--- a/core/src/main/java/org/apache/brooklyn/core/mgmt/entitlement/WebEntitlementContext.java
+++ b/core/src/main/java/org/apache/brooklyn/core/mgmt/entitlement/WebEntitlementContext.java
@@ -19,7 +19,6 @@
 package org.apache.brooklyn.core.mgmt.entitlement;
 
 import org.apache.brooklyn.api.mgmt.entitlement.EntitlementContext;
-import org.apache.brooklyn.util.collections.MutableMap;
 import org.apache.brooklyn.util.javalang.JavaClassNames;
 
 import java.util.HashMap;
@@ -51,7 +50,7 @@ public class WebEntitlementContext implements EntitlementContext {
         this(user, sourceIp, requestUri, requestUniqueIdentifier, new HashMap<>());
     }
 
-    public WebEntitlementContext(String user, String sourceIp, String requestUri, String requestUniqueIdentifier, Map attributes) {
+    public WebEntitlementContext(String user, String sourceIp, String requestUri, String requestUniqueIdentifier, Map<String, Object> attributes) {
         this.user = user;
         this.sourceIp = sourceIp;
         this.requestUri = requestUri;
@@ -59,7 +58,7 @@ public class WebEntitlementContext implements EntitlementContext {
         if(attributes!=null)
             this.attributes = attributes;
         else
-            this.attributes = MutableMap.of();
+            this.attributes = new HashMap<>();
     }
 
     @Override public String user() { return user; }
--- a/core/src/main/java/org/apache/brooklyn/core/mgmt/entitlement/WebEntitlementContext.java
+++ b/core/src/main/java/org/apache/brooklyn/core/mgmt/entitlement/WebEntitlementContext.java
@@ -19,6 +19,7 @@
 package org.apache.brooklyn.core.mgmt.entitlement;
 
 import org.apache.brooklyn.api.mgmt.entitlement.EntitlementContext;
+import org.apache.brooklyn.util.collections.MutableMap;
 import org.apache.brooklyn.util.javalang.JavaClassNames;
 
 import java.util.HashMap;
@@ -38,7 +39,7 @@ public class WebEntitlementContext implements EntitlementContext {
     final String user;
     final String sourceIp;
     final String requestUri;
-    final Map<String, Object> attributes;
+    final Map<String, Object> attributes = MutableMap.of();
     
     /**
      * A mostly-unique identifier for the inbound request, to distinguish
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/apache/brooklyn/core/mgmt/entitlement/WebEntitlementContext.java
```diff
@@ -38,16 +48,25 @@
     final String requestUniqueIdentifier;
     
     public WebEntitlementContext(String user, String sourceIp, String requestUri, String requestUniqueIdentifier) {
+        this(user, sourceIp, requestUri, requestUniqueIdentifier, new HashMap<>());
+    }
+
+    public WebEntitlementContext(String user, String sourceIp, String requestUri, String requestUniqueIdentifier, Map attributes) {
```
duncangrant, line 54: attributes should be parametrised?

core/src/main/java/org/apache/brooklyn/core/mgmt/entitlement/WebEntitlementContext.java
```diff
@@ -38,16 +48,25 @@
     final String requestUniqueIdentifier;
     
     public WebEntitlementContext(String user, String sourceIp, String requestUri, String requestUniqueIdentifier) {
+        this(user, sourceIp, requestUri, requestUniqueIdentifier, new HashMap<>());
```
duncangrant, line 51: There's an inconsistency with the initialisation of attributes - here it's a hashmap and later it's a mutablemap

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/BrooklynWebConfig.java
```diff
@@ -83,6 +83,11 @@
     public final static ConfigKey<String> LDAP_OU = ConfigKeys.newStringConfigKey(
             BASE_NAME_SECURITY+".ldap.ou");
 
+    public final static ConfigKey<Boolean> LDAP_FETCH_USER_GROUPS = ConfigKeys.newBooleanConfigKey(
+            BASE_NAME_SECURITY+".ldap.fetch_user_group",
+            "Whether if user groups should be fetch from LDAP server",
```
duncangrant, line 88: "Whether user groups should be fetched from the LDAP server"

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/security/provider/LdapSecurityProvider.java
```diff
@@ -21,24 +21,32 @@
...
 import com.google.common.base.Joiner;
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Lists;
-import java.util.Arrays;
-import java.util.Hashtable;
-import java.util.List;
-import java.util.function.Supplier;
-import javax.naming.Context;
-import javax.naming.NamingException;
-import javax.naming.directory.InitialDirContext;
```
duncangrant, line 31: Try to avoid moving import statements just because your IDE decided it was a good idea - it makes it hard to see what's been added

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/BrooklynWebConfig.java
jcabrerizo: tks, I'll change this

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/security/provider/LdapSecurityProvider.java
```diff
@@ -99,13 +108,76 @@ public boolean authenticate(HttpServletRequest request, Supplier<HttpSession> se
             env.put(Context.SECURITY_PRINCIPAL, getSecurityPrincipal(user));
             env.put(Context.SECURITY_CREDENTIALS, pass);
 
-            new InitialDirContext(env);  // will throw if password is invalid
+            DirContext ctx = new InitialDirContext(env);// will throw if password is invalid
```
duncangrant, line 111: Feels like this needs a try finally block?

jcabrerizo: I don't think it's needed. If the authentication fails an ti throw an exception, the context won't be created and the user remains unathenticated.
Why do you think is needed @duncangrant ?

core/src/main/java/org/apache/brooklyn/core/mgmt/entitlement/WebEntitlementContext.java
jcabrerizo: I'll use HashMap in `WebEntitlementContext`

duncangrant (approved):
LGTM - thanks @jcabrerizo - I've added a couple of comments - feel free to ignore and merge

core/src/main/java/org/apache/brooklyn/core/mgmt/entitlement/WebEntitlementContext.java
```diff
@@ -21,15 +21,24 @@
...
 
+    public static final String ENTITLEMENTS_ATTRIBUTES = "brooklyn.entitlements.attributes";
+
+    public static final String USER_ROLES = "brooklyn.entitlements.user.roles";
+
+
     final String user;
     final String sourceIp;
     final String requestUri;
+    final Map<String, Object> attributes;
```
ahgittin, line 41: probably `= MutableMap.of()` here

that will use `LinkedHashMap` which preserves order (i never use plain `HashMap`)

and then we can prevent a caller from adding map entries after they are passed in here

core/src/main/java/org/apache/brooklyn/core/mgmt/entitlement/WebEntitlementContext.java
```diff
@@ -38,16 +47,25 @@
     final String requestUniqueIdentifier;
     
     public WebEntitlementContext(String user, String sourceIp, String requestUri, String requestUniqueIdentifier) {
+        this(user, sourceIp, requestUri, requestUniqueIdentifier, new HashMap<>());
```
ahgittin, line 50: last arg `null` for clarity

core/src/main/java/org/apache/brooklyn/core/mgmt/entitlement/WebEntitlementContext.java
```diff
@@ -38,16 +47,25 @@
...
+    }
+
+    public WebEntitlementContext(String user, String sourceIp, String requestUri, String requestUniqueIdentifier, Map<String, Object> attributes) {
         this.user = user;
         this.sourceIp = sourceIp;
         this.requestUri = requestUri;
         this.requestUniqueIdentifier = requestUniqueIdentifier;
+        if(attributes!=null)
+            this.attributes = attributes;
```
ahgittin, line 59: `if (attributes != null) this.attributes.addAll(attributes)` -- and no need for `else` block now

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/security/provider/LdapSecurityProvider.java
ahgittin: it's so annoying, and IJ seems particularly bad about doing this.  there should be a mode "optimize / auto-add imports without rearranging" but i can't find it either.

core/src/main/java/org/apache/brooklyn/core/mgmt/entitlement/WebEntitlementContext.java
```diff
@@ -21,15 +21,24 @@
...
+
 /**
  * Indicates an authenticated web request as the entitlements context;
  * note user may still be null if no authentication was requested
  */
 public class WebEntitlementContext implements EntitlementContext {
 
+    public static final String ENTITLEMENTS_ATTRIBUTES = "brooklyn.entitlements.attributes";
+
+    public static final String USER_ROLES = "brooklyn.entitlements.user.roles";
```
ahgittin, line 35: are these `ROLES` or `GROUPS` ?  when we fetch we call them groups.  should be consistent everywhere.  personally i prefer groups.  maybe worth a comment in the code to indicate these are sometimes in LDAP called roles (if that's true?).

core/src/main/java/org/apache/brooklyn/core/mgmt/entitlement/WebEntitlementContext.java
```diff
@@ -21,15 +21,24 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
 /**
  * Indicates an authenticated web request as the entitlements context;
  * note user may still be null if no authentication was requested
  */
 public class WebEntitlementContext implements EntitlementContext {
 
+    public static final String ENTITLEMENTS_ATTRIBUTES = "brooklyn.entitlements.attributes";
```
ahgittin, line 33: i think remove this, as per below

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/filter/EntitlementContextFilter.java
```diff
@@ -55,30 +58,44 @@ public void filter(ContainerRequestContext requestContext) throws IOException {
...
+            Map<String, Object> entitlementAttributes = null;
+            if (userRoles != null) {
+                entitlementAttributes = ImmutableMap.of(
+                        WebEntitlementContext.ENTITLEMENTS_ATTRIBUTES,
+                        ImmutableMap.of(
+                                WebEntitlementContext.USER_ROLES,
+                                userRoles));
+            }
+
+            WebEntitlementContext entitlementContext = new WebEntitlementContext(userName, remoteAddr, uri, uid, entitlementAttributes);
```
ahgittin, line 84: why the map in a map?  also note `MutableMap.addIfNotNull(...)` -- entire logic can be done replacing `entitlementAttributes` with

    MutableMap.of().addIfNotNull(WebEntitlementContext.USER_GROUPS, getAttributeFromSession(WebEntitlementContext.USER_GROUPS))

core/src/main/java/org/apache/brooklyn/core/mgmt/entitlement/WebEntitlementContext.java
ahgittin: i think untyped is okay

ahgittin:
@jcabrerizo looks good -- a bunch of comments but all minor

core/src/main/java/org/apache/brooklyn/core/mgmt/entitlement/WebEntitlementContext.java
jcabrerizo: Groups is better. Until the RoleResolver is executed there is not direct relation between groups and roles. Changing it.

ahgittin:
looks great

PR #1204, opened 2021-07-16
Introduce lineId for log entries

algairim:


core/src/main/java/org/apache/brooklyn/util/core/logbook/file/FileLogStore.java
```diff
@@ -69,23 +70,28 @@
...
+    private final DateFormat dateFormat;
+
 
     @VisibleForTesting
     public FileLogStore() {
         this.path = null;
         this.filePath = "";
         this.logLinePattern = LOGBOOK_LOG_STORE_REGEX.getDefaultValue();
-        this.logDateFormat = LOGBOOK_LOG_STORE_DATEFORMAT.getDefaultValue();
+        dateFormat = new SimpleDateFormat(LOGBOOK_LOG_STORE_DATEFORMAT.getDefaultValue());
```
jcabrerizo, line 86: I'd add `this.` to dateFormat to keep it consistent

jcabrerizo (commented):
LGTM @algairim

core/src/main/java/org/apache/brooklyn/util/core/logbook/file/FileLogStore.java
ahgittin: personally i don't like the `this.` (except where it is needed)

core/src/test/java/org/apache/brooklyn/util/core/logbook/file/FileLogStoreTest.java
```diff
@@ -48,64 +49,74 @@
             "\tat org.apache.felix.scr.impl.ComponentRegistry.checkComponentName(ComponentRegistry.java:240) ~[?:?]\n" +
             "\tat org.apache.felix.scr.impl.BundleComponentActivator.validateAndRegister(BundleComponentActivator.java:443) ~[?:?]";
 
+    private final AtomicInteger lineCount = new AtomicInteger();
+
     @BeforeTest
     public void setUp() {
-        TimeZone.setDefault(TimeZone.getTimeZone("GMT"));
+        TimeZone.setDefault(TimeZone.getTimeZone("UTC"));
```
ahgittin, line 56: i don't like having to set this in a test but don't have a better idea so go for it

ahgittin:
LGTM

PR #1205, opened 2021-07-16
Be stricter about java lambdas in data which is persisted

ahgittin:
It has always been a bad idea but previously we have allowed it.  This causes obscure but horrible problems in many cases:

* lambdas which don't implement Serializable are written by xstream as `null`, and read back as null, with no warning
* lambdas which do implement Serializable might be okay written by xstream but might not, and can be read differently or not at all by different JVMs, and are even more brittle than inner classes

This PR causes us to be stricter about persisting lambdas e.g. within a config key or part of an effector so that we fail fast.

Default is to FAIL if we attempt to write one which will be treated as null, and WARN otherwise.  This can be configured in brooklyn.cfg using the keys in LambdaPreventionMapper.

In addition this adds a delta-persistence synchronously as part of application/entity creation, so if a problem is detected the creation will fail.  Things are unmanaged now if creation fails.

Internally there is more work done and more API methods to get access to failures during persistence.

### DIFF
--- a/api/src/main/java/org/apache/brooklyn/api/mgmt/EntityManager.java
+++ b/api/src/main/java/org/apache/brooklyn/api/mgmt/EntityManager.java
@@ -76,6 +76,8 @@ public interface EntityManager {
         default boolean isDryRun() { return false; }
         /** whether the item should have a specific identifier; may fail or return existing if it already exists */
         default String getRequiredUniqueId() { return null; }
+        /** whether to persist as part of creation (forces failure if creation fails, in the creating thread) */
+        default boolean persistAfterCreation() { return true; }
     }
 
     /**
--- a/api/src/main/java/org/apache/brooklyn/api/mgmt/rebind/PersistenceExceptionHandler.java
+++ b/api/src/main/java/org/apache/brooklyn/api/mgmt/rebind/PersistenceExceptionHandler.java
@@ -18,6 +18,8 @@
  */
 package org.apache.brooklyn.api.mgmt.rebind;
 
+import com.google.common.annotations.VisibleForTesting;
+import java.util.Set;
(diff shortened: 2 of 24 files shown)


ahgittin:
was pointed out that this breaks on OSGi with

```
org.apache.felix.resolver.reason.ReasonException: Unable to resolve root: missing requirement [root] osgi.identity; osgi.identity=brooklyn-software-winrm; type=karaf.feature; version=0; filter:="(&(osgi.identity=brooklyn-software-winrm)(type=karaf.feature)(version>=0.0.0))" [caused by: Unable to resolve brooklyn-software-winrm/1.1.0.SNAPSHOT: missing requirement [brooklyn-software-winrm/1.1.0.SNAPSHOT] osgi.identity; osgi.identity=org.apache.brooklyn.software-winrm; type=osgi.bundle; version="[1.1.0.SNAPSHOT,1.1.0.SNAPSHOT]"; resolution:=mandatory [caused by: Unable to resolve org.apache.brooklyn.software-winrm/1.1.0.SNAPSHOT: missing requirement [org.apache.brooklyn.software-winrm/1.1.0.SNAPSHOT] osgi.wiring.package; filter:="(&(osgi.wiring.package=org.apache.brooklyn.core.config)(version>=1.1.0)(!(version>=2.0.0)))" [caused by: Unable to resolve org.apache.brooklyn.core/1.1.0.SNAPSHOT: missing requirement [org.apache.brooklyn.core/1.1.0.SNAPSHOT] osgi.wiring.package; filter:="(osgi.wiring.package=com.thoughtworks.xstream.core.util)"]]]
```

last commit should fix that

iuliana (commented):
Currently testing. Will merge afterwards.

iuliana:
This passed locally with tests. I will merge it.

PR #1206, opened 2021-07-19
Exclude 'logbook.log.sample' license to be checked

jcabrerizo:
This fix the build exception due to the lack of license on the file


jcabrerizo:
I miss it was already fixed

PR #1208, opened 2021-07-19
Support logbook multi-phrase query

algairim:


core/src/main/java/org/apache/brooklyn/util/core/logbook/file/FileLogStore.java
```diff
@@ -200,4 +201,15 @@ protected BrooklynLogEntry parseLogLine(String logLine, AtomicInteger lineCount)
         }
         return entry;
     }
+
+    public static boolean containsSearchPhrases(String logMessage, List<String> searchPhrases) {
+        boolean found = true;
+        for (String searchPhrase : searchPhrases) {
+            if (!logMessage.contains(searchPhrase)) {
+                found = false;
```
jcabrerizo, line 209: May returning here `false` makes the code more readable

jcabrerizo (approved):
LGTM @algairim

PR #1209, opened 2021-07-21
Fix partial login when ldap user in no groups

duncangrant:
If ldap user is a member of no groups then they will fail to login but
will have a user associated with the session that cannot easily be
logged out or login as another user.

Fixed by not allowing authentication of a user in no groups.


jcabrerizo (approved):
Thanks @duncangrant LGTM

PR #1210, opened 2021-07-23
Feature/entitlement execute script

jcabrerizo:
Adds a new `Entitlement` to control the Groovy script execution and a new `EntitlementManager` with "**PowerUser**" with the same entitlements as root but the ability to execute groovy scripts

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/mgmt/entitlement/Entitlements.java
+++ b/core/src/main/java/org/apache/brooklyn/core/mgmt/entitlement/Entitlements.java
@@ -284,7 +284,11 @@ public class Entitlements {
         return new EntitlementManager() {
             @Override
             public <T> boolean isEntitled(EntitlementContext context, EntitlementClass<T> permission, T entitlementClassArgument) {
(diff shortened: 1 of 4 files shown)


iuliana (commented):
Looks good to me. And looks good to merge.

core/src/main/java/org/apache/brooklyn/core/mgmt/entitlement/Entitlements.java
```diff
@@ -480,6 +503,8 @@ public static EntitlementManager load(@Nullable ManagementContext mgmt, Brooklyn
             return minimal();
         } else if ("user".equalsIgnoreCase(type)) {
             return user();
+        } else if ("powerUser".equalsIgnoreCase(type) || "power_user".equalsIgnoreCase(type)) {
+            return powerUser();
```
algairim, line 507: Do we have constants for these to re-use?

iuliana (approved):
LGTM

core/src/main/java/org/apache/brooklyn/core/mgmt/entitlement/Entitlements.java
jcabrerizo: No for now. If we need to reuse the name of the entitlement manager we can create, but if not, I think is more clear to see directly the value here.

PR #1212, opened 2021-07-26
persist the actual types of relationships if not standard ones

ahgittin:
this allows properties etc on extension of relationships to be persisted and rebinded


ahgittin:
Without this, Brooklyn only persists the name of relationship types.  This works fine if all relationship types are registered with the management context.  But in some downstream cases ad hoc classes are used as relationship types, sometimes storing properties.  Those relationships were brought back in in a default state.

With this PR we use string names if the relationship type is registered, but if it is not we persist the actual relationship type instance, so it is preserved exactly.  This should ensure relationships are rebinded correctly with no change for the case where it worked, and with the extra metadata for the case where we previously brought them back in a default state.

iuliana (approved):
@ahgittin  LGTM, can be merged.

PR #1213, opened 2021-07-27
Fix disallowed characters in multipart form

duncangrant:
Specifically if posting a multipart form with %s then we would get an
error.


ahgittin:
looks good @duncangrant but i think we need to apply this change also to `createFromYamlAndFormatAndAppId`

ahgittin:
looks good.  think i can tidy up the test so it failed before and passes after, will push separately.

PR #1214, opened 2021-07-30
Feature/blueprint author

jcabrerizo:
Adds a new `EntitlementManager` with the same entitlements than the User except it can not install bundles containing jar or class files. Allowing only adding new models written in yaml


iuliana (commented):
LGTM

PR #1215, opened 2021-08-03
more comments on deprecated REST methods

ahgittin:


rest/rest-api/src/main/java/org/apache/brooklyn/rest/api/CatalogApi.java
```diff
@@ -110,13 +110,16 @@ public Response createFromUpload(
             @QueryParam("forceUpdate") @DefaultValue("false")
                     boolean forceUpdate);
 
+    /** @deprecated since 1.0.0 delete the bundle via DELETE /catalog/bundles/xxx */
```
tbouron, line 113: Should be 1.1.0 right?

rest/rest-api/src/main/java/org/apache/brooklyn/rest/api/CatalogApi.java
```diff
@@ -372,10 +375,13 @@ public CatalogItemSummary getLocation(
         @ApiParam(name = "version", value = "The version identifier of the application to retrieve", required = true)
         @PathParam("version") String version) throws Exception;
 
+    /** @deprecated since 1.0.0 delete the bundle via DELETE /catalog/bundles/xxx */
```
tbouron, line 378: Same as above

tbouron (changes requested):
@ahgittin just 2 tiny comments but otherwise LGTM

tbouron (approved):

PR #1217, opened 2021-08-06
Better scope root evaluation

ahgittin:
scopeRoot logic was too simplistic where catalog types are used extensively.

this fixes it by using the DEPTH_IN_ANCESTOR tag to identify the right blueprint better.

it also has some special logic for root nodes to ensure scopeRoot when used in an ancestor type blueprint refers to the ancestor type, whereas when used on the same entity but in a referencing blueprint it refers to the root of the blueprint which refers to it


iuliana (commented):
Looks good to me. Happy to merge after the tests pass.

algairim (approved):
Looks good.

ahgittin:
updated failing tests, and added notes and test case for the gaps in this imperfect-but-usually-right approach

PR #1219, opened 2021-08-10
Controllin ADD_JAVA and ADD_CATALOG entitlements in BundleResource

jcabrerizo:


utils/common/src/main/java/org/apache/brooklyn/util/io/FileUtil.java
```diff
@@ -201,4 +205,18 @@ public static void deleteDirectory(final File file) {
             }
         }
     }
+
+    public static boolean isJava(InputStreamSource archive) {
```
ahgittin, line 209: a better name would be `doesZipContainJavaBinaries`

ahgittin:
looks great, merging.  method rename could be done later.

PR #1220, opened 2021-08-10
Deprecate AggregationJob.java

tbouron:
This was an experimental feature but nothing is using it. The logic is also very primitive and not fully baked in hence the deprecation

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/enricher/stock/aggregator/AggregationJob.java
+++ b/core/src/main/java/org/apache/brooklyn/enricher/stock/aggregator/AggregationJob.java
@@ -24,6 +24,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 
+import org.apache.brooklyn.util.collections.MutableSet;
 import org.apache.commons.lang3.StringUtils;
 
 import com.google.common.annotations.Beta;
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/apache/brooklyn/enricher/stock/aggregator/AggregationJob.java
```diff
@@ -32,6 +33,8 @@
 import org.apache.brooklyn.core.sensor.BasicAttributeSensorAndConfigKey;
 
 @Beta
+@Deprecated
+/** @Deprecated since 1.0.0 No planed replacement */
```
jcabrerizo, line 37: 1.0.0 was already released, this message should not be since 1.1.0?

core/src/main/java/org/apache/brooklyn/enricher/stock/aggregator/AggregationJob.java
```diff
@@ -24,6 +24,7 @@
 import java.util.List;
 import java.util.Map;
 
+import org.apache.brooklyn.util.collections.MutableSet;
```
jcabrerizo, line 27: Is this import used?

jcabrerizo (approved):
LGTM @tbouron just a couple of very minor comments

core/src/main/java/org/apache/brooklyn/enricher/stock/aggregator/AggregationJob.java
tbouron: Nope indeed. I'll remove

core/src/main/java/org/apache/brooklyn/enricher/stock/aggregator/AggregationJob.java
tbouron: You are absolutely right @jcabrerizo, I'll update that

PR #1221, opened 2021-08-11
Rename method for checking if zip has java binaries

jcabrerizo:


jcabrerizo:
@ahgittin This renames the method as you proposed

PR #1223, opened 2021-08-16
Search for a valid session ID in cookies

jcabrerizo:
Tries to find a valid session ID from the list sent as part of the cookies

When a user logout, the sessions for each of the modules where the used has been are still keep on the browser and that could end in a improper future logout in some scenarios


ahgittin:
(changed the message - merging!)

PR #1224, opened 2021-08-17
fix for 406 http error on post request ha/state

zan-mateusz:
The API '/v1/server/ha/state' failed to work with HTTP error 406 - Not Acceptable.

Changing @produces to @consumes fixes the error.


jcabrerizo (approved):
LGTM.
Thanks @zan-mateusz for spotting this

PR #1225, opened 2021-08-19
Adding new attribute to control save the user groups origin

jcabrerizo:
Adds a new attribute in the WebEntitlementContext sincronized with the user session to keep the origin of the user groups

### DIFF
--- a/rest/rest-resources/src/main/java/org/apache/brooklyn/rest/security/provider/LdapSecurityProvider.java
+++ b/rest/rest-resources/src/main/java/org/apache/brooklyn/rest/security/provider/LdapSecurityProvider.java
@@ -50,6 +50,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import static org.apache.brooklyn.core.mgmt.entitlement.WebEntitlementContext.USER_GROUPS;
+import static org.apache.brooklyn.core.mgmt.entitlement.WebEntitlementContext.USER_GROUPS_ORIGIN;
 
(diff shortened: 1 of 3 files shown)


algairim:
I would like to see some tests!

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/security/provider/LdapSecurityProvider.java
```diff
@@ -69,6 +70,7 @@
     public static final Logger LOG = LoggerFactory.getLogger(LdapSecurityProvider.class);
 
     public static final String LDAP_CONTEXT_FACTORY = "com.sun.jndi.ldap.LdapCtxFactory";
+    public static final String LDAP_USER_GROUPS_ORIGIN = "brooklyn.entitlements.user.groups.origin.LdapSecurityProvider";
```
ahgittin, line 73: i'd make this be `LdapSecurityProvider.class.getName()` -- there is no need for origin in the value here (it is used as a value of a key which says `origin`)

ahgittin:
one trivial comment, ignore or address as you like, then please merge

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/security/provider/LdapSecurityProvider.java
jcabrerizo: Addressed, thanks for the feed back

PR #1232, opened 2021-08-25
Added HA_ADMIN entitlement and added test coverage for it

iuliana:
Added HA_ADMIN entitlement and added test coverage for it.


core/src/test/java/org/apache/brooklyn/core/mgmt/entitlement/EntitlementsTest.java
```diff
@@ -161,6 +163,7 @@ public void testGlobalPowerUserEntitlement() {
         assertTrue(user.isEntitled(null, Entitlements.LOGBOOK_LOG_STORE_QUERY, null));
         assertTrue(user.isEntitled(null, Entitlements.ADD_JAVA, null));
         assertFalse(user.isEntitled(null, Entitlements.EXECUTE_GROOVY_SCRIPT, null));
+        assertFalse(user.isEntitled(null, Entitlements.HA_ADMIN, null));
```
jcabrerizo, line 166: This should be true, powerUser is the same as root but it can not execute groovy

jcabrerizo (changes requested):
Logic seems good but you will need to fix the test

PR #1238, opened 2021-08-30
Removed SCP and SSH mentions from logs and comments

iuliana:
As the title says. Made this change because task summary shows up in the interface and might lead to confusions in regards to this task being a WinRm task, or a Ssh task.


jathanasiou (approved):
Looks good.

PR #1241, opened 2021-08-31
Fix logbook query by EntityId or TaskId

jcabrerizo:
It try to find the Ids in the fields and the message for both, `FileLogStore` and `OpenSearchLogStore`,  instead of only search for it on the message as before.

Requires changes in the UI for sending the Ids as independent fields in the query parameters: 
https://github.com/apache/brooklyn-ui/pull/275

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/util/core/logbook/opensearch/OpenSearchLogStore.java
+++ b/core/src/main/java/org/apache/brooklyn/util/core/logbook/opensearch/OpenSearchLogStore.java
@@ -62,7 +62,6 @@ import java.security.NoSuchAlgorithmException;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
-import java.util.Objects;
 import java.util.stream.Collectors;
 
 import static org.apache.brooklyn.util.core.logbook.LogbookConfig.BASE_NAME_LOGBOOK;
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/apache/brooklyn/util/core/logbook/opensearch/OpenSearchLogStore.java
```diff
@@ -253,17 +250,47 @@ protected String getJSONQuery(LogBookQueryParams params) {
             queryBoolMustListBuilder.add(ImmutableMap.of("range", ImmutableMap.of("timestamp", timestampMapBuilder.build())));
         }
 
+        // Apply search taskId.
+        if (Strings.isNonBlank(params.getTaskId())) {
+            queryBoolMustListBuilder.add(
+                    ImmutableMap.of("bool",
+                            ImmutableMap.of("should",
```
algairim, line 257: `should` is an OR match condition, meaning to look for taskId in the `taskId` filed or `message` field, right?

algairim (approved):
Looks good.

core/src/main/java/org/apache/brooklyn/util/core/logbook/opensearch/OpenSearchLogStore.java
jcabrerizo: Short answer: yes, it how it work when it's inside a "bool" sentence. The docs are here https://www.elastic.co/guide/en/elasticsearch/reference/6.8/query-dsl-bool-query.html but I needed to see some examples to have the query working

PR #1242, opened 2021-09-03
ensuring the server is not terminated after persistence import

zan-mateusz:
Calling the terminate on the temporary management context causes closure of OSGI which results in terminating AMP server after this operation. The removal of the termination call fixes the issue. 
It's fine as the temp management context is garbage collected.


iuliana (approved):

iuliana:
Will merge as soons as the build passes.

PR #1243, opened 2021-09-03
Fix for importing a brooklyn generated export

zan-mateusz:
Removing data subdir as not present in a brooklyn generated persistence export


jcabrerizo (approved):
LGTM @zan-mateusz

PR #1245, opened 2021-09-03
Add user name regex for LDAP authentication

jcabrerizo:
New ConfigKey for set up a regex pattern to be matched by the user names before checking credentials in the LDAP server, avoiding unnecessary request

eg. for  `domain\user` pattern:

```
brooklyn.webconsole.security.ldap.user_name_regex=.*\\.*
```


jcabrerizo:
Doc for this: https://github.com/apache/brooklyn-docs/pull/337

PR #1247, opened 2021-09-06
Revert "BROOKLYN-626: fix jmxmp"

iuliana:
Reverts apache/brooklyn-server#1097


iuliana:
I tested this locally and it works as intended.

PR #1250, opened 2021-09-07
Fix error finding the port

jcabrerizo:


iuliana:
@jcabrerizo  Looks good, let's just wait for the build too pass and merge it.

iuliana (approved):
LGTM

PR #1251, opened 2021-09-07
removing extra subdir causing import issues

zan-mateusz:
Export feature originally captured a zip 'brooklyn-state-id-timestamp' which also contained a subdir named 'web-persistence-id-timestamp' causing the data dir being unable to be found (as it was not in root but in web-persistence...').

Not straightforward to find the subdir name because of extra data like timestamp, but not really necessary for it to be there so removed.

Now, in the root of archive there is a data dir that has the persistence state data.


iuliana (approved):
LGTM

PR #1252, opened 2021-09-07
Separate logic for domain name and user name regex

jcabrerizo:
Two properties for handle a _domain_ name and _username_ regex before validating credentials on the LDAP server:
eg:

```
brooklyn.webconsole.security.ldap.user_name_regex=.*
brooklyn.webconsole.security.ldap.domain_name_regex=DOMAIN
```
will try to validate only usernames with the format `DOMAIN\<user>`


ahgittin:
looks good.  think this is the cleanest solution for the problem where to get a `\` matched by a regex, of course the regex needs `\\` but OSGi ConfigurationAdmin needs each of those two `\` escaped twice, so user name regex ends up something horrible like `DOMAIN\\\\\\\\.*` in the config. this way that ugliness is removed.

PR #1255, opened 2021-09-09
Unmanagement tidy up

ahgittin:
Cleans up various things noticed when using hot_standby and switching management nodes a lot


core/src/main/java/org/apache/brooklyn/core/mgmt/internal/LocalEntityManager.java
```diff
@@ -393,7 +393,8 @@ protected void manageRecursive(Entity e, final ManagementTransitionMode initialM
             if (mode==null) {
                 setManagementTransitionMode(it, mode = initialMode);
             }
-            
+            log.debug("Starting management of {} mode {}", it, mode);
```
iuliana, line 396: of 'node' you mean :)

algairim (approved):
Looks good, tested with local and S3 store setup.

PR #1256, opened 2021-09-09
Added method to use curl with a minimum TLS version specified as

iuliana:
Added method to use curl with a minimum TLS version specified as argument.


zan-mateusz:
LGTM!

PR #1257, opened 2021-09-14
Tidy rebind, scheduled tasks, and related logging

ahgittin:
ScheduledTasks in particular were problematic, continuing to run after rebind cycles (especially bad for hot standby!), and not logging correctly their context.

This fixes both of those, and also sets it up so that IDs of bundles are not changed on a rebind/failover.

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/catalog/internal/CatalogInitialization.java
+++ b/core/src/main/java/org/apache/brooklyn/core/catalog/internal/CatalogInitialization.java
@@ -241,19 +241,14 @@ public class CatalogInitialization implements ManagementContextInjectable {
         }
         synchronized (populatingCatalogMutex) {
             if (hasRunFinalInitialization()) {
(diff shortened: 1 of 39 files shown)


core/src/main/java/org/apache/brooklyn/core/catalog/internal/CatalogInitialization.java
```diff
@@ -241,24 +241,23 @@ public void populateInitialAndPersistedCatalog(ManagementNodeState mode, Persist
...
                 // Multiple calls; will need to reset (only way to clear out the previous persisted state's catalog)
                 if (log.isDebugEnabled()) {
                     String message = "Catalog initialization repeated call to add persisted catalog, resetting catalog (including initial) to start from clean slate; mode="+mode;
-                    if (!ManagementNodeState.isHotProxy(mode)) {
-                        log.debug(message);
-                    } else {
-                        // in hot modes, make this message trace so we don't get too much output then
-                        log.trace(message);
-                    }
+                    log.debug(message);
```
iuliana, line 251: Just put the string value here, no need for the message variable.

ahgittin:
merging so we can test more easily; more reviews welcome

PR #1267, opened 2021-10-28
Groups change policy

duncangrant:
This should now be working.

I've been testing this code with this yaml
```   

    location: localhost
    name: Brooklyn Deployment
    services:
      brooklyn.config:
        example: "dfdfsdfsfs"
        displayName: "MyEntity"
      brooklyn.initializers:
      - type: org.apache.brooklyn.core.sensor.StaticSensor
        brooklyn.config:
          name: tf.resource.type
          static.value: aws_instance

    - type: org.apache.brooklyn.entity.group.DynamicGroup
      name: EC2 instances
      brooklyn.policies:
      - type: org.apache.brooklyn.entity.group.GroupsChangePolicy
        brooklyn.config:
          group: $brooklyn:self()
          member.policies: kdjsldl
          member.initializers:
          - type: org.apache.brooklyn.core.sensor.StaticSensor
            brooklyn.config:
              name: testsensor
              target.type: string
              static.value: $brooklyn:formatString("%s%s","test","sensor")
          member.enrichers:
          - type: org.apache.brooklyn.policy.enricher.HttpLatencyDetector
            brooklyn.config:
              latencyDetector.url: http://localost:8081
              latencyDetector.period: 10s
              latencyDetector.requireServiceUp: false

      brooklyn.config:
        dynamicgroup.entityfilter:
          $brooklyn:object:
            type: org.apache.brooklyn.core.entity.EntityPredicates
            factoryMethod.name: displayNameEqualTo
            factoryMethod.args:
            - "Empty Software Process"



```

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/effector/AddSensorInitializer.java
+++ b/core/src/main/java/org/apache/brooklyn/core/effector/AddSensorInitializer.java
@@ -91,7 +91,7 @@ public class AddSensorInitializer<T> extends EntityInitializers.InitializerPatte
     private Duration period;
     private String type;
     private AttributeSensor<T> sensor;
-    private Object params;
+    private ConfigBag params;
     // introduced in 1.1 for legacy compatibility
     protected Object readResolve() {
         super.readResolve();
--- a/core/src/main/java/org/apache/brooklyn/core/effector/AddSensorInitializer.java
+++ b/core/src/main/java/org/apache/brooklyn/core/effector/AddSensorInitializer.java
@@ -91,7 +91,7 @@ public class AddSensorInitializer<T> extends EntityInitializers.InitializerPatte
     private Duration period;
     private String type;
     private AttributeSensor<T> sensor;
-    private ConfigBag params;
+    private Object params;
     // introduced in 1.1 for legacy compatibility
     protected Object readResolve() {
         super.readResolve();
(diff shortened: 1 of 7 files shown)


ahgittin:
So, changing `params` to be the more general `Object` doesn't allow some of the historic xml deserialization.

A better fix is to annotate `ConfigBag` so that Jackson JSON/YAML can work with it more nicely.  This on that class should do it:

```
@JsonAutoDetect(fieldVisibility = Visibility.ANY, isGetterVisibility = Visibility.NONE, getterVisibility = Visibility.NONE, setterVisibility = Visibility.NONE)
```

(Revert the change of `params`, back to being a `ConfigBag`.)

The test you added to `BRTJST` passes with the above annotation.

I tried the following test also in `BRTJST` which also passes and is probably worth including here (in addition to the test you have here):

```
    @Test
    public void testConfigBagSerialization() throws Exception {
        ConfigBag bag = ConfigBag.newInstance();
        bag.put(ConfigKeys.newConfigKey(String.class, "stringTypedKey"), "foo");
        bag.putStringKey("stringUntypedKey", "bar");
        bag.putStringKey("intUntypedKey", 2);
        bag.getStringKey("stringUntypedKey");

        String out = ser(bag);
        Assert.assertEquals(out, "{\"type\":\"org.apache.brooklyn.util.core.config.ConfigBag\",\"config\":{\"stringTypedKey\":\"foo\",\"stringUntypedKey\":\"bar\",\"intUntypedKey\":2},\"unusedConfig\":{\"stringTypedKey\":\"foo\",\"intUntypedKey\":2},\"live\":false,\"sealed\":false}");

        ConfigBag in = (ConfigBag) deser(out);
        // used and unused is serialized
        Asserts.assertSize(in.getUnusedConfig(), 2);
        Asserts.assertEquals(in.getAllConfig(), bag.getAllConfig());
    }
```

nakomis:
LGTM; Merging

PR #1270, opened 2021-11-03
Bump gson from 2.5 to 2.8.9 due to discovered vulnerability

jcabrerizo:
Bumping gson after snyk identified it as vulnerable: 
https://app.snyk.io/vuln/SNYK-JAVA-COMGOOGLECODEGSON-1730327

PR #1272, opened 2021-11-13
Bump guava 271jre

ahgittin:
In order to bump gson to 2.8.9 per https://github.com/apache/brooklyn-server/pull/1270, it is necessary also to bump jclouds to 2.4.0 which bumps guava to 27.1j.  As jclouds-karaf is not maintained, the definition of the jclouds feature is incorporated here.  This gives us more flexibiilty about start order so we've taken the opportunity to tidy up the feature definitions and optimize startup.

It now starts in 10 seconds with no refreshes!

However for sshj to work, due to an implicit reference to Curve in eddsa.math (which is neither exported from eddsa nor imported to sshj) we need to wrap those to bundles and redeclare the exports/imports.  Not sure how jclouds-karaf managed to do this in a cleaner way, but without this ssh wouldn't connect; now however confirmed that jclouds+ssh is working as part of this.

Note several more downstream PRs which will reference this.


ahgittin:
This closes #1270.

PR #1273, opened 2021-11-15
Add useful initialisers to the catalog

duncangrant:
In order to add initialisers to the catalog I've had to make them non-final and add public no-arg constructors.

### DIFF
--- a/core/src/test/java/org/apache/brooklyn/core/test/BrooklynAppLiveTestSupport.java
+++ b/core/src/test/java/org/apache/brooklyn/core/test/BrooklynAppLiveTestSupport.java
@@ -36,10 +36,17 @@ public class BrooklynAppLiveTestSupport extends BrooklynMgmtUnitTestSupport {
     @Override
     @BeforeMethod(alwaysRun=true)
     public void setUp() throws Exception {
+
         if (mgmt!=null) {
             app = mgmt.getEntityManager().createEntity(newAppSpec());
         } else {
-            mgmt = new LocalManagementContextForTests(BrooklynProperties.Factory.newDefault());
+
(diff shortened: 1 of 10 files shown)


core/src/test/java/org/apache/brooklyn/core/test/BrooklynAppLiveTestSupport.java
```diff
@@ -36,10 +36,17 @@
...
         if (mgmt!=null) {
             app = mgmt.getEntityManager().createEntity(newAppSpec());
         } else {
-            mgmt = new LocalManagementContextForTests(BrooklynProperties.Factory.newDefault());
+
+            LocalManagementContextForTests.Builder builder = LocalManagementContextForTests.builder(true).
+                    setOsgiEnablementAndReuse(true, true);
+            if (true) {
+                builder.useDefaultProperties();
+            }
```
algairim, line 48: Does not look like `if` statement needed.

algairim: I mean, `if (true)`

algairim (approved):
One minor comment, apart from that looks good.

zan-mateusz (approved):
LGTM!

core/src/test/java/org/apache/brooklyn/core/test/BrooklynAppLiveTestSupport.java
duncangrant: I'm not sure where this change has come from - I don't have it locally?

duncangrant: I must have messed up a rebase

PR #1276, opened 2021-11-30
pragmatic fix for AWS default hardware image

ahgittin:
use minCores to trigger t3.micro rather than a1.large, because otherwise default chosen a1 series
(since bump of jclouds to 2.4.0) is incompatible with many images

PR #1278, opened 2021-11-30
Pragmatic fix for AWS default hardware image

grkvlt:
Update to #1276 (see for more details)


ahgittin:
looks good.  as test pass is there any more needed before merge?  (doesn't look like WIP to me anymore!)

nakomis:
Live tests still fail, but that's due to stale imageid etc. Other than that, LGTM, so merging

PR #1280, opened 2021-12-01
Use toURI on resource URLs to fix Windows path issues

grkvlt:
Signed-off-by: Andrew Donald Kennedy <andrew.international@gmail.com>

### DIFF
--- a/utils/common/src/test/java/org/apache/brooklyn/util/io/FileUtilTest.java
+++ b/utils/common/src/test/java/org/apache/brooklyn/util/io/FileUtilTest.java
@@ -18,21 +18,24 @@
  */
 package org.apache.brooklyn.util.io;
 
+import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertFalse;
+import static org.testng.Assert.assertTrue;
+
 import java.io.File;
 import java.io.IOException;
+import java.net.URISyntaxException;
 import java.nio.file.Paths;
+
+import com.google.common.base.Charsets;
+import com.google.common.collect.ImmutableList;
+import com.google.common.io.Files;
 
 import org.apache.brooklyn.util.os.Os;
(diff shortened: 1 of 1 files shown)


utils/common/src/test/java/org/apache/brooklyn/util/io/FileUtilTest.java
```diff
@@ -123,26 +126,30 @@ public void testIsJavaFileNull(){
     }
 
     @Test
-    public void testIsJavaFileText() throws IOException {
-        byte[] bytes = java.nio.file.Files.readAllBytes(Paths.get(this.getClass().getClassLoader().getResource("brooklyn/osgi/brooklyn-osgi-test-a_0.1.0.txt").getPath()));
+    public void testIsJavaFileText() throws IOException, URISyntaxException {
+        byte[] bytes = java.nio.file.Files.readAllBytes(
+               Paths.get(this.getClass().getClassLoader().getResource("brooklyn/osgi/brooklyn-osgi-test-a_0.1.0.txt").toURI()));
         assertFalse(FileUtil.doesZipContainJavaBinaries(InputStreamSource.of("Test bom file", bytes)));
```
ahgittin, line 132: how about

```
        assertFalse(FileUtil.doesZipContainJavaBinaries(InputStreamSource.of("Test bom file",
                Streams.readFully(getClass().getResourceAsStream("/brooklyn/osgi/brooklyn-osgi-test-a_0.1.0.txt")))));
```

AFAIK attempting to read the URL as a file path is not guaranteed anyway, so `URL.toURI()` is a band-aid whereas stream is a fix.  (btw if using the toURI pattern it would be better to wrap all those calls in a `attemptToReadBytesFromAFile(path)` method with a comment that converting to URI increases reliability on windows.)

ahgittin (changes requested):

PR #1281, opened 2021-12-03
Fixes race where multiple RestApiResourceScanner created

grkvlt:
Always call the `install(Collection<Class>)` method, since this has the logic to determine whether we need to create a new scanner. Makes the constructors `private` to prevent this recurring, implements the caveats mentioned in `ScannerInjectHelper` here instead, and make the method `synchronized`.

Signed-off-by: Andrew Donald Kennedy <andrew.international@gmail.com>


ahgittin:
looks good!  merging

PR #1282, opened 2021-12-06
Fix rebind warnings

ahgittin:
fix four warnings discovered when rebinding old state (it succeeds without this, but clean it up!)

(A) default catalog does not cleanly replace previous default catalog

```
2021-12-06T11:41:20,996Z ZRsSuuDG- WARN  207 o.a.b.c.m.r.RebindExceptionHandlerImpl [tures-3-thread-1] No catalog item found with id ThreeTierApp:6.0.0; dangling reference on rebind (subsequent log messages will indicate context)
```

(B) search path not upgraded

```
2021-12-06T11:41:21,020Z ZRsSuuDG- WARN  207 o.a.b.c.c.i.CatalogUtils [tures-3-thread-1] Can't find catalog item org.apache.brooklyn.core:1.1.0.20210922_1514 when searching; a search path may be incomplete and other errors may follo
```

(C) redundant warning about removed bundles

```
2021-12-06T11:41:21,639Z - WARN  207 o.a.b.c.m.p.BrooklynMementoPersisterToObjectStore [tures-3-thread-1] Cannot find managed bundle for added bundle qjmej14cyn; ignoring (probably uninstalled or reinstalled with another OSGi ID; see debug log for contents)
```

(D) guava update renames should include osgi metadata, to avoid

```
2021-12-06T11:41:21,103Z ZRsSuuDG- WARN  207 o.a.b.u.c.ClassLoaderUtils [ooklyn-persister] No class 'com.google.common.collect.EmptyImmutableBiMap' in bundle 'com.google.guava' (using spec 'com.google.guava:com.google.common.collect.EmptyImmutableBiMap'); found using deprecated no-bundle syntax, but this behaviour is deprecated and likely to be unsupported in future. Change so invalid bundle is not supplied.
```

### DIFF
--- a/api/src/main/java/org/apache/brooklyn/api/mgmt/rebind/mementos/BrooklynMementoPersister.java
+++ b/api/src/main/java/org/apache/brooklyn/api/mgmt/rebind/mementos/BrooklynMementoPersister.java
@@ -98,8 +98,8 @@ public interface BrooklynMementoPersister {
     Set<String> getLastErrors();
 
     /** applies a full checkpoint (write) of all state; returns true on success or false on error, with {@link #getLastErrors()} available to get the last errors */
-    boolean checkpoint(BrooklynMementoRawData newMemento, PersistenceExceptionHandler exceptionHandler);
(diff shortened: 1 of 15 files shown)


ahgittin:
confirmed all warnings removed!

PR #1283, opened 2021-12-06
Calling unmanage on a removed child that causes the corresponding dyn…

zan-mateusz:
…amicgroup item to be removed

As above. When removing children that have corresponding items in a group (e.g. DynamicGroup), the corresponding item in the group was not removed. Calling the unmanage() method when removing the child causes necessary event to be triggered, essentially removing the item from the group as required.


iuliana (approved):
LGTM (if you tested and it does what it should, I'm ok to merge)

iuliana (approved):

iuliana:
But also, check if this build failure is caused by this change. I won't merge otherwise.

ahgittin:
Don't think this is right -- caller code should unmanage the child, not ask the parent to remove it.

zan-mateusz:
The test failures are indeed caused by this change. As per discussions this can be achieved in another way without having to modify brooklyn server code, so that is what I will proceed with. Closing this PR.

PR #1286, opened 2021-12-18
Bumped logback from 1.2.3 to 1.2.9 due to CVE-2021-42550

jcabrerizo:
References:
http://logback.qos.ch/news.html
https://nvd.nist.gov/vuln/detail/CVE-2021-42550

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -114,7 +114,7 @@
 
         <!-- Dependency Versions -->
         <jclouds.version>2.4.0</jclouds.version> <!-- JCLOUDS_VERSION -->
-        <logback.version>1.2.3</logback.version>
+        <logback.version>1.2.9</logback.version>
         <slf4j.version>1.7.25</slf4j.version>  <!-- used for java.util.logging jul-to-slf4j interception -->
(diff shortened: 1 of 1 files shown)


geomacy (approved):
LGTM

As an aside, I expect we could get rid of [this](https://github.com/apache/brooklyn-server/blob/97281ed3aa83434a212606fd62dc06299d15fe8d/pom.xml#L210), couldn't we? I think that's likely a historical leftover?
```xml
        <log4j.version>1.2.17</log4j.version>
```

jcabrerizo:
Hi @geomacy , as far i can see it's not used, but I didn't remove it just in case is a project expending the same parent use it and the removal could cause any issue.
I'm closing now this PR in favour of https://github.com/apache/brooklyn-server/pull/1287 that includes also the new version of Karaf.

Thank you for your comment.

geomacy:
> Hi @geomacy , as far i can see it's not used, but I didn't remove it just in case is a project expending the same parent use it and the removal could cause any issue. I'm closing now this PR in favour of #1287 that includes also the new version of Karaf.
> 
> Thank you for your comment.

Guess that makes sense thanks @jcabrerizo. We'll have to figure out a policy that lets us get rid of this sort of thing eventually though.

PR #1287, opened 2021-12-20
Bumping Karaf to the staging version 4.3.5 for update

jcabrerizo:
That release includes Pax Logging 2.0.13 update:
- with logback 1.2.9 upgrade, fixing CVE-2021-42550
- with log4j 2.17.0 upgrade, fixing CVE-2021-45105

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -86,6 +86,8 @@
             <snapshots>
                 <enabled>false</enabled>
             </snapshots>
+        </repository>
+    </repositories>
     <pluginRepositories>
         <pluginRepository>
             <id>testKaraf435-plugin</id>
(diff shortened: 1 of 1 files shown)


geomacy (approved):
LGTM

Karaf 4.3.5 [uses](https://github.com/apache/karaf/blob/karaf-4.3.5/pom.xml#L297)  pax logging 2.0.13, which in turn [uses](https://github.com/ops4j/org.ops4j.pax.logging/blob/logging-2.0.13/pom.xml#L232) Apache Log4j logging 2.17.0, which contains a series of mitigations documented [here](https://logging.apache.org/log4j/2.x/changes-report.html#a2.17.0).

PR #1289, opened 2022-01-06
Allow effector-on-sensor-change subscribing to any entity

algairim:


core/src/main/java/org/apache/brooklyn/policy/InvokeEffectorOnSensorChange.java
```diff
@@ -19,14 +19,17 @@
 package org.apache.brooklyn.policy;
 
 import org.apache.brooklyn.api.effector.Effector;
+import org.apache.brooklyn.api.entity.Entity;
```
algairim, line 22: TODO: remove all unused imports.

duncangrant (approved):
@algairim This looks good to me - this pattern is used on other policies and makes sense here also - and thanks for the tests

PR #1297, opened 2022-02-07
New functionals to apply to an iterable: generic AllEqualsFunction an…

algairim:
…d a sub-functin AllTrueFunction to check if all true in particular

### DIFF
--- a/utils/common/src/main/java/org/apache/brooklyn/util/collections/CollectionFunctionals.java
+++ b/utils/common/src/main/java/org/apache/brooklyn/util/collections/CollectionFunctionals.java
@@ -18,11 +18,7 @@
  */
 package org.apache.brooklyn.util.collections;
 
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
 
 import javax.annotation.Nullable;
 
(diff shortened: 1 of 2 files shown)


utils/common/src/main/java/org/apache/brooklyn/util/collections/CollectionFunctionals.java
```diff
@@ -119,6 +115,35 @@ public T apply(Iterable<? extends T> input) {
...
+        public AllEqualsFunction(Comparable<?> value) {
+            if (Objects.isNull(value)) {
+                throw new IllegalArgumentException("The value to compare all to must be defined");
+            }
+            this.value = value;
+        }
+
+        @Override
+        public Boolean apply(@Nullable Iterable<?> input) {
+            if (Objects.isNull(input) || Iterables.isEmpty(input)) return null;
```
jcabrerizo, line 130: I think it should return `false` if the input is empty or null unless the comparable is an empty iterable or a null value.

Not sure how functional is my logic about that

jcabrerizo (commented):
What do you think about my comment @algairim ?

utils/common/src/main/java/org/apache/brooklyn/util/collections/CollectionFunctionals.java
algairim: Moreover, if there is a DSL watching this `$brooklyn:attributeWhenReady("service.process.isRunning")`, then it will wait until there is a value to read.

algairim: This should address your doubt @jcabrerizo

jcabrerizo: Agree, but for me, it looks like we need to have flag for do not evaluate unless is not empty or not null. 
Yes, the use of `brooklyn:attributeWhenReady` will help, but if only the function used, the transformer will return null if the source sensor is not returning a false value.

I'm happy to approve but please, add some comments explaining why it will return `null` comparing with an empty collection

algairim: Documented in 568af812d307c52c6cb7f1abbb6e363526ddd3ec

jcabrerizo (approved):
Thanks @algairim I'll merge when the tests pass

PR #1298, opened 2022-02-07
Call effector if value of sensor being watched changes only

algairim:

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/policy/InvokeEffectorOnSensorChange.java
+++ b/core/src/main/java/org/apache/brooklyn/policy/InvokeEffectorOnSensorChange.java
@@ -88,7 +88,7 @@ public class InvokeEffectorOnSensorChange extends AbstractInvokeEffectorPolicy i
     public void onEvent(SensorEvent<Object> event) {
         final Object newValue = event.getValue();
         LOG.debug("lastValue='{}', newValue='{}'", lastValue, newValue);
-        if (String.valueOf(newValue).equals(lastValue)) {
+        if (newValue instanceof Comparable && newValue.equals(lastValue)) {
             return;
         }
         lastValue = newValue;
--- a/core/src/main/java/org/apache/brooklyn/policy/InvokeEffectorOnSensorChange.java
+++ b/core/src/main/java/org/apache/brooklyn/policy/InvokeEffectorOnSensorChange.java
@@ -66,6 +66,7 @@ public class InvokeEffectorOnSensorChange extends AbstractInvokeEffectorPolicy i
             .constraint(StringPredicates.isNonBlank())
             .build();
 
+    private final transient Object lock = new Object[0];
     private AttributeSensor<Object> sensor;
     private Object lastValue;
 
@@ -87,11 +88,14 @@ public class InvokeEffectorOnSensorChange extends AbstractInvokeEffectorPolicy i
     @Override
     public void onEvent(SensorEvent<Object> event) {
         final Object newValue = event.getValue();
-        LOG.debug("lastValue='{}', newValue='{}'", lastValue, newValue);
-        if (newValue instanceof Comparable && newValue.equals(lastValue)) {
-            return;
+
+        synchronized (lock) {
+            LOG.debug("lastValue='{}', newValue='{}'", lastValue, newValue);
+            if (newValue instanceof Comparable && newValue.equals(lastValue)) {
+                return;
+            }
+            lastValue = newValue;
         }
-        lastValue = newValue;
 
         if (isBusySensorEnabled()) {
             final Object currentSensorValue = entity.sensors().get(sensor);
--- a/core/src/main/java/org/apache/brooklyn/policy/InvokeEffectorOnSensorChange.java
+++ b/core/src/main/java/org/apache/brooklyn/policy/InvokeEffectorOnSensorChange.java
@@ -46,7 +46,7 @@ import com.google.common.reflect.TypeToken;
  * * allow to be triggered by sensors on members
  */
 public class InvokeEffectorOnSensorChange extends AbstractInvokeEffectorPolicy implements SensorEventListener<Object> {
-    
+
     private static final Logger LOG = LoggerFactory.getLogger(InvokeEffectorOnSensorChange.class);
 
     public static final ConfigKey<Object> SENSOR = ConfigKeys.builder(Object.class)
@@ -67,7 +67,7 @@ public class InvokeEffectorOnSensorChange extends AbstractInvokeEffectorPolicy i
             .build();
 
     private AttributeSensor<Object> sensor;
-
+    private Object lastValue;
 
     @Override
     public void setEntity(EntityLocal entity) {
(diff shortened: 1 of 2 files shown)


duncangrant:
@algairim Why are you making this change?  I think you can already get this behaviour by using an enricher with suppress duplicates.  Is this just to make it easier to do this? 

Should we use a suppress duplicates flag so that anyone relying on the current behaviour doesn't see their blueprint stop working.

core/src/main/java/org/apache/brooklyn/policy/InvokeEffectorOnSensorChange.java
```diff
@@ -67,7 +67,7 @@
             .build();
 
     private AttributeSensor<Object> sensor;
-
+    private Object lastValue;
```
duncangrant, line 70: This won't survive rebind and I'm not sure it's thread safe (it might be - I'd need to dig into the parent's code)

core/src/main/java/org/apache/brooklyn/policy/InvokeEffectorOnSensorChange.java
```diff
@@ -79,18 +79,26 @@ public void setEntity(EntityLocal entity) {
...
         subscriptions().subscribe(producer, sensor, this);
         LOG.debug("{} subscribed to {} events on {}", new Object[]{this, sensor, entity});
     }
 
     @Override
     public void onEvent(SensorEvent<Object> event) {
-        final Effector<?> eff = getEffectorNamed(getConfig(EFFECTOR)).get();
+        final Object newValue = event.getValue();
+        LOG.debug("lastValue='{}', newValue='{}'", lastValue, newValue);
+        if (String.valueOf(newValue).equals(lastValue)) {
```
duncangrant, line 91: I don't understand why you're converting to a string here - sensors can hold any object type.

algairim: - [ ] You are right, I'll cast this to `Comparable`, left this from experimentation.

algairim:
> @algairim Why are you making this change? I think you can already get this behaviour by using an enricher with suppress duplicates. Is this just to make it easier to do this?
> 
> Should we use a suppress duplicates flag so that anyone relying on the current behaviour doesn't see their blueprint stop working.

@duncangrant , which enricher are you referring to and how can I combine it with this policy?

core/src/main/java/org/apache/brooklyn/policy/InvokeEffectorOnSensorChange.java
algairim: Agree, should be synchronised. I completely forgot about rebind!

algairim: Line 82 addresses rebind actually. That is where we know the initial value to ocmpare to new from events.

core/src/main/java/org/apache/brooklyn/policy/InvokeEffectorOnSensorChange.java
algairim: 9cfe27fdca482df87febc49aa6a8b071512a44fa

core/src/main/java/org/apache/brooklyn/policy/InvokeEffectorOnSensorChange.java
```diff
@@ -79,18 +80,29 @@ public void setEntity(EntityLocal entity) {
             LOG.debug("Defaulting to producer==self for {}, on entity {}", this, entity);
             producer = entity;
         }
+        lastValue = entity.sensors().get(sensor);
```
algairim, line 83: This is where last known value is expected to be recovered after rebind.

core/src/main/java/org/apache/brooklyn/policy/InvokeEffectorOnSensorChange.java
algairim: 08aa7c3447658c5c6e5f518bca631b33526fa9d5

core/src/main/java/org/apache/brooklyn/policy/InvokeEffectorOnSensorChange.java
duncangrant: Good point - I missed this

duncangrant:
> > @algairim Why are you making this change? I think you can already get this behaviour by using an enricher with suppress duplicates. Is this just to make it easier to do this?
> > Should we use a suppress duplicates flag so that anyone relying on the current behaviour doesn't see their blueprint stop working.
> 
> @duncangrant , which enricher are you referring to and how can I combine it with this policy?

Propagator with enricher.suppressDuplicates true would work but I don't think that would work with the same target and source entity so maybe there isn't an option that works well here.

duncangrant (changes requested):
From our discussion I think this is useful but I think it needs a suppress duplicates flag in line with enrichers and sensor feeds rather than always do this.

core/src/main/java/org/apache/brooklyn/policy/InvokeEffectorOnSensorChange.java
ahgittin: i think this will be included in rebind anyway (even without being set in `setEntity`).  you could mark it `transient` to exclude it from rebind.

algairim:
> From our discussion I think this is useful but I think it needs a suppress duplicates flag in line with enrichers and sensor feeds rather than always do this.

I did not notice that we have this option in all our enrichers, let me check my use case.

core/src/main/java/org/apache/brooklyn/policy/InvokeEffectorOnSensorChange.java
```diff
@@ -79,18 +80,29 @@ public void setEntity(EntityLocal entity) {
...
         subscriptions().subscribe(producer, sensor, this);
         LOG.debug("{} subscribed to {} events on {}", new Object[]{this, sensor, entity});
     }
 
     @Override
     public void onEvent(SensorEvent<Object> event) {
-        final Effector<?> eff = getEffectorNamed(getConfig(EFFECTOR)).get();
+        final Object newValue = event.getValue();
+
+        synchronized (lock) {
```
ahgittin, line 92: this synch block seems pointless to me, if there were duplicate events coming in concurrently there is no guarantee which one is set first.

however i believe subscription delivery is already synched to ensure in-order delivery anyway.

so remove this synch block.

core/src/main/java/org/apache/brooklyn/policy/InvokeEffectorOnSensorChange.java
```diff
@@ -79,18 +80,29 @@ public void setEntity(EntityLocal entity) {
...
+            LOG.debug("lastValue='{}', newValue='{}'", lastValue, newValue);
+            if (newValue instanceof Comparable && newValue.equals(lastValue)) {
```
ahgittin, line 94: can we add a `ConfigKey SUPPRESS_DUPLICATES` (this is used elsewhere), and only exit here if that is true; it can default to true but at least that way if there is a legacy requirement somewhere to act even if values don't change then we can support this.

ahgittin:
a few changes requested, otherwise looks good:

* `transient lastValue`
* no `synchronized`
* add `SUPPRESS_DUPLICATES` config key

algairim:
No need in this change, I should have used `enricher.suppressDuplicates` option available in all enrichers instead, as @duncangrant suggested.

PR #1320, opened 2022-06-08
Introduce entity adjunct proxy, fix deserialization edge case

ahgittin:
If one BrooklynObject refers to another in an unusual place (eg in a map, not simply policies on entities or vice versa) and the referrer is deserialized before the referent, Apache Brooklyn would give an error.  With this PR, in these cases we introduce an EntityAdjunctProxy, similar to the EntityProxy (long since used for the much more common case of entities referring to other entities).

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/mgmt/rebind/RebindIteration.java
+++ b/core/src/main/java/org/apache/brooklyn/core/mgmt/rebind/RebindIteration.java
@@ -63,6 +63,7 @@ import org.apache.brooklyn.api.mgmt.rebind.mementos.PolicyMemento;
 import org.apache.brooklyn.api.mgmt.rebind.mementos.TreeNode;
 import org.apache.brooklyn.api.objs.BrooklynObject;
 import org.apache.brooklyn.api.objs.BrooklynObjectType;
+import org.apache.brooklyn.api.objs.EntityAdjunct;
 import org.apache.brooklyn.api.policy.Policy;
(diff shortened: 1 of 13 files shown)


core/src/main/java/org/apache/brooklyn/core/mgmt/rebind/RebindIteration.java
```diff
@@ -484,11 +483,23 @@ protected void instantiateLocationsAndEntities() {
...
+
     protected void instantiateMementos() throws IOException {
 
         checkEnteringPhase(4);
 
+        if (!adjunctProxies.isEmpty()) {
+            LOG.warn("Had stale adjunct information when rebinding; ignoring: "+adjunctProxies);
+        }
+        adjunctProxies.clear();
+
```
jcabrerizo, line 499: Clear is only need if is not empty, moving this inside the exiting `if` above will be a minor improve

jcabrerizo (approved):
LGTM @ahgittin, thank you.
I added a minor comment for your consideration. Fell free to ignore it and merge the PR

ahgittin:
Two new commits to handle case where adjunct is in a set and so needs access to hashcode and related methods.  Now uses the ID to compute hashcode, to align with entity and ensure hashcode doesn't change when proxy gets assigned the delegate.  (As hashcodes were not persisted and had no guarantee about the value there should be no issue with changing the adjunct hashcode behaviour.)

jcabrerizo:
Thanks for this @ahgittin

PR #1321, opened 2022-06-20
Adding Kubectl Task factory, Docker effector  and Docker Sensor

iuliana:
**What**

`kubectl` task factory that allows blueprint designers to externalize effectors and sensors  to containers run on a Kubernetes cluster.

**Why**
Because it is practical and higly customizable. 

**Testing**
Sample blueprints are provided. Also tests for sensor, effector and task factory.

### DIFF
--- a/software/base/src/main/java/org/apache/brooklyn/tasks/kubectl/ContainerCommons.java
+++ b/software/base/src/main/java/org/apache/brooklyn/tasks/kubectl/ContainerCommons.java
@@ -19,10 +19,14 @@
 package org.apache.brooklyn.tasks.kubectl;
 
 import com.google.common.collect.Lists;
+import com.google.common.collect.Maps;
 import org.apache.brooklyn.config.ConfigKey;
 import org.apache.brooklyn.core.config.ConfigKeys;
 
+import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 
 @SuppressWarnings({ "rawtypes"})
 public interface ContainerCommons {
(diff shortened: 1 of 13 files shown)


software/base/src/main/java/org/apache/brooklyn/tasks/kubectl/ContainerTaskFactory.java
```diff
@@ -0,0 +1,133 @@
...
+        TaskBuilder<RET> taskBuilder = Tasks.<RET>builder()
+                .displayName(this.summary)
+                .add(buildKubeTask(configBag, "Set Up and Run",
+                        String.format(NAMESPACE_CREATE_CMD,containerName),
+                        String.format(NAMESPACE_SET_CMD,containerName)))
+                .add(runCommandsTask)
+                .add(waitTask)
+                .add(buildKubeTask(configBag, "Retrieve Output", String.format(JOBS_LOGS_CMD,containerName)));
+
+        if(!devMode) {
```
jcabrerizo, line 90: My feeling about `devMode` is it could be a risk allow not to remove a sensor container based on a blueprint parameter. It will create a new container periodically and if is not removed will consume resources and pull down the kubernetes cluster eventually.

Keeping that property in the `booklyn.cfg` file let the Brooklyn admin to allow or not that behaviour at the server level instead the blueprints and I think is safer -> dev server allow dev blueprints in a dev kubernetes cluster, but the production server doesn't

jcabrerizo (changes requested):
HI @iuliana this looks very good. 

My only concern is about the `devMode` as blueprint parameter instead of a server configuration. What do you think? @ahgittin any comments on that?

software/base/src/main/java/org/apache/brooklyn/tasks/kubectl/ContainerTaskFactory.java
ahgittin: I would rename `devMode` as `keepContainersForDebugging` with an additional comment in description that if true, the operator must take responsibility for cleaning up containers.  It should be easy enough, e.g. a cron job which deletes finished containers matching the brooklyn container naming convention, no?

So I don't think we yet need a global property to forbid this, though that could be an enhancement in future.

iuliana: @ahgittin  renamed to `keepContainerForDebugging`

jcabrerizo (approved):
Thanks @iuliana and @ahgittin

ahgittin:
one minor comment, otherwise LGTM

software/base/src/main/java/org/apache/brooklyn/tasks/kubectl/ContainerCommons.java
```diff
@@ -34,6 +38,10 @@ public interface ContainerCommons {
     ConfigKey<List> COMMANDS = ConfigKeys.newConfigKey(List.class,"commands", "Commands to execute for container", Lists.newArrayList());
     ConfigKey<List> ARGUMENTS = ConfigKeys.newConfigKey(List.class,"args", "Arguments to execute for container", Lists.newArrayList());
 
+    ConfigKey<String> WORKING_DIR = ConfigKeys.newStringConfigKey("workingDir", "Location where the container commands are executed");
+    ConfigKey<List<Map<String,String>>> VOLUME_MOUNTS = ConfigKeys.newConfigKey("volumeMounts", "Configuration to  mount a volume  into a container.", Lists.newArrayList());
```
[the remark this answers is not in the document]
iuliana: Addressed comments in the most recent commit. @ahgittin

PR #1322, opened 2022-06-22
fix most reliances on deleteOnExit, deleting when done with the file

ahgittin:
catches leaks to /tmp file system if AMP is long-running or killed non-cleanly

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/mgmt/ha/BrooklynBomOsgiArchiveInstaller.java
+++ b/core/src/main/java/org/apache/brooklyn/core/mgmt/ha/BrooklynBomOsgiArchiveInstaller.java
@@ -96,8 +96,21 @@ public class BrooklynBomOsgiArchiveInstaller {
     private boolean force = false;
     private boolean deferredStart = false;
     private boolean validateTypes = true;
-    
-    private File zipFile;
+
+    public static class FileWithTempInfo<T extends File> {
+        public FileWithTempInfo(T f, boolean isTemp) { this.f = f; this.isTemp = isTemp; }
+        public T f;
+        public boolean isTemp;
+
+        public void deleteIfTemp() {
+            if (isTemp && f!=null) {
+                f.delete();
+                f = null;
+            }
+        }
+    }
+
+    private FileWithTempInfo<File> zipFile;
     private boolean isBringingExistingOsgiInstalledBundleUnderBrooklynManagement = false;
     private Manifest discoveredManifest;
     private VersionedName discoveredBomVersionedName;
@@ -177,7 +190,7 @@ public class BrooklynBomOsgiArchiveInstaller {
         /** set if we anticipate a caller may want to read from a ZIP file;
          * this is for the case where no input stream was supplied but
          * OSGi bundle co-ordinates were supplied and can be used to find a ZIP */
-        public File zipFile;
+        public FileWithTempInfo<File> zipFile;
     }
     public static PrepareInstallResult prepareInstall(ManagementContext mgmt, ManagedBundle suppliedKnownBundleMetadata, ManagedBundle optionalAdditionalInferredMetadata,
                                                       Supplier<InputStream> zipInS, boolean force, @Nullable OsgiBundleInstallationResult resultObject) {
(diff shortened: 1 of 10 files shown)


ahgittin:
Commit-then-review requested @jcabrerizo

core/src/main/java/org/apache/brooklyn/core/mgmt/ha/BrooklynBomOsgiArchiveInstaller.java
```diff
@@ -96,8 +96,21 @@ public class BrooklynBomOsgiArchiveInstaller {
     private boolean force = false;
     private boolean deferredStart = false;
     private boolean validateTypes = true;
-    
-    private File zipFile;
+
+    public static class FileWithTempInfo<T extends File> {
+        public FileWithTempInfo(T f, boolean isTemp) { this.f = f; this.isTemp = isTemp; }
+        public T f;
```
jcabrerizo, line 102: I think `f` should be final to avoid it to be replaced. 
It's not very importand, but I'd feel more comfortable if instead of been public, it's exposed by a getter method, it also will make this code more clear:

```
FileWithTempInfo tempFile = new FileWithTempInfo(file, true);

...
tempFile.getFile();
_vs_
tempFile.f;
```

core/src/main/java/org/apache/brooklyn/util/core/file/ArchiveBuilder.java
```diff
@@ -44,6 +44,7 @@
 import com.google.common.collect.LinkedHashMultimap;
 import com.google.common.collect.Multimap;
 import com.google.common.io.Files;
+import org.apache.brooklyn.util.time.Duration;
```
jcabrerizo, line 47: This new import is not used

core/src/main/java/org/apache/brooklyn/util/core/file/ArchiveUtils.java
```diff
@@ -213,8 +213,10 @@ public static void deploy(String archiveUrl, SshMachineLocation machine, String
      * @see #deploy(Map, String, SshMachineLocation, String, String, String)
      */
     public static void deploy(Map<String, ?> props, String archiveUrl, SshMachineLocation machine, String destDir) {
+        List<File> filesToDelete = MutableList.of();
```
jcabrerizo, line 216: I  can't see why this is is a list of files if it only could content one zip file.
This will a bit faster and consume less memory
```
        File toDelete = null;
        if (Urls.isDirectory(archiveUrl)) {
            File zipFile = ArchiveBuilder.zip().entry(".", Urls.toFile(archiveUrl)).create();
            toDelete = zipFile;
            archiveUrl = zipFile.getAbsolutePath();
        }
        ...
        if(toDelete != null) toDelete.delete();
```

software/base/src/main/java/org/apache/brooklyn/entity/brooklynnode/BrooklynNodeSshDriver.java
```diff
@@ -261,8 +262,10 @@ public void customize() {
             checkNotNull(url, "url");
 
             // If a local folder, then create archive from contents first
+            List<File> filesToDelete = MutableList.of();
```
jcabrerizo, line 265: Same as above

software/base/src/main/java/org/apache/brooklyn/entity/java/VanillaJavaAppSshDriver.java
```diff
@@ -85,9 +85,11 @@ public void customize() {
         SshMachineLocation machine = getMachine();
         VanillaJavaApp entity = getEntity();
         for (String entry : entity.getClasspath()) {
+            List<File> filesToDelete = MutableList.of();
```
jcabrerizo, line 88: Same as above

core/src/main/java/org/apache/brooklyn/util/core/file/ArchiveBuilder.java
ahgittin: good spot

core/src/main/java/org/apache/brooklyn/util/core/file/ArchiveUtils.java
ahgittin: it's the same pattern used elsewhere which i think is nice for consistency, though minor either way imo

jcabrerizo: Makes sense. Thanks for the clarification, I though I was missing something.

PR #1324, opened 2022-06-23
include effector params in the tag on tasks, as returned by REST API

ahgittin:
so effector can be re-invoked


ahgittin:
needed for https://github.com/apache/brooklyn-ui/pull/339

PR #1325, opened 2022-06-24
Fix persistence deletion race

ahgittin:
Build on #1307, fixing the races that could cause persistence files for unmanaged items.  Expands test to cover enrichers also.  Several routes identified and fixed, plus better warnings, per the individual commits.

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/entity/Entities.java
+++ b/core/src/main/java/org/apache/brooklyn/core/entity/Entities.java
@@ -816,6 +816,14 @@ public class Entities {
         return ((EntityInternal)e).getManagementSupport().isNoLongerManaged();
     }
 
+    public static boolean isUnmanaging(Entity e) {
+        return ((EntityInternal)e).getManagementSupport().isUnmanaging();
+    }
+
+    public static boolean isUnmanagingOrNoLongerManaged(Entity e) {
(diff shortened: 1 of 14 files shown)


core/src/main/java/org/apache/brooklyn/core/entity/Entities.java
```diff
@@ -948,10 +955,22 @@ public static ManagementContext startManagement(Application app, ManagementConte
...
         return mgmt;
     }
 
+    private static void premanageRecursively(ManagementContextInternal mgmt, Entity e) {
+        if (Entities.isUnmanagingOrNoLongerManaged(e) || Entities.isManagedActive(e)) {
+            log.warn("Skipping premanagement for "+e+", as it has some form of known management");
+            // skip for active and unmanaged entities
+            return;
+        }
+        mgmt.prePreManage(e);
```
algairim, line 970: Can it be that this condition is required for a particular entity, but not for its children?

algairim: In other words: should we proceed recursively with children?

ahgittin: this method is called `premanageRecursively` so i think the answer is yes.  this is only needed for legacy code (tests), which does want the recursive behaviour.

PR #1329, opened 2022-06-29
Added support for pull policy and docs

iuliana:
As the title says.

### DIFF
--- a/software/base/README.adoc
+++ b/software/base/README.adoc
@@ -1,0 +1,108 @@
+////
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+////
+
+= Apache Brooklyn Software Components
+
+This project contains useful software types for building applications.
+
+== Kubectl components
+
+The `org.apache.brooklyn.tasks.kubectl` package contains types `DockerSensor` and `DockerEffector` that can be used to 
+declare simple actions to be executed on a container managed by a kubernetes cluster. To use them you have to make sure
+the location where the app was deployed has `kubectl` installed and configured to use local or remote Kubernetes cluster. 
+
+For development, we recommend using Minikube locally.
+
+=== The Job Builder
+
+In package `org.apache.brooklyn.tasks.kubectl`  there is also a class named `JobBuilder`. This class takes Brooklyn properties and injects them into a Kubernetes job template. It does so by creating matching Java objects, populating them with the property values, sometimes formatting them and serializing them all into a temporary YAML file that is run with `kubectl`. Files are saved in the temporary system folder.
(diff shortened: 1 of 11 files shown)


algairim (approved):

software/base/src/main/java/org/apache/brooklyn/tasks/kubectl/JobBuilder.java
```diff
@@ -121,6 +130,11 @@ public String build(){
             containerSpec.setWorkingDir(workingDir);
         }
         containerSpec.setImage(imageName);
+        if(Strings.isNonBlank(imagePullPolicy)) {
+            if (PULL_POLICY_ALLOWED_VALUES.contains(imagePullPolicy)) {
+                containerSpec.setImagePullPolicy(imagePullPolicy);
+            } // else stick with default, do not fail
```
geomacy, line 136: surely it would be better to fail?

algairim: @geomacy, the default policy that `kubectl` sets is 'Always'. And now, we simply allow to configure it.

algairim: @iuliana, to remove that code comment `do not fail`?

ahgittin: agree - i'd say either:

* always set the user-supplied value, let kubectl fail if it is invalid (good if they might add new valid args)
* or if we are going to try to duplicate enforcement, use an enum and specify that as the type of the config key (allows type introspection to determine valid values)

silently ignoring is going to confuse a user (esp if kubectl is case insensitive and we just drop things that don't match our arbitrary case!)

software/base/README.adoc
```diff
@@ -0,0 +1,108 @@
...
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+////
+
+= Apache Brooklyn Software Components
```
ahgittin, line 20: this belongs in `brooklyn-docs` project not here, no?

ahgittin:
good additions but a few comments.

i also wonder if there is mileage in an "extra_kubectl_args" list config key for other things we don't support or which get added in future versions?  (that would have allowed pull policy to be specified without code changes here.)

iuliana:
> 

@ahgittin this should like a good idea, but the Job template has multiple levels. There are properties that you can add for `containers`, `spec`, `template` , etc.  `extra_kubectl_args` then would have to be a multi-level map, to cover these cases and this might complicate the `JobBuilder` implementation.  I guess it could be done, if really necessary. Maybe keep this for later after we have the core functionality working.

software/base/src/main/java/org/apache/brooklyn/tasks/kubectl/JobBuilder.java
iuliana: @ahgittin `enum` it is then.

geomacy: > [ahgittin](https://github.com/ahgittin) [2 days ago](https://github.com/apache/brooklyn-server/pull/1329#discussion_r910735955)  silently ignoring is going to confuse a user (esp if kubectl is case insensitive and we just drop things that don't match our arbitrary case!)

That was my thought.

> [iuliana](https://github.com/iuliana) [2 days ago](https://github.com/apache/brooklyn-server/pull/1329#discussion_r911071862)
@geomacy Added enum for imagePullPolicy that validates the value, before the task is even created.

looks great thanks @iuliana

geomacy (approved):
LGTM

PR #1333, opened 2022-07-05
Added proper symbol removal and edited tests to make sure it works as…

iuliana:
What?
Added proper symbol removal(eliminated `redundant` replace calls) and edited tests to make sure it works as intended.
Also reduced the timeout for the negative test.

Why?
Do I really have to explain?  ;)

### DIFF
--- a/software/base/src/main/java/org/apache/brooklyn/tasks/kubectl/ContainerTaskFactory.java
+++ b/software/base/src/main/java/org/apache/brooklyn/tasks/kubectl/ContainerTaskFactory.java
@@ -68,10 +68,8 @@ public class ContainerTaskFactory<T extends ContainerTaskFactory<T,RET>,RET>  im
 
         final String containerName = (Strings.isBlank(containerNameFromCfg)
                 ? ( (Strings.isNonBlank(this.tag) ? this.tag + "-" : "").concat(cleanImageName).concat("-").concat(Strings.makeRandomId(10)))
-                : containerNameFromCfg).replace(" ", "-")
-                .replace("/", "-")
-                .replace("_", "-")
-                .replaceAll("[^a-z0-9-]", "") // remove other symbols
+                : containerNameFromCfg)
+                .replaceAll("[^A-Za-z0-9-]", "") // remove all symbols
                 .toLowerCase();
 
         final String jobYamlLocation =  new JobBuilder()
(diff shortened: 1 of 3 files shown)


software/base/src/main/java/org/apache/brooklyn/tasks/kubectl/ContainerTaskFactory.java
```diff
@@ -68,10 +68,8 @@ public Task<RET> newTask() {
 
         final String containerName = (Strings.isBlank(containerNameFromCfg)
                 ? ( (Strings.isNonBlank(this.tag) ? this.tag + "-" : "").concat(cleanImageName).concat("-").concat(Strings.makeRandomId(10)))
-                : containerNameFromCfg).replace(" ", "-")
-                .replace("/", "-")
-                .replace("_", "-")
-                .replaceAll("[^a-z0-9-]", "") // remove other symbols
+                : containerNameFromCfg)
+                .replaceAll("[^A-Za-z0-9-]", "") // remove all symbols
```
algairim, line 72: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/

software/base/src/test/java/org/apache/brooklyn/tasks/kubectl/ContainerTaskTest.java
```diff
@@ -103,6 +103,7 @@ public void testFailingContainerTask() {
         configBag.put("name", "test-container-task");
         configBag.put("image", "perl");
         configBag.put("commands", commands);
+        configBag.put("timeout", "1m");
```
algairim, line 106: Timeout is a default '1m' in the config key.

iuliana: So... you changes your mind about it yesterday :), got it.

software/base/src/main/java/org/apache/brooklyn/tasks/kubectl/ContainerTaskFactory.java
iuliana: The lowercase call at the end ensures that we have no upper-case characters, but you cannot just remove them, because they might be relevant for something. For example tags can be used to discern between a job being run in a sensor and one in an effector. With your change you removed that completely and I had a lot of pods with `--` in them and their names containing mostly gibberish. ;)

iuliana: I know.   :))

PR #1336, opened 2022-07-12
Fix log viewer glitches

ahgittin:
image

fixes ^ (log messages with no message match any search term), also `NumberFormatException` occasionally displaying, date endpoint logic mistake, and code tidy


jcabrerizo (approved):
LGTM @ahgittin thank you

PR #1337, opened 2022-07-12
better "interesting task" analysis for API when working with limits

ahgittin:
take at least 1 task with each unique name, and the "best" / most recent -- this will ensure 'start' shows up if it is still in memory (unless there are 200+ tasks of different names, which is unusual), instead of showing 20+ other tasks with the same name


algairim (approved):
Looks good.

PR #1338, opened 2022-07-13
Fix import state bundle order

ahgittin:
fixes a bug where state could be exported but not imported, because import didn't do the same gymnastics to allow for b2 referring to b1 in the case where b2 is loaded first.

the `CatalogInitialization` does careful preloading to allow this, on rebind.  it is refactored and used now also for the import case.

test added here fails without the fix commit, and passes now.


algairim (approved):

rest/rest-resources/src/main/java/org/apache/brooklyn/rest/resources/ServerResource.java
```diff
@@ -609,30 +615,23 @@ public Response importPersistenceData(byte[] persistenceData) {
...
-                    } else {
-                        errorMsg = Strings.isNonBlank(bundleInstallResult.getError().getMessage()) ? bundleInstallResult.getError().getMessage() : bundleInstallResult.getError().toString();
-                    }
-                    throw new Exception(errorMsg);
+            Map<VersionedName, CatalogInitialization.InstallableManagedBundle> bundles = new LinkedHashMap<>();
+            for (Map.Entry<String, ByteSource> bundleJar : newMementoRawData.getBundleJars().entrySet()) {
+                ManagedBundleMemento bundleMemento = mementoManifest.getBundle(bundleJar.getKey());
+                ManagedBundle b = RebindIteration.newManagedBundle(bundleMemento);
+                bundles.put(b.getVersionedName(), new RebindIteration.InstallableManagedBundleImpl(bundleMemento, b));
+                log.debug("Installing "+bundleMemento+" for "+b+" as part of persisted state import");
```
jcabrerizo, line 623: moving list line one line up  will help to see which bundle failed to install in case something goes wrong and also be more accurate

jcabrerizo (approved):
Thanks @ahgittin. I added a minor comment, feel free to ignore and merge the code as it is

PR #1342, opened 2022-07-18
Added fix and test for set<map<?,?>> config keys in container task fa…

iuliana:
Added fix and test for set<map<?,?>> config keys in container task factory.


jcabrerizo:
Thanks @iuliana

PR #1343, opened 2022-07-18
Bump SSHJ versions

ahgittin:
Fix issues SSH'ing to recent OSs


jcabrerizo (approved):
LGTM @ahgittin I'll build locally before merge

PR #1345, opened 2022-07-19
rejig container task

ahgittin:
* to track output, gather exit code, and clearer semantics
* also fix dirty detection across all processes


ahgittin:
reviewed live with @iuliana

PR #1347, opened 2022-07-29
Bash curl and certificates configurable

ahgittin:


algairim (approved):

ahgittin:
added ability for trust-all-certs behaviour to be configured

PR #1348, opened 2022-08-01
better check for thread interruption with sshj connections

ahgittin:
sshj clears the interrupted flag and hides the interruption exception inside a ConnectException;
now we look inside it

this also causes Thread.currentThread().interrupt() to be set when we propagate InterruptedException;
previously it was not


iuliana (approved):
This is quite a nice improvement.

iuliana:
Merging since the Jenkins failures do not seem to be related.

PR #1351, opened 2022-08-12
Skip property validation against the constraint if property is null

algairim:
This skips the validation of the property if no constratint is present
that has 'required' word or 'Predicates.notNull()', when the property
value is null.


jcabrerizo:
♻️

jcabrerizo (approved):
LGTM @algairim, I think the test errors it's not related, let me try to run it again before merging

PR #1354, opened 2022-08-23
SSH and SCP executables to get params via env vars

algairim:
This enables configuring a custom SSH adn SCP CLI tools allowing
processing applying password instead of private key for identification,
which is not allowed in SshCliTool by default.


ahgittin:
LGTM

PR #1357, opened 2022-09-07
support `item: URL` in catalog.bom, so a blueprint can be pulled in by reference

ahgittin:
how was this not already supported!?

also expand JAR tests to cover this case.
the test JAR build process needed tweaked to exclude the Force-Bundle headers provided (for some time now) by /parent/pom.xml. an empty such instruction in a pom doesn't work to override so a `-` value is now accepted to mean nothing.


algairim:
This seems to be working okay.

PR #1359, opened 2022-09-12
Workflow step to log messages

algairim:


algairim:
@geomacy, I invite you to look at the potential of workflows (types are not yet bundled for use). I really like the idea and I can imagine how I can wire up sensors and conditions of other entities into a workflow, and capture a re-usable pattern in a catalog. We might need to reconsider the ordering of steps. e.g. natural order of steps as declared, instead of numero-alphabetic ordering. What do you think?

https://user-images.githubusercontent.com/81319331/189756164-18120316-1cca-4086-8da5-5e0811bbcc49.mp4

ahgittin:
looks perfect

i wonder if a convention of `#-name` should be used for steps, even in our tests, to make behaviour clearer

PR #1361, opened 2022-09-13
variables and more step types for workflow

ahgittin:


ahgittin:
This is probably as big a chunk as we should consider in one go.  It builds on #1358 (also including the small #1359 and #1360) to add:

* step types around `{set,clear}-{sensor,config,workflow-variable}`
* inputs and outputs to workflow and individual steps
* expression resolution (using freemarker templates) for inputs, outputs, workflow variables
* a `workflow` for defining custom steps and nested workflow
* tidy up to log framework for testing

The new and updated test cases are probably the best place to start, to understand this.

PR #1369, opened 2022-11-03
Misc fixes for workflow

ahgittin:
Fix various glitches and awkward bits with workflow

* Entities and config wouldn't resolve in immediate scope in workflow
* Add a way to access other applications via DSL
* Fix the target for dsl predicate assertions in some cases
* Expand semantics of `let` step, allowing merge of maps
* Improve how inputs are resolved
* Ensure sensitive data from workflow and tasks are suppressed
* Prevent entity unwrapping in a few cases where it could lose information, and update tests
* Change freemarker default to be computer notation not locale notation (eg not `1,234`)

UI and docs PRs to follow, will reference this.

Proposed release notes:

* Workflow is added as a new way to define effectors, sensors, and policies; this is expected to be preferable to existing techniques in most cases, usually without requiring custom Java, and where custom Java is needed it should be considered switching the Java to be a new workflow step type

* $brooklyn DSL enhancements
  * application(ID) 
  * member(ID) keyword allows traversal through descendants and members
  * corresponding additional scopes for component(SCOPE, ID) per DSL

* Wrapper entities are stricter:  Brooklyn collapses "wrapper entities" in certain cases, such as if a catalog defines an entity wrapped in a 'services' block where that entity does nothing.  There are a few new cases where collapsing is blocked: if different values for display name or config are explicitly supplied given at the wrapper and inner level.

* Freemarker templates (used in workflow) now default to using "computer" syntax (as though the "?c" suffix is appended to numbers), so 1234 formats as "1234" and no longer "1,234" or "1.234".  If the latter locale-specific syntax is desired, use the "?string.number" suffix.


jcabrerizo (approved):
LGTM @ahgittin, thanks

iuliana:
@ahgittin LGTM too, just fix that test and it can be merged.

PR #1370, opened 2022-11-24
Better nested workflow, entitlements, on-error, lock, retention and replay

ahgittin:
This completes the essential workflow functionality.  Specifically:

* `workflow` step allows launching multiple workflows, loops, and concurrently
* a `lock` can be set for mutex behaviour
* a `switch` step is now available
* `replayable` semantics have been improved
* `on-error` lists are more natural (run as a sequence, not first-match-only)
* `retention` (ie `expiry`) is now configurable on a workflow


ahgittin:
The documentation has been updated at https://github.com/apache/brooklyn-docs/tree/master/guide/blueprints/workflow (in master) reflecting all the changes here.

In particular note the `settings.md` and the discussion of `on-error` in `common.md`, plus the `transform` and `switch` statements.

jcabrerizo (commented):
Great job @ahgittin

jcabrerizo (approved):

PR #1372, opened 2022-11-29
Prevent zip slip

jcabrerizo:
Avoid extracting zip files trying to extract files outside the provided path.
https://security.snyk.io/research/zip-slip-vulnerability


core/src/main/java/org/apache/brooklyn/util/core/file/ArchiveUtils.java
```diff
@@ -363,21 +363,38 @@ public static String readFullyString(File sourceFile) {
...
+            File destFile = new File(targetPath, originalName);
+            // validate input path
+            try {
+                String canonicalDestinationDirPath = null;
+                // enforce the file uses the appropriate file separator for the controller SO
+                String sanitizedName = originalName
+                        .replace("\\", File.separator)
+                        .replace("/", File.separator);
+                File sanitizedDestFile = new File(targetPath, sanitizedName);
+                canonicalDestinationDirPath = targetPath.getCanonicalPath();
```
ahgittin, line 381: i thought `getCanonicalPath()` didn't work unless the path existed.  but a quick test reveals i was wrong.  looks good.

core/src/test/java/org/apache/brooklyn/util/core/file/ArchiveUtilsTest.java
```diff
@@ -110,7 +112,31 @@ public void testDeployExplicitDestFile() throws Exception {
         ArchiveUtils.deploy(origJar.getAbsolutePath(), machine, destDir.getAbsolutePath(), destFile);
         assertFilesEqual(new File(destDir, destFile), origJar);
     }
-    
+    @Test(groups="Integration", expectedExceptions = IllegalStateException.class)
+    public void testUnzipFileAccessingPathOutsideTargetFolderEvilWinFormat() throws Exception{
```
ahgittin, line 116: I'd probably refactor these to a `doTestUnzip(url)` method.

i also prefer catching exceptions in the test body and asserting based on them, rather than part of the test annotation which is too easy to overlook and doesn't test as much as we might like.  eg

`Asserts.assertFailsWith(() -> doTestUnzip(url), e -> { Asserts.expectedExceptionContainsIgnoreCase(e, "outside"); return true; })`

Lastly we should probably do a `finally { tempZipFile.delete() }`, eg in case tests are interrupted.

but all comments are minor.  good tests, and good to merge without addressing these comments.

ahgittin:
Looks great.  Minor comments.  Can be merged.

core/src/main/java/org/apache/brooklyn/util/core/file/ArchiveUtils.java
jcabrerizo: Yeah, it work and it also resolve symbolic links that are also a risk

core/src/test/java/org/apache/brooklyn/util/core/file/ArchiveUtilsTest.java
jcabrerizo: Thanks @ahgittin I'll refactor that.

geomacy (commented):
Looks good. I would add a README.md alongsize `evilLinux.zip` to explain what they are for.

Some other minor comments, plus:

`evilLinux.zip` contains ` ..\..\..\..\..\..\..\..\script.sh` as well as the entry with the linux path, so the test actually fails already on this windows path, and never gets to check it handles the linux path OK.

core/src/test/java/org/apache/brooklyn/util/core/file/ArchiveUtilsTest.java
```diff
@@ -159,4 +149,19 @@ private void assertSubFilesEqual(File parentDir, Map<String, String> files) thro
...
+    private void doTestUnzip(String url)  {
+        File tempZipFile = null;
+        InputStream evilZip = ResourceUtils.create(this).getResourceFromUrl(url);
+        try {
+            tempZipFile = File.createTempFile("test-zip", null);
+            tempZipFile.deleteOnExit();
+            java.nio.file.Files.write(tempZipFile.toPath(), ByteStreams.toByteArray(evilZip), StandardOpenOption.TRUNCATE_EXISTING);
+            ArchiveUtils.extractZip(new ZipFile(tempZipFile), destDir.getAbsolutePath());
+        } catch (Exception e) {
+            throw new RuntimeException(e);
```
ahgittin, line 162: `throw Exceptions.propagate(e)` is slightly nicer; it handles wrapping if and only if necessary and some special edge cases

jcabrerizo:
Thank you @geomacy and @ahgittin. I think all comments are addressed now

PR #1373, opened 2022-12-06
add child tags at start of list, in front of parent tags

ahgittin:
this causes the right icon to be found by findFirst (and any other things that use findFirst)


camp/camp-brooklyn/src/main/java/org/apache/brooklyn/camp/brooklyn/spi/creation/BrooklynEntityDecorationResolver.java
```diff
@@ -278,11 +278,15 @@ protected TagsResolver(BrooklynYamlTypeInstantiator.Factory instantiator) {
         public void decorate(EntitySpec<?> entitySpec, ConfigBag attrs, Set<String> encounteredRegisteredTypeIds) {
             Iterable<Object> decorationAttributeJsonValue = getDecorationAttributeJsonValue(attrs);
             if (decorationAttributeJsonValue != null) {
-                entitySpec.tagsAdd(decorationAttributeJsonValue);
+                entitySpec.tagsAddAtStart(decorationAttributeJsonValue);
             }
+
+            // 1) append at start each loop as we descend, so findFirst works <- probably what we intend with tags, child tags should come first
```
zan-mateusz, line 284: shall we remove these comments?

PR #1374, opened 2022-12-06
Workflow startable, and map sensor key access

ahgittin:
Add a `workflow-entity` where `start` / `stop` are defined by workflow.
Also allow `set-sensor service.problems['key'] = "value"` syntax, and same for `clear-sensor`, to allow updating map entries.


jcabrerizo (approved):
LGTM @ahgittin, tks

PR #1378, opened 2023-02-03
Adds iconUrl to bundles

nakomis:


nakomis:
See also https://github.com/apache/brooklyn-docs/pull/373

PR #1379, opened 2023-02-15
Makes WorkflowSensor public and extensible

nakomis:
Removes final designation from WorkflowSensor allowing downstream projects to extend it. WorkflowPollCallable params are made visible to allow comparison of workflowsensors to determine if they've changed (e.g. for a diff)


zan-mateusz (approved):
LGTM

nakomis:
Removes `final` designation from `WorkflowSensor` allowing downstream projects to extend it. `WorkflowPollCallable` params are made visible to allow comparison of workflowsensors to determine if they've changed (e.g. for a diff)

PR #1387, opened 2023-04-10
Swagger fix

iuliana:
Some transitive types schemas generated by swagger are incorrect, this does not make the swagger schema invalid, but it does prevent Swagger from generating a proper HTML REST API page. This method fixes those types as we identify them.

**This fix is necessary until we upgrade Swagger to a version that does not have this bug.**


jcabrerizo (approved):
Thanks for fixing this @iuliana

PR #1390, opened 2023-04-18
adding logging category and level to workflow logs

lukiwlosek:


core/src/main/java/org/apache/brooklyn/core/workflow/steps/flow/LogWorkflowStep.java
```diff
@@ -26,12 +26,16 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import org.apache.brooklyn.core.BrooklynLogging;
+
 public class LogWorkflowStep extends WorkflowStepDefinition {
 
-    private static final Logger LOG = LoggerFactory.getLogger(LogWorkflowStep.class);
+    private static Logger LOG = LoggerFactory.getLogger(LogWorkflowStep.class);
```
duncangrant, line 33: I think global and mutable is probably not a good idea.

iuliana: @duncangrant It cannot be final if the category is configurable.

lukiwlosek: updated

PR #1394, opened 2023-04-20
Workflow entity child and member access pt 184904864

frogfather:
Improves access to child entities in workflow allowing children to be accessed as ${entity.children[n]} in addition to the current syntax $brooklyn:child("id")

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/util/core/text/TemplateProcessor.java
+++ b/core/src/main/java/org/apache/brooklyn/util/core/text/TemplateProcessor.java
@@ -56,6 +56,7 @@ import java.io.*;
 import java.lang.reflect.Method;
 import java.time.Instant;
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.Date;
 import java.util.Map;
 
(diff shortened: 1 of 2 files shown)


duncangrant:
This looks good to me.  Thanks @frogfather .
It does look like there are a couple of tests that are failing intermittently.  I've copied these below and will see if I can tell what's going on.

java.lang.AssertionError: null: lists don't have the same size expected [8] but found [16]

	at org.apache.brooklyn.core.workflow.WorkflowBasicTest.testWorkflowLogging(WorkflowBasicTest.java:381)



testSimpleErrorHandlerOnStep(org.apache.brooklyn.core.workflow.WorkflowPersistReplayErrorsTest)  Time elapsed: 0.276 sec  <<< FAILURE!

java.lang.AssertionError: List of test is different length to list of data to test expected [8] but found [112]

	at org.apache.brooklyn.core.workflow.WorkflowPersistReplayErrorsTest.testSimpleErrorHandlerOnStep(WorkflowPersistReplayErrorsTest.java:599)



testSimpleErrorHandlerOnWorkflow(org.apache.brooklyn.core.workflow.WorkflowPersistReplayErrorsTest)  Time elapsed: 0.271 sec  <<< FAILURE!

java.lang.AssertionError: List of test is different length to list of data to test expected [9] but found [111]

	at org.apache.brooklyn.core.workflow.WorkflowPersistReplayErrorsTest.testSimpleErrorHandlerOnWorkflow(WorkflowPersistReplayErrorsTest.java:630)

ahgittin:
there is no need for `children` to be a field; made a minor code cleanup in `3543694192880438b9cb03153472749a36dc1712`

also:  `members` was not handled, and the proposed map-style lookup `${entity.children["id"]}` was not handled.  these done in `e22c7325a39bfbb8d1ca7872dbfb7915cdaad913`.

PR #1398, opened 2023-05-10
Implements list index specifiers for workflows

nakomis:
Adds ability to specify list index in workflows, e.g.:

```
let list mylist = [1, 2, 3]
let mylist[1] = 4",
return ${mylist[1]}
```


ahgittin:
LGTM

PR #1400, opened 2023-05-22
Workflow - foreach, reducing, update-children, and more

ahgittin:
implementation of the enhancements described in https://github.com/apache/brooklyn-docs/pull/379, specifically:

* `reducing` on `workflow` and on a new `foreach` step
* ability to test if an effector is defined, eg `${entity.effector.start}` for if there is a `start` effector on `entity`
* tidy to forgive a rogue colon and to allow a list of steps to be interpreted as a workflow
* `update-children` (WIP)

### DIFF
--- a/core/src/main/java/org/apache/brooklyn/core/workflow/WorkflowStepResolution.java
+++ b/core/src/main/java/org/apache/brooklyn/core/workflow/WorkflowStepResolution.java
@@ -19,6 +19,7 @@
 package org.apache.brooklyn.core.workflow;
 
 import com.google.common.base.Predicates;
+import com.google.common.collect.Iterables;
 import com.google.common.reflect.TypeToken;
 import org.apache.brooklyn.api.entity.Entity;
 import org.apache.brooklyn.api.mgmt.ManagementContext;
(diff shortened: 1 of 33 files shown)


core/src/main/java/org/apache/brooklyn/core/workflow/WorkflowStepResolution.java
```diff
@@ -90,7 +91,18 @@ static WorkflowStepDefinition resolveStep(ManagementContext mgmt, Object def) {
                 if (s == null) s = defM.remove("shorthand");
                 if (s == null) s = defM.remove("s");
                 if (s == null) {
-                    throw new IllegalArgumentException("Step definition must indicate a `type` or a `step` / `shorthand` / `s` ("+def+")");
+                    if (defM.size()==1) {
+                        // assume the colon caused it accidentally to be a map
+                        s = Iterables.getOnlyElement(defM.keySet());
+                        if (s instanceof String && ((String)s).contains(" ")) {
```
iuliana, line 97: @ahgittin Probably could replace this `if` with an elvis operator.

iuliana (approved):
Looks good, one minor comment.

core/src/main/java/org/apache/brooklyn/core/workflow/WorkflowStepResolution.java
ahgittin: in this (rare!) instance i think the longer-winded syntax is worth it for readability

ahgittin:
`update-children` now completed, including recovering on interrupts with tests, and a few fixes for workflow/task interruption and resumption

ready for final review and merge

PR #1404, opened 2023-06-30
Allow on fire in workflow expressions

duncangrant:
When calling attributeWhenReady using workflow syntax we should not abort when the entity is on fire.  This is different from what we do using DSL expressions.

This gives us more control in pure workflow based entities where we can choose how to respond to the entities status


ahgittin:
Good idea and good impl.  I extended this in b4fd8d634b44981f28d3c323b55acfdf3930478f so that it only applies to workflow, not e.g. to template files being resolved, and merged.

PR #1408, opened 2023-08-29
Added support for command containers in workflow steps, containers ...

iuliana:
...that are run as a command and only require arguments. (e.g. `amazon/aws-cli`,  `hashicorp/terraform`). 
Steps examples: 

```
-  step: "container amazon/aws-cli"
   args: "dynamodb scan --table-name mytable --select COUNT"
   env: 
     AWS_ACCESS_KEY_ID: mytable
     AWS_SECRET_ACCESS_KEY: mysecret
     AWS_REGION: myregion
```
or

```
-  step: "container hashicorp/terraform"
   args: "version"
```

### DIFF
--- a/camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/WorkflowYamlTest.java
+++ b/camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/WorkflowYamlTest.java
@@ -800,10 +800,8 @@ public class WorkflowYamlTest extends AbstractYamlTest {
         Asserts.assertEquals(x2.getTask(false).get().getUnchecked(Duration.seconds(5)), "done");
     }
 
-    // prior to executing this test make sure you create an ec2 instance and tag it with "a2/vm"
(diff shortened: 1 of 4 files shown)


camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/WorkflowYamlTest.java
```diff
@@ -798,6 +800,41 @@ public void testLockReleasedOnCancel() throws Exception {
         Asserts.assertEquals(x2.getTask(false).get().getUnchecked(Duration.seconds(5)), "done");
     }
 
+    // prior to executing this test make sure you create an ec2 instance and tag it with "a2/vm"
```
duncangrant, line 803: Is it worth writing a simpler test that just does echo where the args are string including inline environment variables?

duncangrant (approved):
LGTM thanks @iuliana

camp/camp-brooklyn/src/test/java/org/apache/brooklyn/camp/brooklyn/WorkflowYamlTest.java
iuliana: We wrote simple echo tests and this is how we missed the fact we did not support command containers. :)

But yeah, I see what you mean. I will add an example using the terraform container just to query the version. Thx!

PR #1416, opened 2024-06-21
Lookup user using BrooklynSecurityProviderFilterHelper

duncangrant:
If not available in the context

This fixes a minor issue where the username is logged at the start of a rest request but at the end often fails because it doesn't look it up using BrooklynSecurityProviderFilterHelper.

Per @ahgittin 's suggestion it also adds the username to the logging context.  This means we can log the username with any activities that don't create a task.  This PR https://github.com/apache/brooklyn-dist/pull/186 will modify the log to do this.

It doesn't look like there are any unit tests here so I've just tested locally using the basicauth security provider - I assume this will work with other security providers but I think this will be specific to the provider.  It should at least do the same in both places now.

### DIFF
--- a/rest/rest-resources/src/main/java/org/apache/brooklyn/rest/filter/LoggingResourceFilter.java
+++ b/rest/rest-resources/src/main/java/org/apache/brooklyn/rest/filter/LoggingResourceFilter.java
@@ -46,6 +46,7 @@ import org.slf4j.LoggerFactory;
 
 import com.google.common.base.Joiner;
 import com.google.common.collect.ImmutableSet;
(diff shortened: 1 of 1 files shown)


ahgittin:
This looks good.

Just wondering if we want to set the _user_ (in the REST LoggingFilter) as the `task.id` also, using an `MDC` like we do for tasks, so that if there is no task, you at least see the user id in the log messages. ?

ahgittin:
Looks great.  Merging.

ahgittin:
Docs updated in 4ce267ec3d8ef268a06ac9549a72711aa2cf402b

PR #1424, opened 2025-11-17
Use makeRandomPassword instead of makeRandomId

snyff:
Since makeRandomPassword uses a cryptographically random generator and a larger keyspace, it is a better option than makeRandomId


grkvlt (approved):
lgtm

PR #1428, opened 2026-03-24
suppressing of secrets in complex JSON serializable types

zan-mateusz:
Sanitizer.java
- add a method to parse a JSON object from JSON string and suppress the JSON data if needed
- maintains the current behavior otherwise
- added unit tests as needed

ShellEnvironmentSerializer.java
- handles JSON/YAML string passed in the effector invoked. Previously this would just return the string value as is, leaving the sensitive data unsuppressed
- added unit tests as needed


ahgittin:
good fix - merged

PR #1430, opened 2026-04-21
Fixes/coercion error logbook class filtering dynamic update

zan-mateusz:
3 fixes in separate commits:

1. Type coercion when triggering an effector that contains a JSON list. 
When using a yaml list as input to an effector e.g.:
- hello
- world
Currently it is parsed as a multiline string so you get `-` and newline chars included. Commit 1 fixes this situation so we now get correctly: `"[\"hello\",\"world\"]"`

2. Logbook enhancement to allow filtering based on class/logger name perfix
Enhancement to allow filtering based on the prefix of logger name. It is accompanied by a minor UI tweak in a separate PR in brooklyn-ui. Current search capability does not allow searching for class names, just the message content, this enhancement allows it.


ahgittin:
LGTM - TY

PR #1432, opened 2026-06-19
fix: logbook search case insensitive

zan-mateusz:
Make both the logger prefix and search phrase filters case-insensitive and consistent across the file and OpenSearch backends.

- FileLogStore: lowercase both sides of the clazz prefix and message search-phrase comparisons (Locale.ROOT).
- OpenSearchLogStore: add case_insensitive:true to the class prefix query, which is term-level and otherwise only matches the lowercased indexed terms of the analyzed "class" field.


iuliana (approved):
LGTM

PR #1433, opened 2026-06-19
fix: coerce list-of-custom-type effector inputs supplied as YAML

zan-mateusz:
Invoking an effector whose parameter is a list of a registered (TOSCA) data type failed to coerce the supplied value. Two fixes:

- Parse YAML block-list strings ("- item" lines) into a list of elements for non-string element types, instead of collapsing the whole block into a single string (CommonAdaptorTypeCoercions).
- Run effector parameter coercion in the entity's execution context so registered bean types can resolve their management context, instead of failing with "... without a management context" (Effectors).


ahgittin:
great fix

